PR #394, opened 2021-03-18
Amplification parameter ramping

dmf7z:
This PR adds functionality to ramping the amplification parameter in the pool. Only the pool admin can start an update of the amp. parameter over a period of time or stop it. To reduce storage loading, the amplification data needed for the ramping is packed. 

WIP: tests are missing, but because there is a refactor of stable tests in another branch, will wait for that one to be merged to avoid rework.


contracts/pools/stable/StablePool.sol
```diff
@@ -49,11 +56,114 @@ contract StablePool is BaseGeneralPool, StableMath {
...
+            }
+        } else {
+            isUpdating = false;
+            value = targetValue;
+        }
+    }
+
+    //Admin functions
+
+    function startAmplificationParamaterUpdate(uint256 targetValue, uint256 targetBlock) external authenticate {
```
TomAFrench, line 117:
```suggestion
    function startAmplificationParameterUpdate(uint256 targetValue, uint256 targetBlock) external authenticate {
```

typo

nventuro:
I ported @dmf7z's PR to the new monorepo structure, though some things are still pending. These include:

- tests
- combining with #562 
- deciding whether amp is onlyOwner (I think it is, and I made it this way)
- proper maximum and minimum values, and better dealing with types. I hacked some casts together to not wory about those too much

nventuro:
Merging as is: created issues for pending tasks.

PR #537, opened 2021-04-23
Oracle: Model price pool oracle MVP

facuspagnuolo:


contracts/pools/oracle/PoolPriceOracle.sol
```diff
@@ -0,0 +1,63 @@
...
+     * based on the elapsed time from the initial timestamp of the current sample that was given. In case this diff is
+     * larger than a constant duration, in this case 2 minutes, the updated price information will be written to next
+     * sample in the buffer.
+     *
+     * @param logPairPrice The logarithmic value of the new reported pair price
+     * @param logBptPrice The logarithmic value of the new reported BPT price
+     * @param logInvariant The logarithmic value of the new reported invariant
+     *
+     * @return newSample Tells whether a new sample was created
+     * @return sampleIndex The index of the sample where the updated information was written
```
nventuro, line 48:
```suggestion
     * @dev Processes new price and invariant data, updating the current sample or creating a new one.
     * 
     * Receives the new logarithms of values to store: `logPairPrice`, `logBptPrice` and `logInvariant`, as well the index of 
     * the current sample, and the timestamp of its creation.
     *
     * The return value of `newSample` is true if a new sample was created, in which case `sampleIndex` is its index.
```

contracts/pools/oracle/PoolPriceOracle.sol
```diff
@@ -0,0 +1,63 @@
...
+        uint256 currentIndex,
+        int256 logPairPrice,
+        int256 logBptPrice,
+        int256 logInvariant
+    ) internal returns (bool newSample, uint256 sampleIndex) {
+        // solhint-disable not-rely-on-time
+        bytes32 sample = _samples[currentIndex].update(logPairPrice, logBptPrice, logInvariant, block.timestamp);
+        newSample = block.timestamp - currentSampleInitialTimestamp >= _MAX_SAMPLE_DURATION;
+        sampleIndex = newSample ? ((currentIndex + 1) % _BUFFER_SIZE) : currentIndex;
+        _samples[sampleIndex] = sample;
```
nventuro: * Every time this function is call a sample in the buffer will be updated. The sample to update will be determined
     * based on the elapsed time from the initial timestamp of the current sample that was given. In case this diff is
     * larger than a constant duration, in this case 2 minutes, the updated price information will be written to next
     * sample in the buffer.

```suggestion
        // solhint-disable not-rely-on-time
        
        // Read the last sample, updating it with the newly received data.
        bytes32 sample = _samples[currentIndex].update(logPairPrice, logBptPrice, logInvariant, block.timestamp);
        
        // We create a new sample if more than _MAX_SAMPLE_DURATION seconds have elapsed since the creation of
        // the current one. In other words, no sample accumulates data over a period larger than
        // _MAX_SAMPLE_DURATION.
        newSample = block.timestamp - currentSampleInitialTimestamp >= _MAX_SAMPLE_DURATION;
        sampleIndex = newSample ? ((currentIndex + 1) % _BUFFER_SIZE) : currentIndex;
        
        // Store the updated or new sample.
        _samples[sampleIndex] = sample;
```

contracts/pools/oracle/PoolPriceOracle.sol
```diff
@@ -0,0 +1,107 @@
...
+                currentSampleInitialTimestamp,
+                currentIndex,
+                logPairPrice,
+                logBptPrice,
+                logInvariant
+            );
+    }
+
+    /**
+     * @dev Same as `_processPriceData` but re-using the last log invariant stored in the current sample.
```
nventuro, line 70:
```suggestion
     * @dev Same as `_processPriceData` but re-using the last log invariant stored in the current sample, which doesn't
     * change (significantly) in swaps.
```

contracts/pools/oracle/PoolPriceOracle.sol
```diff
@@ -0,0 +1,107 @@
...
+                currentSampleInitialTimestamp,
+                currentIndex,
+                logPairPrice,
+                logBptPrice,
+                currentSample.lastLogInvariant()
+            );
+    }
+
+    /**
+     * @dev Private function that actually process a price information for a sample.
```
nventuro, line 91:
```suggestion
     * @dev Private function that actually processes new price information for a sample.
```

test/pools/PoolPriceOracle.test.ts
```diff
@@ -0,0 +1,291 @@
...
+
+        it('creates another sample', async () => {
+          const previousSample = await oracle.getSample(index);
+
+          if (elapsed > 0) await advanceTime(elapsed);
+          const tx = useLastInvariant
+            ? await oracle.processPriceDataWithLastInvariant(elapsed, index, newLogPairPrice, newLogBptPrice)
+            : await oracle.processPriceData(elapsed, index, newLogPairPrice, newLogBptPrice, newLogInvariant);
+
+          const expectedIndex = index + 1;
```
nventuro, line 195:
```suggestion
          const expectedIndex = (index + 1) % MAX_BUFFER_SIZE;
```

test/pools/PoolPriceOracle.test.ts
```diff
@@ -0,0 +1,291 @@
...
+          });
+
+          context('when the current timestamp is greater than the initial timestamp by less than 2 minutes', () => {
+            itUpdatesTheExistingSample(index, MINUTE);
+          });
+
+          context('when the current timestamp is greater than the initial timestamp by 2048 minutes', () => {
+            itUpdatesTheExistingSample(index, MINUTE * 3, true);
+          });
+        });
```
nventuro, line 275: I think we're missing this scenario here:

          context('when the current timestamp is greater than the initial timestamp by more than 2 minutes', () => {
            itCreatesAnotherSample(index, MINUTE * 3);
          });

PR #541, opened 2021-04-24
Weighted: Oracle math

nventuro:


contracts/lib/math/SignedFixedPoint.sol
```diff
@@ -0,0 +1,83 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+
+import "./LogExpMath.sol";
+import "../helpers/BalancerErrors.sol";
+
+/* solhint-disable private-vars-leading-underscore */
+
+library SignedFixedPoint {
```
facuspagnuolo, line 22: Wouldn't be possible to support both signed and unsigned ops in the same `FixedPoint` lib? Similarly to how we are doing in `Math`

nventuro: Yes, sorry - we're actually not using this file at all anymore, I forgot to delete it.

PR #542, opened 2021-04-24
Oracle: Add integration tests for weighted pool

facuspagnuolo:


contracts/pools/weighted/WeightedPool2Tokens.sol
```diff
@@ -571,8 +570,10 @@ contract WeightedPool2Tokens is
         _downscaleDownArray(dueProtocolFeeAmounts, scalingFactors);
 
         // Update cached total supply and invariant using the results after the exit that will be used for future
-        // oracle updates.
-        _cacheInvariantAndSupply();
+        // oracle updates, only if the pool was not paused.
```
nventuro, line 573:
```suggestion
        // oracle updates, only if the pool was not paused (to minimize code paths taken while paused).
```

test/pools/weighted/WeightedPool.test.ts
```diff
@@ -42,6 +43,188 @@ describe('WeightedPool', function () {
...
+        it('updates the oracle data', async () => {
+          const previousData = await pool.instance.miscData();
+
+          await advanceTime(MINUTE * 10); // force index update
+          await action(await calcLastChangeBlock(lastChangeBlockOffset));
+
+          const currentMiscData = await pool.instance.miscData();
+          expect(currentMiscData.oracleIndex).to.equal(previousData.oracleIndex.add(1));
+          expect(currentMiscData.oracleSampleInitialTimestamp).to.equal(await currentTimestamp());
+        });
```
nventuro, line 82: Hm, I was thinking of having another `it` here that doesn't advance time and checks the update, but that'd require these tests knowing too much about the update procedure.

nventuro (approved):
These seem just perfect :ok_hand:

PR #543, opened 2021-04-24
Weighted: Add oracle enable method

facuspagnuolo:


contracts/pools/weighted/WeightedPool2Tokens.sol
```diff
@@ -185,20 +188,29 @@ contract WeightedPool2Tokens is
...
+    }
+
     // Caller must be approved by the Vault's Authorizer
     function setPaused(bool paused) external authenticate {
         _setPaused(paused);
     }
 
+    function _isOwnerOnlyAction(bytes32 actionId) internal view override returns (bool) {
+        // This implementation adds support for the enableOracle action identifier.
+        return super._isOwnerOnlyAction(actionId) || actionId == getActionId(WeightedPool2Tokens.enableOracle.selector);
```
nventuro, line 211: The requirement is that governance can _always_ enable the oracle. For simplicity's sake, I thought of not letting the owner do this (which means governance can) - it seems like a weird permission for the owner regardless. If they want it on, they can always turn it on at creation.

(meaning I'd remove this override)

contracts/pools/weighted/WeightedPool2Tokens.sol
```diff
@@ -736,7 +748,7 @@ contract WeightedPool2Tokens is
     // Oracle functions
 
     function _updateOracle(uint256 lastChangeBlock) internal {
-        if (_isOracleEnabled() && block.number > lastChangeBlock) {
+        if (isOracleEnabled() && block.number > lastChangeBlock) {
             bytes32 miscData = _miscData;
```
nventuro, line 752:
```suggestion
        bytes32 miscData = _miscData;
        bool oracleEnabled = _miscData.decodeBool(_MISC_ORACLE_ENABLED_OFFSET);
        if (oracleEnabled() && block.number > lastChangeBlock) {
```

Seems a bit silly to read _miscData twice, no? Regardless, we can ignore this now, and focus on performance once all parts are ready.

contracts/pools/weighted/WeightedPool2Tokens.sol
```diff
@@ -185,20 +188,29 @@ contract WeightedPool2Tokens is
         emit SwapFeePercentageChanged(swapFeePercentage);
     }
 
-    function _isOracleEnabled() private view returns (bool) {
-        return _miscData.decodeBool(_MISC_ORACLE_ENABLED_OFFSET);
+    function enableOracle() external authenticate {
+        _setOracleEnabled(true);
```
nventuro, line 192:
```suggestion
        _setOracleEnabled(true);
        _cacheInvariantAndSupply();
```

contracts/pools/weighted/WeightedPool2Tokens.sol
facuspagnuolo: Yea, I did this initially without taking into account gas reads/writes optimizations for this variable in particular... I have a whole plan to refactor the way we access to it entirely once the integration is finished. I wouldn't spend to much time about it rn.

contracts/pools/weighted/WeightedPool2Tokens.sol
facuspagnuolo: Agreed as we discussed off-line... let's add some tests for it

nventuro (approved):

PR #546, opened 2021-04-26
Weighted: Optimize misc data

facuspagnuolo:


nventuro (approved):

contracts/pools/weighted/WeightedPool2Tokens.sol
```diff
@@ -261,6 +245,7 @@ contract WeightedPool2Tokens is
         // Update price oracle with the pre-swap balances
         bool tokenInIsToken0 = request.tokenIn == _token0;
         _updateOracle(
+            _getMiscData(),
```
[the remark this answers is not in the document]
facuspagnuolo: Yup, it was WIP, just wanted to show you the idea... there are many other places where we need to optimize this read, not just here

contracts/pools/weighted/WeightedPool2Tokens.sol
```diff
@@ -243,17 +239,20 @@ contract WeightedPool2Tokens is
...
             request.lastChangeBlock,
             tokenInIsToken0 ? balanceTokenIn : balanceTokenOut,
             tokenInIsToken0 ? balanceTokenOut : balanceTokenIn
         );
 
         if (request.kind == IVault.SwapKind.GIVEN_IN) {
             // Fees are subtracted before scaling, to reduce the complexity of the rounding direction analysis.
-            request.amount = _upscale(_subtractSwapFeeAmount(request.amount), scalingFactorTokenIn);
+            // This is amount - fee amount, so we round up (favoring a higher fee amount).
+            uint256 feeAmount = request.amount.mulUp(miscData.swapFeePercentage);
```
nventuro, line 254: What do you think about `_subtractSwapFeeAmount(amount, fee)`, to avoid the inline computation?

nventuro (approved):

contracts/pools/weighted/WeightedPool2Tokens.sol
facuspagnuolo: if we are not re-using it anywhere else, I'd keep it like this for now

PR #554, opened 2021-04-28
Weighted: Reduce bytecode size

facuspagnuolo:
It reduces the bytecode of the weighted pool factory for 2 tokens from `24.726` to `23.942`
It also reduces the gas costs of swaps, e.g. for one pool it's being reduced from `133.1k` and `110.2k` to `132.5k` and `110.0k`, for buffer updates and re-use respectively

### DIFF
--- a/test/pools/weighted/WeightedPool.test.ts
+++ b/test/pools/weighted/WeightedPool.test.ts
@@ -344,7 +344,7 @@ describe('WeightedPool', function () {
         });
       });
 
-      describe('oracle setting', () => {
+      describe.only('oracle setting', () => {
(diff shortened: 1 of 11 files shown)


test/pools/weighted/WeightedPool.test.ts
```diff
@@ -344,7 +344,7 @@ describe('WeightedPool', function () {
         });
       });
 
-      describe('oracle setting', () => {
+      describe.only('oracle setting', () => {
```
nventuro, line 347:
```suggestion
      describe('oracle setting', () => {
```

PR #562, opened 2021-05-12
Stable: Avoid fixed-point amp

facuspagnuolo:
By making the Amplification factor non fixed point (i.e. a value of 50.0 is represented as 50 instead of 50e18), we can skip fixed point arithmetic in many of the invariant calculations, saving gas.


nventuro:
Superceded by #609

PR #563, opened 2021-05-17
Stable: Fix token balance getter given invariant

facuspagnuolo:
Part of #572.


nventuro:
We can close this given #609, no?

nventuro:
Superceded by #609

PR #573, opened 2021-05-20
Extract benchmarks and improve deploy

nventuro:
This moves the benchmarks outside of the `v2-core` package, and fixes the (hopefully) last outstanding problem: using `ethers.getContractAt`, which doesn't support loading artifacts from other packages.

I ended up reworking `deploy` and added `deployedAt`, which is a replacement for `ethers.getContracAt` (which we shouldn't use anymore). To reduce verbosity, I made it so the `@balancer-labs` prefix and contract path do not need to be known: all that is passed to these two functions are the (reduced) package name and contract name.


nventuro:
From the documentation of `deploy`:

>// Local contracts are deployed by simply passing the contract name, contracts from other packages must be prefixed by
>// the package name, without the @balancer-labs scope. Note that the full path is never required.
>//
>// For example, to deploy Vault.sol from the package that holds its artifacts, use `deploy('Vault')`. To deploy it from
>// a different package, use `deploy('v2-vault/Vault')`, assuming the Vault's package is @balancer-labs/v2-vault.

TomAFrench (approved):

PR #579, opened 2021-05-24
Fix exec failing silently on Mac OS when deploying

acryptosx:


nventuro (approved):
Thanks a lot @acryptosx! I had totally forgotten about the differences these commands exhibit in different environments.

PR #580, opened 2021-05-25
Added Distributor package for distributing rewards

gtaschuk:
This adds two implementations of a `Distributor` - a contract that distributes tokens such as rewardTokens from liquidity mining earned by an asset manager, or BAL.

**MultiRewards**
Allows users to claim rewards proportional to staked bpt
A user may either
- `getRewards` claim these tokens to an EOA
- `getRewardsAsInternalBalance` claim these tokens to internal balance, for easy reinvestment/swaps

Credit to @TomAFrench who started this work on  `MultiRewards` on the asset-managers repo

**MerkleRedeem**
Allows users to claim rewards per the entries in a merkle tree.  It's based on: https://github.com/balancer-labs/erc20-redeemable/tree/master/merkle/contracts.  An alternative approach for rewards computed off chain would be a `permit` type scheme to facillitate custom distributions calculated off-chain by bLabs or partners

**Open questions for the future:**

*Should a `Distributor` contract hold tokens as ERC20 balance, or as internal balance?*
I think this question boils down to, do we want to prioritize cheap withdrawals to EOAs or cheap withdrawals to vault balance or EOAs:
- Holding them as internal balance would facilitate cheaper claims for LPs to internal balance - useful if you want to immediately reinvest these tokens into a pool
- Holding them as internal balance would make a claim to ERC20 balance more expensive due to vault withdrawal logic
- Holding them in internal balance increases TVL in the vault

For this PR, I implemented distributing from ERC20 balance, as I think it's simpler and cheaper for the standard use case of claiming tokens to an EOA

*Should this be used directly in the pool or into the asset manager*
Which would simplify staking/unstaking, deployment and minimize transfers required when funneling assets to these contracts

*Should we have a singleton distributor that allows you to stake bpt from multiple pools*
This would make the logic more complex but would allow LPs to claim reward tokens across pools in a single transaction

Test helper `expectBalanceChange` now accepts a `vault` arg to track internal balance changes

### DIFF
--- a/pkg/distributors/contracts/MerkleRedeem.sol
+++ b/pkg/distributors/contracts/MerkleRedeem.sol
@@ -1,4 +1,16 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 pragma experimental ABIEncoderV2;
 
--- a/pkg/distributors/contracts/MerkleRedeem.sol
+++ b/pkg/distributors/contracts/MerkleRedeem.sol
@@ -1,18 +1,23 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
 pragma experimental ABIEncoderV2;
 
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/Ownable.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/MerkleProof.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/IERC20.sol";
+import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/SafeERC20.sol";
 
 import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
 import "@balancer-labs/v2-vault/contracts/interfaces/IAsset.sol";
 
+import "./interfaces/IDistributor.sol";
+
 pragma solidity ^0.7.0;
 
-contract MerkleRedeem is Ownable {
-    IERC20 public rewardToken;
+contract MerkleRedeem is IDistributor, Ownable {
+    using SafeERC20 for IERC20;
 
-    event Claimed(address _claimant, uint256 _balance);
+    IERC20 public rewardToken;
 
     // Recorded weeks
     mapping(uint256 => bytes32) public weekMerkleRoots;
@@ -28,7 +33,7 @@ contract MerkleRedeem is Ownable {
 
     function _disburse(address _recipient, uint256 _balance) private {
         if (_balance > 0) {
-            emit Claimed(_recipient, _balance);
+            emit RewardPaid(_recipient, address(rewardToken), _balance);
             require(rewardToken.transfer(_recipient, _balance), "ERR_TRANSFER_FAILED");
         }
     }
@@ -45,9 +50,8 @@ contract MerkleRedeem is Ownable {
                 kind: IVault.UserBalanceOpKind.DEPOSIT_INTERNAL
             });
 
+            emit RewardPaid(_recipient, address(rewardToken), _balance);
             vault.manageUserBalance(ops);
-
-            emit Claimed(_recipient, _balance);
         }
     }
 
@@ -144,8 +148,20 @@ contract MerkleRedeem is Ownable {
         return MerkleProof.verify(_merkleProof, weekMerkleRoots[_week], leaf);
     }
 
-    function seedAllocations(uint256 _week, bytes32 _merkleRoot) external onlyOwner {
+    /**
+     * @notice
+     * Allows the owner to add funds to the contract as a merkle tree, These tokens will
+     * be withdrawn from the sender
+     * These will be pulled from the user
+     */
+    function seedAllocations(
+        uint256 _week,
+        bytes32 _merkleRoot,
+        uint256 amount
+    ) external onlyOwner {
         require(weekMerkleRoots[_week] == bytes32(0), "cannot rewrite merkle root");
+        rewardToken.safeTransferFrom(msg.sender, address(this), amount);
         weekMerkleRoots[_week] = _merkleRoot;
+        emit RewardAdded(address(rewardToken), amount);
     }
 }
--- a/pkg/distributors/contracts/MerkleRedeem.sol
+++ b/pkg/distributors/contracts/MerkleRedeem.sol
@@ -14,6 +14,7 @@
 
 pragma experimental ABIEncoderV2;
 
+import "@balancer-labs/v2-solidity-utils/contracts/math/FixedPoint.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/Ownable.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/MerkleProof.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/IERC20.sol";
@@ -27,6 +28,7 @@ import "./interfaces/IDistributor.sol";
 pragma solidity ^0.7.0;
 
 contract MerkleRedeem is IDistributor, Ownable {
+    using FixedPoint for uint256;
     using SafeERC20 for IERC20;
 
     IERC20 public rewardToken;
@@ -46,7 +48,7 @@ contract MerkleRedeem is IDistributor, Ownable {
     function _disburse(address _recipient, uint256 _balance) private {
         if (_balance > 0) {
             emit RewardPaid(_recipient, address(rewardToken), _balance);
-            require(rewardToken.transfer(_recipient, _balance), "ERR_TRANSFER_FAILED");
+            rewardToken.safeTransfer(_recipient, _balance);
         }
     }
 
@@ -115,7 +117,7 @@ contract MerkleRedeem is IDistributor, Ownable {
                 "Incorrect merkle proof"
             );
 
-            totalBalance += claim.balance;
+            totalBalance = totalBalance.add(claim.balance);
             claimed[claim.week][_liquidityProvider] = true;
         }
 
--- a/pkg/distributors/contracts/MerkleRedeem.sol
+++ b/pkg/distributors/contracts/MerkleRedeem.sol
@@ -31,40 +31,40 @@ contract MerkleRedeem is IDistributor, Ownable {
     using FixedPoint for uint256;
     using SafeERC20 for IERC20;
 
-    IERC20 public rewardToken;
+    IERC20 public immutable rewardToken;
 
     // Recorded weeks
     mapping(uint256 => bytes32) public weekMerkleRoots;
     mapping(uint256 => mapping(address => bool)) public claimed;
 
-    IVault public vault;
+    IVault public immutable vault;
 
-    constructor(address _vault, address _rewardToken) {
-        vault = IVault(_vault);
-        rewardToken = IERC20(_rewardToken);
-        rewardToken.approve(address(vault), type(uint256).max);
+    constructor(IVault _vault, IERC20 _rewardToken) {
+        vault = _vault;
+        rewardToken = _rewardToken;
+        IERC20(_rewardToken).approve(address(_vault), type(uint256).max);
     }
 
-    function _disburse(address _recipient, uint256 _balance) private {
-        if (_balance > 0) {
-            emit RewardPaid(_recipient, address(rewardToken), _balance);
-            rewardToken.safeTransfer(_recipient, _balance);
+    function _disburse(address recipient, uint256 balance) private {
+        if (balance > 0) {
+            emit RewardPaid(recipient, address(rewardToken), balance);
+            rewardToken.safeTransfer(recipient, balance);
         }
     }
 
-    function _disburseToInternalBalance(address payable _recipient, uint256 _balance) private {
-        if (_balance > 0) {
+    function _disburseToInternalBalance(address payable recipient, uint256 balance) private {
+        if (balance > 0) {
             IVault.UserBalanceOp[] memory ops = new IVault.UserBalanceOp[](1);
 
             ops[0] = IVault.UserBalanceOp({
                 asset: IAsset(address(rewardToken)),
-                amount: _balance,
+                amount: balance,
                 sender: address(this),
-                recipient: _recipient,
+                recipient: recipient,
                 kind: IVault.UserBalanceOpKind.DEPOSIT_INTERNAL
             });
 
-            emit RewardPaid(_recipient, address(rewardToken), _balance);
+            emit RewardPaid(recipient, address(rewardToken), balance);
             vault.manageUserBalance(ops);
         }
     }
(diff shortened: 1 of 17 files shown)


TomAFrench:
> _Should the `MultiRewards` contract hold tokens as ERC20 balance, or as internal balance?_

Ideally it would be as an internal balance although I'm aware we shouldn't over-penalise withdrawals to EOAs. I'm sure we have a gas benchmark for withdrawing from internal balances somewhere though. If this cost is <10000 (rough number) gas I think it's worth it - especially as savings from internal balances will increase with 2+ tokens being claimed. 

> _Should this be used directly in the pool or into the asset manager_

I'm of the view that we should try and keep pool logic as separate as possible from staking logic. I don't feel like the benefits of avoiding two transactions when you enter and leave the pool outweigh the additional complexity and headaches of not being able to migrate without doing a full migration of a pool, especially as we're still in the process of rolling out liquidity mining.

We can also offset some of the additional gas costs of a separate staking contract using a relayer.
 
> _Should we have a singleton distributor that allows you to stake bpt from multiple pools_

I'm open to this, however before we went too down this road I'd want to do a comparison with a simpler multicall implementation with separate contracts. I feel much of the savings would be removal of paying the 21000 gas multiple times.

TomAFrench:
I've just added a commit which takes advantage of BPTs supporting permit so that we can avoid users having to approve the staking contract.

TomAFrench:
image

I've put together a quick script to benchmark the costs of performing withdrawals from  internal/external balances. There's about a 17.2k gas difference between storing the erc20 balances within the vault or not, at time of writing this is about $3.

---

This is done using `TestToken` which is an extremely lightweight implementation. I'd expect that internal balance -> internal balance transfer would perform better than in this benchmark when we use tokens such as Aave with balance snapshotting happening on transfers.

Once we use 2 rewards tokens then internal balance transfers start to outperform even with lightweight tokens.
image

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -0,0 +1,284 @@
...
+        IVault.UserBalanceOp[] memory ops = new IVault.UserBalanceOp[](rewardTokens.length);
+
+        for (uint256 i; i < rewardTokens.length; i++) {
+            address _rewardsToken = rewardTokens[i];
+            uint256 reward = rewards[msg.sender][_rewardsToken];
+
+            if (reward > 0) {
+                rewards[msg.sender][_rewardsToken] = 0;
+
+                ops[i] = IVault.UserBalanceOp({
```
gtaschuk, line 192: @nventuro does the vault handle empty `UserBalanceOp`s?

(If there were multiple reward tokens and a user wasn't receiving a reward for all of them)

TomAFrench: That should end up being equivalent to 
```
op = {
  asset: address(0),
  amount: 0,
  sender: address(0),
  recipient: address(0),
  kind: DEPOSIT_INTERNAL
}
```

Which would end up falling afoul of the relayer auth checks. It's probably worth just doing the zero value transfers instead of looping over the assets twice.

TomAFrench: This should only really happen in situations where a rewards token is no longer being fed into the staking contract.

It might be worth maintaining separate arrays of assets - one for all rewards ever and one for "active" rewards. An admin can then set tokens inactive after the rewards period runs out and then users can explicitly claim them. This could be automated to not use an admin but that's implementation details.

TomAFrench: Just noticed that we're doing deposits here rather than internal transfers. I think in this case it's worth looping twice to avoid the external calls to each token contract.

pkg/distributors/test/MerkleRedeem.test.ts
```diff
@@ -0,0 +1,273 @@
...
+    // deploy pool and add liquidity
+    const specialization = GeneralPool;
+    const pool = await deploy('v2-vault/MockPool', { args: [vault.address, specialization] });
+    poolId = await pool.getPoolId();
+
+    await tokens.mint({ to: lp1, amount: tokenInitialBalance });
+    await tokens.approve({ to: vault.address, from: [lp1] });
+
+    const assets = tokens.addresses;
+    const assetManagers = [mockAssetManager.address, mockAssetManager.address];
```
nventuro, line 69:
```suggestion
    const assetManagers = [mockAssetManager.address, mockAssetManager.address];
```

I think we don't need these, I'd just pass the zero address instead.

pkg/distributors/test/MerkleRedeem.test.ts
```diff
@@ -0,0 +1,273 @@
...
+  });
+
+  it('stores an allocation', async () => {
+    const claimBalance = bn('9876');
+
+    const elements = [encodeElement(lp1.address, claimBalance)];
+    const merkleTree = new MerkleTree(elements);
+    const root = merkleTree.getHexRoot();
+
+    await merkleRedeem.connect(admin).seedAllocations(bn(1), root, claimBalance);
```
nventuro, line 90: Let's add another test where we call `seedAllocation` and check that a) the tokens are transferred from the admin into the distributor, and b) the distributor emits an event.

pkg/distributors/test/MerkleRedeem.test.ts
```diff
@@ -0,0 +1,273 @@
...
+    mockAssetManager: SignerWithAddress;
+  let poolId: string;
+  const rewardTokenInitialBalance = bn(100e18);
+  const tokenInitialBalance = bn(200e18);
+
+  before('deploy base contracts', async () => {
+    [admin, lp1, lp2, lp3, mockAssetManager] = await ethers.getSigners();
+  });
+
+  beforeEach('set up tokens and redeem contract', async () => {
```
nventuro, line 44: Using `sharedBeforeEach` here should make the tests much faster.

pkg/distributors/test/MerkleRedeem.test.ts
```diff
@@ -0,0 +1,273 @@
...
+    const proof0 = merkleTree.getHexProof(elements[0]);
+    let result = await merkleRedeem.verifyClaim(lp1.address, 1, claimBalance0, proof0);
+    expect(result).to.equal(true); //"account 0 should have an allocation";
+
+    const proof1 = merkleTree.getHexProof(elements[1]);
+    result = await merkleRedeem.verifyClaim(lp2.address, 1, claimBalance1, proof1);
+    expect(result).to.equal(true); // "account 1 should have an allocation";
+  });
+
+  describe('When a user has an allocation to claim', () => {
```
nventuro, line 135: nit: i'd use the same capitalization in all test descriptions (I believe we're keeping them all lowercase)

pkg/distributors/test/MerkleRedeem.test.ts
```diff
@@ -0,0 +1,273 @@
...
+    const claimBalance = bn('1000');
+    let elements: string[];
+    let merkleTree: MerkleTree;
+
+    beforeEach(async () => {
+      elements = [encodeElement(lp2.address, claimBalance)];
```
nventuro, line 141:
```suggestion
    const week = 1;
    let elements: string[];
    let merkleTree: MerkleTree;
    let lp: SignerWithAddress;

    beforeEach(async () => {
      lp = lp2.address;
      elements = [encodeElement(lp2.address, claimBalance)];
```

I think this will make the tests easier to read.

pkg/distributors/test/MerkleRedeem.test.ts
```diff
@@ -0,0 +1,273 @@
...
+      merkleTree = new MerkleTree(elements);
+      const root = merkleTree.getHexRoot();
+
+      await merkleRedeem.seedAllocations(1, root, claimBalance);
+    });
+
+    it('Allows the user to claimWeek', async () => {
+      const claimedBalance = bn('1000');
```
nventuro, line 149:
```suggestion
```
This is already defined at the top (same in other `it`s).

pkg/distributors/test/MerkleRedeem.test.ts
```diff
@@ -0,0 +1,273 @@
...
+      const merkleProof: BytesLike[] = merkleTree.getHexProof(elements[0]);
+      await merkleRedeem.connect(lp2).claimWeek(lp2.address, 1, claimedBalance, merkleProof, false);
+
+      const result = await rewardToken.balanceOf(lp2.address);
+      expect(result).to.equal(claimedBalance); //"user should have an allocation";
```
nventuro, line 154: I'd replace this for usage of `expectBalanceChange`

pkg/distributors/test/MerkleRedeem.test.ts
```diff
@@ -0,0 +1,273 @@
...
+      merkleTree = new MerkleTree(elements);
+      const root = merkleTree.getHexRoot();
+
+      await merkleRedeem.seedAllocations(1, root, claimBalance);
+    });
+
+    it('Allows the user to claimWeek', async () => {
+      const claimedBalance = bn('1000');
+      const merkleProof: BytesLike[] = merkleTree.getHexProof(elements[0]);
+      await merkleRedeem.connect(lp2).claimWeek(lp2.address, 1, claimedBalance, merkleProof, false);
```
nventuro, line 151: I'd explcitly call this from an account other than `lp` (which i'd just call `other`).

pkg/distributors/test/MerkleRedeem.test.ts
```diff
@@ -0,0 +1,273 @@
...
+
+      const result = await rewardToken.balanceOf(lp2.address);
+      expect(result).to.equal(claimedBalance); //"user should have an allocation";
+
+      const isClaimed = await merkleRedeem.claimed(1, lp2.address);
+      expect(isClaimed).to.equal(true); // "claim should be marked as claimed";
```
nventuro, line 157: WDYT about extracting this into `it('claimed weeks should be marked as claimed')`? It'd let us remove the small comments explaining what each assertion means, and individual `it`s are extremely fast due to the shared setup.

pkg/distributors/test/MerkleRedeem.test.ts
```diff
@@ -0,0 +1,273 @@
...
+        rewardTokens,
+        [{ account: lp2, changes: { DAI: claimedBalance } }],
+        vault
+      );
+
+      const isClaimed = await merkleRedeem.claimed(1, lp2.address);
+      expect(isClaimed).to.equal(true); // "claim should be marked as claimed";
+    });
+
+    it("Doesn't allow a user to claim for another user", async () => {
```
nventuro, line 175: Nit: for consistency's sake I'd also name this test 'reverts when...'

pkg/distributors/test/MerkleRedeem.test.ts
```diff
@@ -0,0 +1,273 @@
...
+
+      const proof1: BytesLike[] = merkleTree1.getHexProof(elements1[0]);
+      const proof2: BytesLike[] = merkleTree2.getHexProof(elements2[0]);
+
+      const merkleProofs = [
+        { week: bn(1), balance: claimedBalance1, merkleProof: proof1 },
+        { week: bn(2), balance: claimedBalance2, merkleProof: proof2 },
+      ];
+
+      await merkleRedeem.connect(lp2).claimWeeks(lp2.address, merkleProofs, false);
```
nventuro, line 245: Let's also add `expectBalanceChange` and event tests here (maybe in different `it`s).

pkg/distributors/test/MerkleRedeem.test.ts
```diff
@@ -0,0 +1,273 @@
...
+
+      expectedResult = [true, false];
+      result = await merkleRedeem.claimStatus(lp2.address, 1, 2);
+      expect(result).to.eql(expectedResult); // "claim status should be accurate"
+    });
+
+    it('Returns an array of merkle roots', async () => {
+      const expectedResult = [root1, root2];
+      const result = await merkleRedeem.merkleRoots(1, 2);
+      expect(result).to.eql(expectedResult); // "claim status should be accurate"
```
nventuro, line 270: Note that it is not the claim status we're testing here. Overall, I'd try to have these tests written in a way such that these comments are not needed because what they're testing is already clear from the test description. 

In this case, I don't think we need the comment at all. In other tests that test multiple things (e.g. `verifyClaim` works, `redeem` works, tokens are transferred, events are emitted, the claim status is updated), I'd consider splitting those in smaller tests that check each individual property.

nventuro (approved):
Looking great, thanks @gtaschuk!

I mostly reviewed `MerkleRedeem` and its tests, didn't get to looking much at `MultiRewards` yet (other than some passing comments). Keeping https://github.com/balancer-labs/balancer-core-v2/pull/582 separate might have worked better, as it seems these two contracts are not very related (other than a shared interface) - we should try to favor smaller PRs to make the review, address issues, merge cycle easier and faster.

nventuro:
(woops, my review was meant to be comment, not approve :sweat_smile:)

pkg/distributors/contracts/MerkleRedeem.sol
```diff
@@ -0,0 +1,163 @@
...
+        uint256 _week,
+        uint256 _claimedBalance,
+        bytes32[] memory _merkleProof,
+        bool internalBalance
+    ) public {
+        require(!claimed[_week][_liquidityProvider], "cannot claim twice");
+        require(verifyClaim(_liquidityProvider, _week, _claimedBalance, _merkleProof), "Incorrect merkle proof");
+
+        claimed[_week][_liquidityProvider] = true;
+        if (internalBalance) {
```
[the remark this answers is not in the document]
gtaschuk: Ah, yeah, I guess now that the user chooses where the rewards go we should disallow other people claiming

pkg/distributors/contracts/MerkleRedeem.sol
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    function _disburse(address _recipient, uint256 _balance) private {
+        if (_balance > 0) {
+            emit RewardPaid(_recipient, address(rewardToken), _balance);
+            require(rewardToken.transfer(_recipient, _balance), "ERR_TRANSFER_FAILED");
+        }
+    }
+
+    function _sweep(address payable _recipient, uint256 _balance) private {
```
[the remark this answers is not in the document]
gtaschuk: yes.  Thats the difference.

renamed sweep -> disburseToInternalBalance

pkg/distributors/test/MerkleRedeem.test.ts
gtaschuk: ah, not quite... `claimBalance` was defined at the top as that is what is the merkle tree, but these claimedBalance vars are what the lp is attempting to claim.  Will rename `claimBalance` to `claimableBalance`

pkg/distributors/test/MultiRewards.test.ts
```diff
@@ -0,0 +1,268 @@
...
+
+  before(async () => {
+    [, , lp, other] = await ethers.getSigners();
+  });
+
+  describe('stakeWithPermit', () => {
+    it('successfully stakes with a permit signature', async () => {
+      const bptBalance = await pool.balanceOf(lp.address);
+
+      const { chainId } = await ethers.provider.getNetwork();
```
nventuro, line 112: There's a helper for this in `v2-helpers/models/misc/signatures` called `signPermit`. Sample usage (from the BPT tests):

```
const { v, r, s } = await signPermit(token, holder, spender, amount);
```

nventuro (approved):

pkg/distributors/contracts/MerkleRedeem.sol
```diff
@@ -0,0 +1,181 @@
...
+
+    /**
+     * @notice Allows a user to claim a particular weeks worth of rewards
+     */
+    function claimWeeks(
+        address payable liquidityProvider,
+        Claim[] memory claims,
+        bool useInternalBalance
+    ) public {
+        require(msg.sender == liquidityProvider, "user must claim own balance");
```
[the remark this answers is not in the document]
gtaschuk: I think we should keep it as an argument, yeah.  Seeing as we think we might reenable the ability to claim on behalf of someone else (our contracts generally avoid using msg.sender as an implicit input)

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -0,0 +1,299 @@
...
+                emit RewardPaid(msg.sender, address(rewardsToken), reward);
+            }
+
+            ops[i] = IVault.UserBalanceOp({
+                asset: IAsset(address(rewardsToken)),
+                amount: reward,
+                sender: address(this),
+                recipient: msg.sender,
+                kind: IVault.UserBalanceOpKind.DEPOSIT_INTERNAL
+            });
```
nventuro, line 218:
```suggestion
                emit RewardPaid(msg.sender, address(rewardsToken), reward);
                ops[i] = IVault.UserBalanceOp({
                    asset: IAsset(address(rewardsToken)),
                    amount: reward,
                    sender: address(this),
                    recipient: msg.sender,
                    kind: IVault.UserBalanceOpKind.DEPOSIT_INTERNAL
                });
            }
```

PR #581, opened 2021-05-25
Add deployment config for Polygon

TomAFrench:
Note: The Polygon RPC connection needs to be activated in the Infura dashboard.

### DIFF
--- a/pkg/deployments/migrations/002_vault.ts
+++ b/pkg/deployments/migrations/002_vault.ts
@@ -13,6 +13,8 @@ export default async function (hre: HardhatRuntimeEnvironment): Promise<void> {
 
   if (chainId == '1') {
     WETH = '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2';
+  } else if (chainId == '137') {
+    WETH = '0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270';
   } else {
     WETH = await (await deployments.get('TestWETH')).address;
   }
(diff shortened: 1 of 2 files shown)


mikemcdonald:
@TomAFrench I got this added this to infura. Existing apps/keys should work with the `polygon-mainnet` and `polygon-mumbai` networks

pkg/deployments/migrations/002_vault.ts
```diff
@@ -13,6 +13,8 @@ export default async function (hre: HardhatRuntimeEnvironment): Promise<void> {
 
   if (chainId == '1') {
     WETH = '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2';
+  } else if (chainId == '137') {
+    WETH = '0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270';
```
nventuro, line 17: For those curious, this is the address of the [`WMATIC`](https://polygon-explorer-mainnet.chainstacklabs.com/address/0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270) token. The code is not verified, but based on [this contract](https://github.com/maticnetwork/contracts/blob/f98188ec16ac1a804ad2f6127a2c75c75e9728c1/contracts/common/tokens/MaticWETH.sol), it seems to be exactly the same as WETH, except on the Polygon network.

nventuro (approved):

pkg/deployments/migrations/002_vault.ts
TomAFrench: The code has been verified here which differs from what's shown on Github (but functionality the same as standard WETH): https://explorer-mainnet.maticvigil.com/address/0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270/contracts

PR #585, opened 2021-05-27
Setup deployment package

nventuro:
This prepares the `deployments` package to be published in npm. I changed a few small things:

- I flattened `IVault.sol` and made a few modifications by hand (e.g. update the solidity pragmas) so people can import a single file and use the interface, without depending on the other packages. This required performing a small change where I had to extract the interface of `ProtocolFeesCollector` to a new file, as IVault was importing the contract (not interface) directly.
- I extracted the `solcInputs` directory to the top level, since a) consumers of this don't need those files, b) they are the same for all networks, and c) even if we were to deploy different contracts to different networks, the inputs are still identified by their unique id.

I also added the artifacts for `WeightedPool2TokensFactory`, which had not been uploaded.


nventuro:
I've been convinced that we should not release the Vault's interface this way, but instead release the source code of all other Solidity packages as well. I've therefore removed the `interfaces` dir from the deployment package.

I'd merge this PR as is - I'll create a new one that sets up the dependencies among packages as actual dependencies so that users have them installed correctly, along with proper versioning.

nventuro:
(reviewed by @EndymionJkb)

PR #586, opened 2021-05-28
Add asset managers to WeightedPool

EndymionJkb:


pkg/pool-weighted/test/WeightedPoolFactory.test.ts
```diff
@@ -26,17 +29,30 @@ describe('WeightedPoolFactory', function () {
...
   sharedBeforeEach('deploy factory & tokens', async () => {
-    const vault = await Vault.create();
+    vault = await Vault.create();
 
     factory = await deploy('WeightedPoolFactory', { args: [vault.address] });
     createTime = await currentTimestamp();
 
     tokens = await TokenList.create(['MKR', 'DAI', 'SNX', 'BAT'], { sorted: true });
+    // Deploy Asset manager
+    assetManager = await deploy('v2-asset-manager-utils/TestAssetManager', {
```
nventuro, line 39: This is fine, but in the future you might want to consider just passing an EOA here to avoid coupling to this contract, and head scratching in the future trying to understand why this test fails if that contract later changes.

pkg/pool-weighted/test/WeightedPool.test.ts
```diff
@@ -602,9 +609,26 @@ describe('WeightedPool', function () {
...
     const weights: BigNumberish[] = WEIGHTS.slice(0, numberOfTokens);
     const initialBalances = INITIAL_BALANCES.slice(0, numberOfTokens);
+    let vault: Contract;
+    let assetManagerContract: Contract;
+    let assetManagers: Array<string>;
 
     async function deployPool(params: RawWeightedPoolDeployment = {}): Promise<void> {
-      params = Object.assign({}, { tokens, weights, swapFeePercentage: POOL_SWAP_FEE_PERCENTAGE, twoTokens }, params);
+      // Deploy Balancer Vault
+      const authorizer = await deploy('v2-vault/Authorizer', { args: [admin.address] });
```
nventuro, line 618: I don't understand this entire change - shouldn't we just createa and pass the `assetManagers` array to `WeightedPool.create()`?

nventuro (commented):
Succint and to the point :ok_hand:

pkg/pool-weighted/test/WeightedPool.test.ts
EndymionJkb: Do you mean just fill it with ZERO_ADDRESS instead of a real contract? (Was just making the factory and pool tests the same.)

nventuro: Ah I now understand what you did. It is wrong however: `WeightedPool.create()` will deploy a new Vault as well and connect the Pool to that Vault. I'd skip the manual deployment of Authorizer, Vault, AssetManagers, and just pass an EOA to `WeightedPool.create()` as the AM.

pkg/pool-weighted/test/WeightedPoolFactory.test.ts
```diff
@@ -26,17 +31,29 @@ describe('WeightedPoolFactory', function () {
...
   sharedBeforeEach('deploy factory & tokens', async () => {
-    const vault = await Vault.create();
+    vault = await Vault.create();
 
     factory = await deploy('WeightedPoolFactory', { args: [vault.address] });
     createTime = await currentTimestamp();
 
     tokens = await TokenList.create(['MKR', 'DAI', 'SNX', 'BAT'], { sorted: true });
+
+    [, assetManager] = await ethers.getSigners();
```
nventuro, line 41: Can you move this into 
```
  before('setup signers', async () => {
    [, assetManager] = await ethers.getSigners();
  });
```

Just so that it looks like every other test file. Thanks!

nventuro (approved):

EndymionJkb:
Will do!

nventuro (approved):

PR #589, opened 2021-05-28
Weightless pool

EndymionJkb:
Factor out weight storage from WeightedPool


pkg/pool-weighted/contracts/BaseWeightedPool.sol
```diff
@@ -0,0 +1,487 @@
...
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/InputHelpers.sol";
+
+import "@balancer-labs/v2-pool-utils/contracts/BaseMinimalSwapInfoPool.sol";
+
+import "./WeightedMath.sol";
+import "./WeightedPoolUserDataHelpers.sol";
+
+// Base class containing WeightedPool logic, but deferring handling of weights (e.g., fixed or mutable)
+// to subclasses
+
```
nventuro, line 28:
```suggestion
/**
 * @dev Base class for WeightedPools containing swap, join and exit logic, but managing the weights themselves.
 * Derived contracts can choose to make weights immutable, mutable, or even dynamic base on local or external
 * logic.
 */
```

pkg/pool-weighted/contracts/BaseWeightedPool.sol
```diff
@@ -0,0 +1,487 @@
...
+        )
+    {
+        // solhint-disable-previous-line no-empty-blocks
+    }
+
+    function _normalizedWeight(IERC20 token) internal view virtual returns (uint256);
+
+    function _normalizedWeights() internal view virtual returns (uint256[] memory);
+
+    function _getMaxWeightTokenIndex() internal view virtual returns (uint256);
```
nventuro, line 68:
```suggestion
    // Virtual functions
    
    /**
     * @dev Returns the normalized weight of `token`. Weights are fixed point numbers that sum to FixedPoint.ONE.
     */ 
    function _normalizedWeight(IERC20 token) internal view virtual returns (uint256);

    /**
     * @dev Returns all normalized weights, in the same order as the Pool's tokens.
     */ 
    function _normalizedWeights() internal view virtual returns (uint256[] memory);

    /**
     * @dev Returns the index of the token with the largest weight.
     */
    function _getMaxWeightTokenIndex() internal view virtual returns (uint256);
    
    // Getters
```

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -15,21 +15,14 @@
...
-
-import "@balancer-labs/v2-pool-utils/contracts/BaseMinimalSwapInfoPool.sol";
-
-import "./WeightedMath.sol";
-import "./WeightedPoolUserDataHelpers.sol";
+import "./BaseWeightedPool.sol";
 
 // This contract relies on tons of immutable state variables to perform efficient lookup, without resorting to storage
 // reads. Because immutable arrays are not supported, we instead declare a fixed set of state variables plus total
 // count, resulting in a large number of state variables.
```
nventuro, line 23:
```suggestion
/**
 * @dev Basic Weighted Pool with immutable weights.
 */
```

The state variables comment was copy-pasted by mistake from `BasePool`: it is meant to be an explanation of why a certain solhint check is disabled there.

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -138,419 +126,7 @@ contract WeightedPool is BaseMinimalSwapInfoPool, WeightedMath {
...
-    }
-
-    /**
-     * @dev This function returns the appreciation of one BPT relative to the
-     * underlying tokens. This starts at 1 when the pool is created and grows over time
-     */
-    function getRate() public view returns (uint256) {
-        // The initial BPT supply is equal to the invariant times the number of tokens.
-        return Math.mul(getInvariant(), _getTotalTokens()).divDown(totalSupply());
+    function _getMaxWeightTokenIndex() internal view virtual override returns (uint256) {
```
nventuro, line 129: Nit: for consistency with the other two functions I'd consider renaming this to `_maxWeightTokenIndex()`, or adding the 'get' preffix to the weight functions.

nventuro (commented):
Looks great, thanks a lot @EndymionJkb!

pkg/pool-weighted/contracts/WeightedPool.sol
EndymionJkb: I made them all get; otherwise maxWeightTokenIndex conflicts with the state variable. (I tried the state variable override trick, but it only works with public variables.)

nventuro (approved):

PR #592, opened 2021-05-30
LBP pool prep

EndymionJkb:
As a first step toward the LBP, this is a regular weighted pool (same tests) - just with packed, mutable weights.

I've put it in a new pool-smart workspace.

Nearly all LBPs have two tokens, but there have been a couple with 3 and one with 4. I don't think more than 4 makes sense (and want to save bytecode), so I've limited it to 4 tokens.

When the weight handling approach is validated, I'll add the actual LBP logic


pkg/pool-smart/contracts/LiquidityBootstrappingPool.sol
```diff
@@ -0,0 +1,130 @@
...
+        uint256 pauseWindowDuration,
+        uint256 bufferPeriodDuration,
+        address owner
+    )
+        BaseWeightedPool(
+            vault,
+            name,
+            symbol,
+            tokens,
+            new address[](tokens.length), // LBPs can't have asset managers
```
nventuro, line 54:
```suggestion
            new address[](tokens.length), // Pass the zero address as LBPs can't have asset managers
```

pkg/pool-smart/contracts/LiquidityBootstrappingPool.sol
```diff
@@ -0,0 +1,130 @@
...
+            bufferPeriodDuration,
+            owner
+        )
+    {
+        uint256 numTokens = tokens.length;
+        _require(numTokens <= _MAX_LBP_TOKENS, Errors.MAX_TOKENS);
+
+        InputHelpers.ensureInputLengthMatch(numTokens, normalizedWeights.length);
+
+        // Ensure  each normalized weight is above them minimum and find the token index of the maximum weight
```
nventuro, line 66:
```suggestion
        // Ensure each normalized weight is above the minimum, and find the token index of the maximum weight
```

pkg/pool-smart/test/LiquidityBootstrappingPool.test.ts
```diff
@@ -0,0 +1,753 @@
...
+    });
+  });
+
+  context('for a 2 token pool', () => {
+    itBehavesAsWeightedPool(2);
+  });
+
+  context('for a 3 token pool', () => {
+    itBehavesAsWeightedPool(3);
+  });
```
nventuro, line 46: Let's also add a test for a 4 token pool, since that fully test the weight packing logic.

pkg/pool-smart/test/LiquidityBootstrappingPool.test.ts
```diff
@@ -0,0 +1,753 @@
...
+import LiquidityBootstrappingPool from '@balancer-labs/v2-helpers/src/models/pools/smart/LiquidityBootstrappingPool';
+import { RawLiquidityBootstrappingPoolDeployment } from '@balancer-labs/v2-helpers/src/models/pools/smart/types';
+
+describe('LiquidityBootstrappingPool', function () {
+  let allTokens: TokenList;
+  let trader: SignerWithAddress, recipient: SignerWithAddress;
+  let other: SignerWithAddress, lp: SignerWithAddress, owner: SignerWithAddress;
+
+  const POOL_SWAP_FEE_PERCENTAGE = fp(0.01);
+  const WEIGHTS = [fp(30), fp(70), fp(5), fp(5)];
```
nventuro, line 19:
```suggestion
  const WEIGHTS = [fp(30), fp(65), fp(10), fp(5)];
```

Let's make all weights different so that we can be sure they are being stored and read correctly.

pkg/pool-smart/test/LiquidityBootstrappingPool.test.ts
```diff
@@ -0,0 +1,753 @@
...
+    it('reverts if there are too many tokens', async () => {
+      // The maximum number of tokens is 4
+      const tokens = await TokenList.create(5, { sorted: true });
+      const weights = new Array(5).fill(fp(1));
+
+      await expect(LiquidityBootstrappingPool.create({ tokens, weights })).to.be.revertedWith('MAX_TOKENS');
+    });
+  });
+
+  function itBehavesAsWeightedPool(numberOfTokens: number) {
```
nventuro, line 58: These tests are not wrong per se, but what we're really testing here is (for the most part) `BaseWeightedPool`. It would be good to refactor the `WeightedPool` tests into tests for `BaseWeighted` (most of them), and the actual `WeightedPool` (checking that the weights are correctly stored in the different `immutable` slots). Then, we would only actually tests the contents of `LBPool` here (for now, testing that the weights are correctly stored).

nventuro (commented):
Fantastic! Loved the usage of `WordCoded` :grin: :ok_hand: :ok_hand:

pkg/pool-smart/test/LiquidityBootstrappingPool.test.ts
EndymionJkb: Ok; I had that and it seemed to be failing for some other reason - but will add it. Of course it should work with 4 tokens

pkg/pool-smart/test/LiquidityBootstrappingPool.test.ts
EndymionJkb: Agreed - this isn't the LBP pool yet. I can make LBP-specific tests when I have the logic (and correct structure)

EndymionJkb:
Still needs tests, but here is the base logic

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -0,0 +1,501 @@
...
+
+    // Type declarations
+
+    // Use timestamps instead of blocks (used in V1 ConfigurableRightsPool)
+    // End weights are packed as 4 uint64's, like the _normalizedWeights
+    struct GradualUpdateParams {
+        uint256 startTime;
+        uint256 endTime;
+        bytes32 endWeights;
+    }
```
nventuro, line 41: I think we could pack this into a single storage slot: timestamps can be very reasonably be represented in 32 bits (we could even change `endTime` for `duration` if we wanted to, and use even fewer bits, as no LBP will run for over say 2 years).

This would save two storage accesses on each swap, or ~5k gas.

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -0,0 +1,501 @@
...
+
+    // State variables
+
+    // Minimum time over which to compute a gradual weight change (i.e., seconds between timestamps)
+    uint256 public immutable minWeightChangeDuration;
+
+    // Setting this to false pauses trading
+    bool public publicSwapEnabled;
```
nventuro, line 49:
```suggestion
    bool public swapEnabled;
```

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -0,0 +1,501 @@
...
+    struct GradualUpdateParams {
+        uint256 startTime;
+        uint256 endTime;
+        bytes32 endWeights;
+    }
+
+    // State variables
+
+    // Minimum time over which to compute a gradual weight change (i.e., seconds between timestamps)
+    uint256 public immutable minWeightChangeDuration;
```
nventuro, line 46: What's the intended goal of this value?

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -0,0 +1,501 @@
...
+
+    // External functions
+
+    /**
+     * @dev Can pause/unpause trading
+     */
+    function setPublicSwap(bool publicSwap) external onlyOwner whenNotPaused nonReentrant {
+        publicSwapEnabled = publicSwap;
+
+        emit PublicSwapSet(publicSwap);
```
nventuro, line 145: Let's also emit this event in construction.

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -0,0 +1,501 @@
...
+        // Enforce a minimum time over which to make the changes
+        // The also prevents endBlock <= startBlock
+        _require(
+            endTime.sub(_gradualUpdateParams.startTime) >= minWeightChangeDuration,
+            Errors.WEIGHT_CHANGE_TIME_BELOW_MIN
+        );
+
+        // Must specify normalized weights for all tokens
+        uint256 numTokens = _getTotalTokens();
+        InputHelpers.ensureInputLengthMatch(numTokens, endWeights.length);
```
nventuro, line 182: I'd move this check to the top of this function

nventuro (commented):
Only left a few comments since I think we might be able to find a simpler way to handle the ramping, discussing that via DM.

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
EndymionJkb: Ok - just copied it from V1 :)

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
EndymionJkb: As in V1 - if the controller could change the weights (and therefore prices) instantly, it would require complete trust from users. Longer values = closer to trustless. The users can be sure the weights can change in say, less than several hours.

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
EndymionJkb: Ok

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
EndymionJkb: I can; was trying to save gas by doing the checks most likely to fail first, and the "error" ones last

EndymionJkb: But yes; I can move it up before some of the heavier weight operations

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
alexvansande: >  no LBP will run for over say 2 years

I beg to differ! In fact I came here precisely because the Gitcoin Akita LBP, is planned to last precisely 2 years!
https://snapshot.org/#/gitcoindao.eth/proposal/QmR1mSZaQP8HEMU8nSMeBM5pVvkGoGG3TVu2H5aDuXyB8m

It's more akin to a slow sale or a stream of this very illiquid token they own, so we should not make assumptions that all LBPs are quick sales

PR #593, opened 2021-05-31
AAVE - Single pool asset manager

gtaschuk:
- Introduces a single pool asset manager that only handles the funds of 1 pool's token
- Uses this for an initial AAVE asset manager implementation

Credit to @TomAFrench who got this started on the asset-manager repo


pkg/asset-manager-utils/contracts/aave/DataTypes.sol
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: agpl-3.0
```
TomAFrench, line 1: We don't seem to use this file anywhere. Shall we remove it for now?

pkg/asset-manager-utils/contracts/test/SinglePoolTestAssetManager.sol
```diff
@@ -0,0 +1,74 @@
...
+
+pragma experimental ABIEncoderV2;
+
+import "../SinglePoolAssetManager.sol";
+
+pragma solidity ^0.7.0;
+
+// solhint-disable no-empty-blocks
+// solhint-disable var-name-mixedcase
+// solhint-disable private-vars-leading-underscore
```
TomAFrench, line 23: We seem to be out of sync with the rest of the monorepo codebase on this. A lot of the asset manager code used underscores to denote function args whereas other packages use them for private variables.

gtaschuk: Yeah.  Personally I like underscores for function args, I think it makes it clear what is coming from calldata and what is allocated.

This `solhint-disable` line is for something else though - it's so we can have const `ONE` and not `_ONE`

pkg/asset-manager-utils/contracts/aave/DataTypes.sol
gtaschuk: It's used by `ILendingPool.sol` (for various reserve config arguments)

TomAFrench: Ah gotcha, sorry that file was folded.

pkg/asset-manager-utils/contracts/test/SinglePoolTestAssetManager.sol
TomAFrench: Ah fair enough, this is probably something for another PR so we can hit both single and multi-pool AMs but we should probably move these to be more selective and just target those lines.

pkg/asset-manager-utils/contracts/aave/SinglePoolAaveATokenAssetManager.sol
```diff
@@ -0,0 +1,119 @@
...
+        IERC20(_token).approve(_lendingPool, type(uint256).max);
+    }
+
+    /**
+     * @dev Should be called in same transaction as deployment through a factory contract
+     * @param pId - the id of the pool
+     * @param rewardsContract - the address of the rewards contract (to distribut stkAAVE)
+     */
+    function initialise(bytes32 pId, address rewardsContract) public {
+        require(poolId == bytes32(0), "Already initialised");
```
nventuro, line 63: It is very unfortunate that this has to be this way. I don't think we'll even be able to have a single factory contract, since each contract is so large. It'll likely end up being something like the following:

```
contract MegaFactory {
  function createPool(...) {
    assetManager = assetManagerFactory.create(...)
    pool = poolFactory.create(assetMananger, ...);
    assetManager.initialize(pool)
  }
}
```

Even if we end up following that pattern though, I think it may be good to have internal `_initialize` functions for the base contracts (e.g. setting `poolId`) that we call from the derived ones, and remove these arguments (such as `poolId`) from their constructors.

pkg/asset-manager-utils/contracts/aave/SinglePoolAaveATokenAssetManager.sol
```diff
@@ -0,0 +1,119 @@
...
+        poolId = pId;
```
nventuro, line 64: We should also check that `pId != 0` to avoid multiple initializations.

pkg/asset-manager-utils/contracts/aave/SinglePoolAaveATokenAssetManager.sol
```diff
@@ -0,0 +1,119 @@
...
+        distributor = IMultiRewards(rewardsContract);
+        IERC20(stkAave).approve(rewardsContract, type(uint256).max);
+    }
+
+    /**
+     * @dev To be used in the case of a staking contract migration
+     */
+    function updateStakingContract(address rewardsContract) external {
+        require(msg.sender == address(distributor), "Must be called by current staking contract");
```
nventuro, line 73: It seems like we're using three different names here ('staking contract', 'rewards contract' and 'distributor') to refer to the same entity - we should pick a single one. 

Out of these, I'd go with either 'rewards contract' or 'distributor' (maybe 'rewards distributor' is the most clear one), as 'staking' refers to an implementation detail and not the goal of that component.

pkg/asset-manager-utils/contracts/aave/SinglePoolAaveATokenAssetManager.sol
```diff
@@ -0,0 +1,119 @@
...
+     * @param amount - the amount of tokens being deposited
+     * @param aum - the current assets under management of this asset manager
+     * @return the number of shares to mint for the pool
+     */
+    function _invest(uint256 amount, uint256 aum) internal override returns (uint256) {
+        lendingPool.deposit(address(token), amount, address(this), REFERRAL_CODE);
+        if (aum == 0) {
+            return amount;
+        }
+        return amount / aum;
```
nventuro, line 91: I don't understand this division, can you expand on it? The documentation here talks about Pool shares, but SinglePoolAM has no shares, and the docs in `SinglePoolAM` say this should return 'the number of deposited tokens', which seems to be just `amount`.

pkg/asset-manager-utils/contracts/aave/SinglePoolAaveATokenAssetManager.sol
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * @dev Withdraws capital out of Aave
+     * @param amount - the amount to withdraw
+     * @return the number of tokens to return to the vault
+     */
+    function _divest(uint256 amount, uint256) internal override returns (uint256) {
+        return lendingPool.withdraw(address(token), amount, address(this));
```
nventuro, line 100: Aave's docs just say that this returns 'The final amount withdrawn'. Are there scenarios under which we'd withdraw less than `amount`?

pkg/asset-manager-utils/contracts/aave/IAaveIncentivesController.sol
```diff
@@ -0,0 +1,33 @@
...
+interface IAaveIncentivesController {
+    function handleAction(
+        address asset,
+        uint256 userBalance,
+        uint256 totalSupply
+    ) external;
+
+    function getRewardsBalance(address[] calldata assets, address user) external view returns (uint256);
+
+    function claimRewards(
```
nventuro, line 27: Weird, these functions are not in this interface in version 1.0.1 of the protocol-v2 package. Where did you retrieve them from?

pkg/asset-manager-utils/test/SinglePoolAaveATokenAssetManager.test.ts
```diff
@@ -0,0 +1,330 @@
...
+import { encodeJoinWeightedPool } from '@balancer-labs/v2-helpers/src/models/pools/weighted/encoding';
+
+const OVER_INVESTMENT_REVERT_REASON = 'investment amount exceeds target';
+const UNDER_INVESTMENT_REVERT_REASON = 'withdrawal leaves insufficient balance invested';
+
+const tokenInitialBalance = bn(200e18);
+const amount = bn(100e18);
+
+const setup = async () => {
+  const [, admin, lp, other] = await ethers.getSigners();
```
nventuro, line 24: I'd have this function inlined inside the first `sharedBeforeEach` (and maybe split in smaller parts). That would also prevent mistakes caused by calling `getSigners` multiple times.

pkg/asset-manager-utils/test/SinglePoolAaveATokenAssetManager.test.ts
```diff
@@ -0,0 +1,330 @@
...
+
+  // Deploy Asset manager
+  const assetManager = await deploy('SinglePoolAaveATokenAssetManager', {
+    args: [
+      vault.address,
+      tokens.DAI.address,
+      lendingPool.address,
+      daiAToken.address,
+      aaveRewardsController.address,
+      stkAave.address,
```
nventuro, line 48: In the real system these two are not the same contract, right? Meaning stkAave is standalone, separate from the rewards controller.

pkg/asset-manager-utils/test/SinglePoolAaveATokenAssetManager.test.ts
```diff
@@ -0,0 +1,330 @@
...
+    assetManagers,
+    fp(0.0001),
+    0,
+    0,
+    admin.address,
+  ];
+
+  const pool = await deploy('v2-pool-weighted/WeightedPool', {
+    args,
+  });
```
nventuro, line 69: Using the WeightedPool model instead of calling the contracts directly will not only greatly simplify this, but also joins, etc.

pkg/asset-manager-utils/test/SinglePoolAaveATokenAssetManager.test.ts
```diff
@@ -0,0 +1,330 @@
...
+    tokens = contracts.tokens;
+    vault = contracts.vault;
+  });
+
+  describe('deployment', () => {
+    it('different managers can be set for different tokens', async () => {
+      expect((await vault.getPoolTokenInfo(poolId, tokens.DAI.address)).assetManager).to.equal(assetManager.address);
+      expect((await vault.getPoolTokenInfo(poolId, tokens.MKR.address)).assetManager).to.equal(other.address);
+    });
+  });
```
nventuro, line 147: I'd remove this, since this is just a `BasePool` test
```suggestion
```

pkg/asset-manager-utils/test/SinglePoolAaveATokenAssetManager.test.ts
```diff
@@ -0,0 +1,330 @@
...
+
+  describe('setPoolConfig', () => {
+    it('allows a pool controller to set the desired target investable %', async () => {
```
nventuro, line 150:
```suggestion
    it.skip('allows a pool controller to set the desired target investable %', async () => {
```

`.skip` makes it really obvious that some tests are not really running. Similarly, I'd create another (empty for now) `it` that does 'reverts if caller is not authorized'.

pkg/asset-manager-utils/test/SinglePoolAaveATokenAssetManager.test.ts
```diff
@@ -0,0 +1,330 @@
...
+    it('distributes the reward according to the fraction of staked LP tokens', async () => {
+      await assetManager.claimRewards();
+      await advanceTime(10);
+
+      const expectedReward = fp(0.75);
+      const actualReward = await stakingContract.earned(lp.address, stkAave.address);
+      expect(expectedReward.sub(actualReward).abs()).to.be.lte(100);
+    });
+  });
+});
```
nventuro, line 330: We seem to be missing tests for `updateStakingContract()` and `readAUM()` here. I believe the goal is to rework the notion of AUM, cache and so on? If so, maybe it doesn't make sense to spend time working on these now.

nventuro (commented):
Great to see how the whole thing comes together :grin: Thanks for bringing this over in small pieces @gtaschuk!

Re. the fixes for #598, it seems some of the comments about the tests were not addressed? Some also apply here, e.g. avoiding doing `-bn`.

pkg/asset-manager-utils/test/SinglePoolAaveATokenAssetManager.test.ts
gtaschuk: correct

pkg/asset-manager-utils/test/SinglePoolAaveATokenAssetManager.test.ts
gtaschuk: removed `updateStakingContract` for now - I think it's likely that we would be doing an upgrade from a merkle drop to a staking contract, if anything (hoping we can just start use MultiMultiRewards from the beginning)

pkg/asset-manager-utils/test/SinglePoolAaveATokenAssetManager.test.ts
gtaschuk: well, this test does actually setPoolConfig.  I added an `xit` for the reverts if caller not authorized test.

pkg/asset-manager-utils/test/SinglePoolAaveATokenAssetManager.test.ts
gtaschuk: Hmm, I saw that but I actually don't find the helper very helpful in this case.   It feels more illustrative to write these specs around a contract instance than a helper model.

Generally I think I would prefer a helper that serves more as a factory to return a contract instance than the `WeightedPool` helper which returns a decorated contract instance.  For example, bubbling all the instance methods up to the `WeightedPool` model seems like an unnecessary abstraction/source of indirection imho

Up to you.  If you insist, I'll use it.

pkg/asset-manager-utils/contracts/aave/IAaveIncentivesController.sol
TomAFrench: I pulled these from the verified contracts on Etherscan for the rewards contract. Strangely looking back I can't re-find these so we should drop the final param of `claimRewards`.

We should drop the other two functions as they are unused.

pkg/asset-manager-utils/test/SinglePoolAaveATokenAssetManager.test.ts
nventuro: Huh you're right, just skipping the other one ought to be enough.

pkg/asset-manager-utils/contracts/SinglePoolAaveATokenAssetManager.sol
```diff
@@ -0,0 +1,105 @@
...
+        return aToken.balanceOf(address(this));
+    }
+
+    function claimRewards() public {
+        // Claim stkAave from incentives controller
+        address[] memory assets = new address[](1);
+        assets[0] = address(aToken);
+        aaveIncentives.claimRewards(assets, type(uint256).max, address(this));
+
+        // Forward to staking contract
```
nventuro, line 102:
```suggestion
        // Forward to distributor
```

pkg/asset-manager-utils/contracts/aave/IAaveIncentivesController.sol
nventuro: Let's double check with the Aave team that these are the ones we should be using then.

pkg/asset-manager-utils/test/SinglePoolAaveATokenAssetManager.test.ts
nventuro: Yeah I know what you mean, especially regarding them possibly obscuring things. I gave this a try locally to see what the result would be, and sadly they're a little too magical at the moment and e.g. deploy their own Vault. We might improve them in the future if the need arises.

pkg/asset-manager-utils/contracts/aave/SinglePoolAaveATokenAssetManager.sol
gtaschuk: I don't think there are.  Is the final amount withdrawn different from the amount specified?

nventuro (approved):

PR #594, opened 2021-06-01
Permit relayer

TomAFrench:
This PR adds a relayer which allows users to atomically call permit on a number of tokens before calling one (or multiple) functions on the vault.

To maintain signature based authorisation I've needed to fall back to low-level calls so that the extra calldata isn't stripped out. 
I wanted to avoid allowing people to call arbitrary functions on the Vault so they may only call wrapped copies of the main Vault functions which take the authorisation signature + value of ETH to send as extra arguments (as opposed to allowing them to supply arbitrary data).

TODO:
- update new `solidity-utils` contracts to use `_require`
- add some sanity check tests for multiple vault interactions.


pkg/standalone-utils/contracts/base/VaultPermit.sol
```diff
@@ -0,0 +1,56 @@
...
+import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
+import "../interfaces/IPermit.sol";
+import "../interfaces/IPermitDAI.sol";
+
+/**
+ * @notice Functionality to call permit on any EIP2612 token or DAI for interactions with the Vault.
+ * @dev These functions are made payable as they are expected to be delegate called into
+ *      within a multicall which may receive ether.
+ */
+contract VaultPermit {
```
nventuro, line 27: I'd consider merging this with `PermitRelayer`: by itself this doesn't add much and I found it harder to understand the whole contract with this added indirection.

pkg/standalone-utils/contracts/PermitRelayer.sol
```diff
@@ -0,0 +1,126 @@
...
+import "./base/Multicall.sol";
+import "./base/VaultPermit.sol";
+import "./interfaces/IPermit.sol";
+import "./interfaces/IPermitDAI.sol";
+
+/**
+ * @title PermitRelayer
+ * @notice Allows users to atomically call permit on multiple tokens and then interact with the Balancer Vault
+ * @dev This relayer is extremely powerful and has full control over users' assets should they authorise it.
+ * For this reason, users should only perform signature-based approvals to allow it to perform only a single action.
```
nventuro, line 28: I'd frame this as a vault multicall contract that lets the caller perform multiple Vault actions (out of a subset) in sequence, along with token approvals, allowing for signature-based single transaction interactions (including onboarding!).

pkg/standalone-utils/contracts/PermitRelayer.sol
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * WARNING: Do not call setRelayerApproval on the Vault for this relayer - doing so WILL result in loss of funds.
```
nventuro, line 30: We're always spending `msg.sender`'s funds, why would it be an issue for a user to grant approval to this contract?

pkg/standalone-utils/contracts/test/TestTokenPermitDAI.sol
```diff
@@ -0,0 +1,43 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+
+import "./TestTokenPermit.sol";
+import "../interfaces/IPermitDAI.sol";
+
+contract TestTokenPermitDAI is TestTokenPermit, IPermitDAI {
```
nventuro, line 20: I'd consider renaming this to TestDAI, mentioning this simulates DAI's permit

pkg/standalone-utils/test/VaultPermit.test.ts
```diff
@@ -0,0 +1,70 @@
...
+describe('VaultPermit', function () {
+  let tokens: TokenList, vaultPermit: Contract;
+  let admin: SignerWithAddress, vault: SignerWithAddress;
+
+  before('deploy base contracts', async () => {
+    [, admin, vault] = await ethers.getSigners();
+  });
+
+  sharedBeforeEach('set up asset manager', async () => {
+    const { contracts } = await setup();
```
nventuro, line 42: Why not inline `setup` here?

TomAFrench:
I've had to rename `MultiCall` to `MultiDelegateCall` to avoid a name collision between two artifacts in the deployments package. I'm open to better names though.

pkg/standalone-utils/test/VaultPermit.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    vaultPermit = contracts.vaultPermit;
+  });
+
+  describe('vaultPermit', () => {
+    it("approves the Vault to use the caller's tokens", async () => {
+      const mkr = tokens.MKR.instance;
+      const approvalAmount = 100;
+      const { v, r, s, deadline } = await signPermit(mkr, admin, vault, approvalAmount);
+
+      await vaultPermit.connect(admin).vaultPermit(mkr.address, approvalAmount, deadline, v, r, s);
```
nventuro, line 54: I'd use another account (e.g. `signer` or `holder`) for this, instead of `admin`

pkg/standalone-utils/test/VaultPermit.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  sharedBeforeEach('set up asset manager', async () => {
+    const { contracts } = await setup();
+
+    tokens = contracts.tokens;
+    vaultPermit = contracts.vaultPermit;
+  });
+
+  describe('vaultPermit', () => {
+    it("approves the Vault to use the caller's tokens", async () => {
```
nventuro, line 49: If we reused a base layer ERC20Permit as mentioned on some of the comments above, we'd be able to rely on the tests for e.g. invalid nonce (which are currently BalancerPoolToken tests)

pkg/standalone-utils/test/PermitRelayer.test.ts
```diff
@@ -0,0 +1,432 @@
...
+
+  const relayer = await deploy('PermitRelayer', { args: [vault.address] });
+
+  // Authorise relayer to perform actions.
+  const single = await actionId(vault, 'swap');
+  const batch = await actionId(vault, 'batchSwap');
+  const manageUserBalance = await actionId(vault, 'manageUserBalance');
+  const joinPool = await actionId(vault, 'joinPool');
+  const exitPool = await actionId(vault, 'exitPool');
+  await authorizer.connect(admin).grantRoles([single, batch, manageUserBalance, joinPool, exitPool], relayer.address);
```
nventuro, line 70: I'd try to split up this very large setup into smaller parts, as we'd likely want to have at least some tests for partial setup (e.g. what happens if the contract has not been authorized), using `context` and `sharedBeforeEach`

pkg/standalone-utils/test/PermitRelayer.test.ts
```diff
@@ -0,0 +1,432 @@
...
+
+  const dai = await deploy('TestTokenPermitDAI', { args: [admin.address, 'DAI', 'DAI', 18] });
+  const mkr = await deploy('TestTokenPermit', { args: [admin.address, 'MKR', 'MKR', 18] });
+
+  const tokenList = [await Token.fromContract(dai), await Token.fromContract(mkr), await Token.create('WETH')];
+  const tokens = new TokenList(tokenList.sort((a, b) => a.compare(b)));
+
+  // Deploy Balancer Vault
+  const authorizer = await deploy('v2-vault/Authorizer', { args: [admin.address] });
+  const vault = await deploy('v2-vault/Vault', { args: [authorizer.address, tokens.WETH.address, 0, 0] });
```
nventuro, line 39: If we remove `TestTokenPermit` (and just make that `TestToken`), we should be able to simplifly this greatly by using TokenList.create and Vault.create.

pkg/standalone-utils/test/PermitRelayer.test.ts
```diff
@@ -0,0 +1,432 @@
...
+
+      const daiPermit = relayer.interface.encodeFunctionData('vaultPermitDAI', [
+        dai.address,
+        daiSig.nonce,
+        MAX_UINT256,
+        true,
+        daiSig.v,
+        daiSig.r,
+        daiSig.s,
+      ]);
```
nventuro, line 120: I'd consider reducing repeated code and verboseness by e.g. extracting thise into the following:

```
context('with signed permit', () => {
  let daIPermit: string;
     daiPermit = await ...
  });
  
  it('can call permit on multiple tokens', async () => {
     const result = await multicall();
     expect(result).to.be.ok;
  })
})
```

It otherwise becomes harder to review each individual test, and how they are different from each other.

pkg/standalone-utils/test/PermitRelayer.test.ts
```diff
@@ -0,0 +1,432 @@
...
+
+        expect(await ethers.provider.getBalance(vault.address)).to.be.eq(0);
+        expect(await ethers.provider.getBalance(relayer.address)).to.be.eq(0);
+
+        expectEvent.inIndirectReceipt(receipt, vault.interface, 'PoolBalanceChanged', {
+          poolId,
+          liquidityProvider: lp.address,
+          tokens: [dai.address, weth.address, mkr.address],
+          deltas: [0, 100, 0],
+          protocolFeeAmounts: [0, 0, 0],
```
nventuro, line 386: What do you think about splitting this up a bit? I find it both easier to maintain and more readable to have multiple `it` clauses that each test a simpler thing (e.g. that the event was emitted, or that tokens went to the correct place), with some shared setup. See e.g. some of the `it`s of `itJoinsCorrectly` in the Vault's JoinPool.test.ts.

nventuro (commented):
Great work, thanks a lot @TomAFrench! Most of my comments go in the general direction of simplifying the tests and contracts a bit, and trying to avoid introducing new contracts where we can reuse existing ones.

Also, can you run `yarn version minor --deferred` in the `standalone-utils` package? That will create a file that keeps track of the need to bump the minor version of this package (accounting for this new feature) once we make a new release. You can read more about that [here](https://yarnpkg.com/features/release-workflow#deferred-versioning).

pkg/standalone-utils/contracts/PermitRelayer.sol
TomAFrench: It's not. I wrote this line before I added more checks to restrict `funds.sender`.

nventuro: In that case, wouldn't it make sense to let users use such a relayer _without_ having to provide signatures? This otherwise has you sign n messages, plus the final transaction.

nventuro: (as an unrelated comment, we shouldn't ever have governance approved contracts that are not safe for a user to approve, that is, that must only be used via signatures. This would be difficult to communicate, and we shouldn't have hard to understand _and_ dangerous function calls available.

If such a use case existed (like we had here before that change), then what the relayer should do is `require(!vault.hasApprovedRelayer(user, address(this)))` to be absolutely sure that any successful transactions happen via signatures)

TomAFrench: Sorry, I'm agreeing with you. This line is a holdover from when things were less secure so that would have been an issue.

TomAFrench: We should add another function to allow setting an approval atomically with the first signature based interaction.

nventuro: Ah I must not have expressed myself correctly, I did get that you agreed with me :sweat_smile: My two comments above mean:
 1) would it make sense to have a multicall relayer that doesn't use signatures, for simpler UX?
 2) we should never make a relayer that is unsafe to approve by a user

pkg/standalone-utils/test/PermitRelayer.test.ts
TomAFrench: Ah, I wasn't aware of Vault.create. I'll change to use that.

`TokenList.create` is a little trickier as we need to handle `TestDAI` as well and I don't really want to add both styles of permit to `TestToken`

TomAFrench: These also don't really play nicely together as I'll need to extract WETH from the vault and create a `Token` manually for it.

nventuro: I think you're right. Let's leave as is here then, and get back to it in the future if we encounter this again (which I doubt, as I don't think we'll need to test DAI's custom permit often).

TomAFrench:
One question on this is whether now that we support setting a permanent approval within a multicall, should we even include the `authorisation` parameter on the other functions? We could do away with the lowlevel calls in this case and just call functions on the vault normally.

pkg/standalone-utils/test/PermitRelayer.test.ts
```diff
@@ -0,0 +1,432 @@
...
+
+import { bn } from '@balancer-labs/v2-helpers/src/numbers';
+import { MAX_UINT256, ZERO_ADDRESS } from '@balancer-labs/v2-helpers/src/constants';
+import Token from '@balancer-labs/v2-helpers/src/models/tokens/Token';
+import Vault from '@balancer-labs/v2-helpers/src/models/vault/Vault';
+
+const tokenInitialBalance = bn(200e18);
+
+const setup = async () => {
+  const [minter, lp] = await ethers.getSigners();
```
nventuro, line 31:
```suggestion
  const [, minter, lp] = await ethers.getSigners();
```

pkg/standalone-utils/test/PermitRelayer.test.ts
```diff
@@ -0,0 +1,432 @@
...
+  let poolId: string;
+
+  let daiPermit: string, mkrPermit: string;
+  let dai: Token, weth: Token, mkr: Token;
+
+  before('deploy base contracts', async () => {
+    [, signer] = await ethers.getSigners();
+  });
+
+  sharedBeforeEach('set up asset manager', async () => {
```
nventuro, line 94:
```suggestion
  sharedBeforeEach('set up relayer', async () => {
```

pkg/standalone-utils/test/PermitRelayer.test.ts
```diff
@@ -0,0 +1,432 @@
...
+
+  context('when relayer is authorised by governance', () => {
+    sharedBeforeEach('authorise relayer', async () => {
+      const single = await actionId(vault.instance, 'swap');
+      const batch = await actionId(vault.instance, 'batchSwap');
+      const manageUserBalance = await actionId(vault.instance, 'manageUserBalance');
+      const joinPool = await actionId(vault.instance, 'joinPool');
+      const exitPool = await actionId(vault.instance, 'exitPool');
+      const setApproval = await actionId(vault.instance, 'setRelayerApproval');
+      return await vault.authorizer?.grantRoles(
```
nventuro, line 134:
```suggestion

      await vault.authorizer?.grantRoles(
```

nventuro:
>One question on this is whether now that we support setting a permanent approval within a multicall, should we even include the authorisation parameter on the other functions? We could do away with the lowlevel calls in this case and just call functions on the vault normally.

Yes, this seems like a better approach overall if we intend for this relayer to be the singleton 'multicall' building block.

pkg/standalone-utils/contracts/PermitRelayer.sol
```diff
@@ -0,0 +1,149 @@
...
+        vault.manageUserBalance{ value: value }(ops);
+    }
+
+    function joinPool(
+        bytes32 poolId,
+        address recipient,
+        IVault.JoinPoolRequest calldata request,
+        uint256 value
+    ) external payable {
+        vault.joinPool{ value: value }(poolId, msg.sender, recipient, request);
```
nventuro, line 103: I'd receive `sender` and check that it matches `msg.sender`, to be consistent with the other functions.

pkg/standalone-utils/contracts/PermitRelayer.sol
```diff
@@ -0,0 +1,149 @@
...
+
+        return result;
+    }
+
+    function setRelayerApproval(
+        address relayer,
+        bool approved,
+        bytes calldata authorisation
+    ) external payable {
+        bytes memory data = abi.encodePacked(
```
nventuro, line 59:
```suggestion
        require(relayer == address(this));
        bytes memory data = abi.encodePacked(
```

Not sure about this, what do you think?

pkg/standalone-utils/contracts/base/MultiDelegatecall.sol
```diff
@@ -0,0 +1,47 @@
...
+            (bool success, bytes memory result) = address(this).delegatecall(data[i]);
+
+            if (!success) {
+                // Next 6 lines from https://ethereum.stackexchange.com/a/83577
+                if (result.length < 68) revert("MULTICALL_FAILED");
+                // solhint-disable-next-line no-inline-assembly
+                assembly {
+                    result := add(result, 0x04)
+                }
+                revert(abi.decode(result, (string)));
```
nventuro, line 36:
```suggestion
                assembly {
                    returndatacopy(0, 0, returndatasize())
                    revert(0, returndatasize())
                }
```

We should have tests for reverts that happen both in the multicall (the sender `require`s) and in the Vault to test the the revert reason bubble-up mechanism works correctly.

nventuro (commented):
I really really like how this is turning up, great work!

pkg/standalone-utils/contracts/PermitRelayer.sol
TomAFrench: Quite probably worth it. It prevents a malicious frontend opting the user in to a relayer without their knowledge by including it in a multicall and it's unlikely that a user will want to approve a batch of relayers without directly interacting with them at the same time.

We should probably have a `approveRelayerAndCall` function which activates the relayer and then delegatecalls into another of its functions. This could then be standard on all relayers.

TomAFrench: Can then decide whether we want to use `approveRelayerAndCall` on this relayer as it's an unnecessary abstraction on top of a multicall.

TomAFrench: Addressed in #671

pkg/standalone-utils/contracts/base/MultiDelegatecall.sol
TomAFrench: Addressed in #671

nventuro:
Marking as 'on hold' until design on #671 finishes and it is merged.

PR #595, opened 2021-06-01
Asset manager rebalancing updates/fixes

TomAFrench:
This PR updates `AssetManager.sol` to use the new rebalancing model with an upper and lower critical percentage with a cooldown period between paying out fees. It also includes a number of fixes involving the correct payout of fees.

Changelog:
- remove `totalAUM` and rely on caching value in memory or rereading
- Enforce that entirety of rebalance fee is used as input to swap in `rebalanceAndSwap`
- Correctly update the pool's balance after extracting tokens to pay fees.
- Modify rebalance amount to correctly account for the effect of fees on the pool's balance.
- Switch calculating of fees based on a "two critical levels" model
- Rebalancing fees are only paid out per pool once per day


pkg/asset-manager-utils/contracts/AssetManager.sol
```diff
@@ -65,8 +67,16 @@ abstract contract AssetManager is IAssetManager {
...
      */
     function balanceOf(bytes32 poolId) public view returns (uint256) {
+        return _poolManaged(poolId, readAUM());
+    }
+
+    /**
+     * @param poolId - The id of the pool of interest
+     * @return The amount of the underlying tokens which are owned by the specified pool
+     */
+    function _poolManaged(bytes32 poolId, uint256 aum) public view returns (uint256) {
```
gtaschuk, line 77: This puts a leading underscore in the interface, which we probably don't want

pkg/asset-manager-utils/test/AssetManager/AssetManager.test.ts
```diff
@@ -0,0 +1,259 @@
+import { ethers } from 'hardhat';
```
gtaschuk, line 1: What's the rationale for moving this file?  Seems the test folder should have the same, or flatter, structure than the contracts folder.

tbh just mentioning cause git isn't recognizing as a file move so it's hard to tell what's changed

pkg/asset-manager-utils/test/helpers/setup.ts
```diff
@@ -0,0 +1,78 @@
...
+export const amount = bn(100e18);
+
+export const setup = async () => {
+  const [, admin, lp, other] = await ethers.getSigners();
+
+  const tokens = await TokenList.create(['DAI', 'MKR'], { sorted: true });
+
+  // Deploy Balancer Vault
+  const authorizer = await deploy('v2-vault/Authorizer', { args: [admin.address] });
+  const vault = await deploy('v2-vault/Vault', { args: [authorizer.address, tokens.DAI.address, 0, 0] });
```
gtaschuk, line 22: Can use `Vault.create(...)` for consistency

pkg/asset-manager-utils/contracts/AssetManager.sol
nventuro: Perhaps this function was meant to be internal?

TomAFrench: Whoops, this is accidental. I've just made this internal.

nventuro (commented):
Overall the code looks very solid, my suggestions are quite minimal (though I haven't gone over the entire thing).

However, I find reviewing this hard due to the sheer size of the PR caused by seemingly unrelated changes. In fact, the changelog you listed in the PR description seems like great candidates of what would be more surgical PRs that address each individual issue at once along with associated tests.

I find the extra effort of opening multiple self-contained PRs to be well worth it, leading to more detailed discussion on each topic and overall faster review and merge turnaround. If this PR's commits were made in a way such that it'd be easy to split them in different branches, I'd consider giving that a try. 

If its all mixed up however and too much work to untangle, we can go ahead and continue reviewing this as is - though I'd very much appreciate trying to reduce scope in the future.

Thanks!

TomAFrench:
> I find the extra effort of opening multiple self-contained PRs to be well worth it, leading to more detailed discussion on each topic and overall faster review and merge turnaround. If this PR's commits were made in a way such that it'd be easy to split them in different branches, I'd consider giving that a try.

Apologies. This is also my preference but this branch ran away a bit.

TomAFrench:
Thanks for the reviews. I'm going to close this PR and replace it with a few more targeted ones while addressing any review issues.

nventuro:
Thanks a lot for that Tom!

PR #601, opened 2021-06-02
Multipool support for multirewards

gtaschuk:
This allows you to stake bpt from multiple pools to earn multiple reward tokens so we can have a singleton staking contract

TODO
- Batch ERC20/vault transfers for lower gas use (as is this does multiple erc20 transfers if you were to claim BAL from multiple pools)
- Rename to avoid confusion with Curve's `MultiRewards` contract.  Accepting suggestions better than `MultiMultiRewards`

First step of https://github.com/balancer-labs/balancer-v2-monorepo/issues/602

### DIFF
--- a/pkg/asset-manager-utils/contracts/SinglePoolAaveATokenAssetManager.sol
+++ b/pkg/asset-manager-utils/contracts/SinglePoolAaveATokenAssetManager.sol
@@ -95,6 +95,7 @@ contract SinglePoolAaveATokenAssetManager is SinglePoolAssetManager {
         aaveIncentives.claimRewards(assets, type(uint256).max, address(this));
 
         // Forward to distributor
-        distributor.notifyRewardAmount(stkAave, stkAave.balanceOf(address(this)));
+        IERC20 poolAddress = IERC20((uint256(poolId) >> (12 * 8)) & (2**(20 * 8) - 1));
+        distributor.notifyRewardAmount(poolAddress, stkAave, stkAave.balanceOf(address(this)));
     }
 }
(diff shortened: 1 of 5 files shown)


TomAFrench:
> Accepting suggestions better than MultiMultiRewards

If we're planning on this being the main source of rewards from liquidity mining + asset managers then just `BalancerRewards` should be fine.

TomAFrench:
How open are we wanting to make this staking contract? Ideally we would allow permissionless rewarding of various pools without having to go through Balancer itself.

gtaschuk:
> How open are we wanting to make this staking contract? Ideally we would allow permissionless rewarding of various pools without having to go through Balancer itself.

I propose we add a per-pool whitelist of reward tokens.  We just need to be mindful of rewardTokens being added to a pool incurring additional gas costs to claim, since the LP is doing an all or nothing claim on a per-pool basis.

It seems like ultimately it should be up to the pool what tokens it should allow?

TomAFrench:
> I propose we add a per-pool whitelist of reward tokens. We just need to be mindful of rewardTokens being added to a pool incurring additional gas costs to claim, since the LP is doing an all or nothing claim on a per-pool basis.

This is my concern. Permissionless adding of rewards is a DoS attack vector while we're iterating through the entire list however we don't want to have endless votes on distributing new tokens.

Would providing a 2D array of the tokens the user wants to claim for under each pool work? I don't see a way in which a staker could manipulate the contract to get *more* rewards.

nventuro (approved):
The points raised by @TomAFrench are very valid: it'd be great if we could leave this entirely open, while at the same time not risking a DoS.

I propose the following:
 - merge this as-is, with `addRewards` remaining `onlyOwner`, so we already have something that works and can start using in other places, such as the SPAM
 - in parallel, continue work on this contract and its tests, both to make sure making it multi-pool doesn't introduce any issues, and to look into ways we might be able to get rid of `Ownable`

PR #605, opened 2021-06-02
Add basic LogExpMath tests

nventuro:
These tests existed in a private repo and had not been ported here. They are quite simple, but might be useful as a starting point for more detailed tests in the future.


dmf7z (approved):
As initial tests they are ok. In the future, we should also test log function because we are using it in the oracle. And also do tests of this sort:  `log(exp(x)) = x`

nventuro:
Indeed, these are very much not comprehensive: they are simply a port of tests we already had and wanted to preserve them (and serve as a foundation for future tests).

PR #606, opened 2021-06-02
Remove multi-tenant asset manager

TomAFrench:
Based on conversations where it was agreed that the `SinglePoolAssetManager` would be the default asset manager for most deployments, I've removed the multi-tenant asset manager and renamed `SinglePoolAssetManager` as `AssetManager`.

I've maintained the `rebalanceAndSwap` function in `AssetManager`


gtaschuk:
Nico and I were discussing this and we decided that the multitenant asset manager would likely still be useful for cases in which an asset manager didn't need to be able to distribute rewards.

I vote we keep it for ease of reuse and deployment

TomAFrench:
I agree there's a place for the muiltenant asset manager in the future but I think for the time being we can focus on the single pool version and pull back the other version when needed. While we're pushing to get out the initial version I don't think it's worth trying to maintain parity across two versions.

nventuro:
Yup I agree with @TomAFrench's take. We both have the commit history and the knowledge of how to implement a multi-tenant version when (if) we need to. YAGNI applies quite well here.

I would consider however renaming SPAM to `RewardsAssetManager` instead, since that is what it really is: an Asset Manager that is capable of handling rewards, is connected to the reward distributor, etc. The fact that it can only handle a single Pool is an implementation detail, not a goal in itself.

gtaschuk:
hmm, I think it's quite likely that we'll need it and they aren't coupled in any way that requires any work to keep the multipool asset manager up to date.  I don't think it does any harm and is a good base class for someone who wants to launch an asset manager.

Can we wait to merge this one until #593 is in?  They conflict

nventuro (approved):

PR #608, opened 2021-06-04
Combine getNormalizedWeights and getMaxWeightTokenIndex

EndymionJkb:
In talking through how to make mutable weight pools efficient, we noted that the max token index was only used for calculating fees on joins/exits, and those functions also needed all the token weights. So combining the max weight token index and normalizedWeights getters could be more efficient. The weights can be used on their own (e.g., in the invariant calculations) sometimes, but the max index is only needed for fees.

### DIFF
--- a/pkg/pool-weighted/contracts/BaseWeightedPool.sol
+++ b/pkg/pool-weighted/contracts/BaseWeightedPool.sol
@@ -71,14 +71,10 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool, WeightedMath {
     function _getNormalizedWeight(IERC20 token) internal view virtual returns (uint256);
 
     /**
-     * @dev Returns all normalized weights, in the same order as the Pool's tokens.
+     * @dev Returns all normalized weights, in the same order as the Pool's tokens, along with the index of the token
+     * with the highest weight
      */
-    function _getNormalizedWeights() internal view virtual returns (uint256[] memory);
-
-    /**
-     * @dev Returns the index of the token with the largest weight.
-     */
-    function _getMaxWeightTokenIndex() internal view virtual returns (uint256);
+    function _getNormalizedWeightsAndMaxWeightIndex() internal view virtual returns (uint256[] memory, uint256);
 
     function getLastInvariant() external view returns (uint256) {
         return _lastInvariant;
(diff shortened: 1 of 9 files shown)


nventuro (commented):
Thanks a lot @EndymionJkb!

pkg/pool-weighted/contracts/BaseWeightedPool.sol
```diff
@@ -94,8 +99,8 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool, WeightedMath {
         return WeightedMath._calculateInvariant(normalizedWeights, balances);
     }
 
-    function getNormalizedWeightsAndMaxWeightIndex() external view returns (uint256[] memory, uint256) {
```
nventuro: Sorry, I wasn't clear in my previous message :sweat-smile:

I meant let's keep `getWeightsAndMaxIndex` in the base class, but in the special case of `WeightedPool`, implement it by calling `getWeights` and then also returning `maxIndex`, to keep that implementation clean.

pvt/helpers/src/models/pools/weighted/WeightedPool.ts
```diff
@@ -171,10 +176,16 @@ export default class WeightedPool {
...
+  // Valid on Oracle pools
   async getNormalizedWeights(): Promise<BigNumber[]> {
     return this.instance.getNormalizedWeights();
   }
 
+  // Valid on BaseWeightedPool subclasses
+  async getNormalizedWeightsAndMaxWeightIndex(): Promise<[BigNumber[], BigNumber]> {
+    return this.instance.getNormalizedWeightsAndMaxWeightIndex();
+  }
+
```
nventuro, line 188:
```suggestion
```

This doesn't exist now

pvt/helpers/src/models/pools/weighted/WeightedPool.ts
```diff
@@ -171,10 +176,16 @@ export default class WeightedPool {
     return this.instance.getSwapFeePercentage();
   }
 
+  // Valid on Oracle pools
```
nventuro, line 179:
```suggestion
```

nventuro (approved):

nventuro (approved):

PR #609, opened 2021-06-04
Stable: Improve math roundings

facuspagnuolo:


pkg/core/contracts/pools/BasePool.sol
```diff
@@ -453,7 +453,7 @@ abstract contract BasePool is IBasePool, BasePoolAuthorization, BalancerPoolToke
      */
     function _addSwapFeeAmount(uint256 amount) internal view returns (uint256) {
         // This returns amount + fee amount, so we round up (favoring a higher fee amount).
-        return amount.divUp(_swapFeePercentage.complement());
+        return amount.divUp(FixedPoint.ONE.sub(_swapFeePercentage));
```
nventuro, line 456: This is just more clear, but it doesn't change the behavior, right?

pkg/core/contracts/pools/stable/StableMath.sol
```diff
@@ -106,7 +111,7 @@ contract StableMath {
 
         // Amount out, so we round down overall.
 
-        uint256 invariant = _calculateInvariant(amplificationParameter, balances);
+        uint256 invariant = _calculateInvariant(amplificationParameter, balances, true);
```
nventuro, line 114: The comment says we're rounding down, but we're rounding up.

pkg/core/contracts/pools/stable/StableMath.sol
```diff
@@ -196,43 +198,42 @@ contract StableMath {
...
+        uint256 currentInvariant = _calculateInvariant(amp, balances, true);
+        uint256 newInvariant = _calculateInvariant(amp, newBalances, false);
+        uint256 invariantRatio = newInvariant.divDown(currentInvariant);
 
-        // return amountBPTOut
-        return bptTotalSupply.mulDown(newInvariant.divDown(currentInvariant).sub(FixedPoint.ONE));
+        if (invariantRatio > FixedPoint.ONE) {
+            return bptTotalSupply.mulDown(invariantRatio.sub(FixedPoint.ONE));
+        } else {
+            return 0;
```
nventuro, line 235: This should only happen due to rounding, correct? Can we have a small comment explaining that if this is the case?

pkg/core/contracts/pools/stable/StableMath.sol
```diff
@@ -357,34 +355,40 @@ contract StableMath {
         uint256 bptTotalSupply,
         uint256 swapFeePercentage
     ) internal pure returns (uint256) {
+        // Token out, so we round down overall.
+
         // Get the current invariant
-        uint256 currentInvariant = _calculateInvariant(amp, balances);
+        uint256 currentInvariant = _calculateInvariant(amp, balances, true);
```
nventuro, line 361: If we're rounding down, why is the invariant rounded up? It'd be good to have a short comment about this

pkg/core/contracts/pools/stable/StablePool.sol
```diff
@@ -409,23 +413,34 @@ contract StablePool is BaseGeneralPool, StableMath {
     }
 
     function _invariantAfterJoin(uint256[] memory balances, uint256[] memory amountsIn) private view returns (uint256) {
-        for (uint256 i = 0; i < _getTotalTokens(); ++i) {
-            balances[i] = balances[i].add(amountsIn[i]);
-        }
-
-        return StableMath._calculateInvariant(_amplificationParameter, balances);
+        _mutateAmounts(balances, amountsIn, FixedPoint.add);
+        // This invariant is used only to calculate the due protocol fees, then we round up
```
nventuro, line 417: Shouldn't protocol fees be rounded down?

pkg/core/contracts/pools/stable/StablePool.sol
```diff
@@ -434,6 +449,10 @@ contract StablePool is BaseGeneralPool, StableMath {
      */
     function getRate() public view returns (uint256) {
         (, uint256[] memory balances, ) = getVault().getPoolTokens(getPoolId());
-        return StableMath._calculateInvariant(_amplificationParameter, balances).divDown(totalSupply());
+
+        // When calculating the current BPT rate, we may not have paid the protocol fees, therefore
+        // the invariant should be smaller than it's current value. Then, we round down overall.
```
nventuro, line 454:
```suggestion
        // the invariant should be smaller than its current value. Then, we round down overall.
```

nventuro (commented):
Excelent :raised_hands: Thank you very much for updating the comments and bringing this more in line with the Weighted Pool!

I did an initial pass asking a few questions about things that seemed odd, I'll do another more detailed pass later today.

pkg/core/contracts/pools/BasePool.sol
facuspagnuolo: We did change the behavior, but the swap fee pct is guaranteed to be lower than or equal to 50%

pkg/pool-stable/contracts/StablePool.sol
```diff
@@ -410,23 +414,34 @@ contract StablePool is BaseGeneralPool, StableMath {
     }
 
     function _invariantAfterJoin(uint256[] memory balances, uint256[] memory amountsIn) private view returns (uint256) {
-        for (uint256 i = 0; i < _getTotalTokens(); ++i) {
-            balances[i] = balances[i].add(amountsIn[i]);
-        }
-
-        return StableMath._calculateInvariant(_amplificationParameter, balances);
+        _mutateAmounts(balances, amountsIn, FixedPoint.add);
+        // This invariant is used only to compute the final balance when calculating the protocol fees, then we round up
```
nventuro, line 418:
```suggestion
        // This invariant is used only to compute the final balance when calculating the protocol fees. These are rounded
        // down, so we round the invariant up.
```

pkg/pool-stable/contracts/StablePool.sol
```diff
@@ -410,23 +414,34 @@ contract StablePool is BaseGeneralPool, StableMath {
...
+        return StableMath._calculateInvariant(_amplificationParameter, balances, true);
     }
 
     function _invariantAfterExit(uint256[] memory balances, uint256[] memory amountsOut)
         private
         view
         returns (uint256)
     {
+        _mutateAmounts(balances, amountsOut, FixedPoint.sub);
+        // This invariant is used only to compute the final balance when calculating the protocol fees, then we round up
```
nventuro, line 428:
```suggestion
        // This invariant is used only to compute the final balance when calculating the protocol fees. These are rounded
        // down, so we round the invariant up.
```

nventuro (approved):

PR #610, opened 2021-06-04
Consolidate asset manager tests styles

TomAFrench:
We have multiple styles of tests currently for asset managers, I've converted all of them to the most up to date format.

Waiting on #606 to be merged for a rebase.

### DIFF
--- a/pkg/asset-manager-utils/test/AaveATokenAssetManager.test.ts
+++ b/pkg/asset-manager-utils/test/AaveATokenAssetManager.test.ts
@@ -1,5 +1,5 @@
 import { ethers } from 'hardhat';
-import { Contract } from 'ethers';
+import { BigNumber, Contract } from 'ethers';
 import { expect } from 'chai';
 import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/dist/src/signer-with-address';
 
@@ -9,10 +9,9 @@ import { bn, fp } from '@balancer-labs/v2-helpers/src/numbers';
 import { MAX_UINT256 } from '@balancer-labs/v2-helpers/src/constants';
 
 import { deploy } from '@balancer-labs/v2-helpers/src/contract';
-import Vault from '@balancer-labs/v2-helpers/src/models/vault/Vault';
 import { expectBalanceChange } from '@balancer-labs/v2-helpers/src/test/tokenBalance';
-import { advanceTime } from '@balancer-labs/v2-helpers/src/time';
 import { encodeJoinWeightedPool } from '@balancer-labs/v2-helpers/src/models/pools/weighted/encoding';
+import { advanceTime } from '@balancer-labs/v2-helpers/src/time';
 
 const OVER_INVESTMENT_REVERT_REASON = 'investment amount exceeds target';
 const UNDER_INVESTMENT_REVERT_REASON = 'withdrawal leaves insufficient balance invested';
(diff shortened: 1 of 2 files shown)


pkg/asset-manager-utils/test/AaveATokenAssetManager.test.ts
```diff
@@ -9,104 +9,134 @@ import { bn, fp } from '@balancer-labs/v2-helpers/src/numbers';
...
-    pool: Contract,
-    vault: Contract,
-    lendingPool: Contract,
-    assetManager: Contract,
-    distributor: Contract,
-    stkAave: Contract;
-  let admin: SignerWithAddress, lp: SignerWithAddress, other: SignerWithAddress;
-  let poolId: string;
+const setup = async () => {
+  const [, admin, lp, other] = await ethers.getSigners();
```
nventuro, line 23: I think not having the `setup` function and having this in one (or many) `sharedBeforeEach`s may actually be the better approach (and consistent with all other test files we have). For one, it avoids having to keep the `getSigners()` calls in sync.

pkg/asset-manager-utils/test/AaveATokenAssetManager.test.ts
```diff
@@ -9,104 +9,134 @@ import { bn, fp } from '@balancer-labs/v2-helpers/src/numbers';
...
-  let tokens: TokenList,
-    pool: Contract,
-    vault: Contract,
-    lendingPool: Contract,
-    assetManager: Contract,
-    distributor: Contract,
-    stkAave: Contract;
-  let admin: SignerWithAddress, lp: SignerWithAddress, other: SignerWithAddress;
-  let poolId: string;
+const setup = async () => {
```
gtaschuk, line 22: https://github.com/balancer-labs/balancer-v2-monorepo/pull/593#discussion_r644091580 :rofl:

nventuro (approved):

gtaschuk:
Honestly I don't have a problem mixing use of `setup` and `sharedBeforeEach` setup in the repo, especially as it grows.   Much prefer documented conventions and automated tools to reviews that say "oh we do it this way elsewhere, can you change it here?"  Similarly refactors like this should only be done when they really add value as they take time to review and pollute git history.

The new tests are useful for sure!

TomAFrench:
I'm happy to move towards more modular `sharedBeforeEach` approach (probably in a later PR).

For context, my main motivation for `setup` was separation of the "boring" setup like deploying tokens and the vault: I noticed that when we had this mixed in with the tests it tended to bloat them such that you had to scroll back and forth over a block of boilerplate to check variables at the top of the `describe` block. It also led to extraneous variables which were only being used in setup being made available to the whole set of tests.

Of course now that we have initialisation of asset managers we should move this out of `setup` so that we can test improper initialisation, etc. but otherwise I'd like to keep the plumbing which isn't relevant to what we're testing out of the way.

gtaschuk (changes requested):
could you back out the `setup()` change?  Mostly asking cause I just did the work of inlining it yesterday based on review

EDIT: thanks for providing rationale (we commented at the same time).   Still slightly prefer we limit this PR to just new tests and keep refactors to separate PRs, but I'm fine with this pattern/merging as-is.

pkg/asset-manager-utils/test/RewardsAssetManager.test.ts
```diff
@@ -131,6 +131,8 @@ describe('Rewards Asset manager', function () {
         'Fee on critical rebalances must be less than 10%'
       );
     });
+
+    it('prevents an unauthorized user from setting the pool config');
```
gtaschuk, line 135: better to explicitly mark as pending with `xit` imho.  If tests with empty blocks are marked as pending than disregard!

TomAFrench: Yep, `it` blocks without an associated function show as pending so this is a no-op.

TomAFrench:
> could you back out the `setup()` change? Mostly asking cause I just did the work of inlining it yesterday based on review

I'm open to inlining the setup in future (happy to take this on myself) but I'd quite like to get this merged as-is so I can share my other PRs as modifying the structure of the test files will add quite a bit of work late in the day.

PR #611, opened 2021-06-04
Fix updating pool balances to account for fees

TomAFrench:
Previously we were just performing a withdrawal from the Vault to retrieve fees to be paid out to the keeper. The pool then thinks that it still has those funds in its managed balance.

We now update the managed balance to account for these fees being lost.

### DIFF
--- a/pkg/asset-manager-utils/contracts/RewardsAssetManager.sol
+++ b/pkg/asset-manager-utils/contracts/RewardsAssetManager.sol
@@ -207,6 +207,19 @@ abstract contract RewardsAssetManager is IAssetManager {
     }
 
     /**
+     * @dev When withdrawing `amount` will be moved from the pool's cash to managed balance
+     * As these funds are to be paid as fees (and so lost) we cache the previous managed balance and restore it.
+     */
+    function _removeFeesFromVault(uint256 amount) private {
+        // Pull funds from the vault and update balance to reflect that the fee is no longer part of managed funds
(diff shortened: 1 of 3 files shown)


pkg/asset-manager-utils/contracts/RewardsAssetManager.sol
```diff
@@ -206,6 +206,20 @@ abstract contract RewardsAssetManager is IAssetManager {
...
+    /**
+     * @notice withdraw `amount` of cash from the Vault, reducing the pool's TVL
+     * @dev When withdrawing `amount` will be moved from the pool's cash to managed balance
+     * As these funds are to be paid as fees (and so lost) we cache the previous managed balance and restore it.
+     */
+    function _withdrawCashFromVault(uint256 amount) private {
+        // Pull funds from the vault and update balance to reflect that the fee is no longer part of managed funds
+        IVault.PoolBalanceOp[] memory ops = new IVault.PoolBalanceOp[](2);
+        ops[0] = IVault.PoolBalanceOp(IVault.PoolBalanceOpKind.WITHDRAW, poolId, token, amount);
+        ops[1] = IVault.PoolBalanceOp(IVault.PoolBalanceOpKind.UPDATE, poolId, token, readAUM().sub(amount));
```
nventuro, line 218: The comments and implementation don't match: we're not restoring the previous value, but rather syncing the Vault's `managed` value with the actual current value of AUM.

nventuro (approved):

PR #613, opened 2021-06-04
Add rebalanceAndSwap testcases

TomAFrench:
I've changed `rebalanceAndSwap` to take a `BatchSwap` struct as converting only `swaps` to being in memory led to stack errors.


nventuro (approved):
Looking good! I removed the AaveAM tests as it inherits from RewardAM - no point in having these tests duplicated.

In future PRs we should also try to avoid much duplication inside the tests themsevels - not because duplication is bad by itself here but because higher level tests (e.g. `itRebalancesCorrectly()`) make the overall structure and tested behavior easier to understand.

Issue #616, opened 2021-06-04
Fix computation of protocol fees in StablePool


nventuro:
https://github.com/balancer-labs/balancer-v2-monorepo/pull/394 introduced a way to change the amplification factor, which in turn changes the computation of the invarant. This can lead to errors when computing due protocol fees.

We decided the cleanest way to handle this is to store the current Amp value along with `_lastInvariant`, and to use it to compute the current invariant when computing due fees. This essentially means that, in the worst case, the Pool is performing an internal swap of its collected fees using a slightly old amplification factor, which is deemed a non-issue.

PR #619, opened 2021-06-07
Fix Protocol Fees computation on changing amp

nventuro:
Fixes https://github.com/balancer-labs/balancer-v2-monorepo/issues/616

This adds an extra SSTORE to all joins and exits (~5k). If we really wanted to, we could merge this with `BasePool`'s `_swapFeePercentage` to avoid the extra gas, but that would require changes to this other contract.


dmf7z (commented):
Looks good 👍

PR #621, opened 2021-06-08
Stable: Increase amp precision

facuspagnuolo:


pkg/pool-stable/contracts/StablePool.sol
```diff
@@ -20,6 +20,7 @@ import "@balancer-labs/v2-solidity-utils/contracts/helpers/InputHelpers.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
 
 import "@balancer-labs/v2-pool-utils/contracts/BaseGeneralPool.sol";
+import "hardhat/console.sol";
```
dmf7z, line 23: :)

pkg/pool-stable/contracts/StableMath.sol
```diff
@@ -28,6 +28,7 @@ contract StableMath {
 
     uint256 internal constant _MIN_AMP = 1;
     uint256 internal constant _MAX_AMP = 5000;
+    uint256 internal constant _AMP_PRECISION = 1e4;
```
dmf7z, line 31: Curve has it in 1e2 (100), However,I think we can leave it at 1e4 (10000) and tests if there are no overflows for 5 tokens pool.

dmf7z (commented):
Math looks good!

pkg/pool-stable/test/StablePool.test.ts
```diff
@@ -115,7 +116,7 @@ describe('StablePool', function () {
 
         it('sets amplification', async () => {
           const { value, isUpdating } = await pool.getAmplificationParameter();
-          expect(value).to.be.equal(AMPLIFICATION_PARAMETER);
+          expect(value).to.be.equal(AMPLIFICATION_PARAMETER.mul(AMP_PRECISION));
```
dmf7z, line 119: we could use the precision returned by getAmplificationParameter

dmf7z (approved):

PR #626, opened 2021-06-09
Delete CODEOWNERS

nventuro:
It seems this feature isn't proving as useful as I expected it would - in this early stage and given team size it is difficult to assign static owners to directories due to how we assign work internally. 

The standard approach of 'PR creator asks people for review manually' seems better suited for now.


dmf7z (approved):
Sounds good!

PR #627, opened 2021-06-11
Claim from distributor to a weighted pool

gtaschuk:
This implements a reinvestment flow wherein a reward is claimed directly into a weighted pool

This allows a LP to reinvest reward tokens in a single transaction which could otherwise require 3 transactions:
- Withdraw reward tokens (ie to an EOA)
- Approve reward token to the vault (if claimed to an EOA)
- join pool with the reward token

Presenting this as a proof of concept.  If it seems sensible, I'll add more tests and checks...


gtaschuk:
Closing in favor of approach in #659

PR #628, opened 2021-06-11
Stable: Emit AmpUpdateStopped in construction

nventuro:
By having the initial value in an event, amplification parameter values can be computed by just looking at events.

### DIFF
--- a/pkg/pool-stable/contracts/StablePool.sol
+++ b/pkg/pool-stable/contracts/StablePool.sol
@@ -87,7 +87,6 @@ contract StablePool is BaseGeneralPool, StableMath {
 
         uint256 initialAmp = Math.mul(amplificationParameter, _AMP_PRECISION);
         _setAmplificationData(initialAmp);
-        emit AmpUpdateStopped(initialAmp);
     }
 
     // Base Pool handlers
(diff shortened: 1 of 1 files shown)


pkg/pool-stable/contracts/StablePool.sol
```diff
@@ -84,7 +84,10 @@ contract StablePool is BaseGeneralPool, StableMath {
 
         _require(amplificationParameter >= _MIN_AMP, Errors.MIN_AMP);
         _require(amplificationParameter <= _MAX_AMP, Errors.MAX_AMP);
-        _setAmplificationData(Math.mul(amplificationParameter, _AMP_PRECISION));
+
+        uint256 initialAmp = Math.mul(amplificationParameter, _AMP_PRECISION);
+        _setAmplificationData(initialAmp);
+        emit AmpUpdateStopped(initialAmp);
```
facuspagnuolo, line 90: Nit: We can extract a few lines from `stopAmplificationParameterUpdate` into an internal method and re-use it from here

facuspagnuolo (approved):

pkg/pool-stable/contracts/StablePool.sol
nventuro: You're right, I don't know why I didn't do that.

Issue #571, opened 2021-05-19
Reduce bytecodesize of BasePoolFactory


nventuro:
Even though the bytecode size of the Pool themselves is fine (~19kb), that of the Factory is very close to the 24kb limit. This is because the constructors of the Pools are very large: `immutable` state variables in particular are expensive to initialize.

Given we're thinking of adding new features to the Pools (including Asset Manager support), we should look for alternative ways to create Pools.

Possible approaches include:
 - initializer functions that are delegatecalled into
 - merge `immutable` variables (e.g. the token weights)
 - somehow store the code in multiple places and stitch it together at deployment

PR #629, opened 2021-06-11
Add PoolSplitCodeFactory

nventuro:
This introduces a new base factory, which I expect will fully replace the current BasePoolFactory. This new factory is much less constrained by contract creation code: it supports creation codes of up to 48kB (which ought to be enough for all contracts).

The way it achieves this is by not storing the creation code itself, but instead deploying two secondary contracts that each hold half of the target creation code as their own code. It then fetches this code via `extcodesize` to recreate the original complete creation code during contract creation.

This is all fully implemented inside Solidity and requires no external tooling. Using it is quite similar to the current BasePoolFactory.

```diff
-contract WeightedPoolFactory is BasePoolFactory, FactoryWidePauseWindow {
+contract WeightedPoolFactory is BasePoolSplitCodeFactory, FactoryWidePauseWindow {

-   constructor(IVault vault) BasePoolFactory(vault) { }
+   constructor(IVault vault) BasePoolSplitCodeFactory(vault, type(WeightedPool).creationCode) { }
    
    function create(string memory name, address owner) external returns (address) {
        (uint256 pauseWindowDuration, uint256 bufferPeriodDuration) = getPauseConfiguration();

-       address pool = address(new WeightedPool(getVault(), name, owner));
+       address pool = _create(abi.encode(name, owner));
        return pool;
    }
}
```

The only downside of this approach is that we lose type-safety of the constructor arguments: it is possible to generate incorrect calls to `_create`, or to pass arguments in the wrong order to `abi.encode`. Basic unit testing should detect these issues however.

---

The implementation is sort of tricky, as it relies heavily on `type(Foo).creationCode`, for which Solidity unfortunately does not provide to indicate _where_ we want for this code to be stored. Since the creation code will be very large (>24kB), we can't move it around: all manipulation has to happen in-place.

I created a `CodeDeployer` library in `solidity-utils`, which simply stores data at an address as code. That is then used by `BaseSplitCodeFactory` to implement a split-code factory. `BasePoolSplitCodeFactory` is a thin layer on top, adding a created Pools registry, events, etc. Each of these contracts has their own independent tests, and **I recommend reviewing them in that order**.

Fixes https://github.com/balancer-labs/balancer-v2-monorepo/issues/571


nventuro:
I tried out porting `WeightedPoolFactory` to this new model and everything worked perfectly, while staying well below the bytecode size limits. I plan to include these changes in a later PR.

pkg/solidity-utils/contracts/helpers/BaseSplitCodeFactory.sol
```diff
@@ -0,0 +1,208 @@
...
+    address private immutable _creationCodeStorageB;
+    uint256 private immutable _creationCodeSizeB;
+
+    /**
+     * @dev The creation code of a contract Foo can be obtained inside Solidity with `type(Foo).creationCode`.
+     */
+    constructor(bytes memory creationCode) {
+        uint256 creationCodeSize = creationCode.length;
+
+        // We are going to deploy two contracts: one with the approximately the first half of `creationCode`'s contents
```
EndymionJkb, line 47:
```suggestion
        // We are going to deploy two contracts: one with approximately the first half of `creationCode`'s contents
```

pkg/solidity-utils/contracts/helpers/BaseSplitCodeFactory.sol
```diff
@@ -0,0 +1,208 @@
...
+        uint256 creationCodeSizeB = creationCodeSize - creationCodeSizeA;
+        _creationCodeSizeB = creationCodeSizeB;
+
+        // To deploy the contracts, we're going to use `CodeDeployer.deploy()`, which expects a memory array with
+        // the code to deploy. Note that we cannot simply create arrays for A and B's code by copying or moving
+        // `creationCode`'s contents as they are expected to be very large (> 24kB), so we must operate in-place.
+
+        // Memory: [ code length ] [ A.data ] [ B.data ]
+
+        // Creating A's array is simple: we simply replace `creactionCode`'s length with A's length. We'll later restore
```
EndymionJkb, line 63:
```suggestion
        // Creating A's array is simple: we simply replace `creationCode`'s length with A's length. We'll later restore
```

EndymionJkb (changes requested):
Just straightening the frame of the Mona Lisa here with a couple typos. Will write more (external) docs to make sure I understand it

pkg/solidity-utils/contracts/helpers/BaseSplitCodeFactory.sol
```diff
@@ -0,0 +1,208 @@
...
+    // means this factory supports contracts with creation code of up to 48kB.
+    // We rely on inline-assembly to achieve this, both to make the entire operation highly gas efficient, and because
+    // `extcodecopy` is not avaiable in Solidity.
+
+    // solhint-disable no-inline-assembly
+
+    address private immutable _creationCodeStorageA;
+    uint256 private immutable _creationCodeSizeA;
+
+    address private immutable _creationCodeStorageB;
```
facuspagnuolo, line 38: I'd remove the word "storage" from these

pkg/solidity-utils/contracts/helpers/BaseSplitCodeFactory.sol
```diff
@@ -0,0 +1,208 @@
...
+        // Memory: [ A.length ] [ A.data[ : -1] ] [ B.length ][ B.data ]
+        //         ^ creationCodeA                ^ creationCodeB
+
+        _creationCodeStorageB = CodeDeployer.deploy(creationCodeB);
+
+        // We now restore the original contents of `creationCode` by writing back the original length and A's last byte.
+        assembly {
+            mstore(creationCodeA, creationCodeSize)
+            mstore(creationCodeB, lastByteA)
+        }
```
facuspagnuolo, line 101: Why do you need to restore these contents?

facuspagnuolo: I think this can be simplified in a few lines if you don't care that much about the final memory layout

pkg/solidity-utils/contracts/helpers/BaseSplitCodeFactory.sol
```diff
@@ -0,0 +1,208 @@
...
+        // Finally, we copy the constructorArgs to the end of the array. Unfortunately there is no way to avoid this
+        // copy, as it is not possible to tell Solidity where to store the result of `abi.encode()`.
+        uint256 constructorArgsDataPtr;
+        uint256 constructorArgsCodeDataPtr;
+        assembly {
+            constructorArgsDataPtr := add(constructorArgs, 32)
+            constructorArgsCodeDataPtr := add(add(code, 32), creationCodeSize)
+        }
+
+        _memcpy(constructorArgsCodeDataPtr, constructorArgsDataPtr, constructorArgsSize);
```
facuspagnuolo, line 165: :+1:

facuspagnuolo (approved):
marvelous 👏

pkg/solidity-utils/contracts/helpers/BaseSplitCodeFactory.sol
nventuro: We don't _need_ to restore them - in fact, in the expected use case where we do `SplitCodeFactory(type(Foo).creationCode)` the reference to the original creation code will be lost anyway.

But restoring the original value is so simple that I'd rather we don't mutate it at all and can not worry about it.

PR #630, opened 2021-06-12
LBPool implementation

EndymionJkb:
Full implementation, with initial tests


pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -59,73 +76,351 @@ contract LiquidityBootstrappingPool is BaseWeightedPool, ReentrancyGuard {
...
+        uint256 currentTime = block.timestamp;
+
+        _require(currentTime < endTime, Errors.GRADUAL_UPDATE_TIME_TRAVEL);
+
+        // Must specify normalized weights for all tokens
+        uint256 totalTokens = _getTotalTokens();
+        InputHelpers.ensureInputLengthMatch(totalTokens, endWeights.length);
+
+        // If the start time is in the past, "fast forward" to start now
+        // This prevents circumventing the minimum weight change duration
```
TomAFrench, line 170: We seem to allow zero duration weight changes so this comment doesn't look correct. Was the minimum removed?

EndymionJkb: Leftover comment, thanks! Discussed this with Nico, who made an analogy with Stable Pools, and said a minimum duration might give a false sense of security. First, it can be set to zero anyway (and in fact, most V1 smart pools left it at the 2 minute GUI default, effectively zero). They all could have changed the weights drastically more or less instantly = arbitrary price manipulation, and the users didn't care (and nobody actually did that).

It could possibly be effective if you made it very long, or there were a hard-coded minimum duration (>> 0). That's all up for discussion - it does require trust now, as it does on V1. But for now, there is no minimum duration, so I'll remove that stale comment.

One improvement in V2 over V1 is you can now see what the end weights are, and the start/end times.

EndymionJkb: There is still a reason to do this - preventing sudden discontinuities. I updated the comment.

pkg/pool-weighted/test/LiquidityBootstrappingPool.test.ts
```diff
@@ -1,2 +1,161 @@
...
 // eslint-disable-next-line @typescript-eslint/no-empty-function
-describe('LiquidityBootstrappingPool', function () {});
+describe('LiquidityBootstrappingPool', function () {
+  describe('as a 2-token weighted pool', () => {
+    itBehavesAsWeightedPool(2);
+  });
+
+  describe('as a 3-token weighted pool', () => {
+    itBehavesAsWeightedPool(3);
+  });
```
nventuro, line 22: This is just testing regular weighted pools. Did you mean to add an extra argument?

pkg/pool-weighted/test/LiquidityBootstrappingPool.test.ts
```diff
@@ -1,2 +1,161 @@
...
+    const params = { tokens, weights, owner, lbp: true, swapEnabledOnStart: true };
+    pool = await WeightedPool.create(params);
+  });
+
+  /*const initializePool = () => {
+    sharedBeforeEach('initialize pool', async () => {
+      await pool.init({ initialBalances, recipient: lp });
+    });
+  };*/
+
```
nventuro, line 51:
```suggestion
```

pkg/pool-weighted/test/LiquidityBootstrappingPool.test.ts
```diff
@@ -1,2 +1,161 @@
...
+    tokens = await TokenList.create(['MKR', 'DAI', 'SNX', 'BAT'], { sorted: true });
+    await tokens.mint({ to: [lp, trader], amount: fp(200) });
+  });
+
+  let pool: WeightedPool;
+  const weights = [fp(0.3), fp(0.55), fp(0.1), fp(0.05)];
+  const initialBalances = [fp(0.9), fp(1.8), fp(2.7), fp(3.6)];
+
+  sharedBeforeEach('deploy pool', async () => {
+    const params = { tokens, weights, owner, lbp: true, swapEnabledOnStart: true };
```
nventuro, line 42: For now this might be fine, but it's weird that `swapEnabledOnStart` does nothing if `lbp` is not set. Maybe we could have a `type` field that can be set to `LBP`, and then associated settings? e.g.

```
type LBPConfig = {
    kind: 'LBP';
    swapEnabled?: boolean;
}

type OracleConfig = {
    kind: 'Oracle';
    duration?: Number;
}

type MiscConfig = LBPConfig | OracleConfig;
```

(I'm using oracle as an example of another 'variant' - I'd start with just LBP for now)

pkg/pool-weighted/test/LiquidityBootstrappingPool.test.ts
```diff
@@ -1,2 +1,161 @@
...
+      await expect(pool.swapGivenIn({ in: 1, out: 0, amount: fp(0.1) })).to.be.revertedWith('SWAPS_PAUSED');
+
+      await pool.setPublicSwap(owner, true);
+
+      expect(await pool.instance.swapEnabled()).to.be.true;
+    });
+
+    it('LP cannot pause swaps', async () => {
+      await expect(pool.setPublicSwap(lp, false)).to.be.revertedWith('SENDER_NOT_ALLOWED');
+    });
```
nventuro, line 81: What do you think about splitting these up in a few more concrete tests? e.g.

```
context when the sender is the owner
   it owner can pause
   it pause emits an event
   context when paused
       before each pause
       it swap reverts
       it owner can unpause
       it non owner cannot unpause
       it unpause emits an event
context when the sender is not the owner
   it non owner cannot unpaused
```
Also I'd use a signer called `other` instead of `lp` for these things, as being an LP doesn't really have anything to do with this.

pkg/pool-weighted/test/LiquidityBootstrappingPool.test.ts
```diff
@@ -1,2 +1,161 @@
...
+      //await pool.vault.grantRole(action, owner);
+    });
+
+    it('no public LPs', async () => {
+      await expect(pool.joinGivenIn({ recipient: lp, amountsIn: initialBalances })).to.be.revertedWith(
+        'SENDER_NOT_ALLOWED'
+      );
+    });
+
+    it('allows owner to join', async () => {
```
nventuro, line 98:
```suggestion
    it('non-owners cannot join', async () => {
      await expect(pool.joinGivenIn({ recipient: lp, amountsIn: initialBalances })).to.be.revertedWith(
        'SENDER_NOT_ALLOWED'
      );
    });

    it('owner can join', async () => {
```

Also, isn't `sender` what the contract tests for? We're setting `recipient` here

pkg/pool-weighted/test/LiquidityBootstrappingPool.test.ts
```diff
@@ -1,2 +1,161 @@
...
+    });
+  });
+
+  describe('update weights', () => {
+    sharedBeforeEach('deploy tokens', async () => {
+      const action = await actionId(pool.instance, 'updateWeightsGradually');
+      await pool.vault.grantRole(action, owner);
+    });
+
+    it('only allows owner to update weights', async () => {
```
nventuro, line 109:
```suggestion
    it('non-owner cannot update weights', async () => {
```

I find it's better to say what the test actually does ('if `other` tries to update it reverts'), than what it attempts to prove ('only the owner can update'). It makes the tests easier to read, and the test case summary more honest.

pkg/pool-weighted/test/LiquidityBootstrappingPool.test.ts
```diff
@@ -1,2 +1,161 @@
...
+      });
+
+      it('gets start weights if called early', async () => {
+        const normalizedWeights = await pool.getNormalizedWeights();
+
+        // Need to decrease precision
+        expect(normalizedWeights).to.equalWithError(pool.normalizedWeights, 0.0001);
+      });
+
+      it('gets end weights if called late', async () => {
```
nventuro, line 144:
```suggestion
      it('gets start weights if called before the start time', async () => {
        const normalizedWeights = await pool.getNormalizedWeights();

        // Need to decrease precision
        expect(normalizedWeights).to.equalWithError(pool.normalizedWeights, 0.0001);
      });

      it('gets end weights if called after the end time' async () => {
```

pkg/pool-weighted/test/LiquidityBootstrappingPool.test.ts
```diff
@@ -1,2 +1,161 @@
...
+        await advanceTime(MINUTE * 100);
```
nventuro, line 145: It'd be good to have this be tied to the update duration more explicitly instead of it being a magic value

pkg/pool-weighted/test/LiquidityBootstrappingPool.test.ts
```diff
@@ -1,2 +1,161 @@
...
+        const normalizedWeights = await pool.getNormalizedWeights();
+
+        // Need to decrease precision
+        expect(normalizedWeights).to.equalWithError(endWeights, 0.0001);
+      });
+
+      it('gets intermediate weights if called halfway through', async () => {
+        await advanceTime(MINUTE * 40);
```
nventuro, line 153: Same here, it is not obvious that this is the halfway point. 

Other test scenarios we should have:
 - some more percentages (e.g. 30 and 70): 50% is a particularly tricky one because it doesn't detect errors where you're taking the wrong end
 - calling 'start update' after start time

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -16,25 +16,41 @@ pragma solidity ^0.7.0;
...
-    // Not worth packing - only referenced on join/exits
-    uint256 private _maxWeightTokenIndex;
+    // Setting this to false pauses trading
+    bool public swapEnabled;
+
+    // For gas optimization, store start/end weights and timestamps in one bytes32
+    // Start weights need to be high precision, since restarting the update resets them to "spot"
+    // values. Target end weights do not need as much precision.
+    // <---     128 bits     ---|---     64 bits     ---|---  32 bits ----|--  32 bits ----|
+    // 4 x 32bit start weights  | 4 x 16bit end weights | start timestamp | end timestamp  |
```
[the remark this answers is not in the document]
EndymionJkb: Ok - will copy this format

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -16,25 +16,41 @@ pragma solidity ^0.7.0;
...
+    bytes32 private _poolState;
+
+    // Event declarations
+
+    event PublicSwapSet(bool swapEnabled);
```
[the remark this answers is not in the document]
EndymionJkb: Yes; name left over from V1

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -59,73 +76,352 @@ contract LiquidityBootstrappingPool is BaseWeightedPool, ReentrancyGuard {
...
+        _require(currentTime < endTime, Errors.GRADUAL_UPDATE_TIME_TRAVEL);
+
+        // Must specify normalized weights for all tokens
+        uint256 totalTokens = _getTotalTokens();
+        InputHelpers.ensureInputLengthMatch(totalTokens, endWeights.length);
+
+        // If the start time is in the past, "fast forward" to start now
+        // This avoids discontinuities in the weight curve. Otherwise, if you set the start/end times with
+        // only 10% of the period in the future, the weights would immediately jump 90%
+        uint256 effectiveStartTime = currentTime > startTime ? currentTime : startTime;
```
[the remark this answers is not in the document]
EndymionJkb: This is from V1. I suppose it could be because of gas too, but the usual use case for this was when people wanted to reissue the same command they did before to just adjust the end weights, so they wouldn't have to recalculate anything, and so that the GUI would still show the original start time.

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -59,73 +76,352 @@ contract LiquidityBootstrappingPool is BaseWeightedPool, ReentrancyGuard {
...
+     */
+    function updateWeightsGradually(
+        uint256 startTime,
+        uint256 endTime,
+        uint256[] memory endWeights
+    ) external authenticate whenNotPaused nonReentrant {
+        // solhint-disable-next-line not-rely-on-time
+        uint256 currentTime = block.timestamp;
+
+        _require(currentTime < endTime, Errors.GRADUAL_UPDATE_TIME_TRAVEL);
```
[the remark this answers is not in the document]
EndymionJkb: I reordered from V1, but forgot that one

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -59,73 +76,352 @@ contract LiquidityBootstrappingPool is BaseWeightedPool, ReentrancyGuard {
...
+        internal
+        override
+        whenNotPaused
+        returns (
+            uint256,
+            uint256[] memory,
+            uint256[] memory
+        )
+    {
+        _require(sender == getOwner(), Errors.SENDER_NOT_ALLOWED);
```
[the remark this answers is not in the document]
EndymionJkb: That's a good catch - don't want somebody front-running the funding. Will comment

pkg/pool-weighted/test/LiquidityBootstrappingPool.test.ts
EndymionJkb: No, I meant to make sure nothing in the derived class broke any standard behavior (since it does override things). Unnecessary?

pkg/pool-weighted/test/LiquidityBootstrappingPool.test.ts
EndymionJkb: Oracle is another type - oracleEnabled also does nothing if twoTokens isn't set. I was trying not to make a new model.

pkg/pool-weighted/test/LiquidityBootstrappingPool.test.ts
EndymionJkb: Ok; the tests are in better shape. Will look into generalizing further (to different counts and weights) later

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -45,7 +70,8 @@ contract LiquidityBootstrappingPool is BaseWeightedPool, ReentrancyGuard {
         uint256 swapFeePercentage,
         uint256 pauseWindowDuration,
         uint256 bufferPeriodDuration,
-        address owner
+        address owner,
```
TomAFrench, line 73: I would add a requirement that this isn't the zero address (potentially not event delegated ownership?). It would prevent headaches from someone shelling out $ on a useless pool.

TomAFrench: Maybe not on disallowing delegated ownership. That's something that's much harder to do by accident.

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -59,73 +85,345 @@ contract LiquidityBootstrappingPool is BaseWeightedPool, ReentrancyGuard {
...
-                maxWeightTokenIndex = i;
-                maxNormalizedWeight = normalizedWeight;
-            }
         }
         // Ensure that the normalized weights sum to ONE
         _require(normalizedSum == FixedPoint.ONE, Errors.NORMALIZED_WEIGHT_INVARIANT);
 
-        // Initial value; can change later if weights change
-        _maxWeightTokenIndex = maxWeightTokenIndex;
+        // Write initial pool state - all zeros except for the start weights
```
TomAFrench, line 111:
```suggestion
        // Write initial pool state
```

We now fill in both start and end weights so this is stale. I don't think we need to describe what the state looks like as it isn't an atypical state.

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
EndymionJkb: The CRP did this (prevented transferring control to 0 - it was initially set to the sender, which can't be zero, so it didn't have the same initialization issue).

Of course, you can always defeat it by setting the owner to a non-zero burn address, but that would be deliberate. An error in a script could cause the owner to be set to zero accidentally. It's not much gas, and only on construction.

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
EndymionJkb: Good catch!

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -59,73 +85,345 @@ contract LiquidityBootstrappingPool is BaseWeightedPool, ReentrancyGuard {
...
 
+        bytes32 poolState = _poolState;
+
         // prettier-ignore
         {
-            if (totalTokens > 0) { normalizedWeights[0] = _getNormalizedWeight(0); } else { return normalizedWeights; }
-            if (totalTokens > 1) { normalizedWeights[1] = _getNormalizedWeight(1); } else { return normalizedWeights; }
-            if (totalTokens > 2) { normalizedWeights[2] = _getNormalizedWeight(2); } else { return normalizedWeights; }
-            if (totalTokens > 3) { normalizedWeights[3] = _getNormalizedWeight(3); } else { return normalizedWeights; }
+            if (totalTokens > 0) { normalizedWeights[0] = _getNormalizedWeightByIndex(0, poolState);
```
TomAFrench, line 248: Facu used a nice format for this sort of thing in #673. This block would then look like
```
if (totalTokens == 0)  return normalizedWeights;
normalizedWeights[0] = _getNormalizedWeightByIndex(0, poolState);
if (totalTokens == 1  return normalizedWeights;
normalizedWeights[1] = _getNormalizedWeightByIndex(1, poolState);
if (totalTokens == 2)  return normalizedWeights;
normalizedWeights[2] = _getNormalizedWeightByIndex(2, poolState);
if (totalTokens == 3)  return normalizedWeights;
normalizedWeights[3] = _getNormalizedWeightByIndex(3, poolState);
```

We could also remove some of the redundant checks as so:
```
normalizedWeights[0] = _getNormalizedWeightByIndex(0, poolState);
normalizedWeights[1] = _getNormalizedWeightByIndex(1, poolState);
if (totalTokens == 2)  return normalizedWeights;
normalizedWeights[2] = _getNormalizedWeightByIndex(2, poolState);
if (totalTokens == 3)  return normalizedWeights;
normalizedWeights[3] = _getNormalizedWeightByIndex(3, poolState);
```

EndymionJkb: Ok; will update

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -59,73 +85,345 @@ contract LiquidityBootstrappingPool is BaseWeightedPool, ReentrancyGuard {
...
+    /**
+     * @dev When calling updateWeightsGradually again during an update, reset the start weights to the current weights,
+     * if necessary. Since it is called from a view function, read from storage and call in memory
+     */
+    function _initializeGradualWeightUpdate(
+        uint256 totalTokens,
+        uint256 startTime,
+        uint256 endTime
+    ) private view returns (bytes32, uint256[] memory) {
+        // A weight change is (or was) in progress, we need to nodify the start weights
```
TomAFrench, line 395:
```suggestion
        // A weight change is (or was) in progress, we need to notify the start weights
```

EndymionJkb: Todally missed that one. Thanks!

pkg/pool-weighted/test/LiquidityBootstrappingPool.test.ts
```diff
@@ -1,2 +1,221 @@
...
+        await pool.updateWeightsGradually(owner, startTime, endTime, endWeights);
+      });
+
+      it('stores the params', async () => {
+        const updateParams = await pool.getGradualWeightUpdateParams();
+
+        expect(updateParams.startTime).to.equalWithError(startTime, 0.001);
+        expect(updateParams.endTime).to.equalWithError(endTime, 0.001);
+        expect(updateParams.endWeights).to.equalWithError(endWeights, 0.001);
+      });
```
nventuro, line 195: We should have a test where start time is in the past, and it is fast-forwarded to now.

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -19,22 +19,47 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/ReentrancyGuard.
...
 
-    // For gas optimization, store the 4 weights as uint64's in a single state variable
-    bytes32 private _normalizedWeights;
+    // State variables
 
-    // The protocol fees will always be charged using the token associated with the max weight in the pool.
-    // Not worth packing - only referenced on join/exits
-    uint256 private _maxWeightTokenIndex;
+    // All swaps fail while this is false
+    bool public swapEnabled;
```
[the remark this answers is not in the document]
EndymionJkb: Ok; you mean make it private and add a getter

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -19,22 +19,47 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/ReentrancyGuard.
...
+    // Start weights need to be high precision, since restarting the update resets them to "spot"
+    // values. Target end weights do not need as much precision.
+    // [     32 bits   |     32 bits     |      64 bits     |      128 bits      |
+    // [ end timestamp | start timestamp | 4x16 end weights | 4x32 start weights |
+    // |MSB                                                                   LSB|
+
+    bytes32 private _poolState;
+
+    // Offsets for data elements in _poolState
+    // Start weights begin at offset 0
```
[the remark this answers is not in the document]
EndymionJkb: Ok; the optimizer should take care of it so it doesn't add bytecode (e.g., i * 32 vs 0 + i * 32

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -59,73 +85,343 @@ contract LiquidityBootstrappingPool is BaseWeightedPool, ReentrancyGuard {
...
-        uint256 maxWeightTokenIndex = 0;
-        uint256 maxNormalizedWeight = 0;
-        for (uint8 i = 0; i < numTokens; i++) {
+        for (uint8 i = 0; i < totalTokens; i++) {
             uint256 normalizedWeight = normalizedWeights[i];
             _require(normalizedWeight >= _MIN_WEIGHT, Errors.MIN_WEIGHT);
 
-            _setNormalizedWeight(normalizedWeight, i);
+            // Insert "start weights" and "end weights" into poolState
+            // Before the first update, start=end, which avoids initialization edge cases
```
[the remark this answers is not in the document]
EndymionJkb: Well, there is no weight update until they call updateWeightsGradually (which they theoretically might never do). I also found I didn't really need to initialize the times (it just checks if they're equal, so 0 works as well as any other value, and saves the gas/bytecode of writing values that are never used).

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -59,73 +85,343 @@ contract LiquidityBootstrappingPool is BaseWeightedPool, ReentrancyGuard {
...
+        uint256 totalTokens = _getTotalTokens();
+        endWeights = new uint256[](totalTokens);
+
+        // prettier-ignore
+        {
+            endWeights[0] = poolState.decodeUint16(_END_WEIGHT_OFFSET).uncompress16();
+            endWeights[1] = poolState.decodeUint16(_END_WEIGHT_OFFSET + 16).uncompress16();
+            if (totalTokens > 2) { endWeights[2] = poolState.decodeUint16(_END_WEIGHT_OFFSET + 32).uncompress16(); }
+            if (totalTokens > 3) { endWeights[3] = poolState.decodeUint16(_END_WEIGHT_OFFSET + 48).uncompress16(); }
+        }
```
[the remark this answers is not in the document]
EndymionJkb: This layout came from Facu (via Tom); I changed it after previous comments :) It would use less gas this way, since 95% of LBPs will be two tokens.

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -59,73 +85,343 @@ contract LiquidityBootstrappingPool is BaseWeightedPool, ReentrancyGuard {
...
-            if (totalTokens > 0) { normalizedWeights[0] = _getNormalizedWeight(0); } else { return normalizedWeights; }
-            if (totalTokens > 1) { normalizedWeights[1] = _getNormalizedWeight(1); } else { return normalizedWeights; }
-            if (totalTokens > 2) { normalizedWeights[2] = _getNormalizedWeight(2); } else { return normalizedWeights; }
-            if (totalTokens > 3) { normalizedWeights[3] = _getNormalizedWeight(3); } else { return normalizedWeights; }
+            normalizedWeights[0] = _getNormalizedWeightByIndex(0, poolState);
+            normalizedWeights[1] = _getNormalizedWeightByIndex(1, poolState);
+            if (totalTokens == 2) return normalizedWeights;
+            normalizedWeights[2] = _getNormalizedWeightByIndex(2, poolState);
+            if (totalTokens == 3) return normalizedWeights;
+            normalizedWeights[3] = _getNormalizedWeightByIndex(3, poolState);
```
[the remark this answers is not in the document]
EndymionJkb: See the comment above about copying this pattern from Facu. I agree we should decide how we want to do these repeated things and make them all the same.

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -59,73 +85,343 @@ contract LiquidityBootstrappingPool is BaseWeightedPool, ReentrancyGuard {
...
+
+    function _onInitializePool(
+        bytes32 poolId,
+        address sender,
+        address recipient,
+        uint256[] memory scalingFactors,
+        bytes memory userData
+    ) internal override whenNotPaused returns (uint256, uint256[] memory) {
+        // Only the owner can initialize the pool
+        _require(sender == getOwner(), Errors.SENDER_NOT_ALLOWED);
```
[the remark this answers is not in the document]
EndymionJkb: There was already a CALLER_IS_NOT_OWNER. Is that ok?

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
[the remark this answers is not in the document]
nventuro: Sure, but imagine you develop an appliation using events, or e.g. do a subgraph. You wouldn't be able to compute the weights from events alone because there is none for the initial weights: you'd require some initial state. 

This is why we emit events in the constructor for all of these things, it makes those applications much easier to develop, and the cost of emitting the events/writing to storage are negligible compared to deploying the Pool. Same with setting the start and end times: construction is equivalent to a zero-duration weight change, and making it be that way means there is no edge-case of 'zero start time' that needs to be handled.

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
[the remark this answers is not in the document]
nventuro: Sure, but this is an external view function, likely will never be called on-chain - I'd favor readability.

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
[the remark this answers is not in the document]
nventuro: I'd follow the same patterns used in e.g. `BasePool`.

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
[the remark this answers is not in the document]
nventuro: No, that's an `Ownable` error. This is something unique and different - attempting to join an LBP from a non-owner account.

pkg/pool-weighted/test/LiquidityBootstrappingPool.test.ts
```diff
@@ -1,2 +1,221 @@
...
+    const params = { tokens, weights, owner, lbp: true, swapEnabledOnStart: true };
+    pool = await WeightedPool.create(params);
+  });
+
+  describe('weights', () => {
+    it('sets token weights', async () => {
+      const normalizedWeights = await pool.getNormalizedWeights();
+
+      // Need to decrease precision
+      expect(normalizedWeights).to.equalWithError(pool.normalizedWeights, 0.0001);
```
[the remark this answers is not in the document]
EndymionJkb: They don't

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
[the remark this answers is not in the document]
EndymionJkb: Yes - this plus refactoring the initialize weight change thing makes it very clean!

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -59,73 +88,315 @@ contract LiquidityBootstrappingPool is BaseWeightedPool, ReentrancyGuard {
...
+        // only 10% of the period in the future, the weights would immediately jump 90%
+        uint256 currentTime = block.timestamp;
+        startTime = currentTime > startTime ? currentTime : startTime;
 
-        // Initial value; can change later if weights change
-        _maxWeightTokenIndex = maxWeightTokenIndex;
+        _require(startTime <= endTime, Errors.GRADUAL_UPDATE_TIME_TRAVEL);
+
+        // Cannot call _getNormalizedWeights() inside _startGradualWeightChange, because it reads from immutable storage
+        // and is also called from the constructor
```
nventuro, line 166:
```suggestion
```
Ah, interesting! But I don't think we need to explain this to the reader

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -59,73 +88,315 @@ contract LiquidityBootstrappingPool is BaseWeightedPool, ReentrancyGuard {
...
-    // Assumes i is in range
-    function _getNormalizedWeight(uint8 i) private view returns (uint256) {
-        return _normalizedWeights.decodeUint64(i * 64);
+    function _interpolateWeight(
+        uint256 startWeight,
+        uint256 endWeight,
+        uint256 pctProgress
+    ) private pure returns (uint256 finalWeight) {
+        if (pctProgress == 0) return startWeight;
+        if (pctProgress == FixedPoint.ONE) return endWeight;
```
nventuro, line 384:
```suggestion
        if (pctProgress >= FixedPoint.ONE) return endWeight;
```

Shouldn't be necessary, but will prevent ever returning values larger than `endWeight`.

pkg/pool-weighted/test/LiquidityBootstrappingPool.test.ts
```diff
@@ -17,275 +17,319 @@ describe('LiquidityBootstrappingPool', function () {
...
+            expect(bptAfterExit).to.lt(bptAfterJoin);
+          });
+
+          describe('update weights gradually', () => {
+            const UPDATE_DURATION = MINUTE * 60;
+      
+            sharedBeforeEach('grant permission', async () => {
+              const action = await actionId(pool.instance, 'updateWeightsGradually');
+              await pool.vault.grantRole(action, owner);
+            });
```
nventuro, line 161: I think we don't need this here, as this is `onlyOwner`. Also, given that the entire only owner / delegated owner flow is so common in all pools, I'd try to eventually find a more concise way of testing that independently - let's not worry about it here.

pkg/pool-weighted/test/LiquidityBootstrappingPool.test.ts
```diff
@@ -17,275 +17,319 @@ describe('LiquidityBootstrappingPool', function () {
...
-        it('gets intermediate weights if called 80% through', async () => {
-          await advanceTime(START_DELAY + UPDATE_DURATION * 0.8);
-          const normalizedWeights = await pool.getNormalizedWeights();
-
-          // Need to decrease precision
-          expect(normalizedWeights).to.equalWithError(fourFifthWeights, 0.001);
-        });
+        it('anyone can query the weight update params', async () => {
+          await expect(pool.getGradualWeightUpdateParams(sender)).to.not.be.reverted;
+        })
```
nventuro: I'd remove the tests for the `view` functions, it is expected that they don't depend on the caller.

nventuro (approved):
The updated tests look _g r e a t_ :tada:

nventuro (approved):

PR #631, opened 2021-06-14
Add an upper critical level for rebalancing

TomAFrench:
Adds an upper critical level to ensure that AM is rebalanced when overinvested without having to wait for an arb opportunity.

Tests have been changed only as much to get them to pass again. I have another PR which adds full tests for this and improves handling of rebalancing tests

### DIFF
--- a/pkg/asset-manager-utils/test/AaveATokenAssetManager.test.ts
+++ b/pkg/asset-manager-utils/test/AaveATokenAssetManager.test.ts
@@ -172,7 +172,7 @@ describe('Aave Asset manager', function () {
     it('reverts when setting upper critical over 100%', async () => {
       const badPoolConfig = {
         targetPercentage: 0,
-        upperCriticalPercentage: fp(1.1),
+        upperCriticalPercentage: fp(1).add(1),
         lowerCriticalPercentage: 0,
         feePercentage: 0,
       };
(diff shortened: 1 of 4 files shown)


pkg/asset-manager-utils/test/AaveATokenAssetManager.test.ts
```diff
@@ -154,33 +154,66 @@ describe('Aave Asset manager', function () {
...
       );
     });
 
     it('reverts when setting fee percentage over 100%', async () => {
-      const badPoolConfig = { targetPercentage: 0, criticalPercentage: 0, feePercentage: fp(1.01) };
+      const badPoolConfig = {
+        targetPercentage: 0,
+        upperCriticalPercentage: 0,
+        lowerCriticalPercentage: 0,
+        feePercentage: fp(1.01),
```
nventuro, line 213:
```suggestion
        feePercentage: fp(0.1).add(1),
```

1.01 would be a fee percentage of 101%, not 10%

pkg/asset-manager-utils/test/RewardsAssetManager.test.ts
```diff
@@ -121,33 +121,66 @@ describe('Rewards Asset manager', function () {
...
       );
     });
 
     it('reverts when setting fee percentage over 100%', async () => {
-      const badPoolConfig = { targetPercentage: 0, criticalPercentage: 0, feePercentage: fp(1.01) };
+      const badPoolConfig = {
+        targetPercentage: 0,
+        upperCriticalPercentage: 0,
+        lowerCriticalPercentage: 0,
+        feePercentage: fp(1.01),
```
nventuro, line 180: Same as the other comment (as these tests are duplicated :stuck_out_tongue:)

nventuro (approved):

Issue #623, opened 2021-06-09
Prevent sandwich attacks on AM rebalance fee


nventuro:
It is possible to atomically perform a swap that unbalances an Asset Manager, rebalance it, and then undo the swap, potentially profiting from the reward.

We should use `lastBlockChange` to only pay out rewards if the rebalancing interaction is the first one in the transaction, thereby avoiding this type of attack.

nventuro:
@TomAFrench and I discussed this in light of the recent decision to remove rebalance rewards, and decided a) we don't need this if no rewards are at stake, and b) it may cause issues with #604.

Issue #604, opened 2021-06-02
Create Rebalancing Relayer


nventuro:
https://github.com/balancer-labs/balancer-v2-monorepo/issues/603 will restrict Pools to be joined/exited by a single relayer. This relayer will be the contract used by LPs to actually join and exit Pools that are restricted this way (Pools with auto-balancing Asset Managers).

At the very least, we'll need:

on join:
 - call vault.joinPool, with `msg.sender` as the sender (this requires relayer approval or signature) and self as the recipient
 - transfer received BPT to `msg.sender`
 - query the Asset Manager of each of the Pool's tokens, and call `rebalance` on non-zero managers

on exit:
 - transfer tokens from `msg.sender` into self (this requires allowance, which can be obtained via `permit`)
 - call vault.exitPool, with self as the sender and `msg.sender` as the recipient
 - query the Asset Manager of each of the Pool's tokens, and call `rebalance` on non-zero managers

An additional consideration is that #603 doesn't restrict users from joining Pools directly with no relayer, sending the BPT to them. We must be careful to ensure the relayer works on BPT balance deltas and doesn't use absolute balance, to prevent this from becoming a way to sidestep this mechanism.

These features can be improved in a number of ways, but seem like a sensible starting place. Potential improvements that would be nice to have include:
 - ask for a signature from the LP on the first join ('Pool onboarding') and use it to approve the relayer in the same join transaction
 - use `permit` to grant allowance in the first exit (or maybe on the first join to have the entire 'onboarding' happen at once)
 - improve `rebalance` so that it only actually rebalances if the current percentage is far from the target (but don't let it get too close to the critical thresholds)

nventuro:
As mentioned in [this comment](https://github.com/balancer-labs/balancer-v2-monorepo/issues/603#issuecomment-859044980), it should be possible to come up with a simpler mechanism where we can avoid unnecessary BPT transfers from relayer to the LP.

Additionally, using this mechanism means that in the case of exits we actually _don't_ need to use `permit`, allowances or anything of the sort. The Pool burns the BPT of `sender`, and the Vault checks that either `sender == msg.sender` or `msg.sender` is a relayer for `sender`, meaning the relayer is allowed to call 'exitPool' as the user, using their BPT.

PR #632, opened 2021-06-14
Remove AssetManager rebalance rewards

nventuro:
As previously discussed, we're going to rely on #604 for the most part to keep Asset Managers close to the target balance, and expect manual rebalancing to only be required infrequently. Dropping this feature removes a large potential attack vector.

I've kept this PR as minimal as possible so that, if we were to decide to revisit this approach in the future, it'll be simple to see what it entailed. 

_Note to future people: if you _do_ revive this, don't forget to also revive #623!_


nventuro:
Closing in favor of #640

PR #634, opened 2021-06-15
Offchain fees calculation

TomAFrench:
This PR essentially just does some standardisation of calculating rebalance amounts + fees which makes it easier to refactor into a `itRebalancesCorrectly` set of tests in future.

The only change that would survive #632 would be exposing a public view function to return `_getPoolBalances`.

### DIFF
--- a/pkg/asset-manager-utils/test/AaveATokenAssetManager.test.ts
+++ b/pkg/asset-manager-utils/test/AaveATokenAssetManager.test.ts
@@ -359,7 +359,7 @@ describe('Aave Asset manager', function () {
         const amountToWithdraw = maxInvestableBalance.abs();
 
         await expectBalanceChange(() => assetManager.connect(lp).capitalOut(poolId, amountToWithdraw), tokens, [
-          { account: lendingPool.address, changes: { DAI: ['near', -amountToWithdraw] } },
+          { account: lendingPool.address, changes: { DAI: ['near', amountToWithdraw.mul(-1)] } },
           { account: vault.address, changes: { DAI: ['near', amountToWithdraw] } },
         ]);
       });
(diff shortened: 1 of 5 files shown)


pkg/asset-manager-utils/test/AaveATokenAssetManager.test.ts
```diff
@@ -491,14 +492,12 @@ describe('Aave Asset manager', function () {
...
-        const targetInvestmentAmount = poolTVL.mul(poolConfig.targetPercentage).div(fp(1));
-        const expectedRebalanceAmount = managed.sub(targetInvestmentAmount);
+        const { poolCash, poolManaged } = await assetManager.getPoolBalances(poolId);
+        const expectedRebalanceAmount = calcRebalanceAmount(poolCash, poolManaged, poolConfig);
 
         await expectBalanceChange(() => assetManager.rebalance(poolId), tokens, [
-          { account: lendingPool.address, changes: { DAI: ['very-near', -expectedRebalanceAmount] } },
-          { account: vault.address, changes: { DAI: ['very-near', expectedRebalanceAmount] } },
+          { account: lendingPool.address, changes: { DAI: ['very-near', expectedRebalanceAmount] } },
+          { account: vault.address, changes: { DAI: ['very-near', -expectedRebalanceAmount] } },
```
nventuro, line 500:
```suggestion
          { account: vault.address, changes: { DAI: ['very-near', expectedRebalanceAmount,mul(-1)] } },
```

This only works for small BNs.

pkg/asset-manager-utils/test/RewardsAssetManager.test.ts
```diff
@@ -413,22 +451,16 @@ describe('Rewards Asset manager', function () {
...
-        };
         sharedBeforeEach(async () => {
-          poolController = lp; // TODO
+          const poolController = lp; // TODO
 
           await assetManager.connect(poolController).setPoolConfig(poolId, poolConfig);
-          targetInvestmentAmount = await assetManager.maxInvestableBalance(poolId);
         });
 
         it('returns the expected fee', async () => {
```
nventuro, line 460: Not sure what sort of refactor you have in mind, but for these I'd consider not having a full `describe` just for `getRebalanceFee` (with their complex setup and difference scenarios), but instead add these as an `it` in the multiple contexts of `describe('rebalance')`, e.g. `it('returns the expected fee'); it('emits an event');`, etc.

nventuro (approved):
Looks very good, thanks!

pkg/asset-manager-utils/test/AaveATokenAssetManager.test.ts
nventuro: There's a few more instances of this

Issue #603, opened 2021-06-02
Create BasePool that restricts joiners and exiters


nventuro:
In order to have Pools that auto-rebalance their Asset Managers on joins and exits, we will have a relayer that rebalances on these actions. It _has_ to be a relayer, as the Pool hooks are executed within the Vault's execution, which is non-reentrant.

As we lack a way to query which account triggered the Vault call from the Pool hook, we will rely on `sender` and `recipient`: joins will only be accepted if the `recipient` is this relayer contract, and exits only accepted if the `sender` is the relayer. To begin with, we could receive the relayer as a constructor argument and have it be `immutable`.

nventuro:
@facuspagnuolo proposed an alternative approach that seems simpler and might even lead to reduced gas costs. Instead of imposing conditions on `sender` and `recipient`, we perform a view call to the relayer that must return a magic value:

```
// pool

fn onJoin/onExit() override {
  require(relayer.isJoiningOrExiting());
  super.onJoin/onExit();
}

// relayer

fn join/exit() {
  _isJoiningOrExiting = true;
  _doJoinOrExit();
    _isJoiningOrExiting = false;
}

fn isJoiningOrExiting() {
  return _isJoiningOrExiting;
}
```

Because of [EIP 2200](https://eips.ethereum.org/EIPS/eip-2200), resetting a storage slot to its original value is inexpensive, and because of [EIP 2929}(https://eips.ethereum.org/EIPS/eip-2929) calling back into the relayer and reading `_isJoiningOrExiting` is also inexpensive. This may be cheaper than forcing the relayer to perfrom extra ERC20 transfers.

It should be noted that [EIP 3529](https://eips.ethereum.org/EIPS/eip-3529#rationale) (scheduled for London) reduces refunds, but not for this specific use case, which are scenarios 2 and 3 of the 'rationale' section.

>Of these three, only (1) enables gastokens (...). This EIP deals with case (1).

PR #635, opened 2021-06-15
Pools: Mark external virtual functions as public

facuspagnuolo:
External functions cannot be called using `super` if we want to extend it, using `public` makes it possible. Required for #603

### DIFF
--- a/pkg/pool-utils/contracts/BasePool.sol
+++ b/pkg/pool-utils/contracts/BasePool.sol
@@ -173,7 +173,7 @@ abstract contract BasePool is IBasePool, BasePoolAuthorization, BalancerPoolToke
         return _swapFeePercentage;
     }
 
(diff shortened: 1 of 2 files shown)


nventuro:
I think the errors are a bug in the compiler. I'll ask about it.

Meanwhile, doing `this.foo.selector` seems to serve as a workaround.

nventuro (approved):
Dicussed this with the Solidity team, seems like this is the way to go:

> that's not really a hack. At least the way it was designed. Would be convenient if foo.selector worked where possible too but selectors are only guaranteed to be available on stuff that is external so it's "more correct" to do this.foo.selector.

If we were to use it in a constructor (which we might to initialize `immutable` variables), we'd run into a warning about using `this`, but it'd be a false positive: https://github.com/ethereum/solidity/issues/11540

PR #636, opened 2021-06-15
Pools: Implement relayed base pool

### LINKED ISSUE #603: Create BasePool that restricts joiners and exiters

facuspagnuolo:
Fixes #603


pkg/pool-utils/contracts/RelayedBasePool.sol
```diff
@@ -0,0 +1,71 @@
...
+import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/Address.sol";
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/BalancerErrors.sol";
+
+import "./BasePool.sol";
+import "./interfaces/IBasePoolRelayer.sol";
+
+/**
+ * @dev Base Pool associated with a relayer that guarantees it can only be joined/exited from the relayer itself.
+ * This contract as a simple mixin for pools. Implementing pools must make sure to call the BasePool's constructor
+ * properly, and validate that the given
```
nventuro, line 27: incomplete sentence?

pkg/pool-utils/contracts/RelayedBasePool.sol
```diff
@@ -0,0 +1,71 @@
...
+
+    modifier ensureRelayer(bytes32 poolId) {
+        // Note that this modifier is relying on an input parameter, which is usually given by the Vault which is
+        // considered a trusted party. It's up to the implementing pool whether to validate this value or not.
+        _require(relayer.hasCalledPool(poolId), Errors.BASE_POOL_RELAYER_NOT_CALLED);
+        _;
+    }
+
+    constructor(IBasePoolRelayer _relayer) {
+        _require(address(_relayer).isContract(), Errors.RELAYER_NOT_CONTRACT);
```
nventuro, line 42: Is this just to prevent accidental misuse?

pkg/pool-utils/contracts/RelayedBasePool.sol
```diff
@@ -0,0 +1,71 @@
...
+ * properly, and validate that the given
+ */
+abstract contract RelayedBasePool is BasePool {
+    using Address for address;
+
+    IBasePoolRelayer public immutable relayer;
+
+    modifier ensureRelayer(bytes32 poolId) {
+        // Note that this modifier is relying on an input parameter, which is usually given by the Vault which is
+        // considered a trusted party. It's up to the implementing pool whether to validate this value or not.
```
nventuro, line 36: I get the intent behind this comment, but it is not entirely accurate. We're inheriting from `BasePool` here and calling its implementation, which _does_ validate `poolId`. I'd consider removing it.

pkg/pool-utils/contracts/RelayedBasePool.sol
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * @dev Base Pool associated with a relayer that guarantees it can only be joined/exited from the relayer itself.
+ * This contract as a simple mixin for pools. Implementing pools must make sure to call the BasePool's constructor
+ * properly, and validate that the given
+ */
+abstract contract RelayedBasePool is BasePool {
+    using Address for address;
+
+    IBasePoolRelayer public immutable relayer;
```
nventuro, line 32: All other getters start with `get` e.g. `getPoolId()`. I don't personally mind either having or not having it, but given those alredy exist I'd be consistent with them.

pkg/pool-utils/contracts/RelayedBasePool.sol
```diff
@@ -0,0 +1,71 @@
...
+
+    modifier ensureRelayer(bytes32 poolId) {
```
nventuro, line 34:
```suggestion
    modifier ensureRelayerCall(bytes32 poolId) {
```

nventuro: `ensureRelayer(poolId)` seemed a bit too cryptic

nventuro (approved):
Lovely, thanks a bunch!

PR #637, opened 2021-06-15
Assets manager: Implement rebalancing relayer

### LINKED ISSUE #604: Create Rebalancing Relayer

facuspagnuolo:
Fixes #604 
Depends on #636


pkg/asset-manager-utils/contracts/RebalancingRelayer.sol
```diff
@@ -0,0 +1,62 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
+
+import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
+import "@balancer-labs/v2-pool-utils/contracts/interfaces/IBasePoolRelayer.sol";
```
nventuro, line 19: This name doesn't really ilustrate what is going on, but I'm having trouble coming up with a better one :/

pkg/asset-manager-utils/contracts/RebalancingRelayer.sol
```diff
@@ -0,0 +1,62 @@
...
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/BalancerErrors.sol";
+
+import "./IAssetManager.sol";
+
+contract RebalancingRelayer is IBasePoolRelayer {
+    bytes32 constant internal _EMPTY_CALLED_POOL = bytes32(0x0000000000000000000000000000000000000000000000000000000000000001);
```
nventuro, line 25:
```suggestion
    // We start at a non-zero value to make EIP2200 refunds lower, meaning there'll be a higher chance of them being
    // fully effective.
    bytes32 constant internal _EMPTY_CALLED_POOL = bytes32(0x0000000000000000000000000000000000000000000000000000000000000001);
```

pkg/asset-manager-utils/contracts/RebalancingRelayer.sol
```diff
@@ -0,0 +1,62 @@
...
+    function hasCalledPool(bytes32 poolId) external view override returns (bool) {
+        return _calledPool == poolId;
+    }
+
+    function joinPool(bytes32 poolId, IVault.JoinPoolRequest memory request) external payable rebalance(poolId) {
+        vault.joinPool(poolId, msg.sender, msg.sender, request);
+    }
+
+    function exitPool(bytes32 poolId, IVault.ExitPoolRequest memory request) external rebalance(poolId) {
+        vault.exitPool(poolId, msg.sender, msg.sender, request);
```
nventuro, line 52: The sender sure, but could we not accept `recipient` as an argument?

pkg/asset-manager-utils/contracts/RebalancingRelayer.sol
```diff
@@ -0,0 +1,62 @@
...
+        vault = _vault;
+        _calledPool = _EMPTY_CALLED_POOL;
+    }
+
+    function hasCalledPool(bytes32 poolId) external view override returns (bool) {
+        return _calledPool == poolId;
+    }
+
+    function joinPool(bytes32 poolId, IVault.JoinPoolRequest memory request) external payable rebalance(poolId) {
+        vault.joinPool(poolId, msg.sender, msg.sender, request);
```
nventuro, line 48: Shouldn't this be `joinPool{ value: msg.value }(...args)`?

pkg/asset-manager-utils/contracts/RebalancingRelayer.sol
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    function exitPool(bytes32 poolId, IVault.ExitPoolRequest memory request) external rebalance(poolId) {
+        vault.exitPool(poolId, msg.sender, msg.sender, request);
+    }
+
+    function _rebalance(bytes32 poolId) internal {
+        (IERC20[] memory tokens, , ) = vault.getPoolTokens(poolId);
```
nventuro, line 56: We could skip this as both join and exit pool requests have an `assets` field with all of the tokens: we could pass that to `rebalance`.

pkg/asset-manager-utils/contracts/RebalancingRelayer.sol
```diff
@@ -0,0 +1,62 @@
...
+        for (uint256 i = 0; i < tokens.length; i++) {
+            (,,, address assetManager) = vault.getPoolTokenInfo(poolId, tokens[i]);
+            IAssetManager(assetManager).rebalance(poolId);
```
nventuro, line 59:
```suggestion
            if (assetManager != address(0)) {
                IAssetManager(assetManager).rebalance(poolId);
            }
```

pkg/asset-manager-utils/test/RebalancingRelayer.test.ts
```diff
@@ -0,0 +1,234 @@
...
+import { actionId } from '@balancer-labs/v2-helpers/src/models/misc/actions';
+import { GeneralPool } from '@balancer-labs/v2-helpers/src/models/vault/pools';
+import { MAX_UINT256, ZERO_ADDRESS } from '@balancer-labs/v2-helpers/src/constants';
+import { BigNumberish, fp } from '@balancer-labs/v2-helpers/src/numbers';
+import {
+  encodeExitWeightedPool,
+  encodeJoinWeightedPool,
+} from '@balancer-labs/v2-helpers/src/models/pools/weighted/encoding';
+
+describe.only('RebalancingRelayer', function () {
```
nventuro, line 19:
```suggestion
describe('RebalancingRelayer', function () {
```

pkg/asset-manager-utils/test/RebalancingRelayer.test.ts
```diff
@@ -0,0 +1,234 @@
...
+    tokens = await TokenList.create(['DAI', 'MKR'], { sorted: true });
+    assetManager = await deploy('MockAssetManager');
+    pool = await deploy('v2-pool-utils/MockRelayedBasePool', {
+      args: [
+        vault.address,
+        GeneralPool,
+        'BPT',
+        'BPT',
+        tokens.addresses,
+        Array(tokens.length).fill(assetManager.address),
```
nventuro, line 43: Let's have at least two different asset managers

pkg/asset-manager-utils/test/RebalancingRelayer.test.ts
```diff
@@ -0,0 +1,234 @@
...
+
+          it('reverts', async () => {
+            await expect(relayer.connect(user).joinPool(poolId, request)).to.be.revertedWith(
+              'USER_DOESNT_ALLOW_RELAYER'
+            );
+          });
+        });
+
+        context('when the relayer is allowed to join', () => {
+          sharedBeforeEach('allow relayer', async () => {
```
nventuro, line 117:
```suggestion
        context('when the relayer is not allowed to join', () => {
          sharedBeforeEach('revoke relayer', async () => {
```

pkg/asset-manager-utils/test/RebalancingRelayer.test.ts
```diff
@@ -0,0 +1,234 @@
...
+        context('when the user did allow the relayer', () => {
+          sharedBeforeEach('allow relayer', async () => {
+            await vault.connect(user).setRelayerApproval(user.address, relayer.address, true);
+          });
+
+          it('joins the pool', async () => {
+            const previousUserBalance = await pool.balanceOf(user.address);
+            const previousRelayerBalance = await pool.balanceOf(relayer.address);
+
+            await relayer.connect(user).joinPool(poolId, request);
```
nventuro, line 89: What do you think about emitting an event in the pool with the join arguments it received? Overkill?

pkg/asset-manager-utils/test/RebalancingRelayer.test.ts
```diff
@@ -0,0 +1,234 @@
...
+
+          it('reverts', async () => {
+            await expect(relayer.connect(user).exitPool(poolId, exitRequest)).to.be.revertedWith(
+              'USER_DOESNT_ALLOW_RELAYER'
+            );
+          });
+        });
+
+        context('when the relayer is allowed to exit', () => {
+          sharedBeforeEach('allow relayer', async () => {
```
nventuro, line 214:
```suggestion
        context('when the relayer is not allowed to exit', () => {
          sharedBeforeEach('revoke relayer', async () => {
```

pkg/asset-manager-utils/test/RebalancingRelayer.test.ts
```diff
@@ -0,0 +1,284 @@
...
+            const receipt = await relayer.connect(sender).joinPool(poolId, recipient.address, request);
+
+            expectEvent.inIndirectReceipt(await receipt.wait(), assetManagers[0].interface, 'Rebalanced', {
+              poolId,
+              assetManager: assetManagers[0].address,
+              token: tokens.first.address,
+              force: false,
+            });
+
+            expectEvent.inIndirectReceipt(await receipt.wait(), assetManagers[0].interface, 'Rebalanced', {
```
nventuro, line 125:
```suggestion
            expectEvent.inIndirectReceipt(await receipt.wait(), assetManagers[1].interface, 'Rebalanced', {
```

Really the same, but slightly more correct :)

pkg/asset-manager-utils/contracts/RebalancingRelayer.sol
```diff
@@ -0,0 +1,79 @@
...
+    function hasCalledPool(bytes32 poolId) external view override returns (bool) {
+        return _calledPool == poolId;
+    }
+
+    function joinPool(
+        bytes32 poolId,
+        address recipient,
+        IVault.JoinPoolRequest memory request
+    ) external payable rebalance(poolId, request.assets) {
+        vault.joinPool{ value: msg.value }(poolId, msg.sender, recipient, request);
```
nventuro, line 59: We missed something here: from the Vault docs
```
// ETH can be used by passing the ETH sentinel value as the asset and forwarding ETH in the call: it will be wrapped
// and deposited as WETH. Any ETH amount remaining will be sent back to the caller (not the sender, which is
// relevant for relayers).
```

Any leftover ETH needs to be sent back to the sender. Let's also add a test for joins and exits with ETH.

pkg/asset-manager-utils/test/RebalancingRelayer.test.ts
```diff
@@ -0,0 +1,284 @@
...
+
+          it('rebalances the pool', async () => {
+            const receipt = await relayer.connect(sender).exitPool(poolId, recipient.address, exitRequest);
+
+            expectEvent.inIndirectReceipt(await receipt.wait(), assetManagers[0].interface, 'Rebalanced', {
+              poolId,
+              token: tokens.first.address,
+            });
+
+            expectEvent.inIndirectReceipt(await receipt.wait(), assetManagers[0].interface, 'Rebalanced', {
```
nventuro, line 242:
```suggestion
            expectEvent.inIndirectReceipt(await receipt.wait(), assetManagers[1].interface, 'Rebalanced', {
```

pkg/asset-manager-utils/contracts/RebalancingRelayer.sol
```diff
@@ -51,12 +54,23 @@ contract RebalancingRelayer is IBasePoolRelayer, AssetHelpers {
         return _calledPool == poolId;
     }
 
+    receive() external payable {
+        // Accept ETH transfers only coming from the Vault. This is only expected to happen when joining a pool,
+        // any remaining ETH value will be transferred back to this contract and forwarded back to the original sender.
+        _require(msg.sender == address(vault), Errors.ETH_TRANSFER);
+    }
```
nventuro, line 61: This can actually also happen in an exit if the pool holds weth. Because the vault sends eth back to the caller (not `sender`), we'd need to have the same `if balance(this) > 0` check in `exitPool`.

pkg/asset-manager-utils/contracts/RebalancingRelayer.sol
```diff
@@ -0,0 +1,79 @@
...
+
+contract RebalancingRelayer is IBasePoolRelayer, AssetHelpers {
+    // We start at a non-zero value to make EIP2200 refunds lower, meaning there'll be a higher chance of them being
+    // fully effective.
+    bytes32 internal constant _EMPTY_CALLED_POOL = bytes32(
+        0x0000000000000000000000000000000000000000000000000000000000000001
+    );
+
+    modifier rebalance(bytes32 poolId, IAsset[] memory assets) {
+        _require(_calledPool == _EMPTY_CALLED_POOL, Errors.REBALANCING_RELAYER_REENTERED);
```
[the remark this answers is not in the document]
facuspagnuolo: Exactly

nventuro (approved):

PR #638, opened 2021-06-15
prevent possible underflow in capitalOut

TomAFrench:
Previously we would deposit funds back into the vault and then update the pool's managed balance with the new AUM after the withdrawal. This can result in an issue where if the AM gets a large amount of gains such that it's depositing more than the current managed balance as recorded on the Vault then it can result in an underflow.

I've swapped these two operations so that we would update that it's aware of the new higher managed balance before performing the deposit.

### DIFF
--- a/pkg/asset-manager-utils/test/RewardsAssetManager.test.ts
+++ b/pkg/asset-manager-utils/test/RewardsAssetManager.test.ts
@@ -334,8 +334,8 @@ describe('Rewards Asset manager', function () {
         // await assetManager.connect(poolController).setInvestablePercent(poolId, fp(0));
 
         await expectBalanceChange(() => assetManager.connect(lp).capitalOut(poolId, amountToWithdraw), tokens, [
-          { account: assetManager.address, changes: { DAI: ['very-near', amountToWithdraw.mul(-1)] } },
-          { account: vault.address, changes: { DAI: ['very-near', amountToWithdraw] } },
+          { account: assetManager.address, changes: { DAI: amountToWithdraw.mul(-1) } },
+          { account: vault.address, changes: { DAI: amountToWithdraw } },
         ]);
       });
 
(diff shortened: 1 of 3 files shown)


nventuro (approved):
Great catch! Yes, updates should always happen before deposits or withdrawals. 

Can you add a simple test for this?

TomAFrench:
I've added a quick test for this which isn't ideal but would be fixed up in #639

pkg/asset-manager-utils/test/RewardsAssetManager.test.ts
```diff
@@ -363,6 +363,18 @@ describe('Rewards Asset manager', function () {
...
       });
+
+      it('allows withdrawing returns which are greater than the current managed balance', async () => {
+        await tokens.DAI.mint(assetManager.address, tokenInitialBalance.mul(10));
+        await assetManager.connect(lp).setUnrealisedAUM(amountToDeposit.add(tokenInitialBalance.mul(10)));
+
+        const amountToWithdraw = (await assetManager.maxInvestableBalance(poolId)).mul(-1);
+
+        await expectBalanceChange(() => assetManager.connect(lp).capitalOut(poolId, amountToWithdraw), tokens, [
+          { account: assetManager.address, changes: { DAI: ['very-near', -amountToWithdraw] } },
```
nventuro, line 374:
```suggestion
          { account: assetManager.address, changes: { DAI: ['very-near', amountToWithdraw.mul(-1)] } },
```

nventuro: Do we need 'very-near' here? Should it be exactly equal?

TomAFrench: Yes, I think we can make this a strict equality. `very-near` was necessary when we were using shares but in `RewardsAssetManager` we can remove this.

PR #639, opened 2021-06-15
Remove use of capitalIn from test setup

TomAFrench:
Use of `capitalIn` has been removed from the setup for tests in `RewardsAssetManager`. Instead, tokens are minted directly to the AM to increase its AUM.

`AaveATokenAssetManager` tests have not been updated to match.

### DIFF
--- a/pkg/asset-manager-utils/contracts/test/TestAssetManager.sol
+++ b/pkg/asset-manager-utils/contracts/test/TestAssetManager.sol
@@ -23,7 +23,6 @@ pragma solidity ^0.7.0;
 // solhint-disable private-vars-leading-underscore
 contract TestAssetManager is RewardsAssetManager {
     using Math for uint256;
-    uint256 public nextAUM;
 
     constructor(
         IVault _vault,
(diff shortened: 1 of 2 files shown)


TomAFrench:
Ah, the test failures are due to me separating this from #638. I can rebase this once that is merged and it should go green.

pkg/asset-manager-utils/contracts/test/TestAssetManager.sol
```diff
@@ -43,31 +42,22 @@ contract TestAssetManager is RewardsAssetManager {
...
     function readAUM() public view override returns (uint256) {
-        return nextAUM;
-    }
-
-    /**
-     * @dev Sets the next value to be read by `readAUM`
-     */
-    function setUnrealisedAUM(uint256 aum) external {
-        nextAUM = aum;
+        return token.balanceOf(address(this));
```
nventuro, line 61: This is fantastic, much more straightforward to understand.

If we add a `burn` function to `TestToken`, we can do the same thing for `capitalOut`.

nventuro (approved):

PR #641, opened 2021-06-16
Soft rebalance

TomAFrench:
This PR adds an initial implementation of "soft" rebalances to prevent the rebalance relayer using extra gas when not necessary.

The critical levels which were used for fees are now used to determine whether a "soft" rebalance should occur or not.

### DIFF
--- a/pkg/asset-manager-utils/contracts/IAssetManager.sol
+++ b/pkg/asset-manager-utils/contracts/IAssetManager.sol
@@ -73,8 +73,14 @@ interface IAssetManager {
 
     /**
      * @notice Rebalances funds between the pool and the asset manager to maintain target investment percentage.
-     * If the pool is below its critical threshold for the amount invested then calling this will send a small reward
-     * to the sender
+     * @param poolId - the poolId of the pool to be rebalanced
      */
     function rebalance(bytes32 poolId) external;
+
+    /**
+     * @notice Rebalances funds between the pool and the asset manager to maintain target investment percentage.
+     * @param poolId - the poolId of the pool to be rebalanced
+     * @param force - a boolean representing whether a rebalance should be forced even when the pool is near balance
+     */
+    function rebalance(bytes32 poolId, bool force) external;
 }
--- a/pkg/asset-manager-utils/contracts/IAssetManager.sol
+++ b/pkg/asset-manager-utils/contracts/IAssetManager.sol
@@ -79,12 +79,6 @@ interface IAssetManager {
     /**
      * @notice Rebalances funds between the pool and the asset manager to maintain target investment percentage.
      * @param poolId - the poolId of the pool to be rebalanced
-     */
-    function rebalance(bytes32 poolId) external;
-
-    /**
-     * @notice Rebalances funds between the pool and the asset manager to maintain target investment percentage.
-     * @param poolId - the poolId of the pool to be rebalanced
      * @param force - a boolean representing whether a rebalance should be forced even when the pool is near balance
      */
     function rebalance(bytes32 poolId, bool force) external;
--- a/pkg/asset-manager-utils/contracts/IAssetManager.sol
+++ b/pkg/asset-manager-utils/contracts/IAssetManager.sol
@@ -23,6 +23,11 @@ interface IAssetManager {
     }
 
     /**
+     * @notice Emitted when asset manager is rebalanced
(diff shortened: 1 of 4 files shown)


pkg/asset-manager-utils/contracts/IAssetManager.sol
```diff
@@ -73,8 +78,14 @@ interface IAssetManager {
 
     /**
      * @notice Rebalances funds between the pool and the asset manager to maintain target investment percentage.
-     * If the pool is below its critical threshold for the amount invested then calling this will send a small reward
-     * to the sender
+     * @param poolId - the poolId of the pool to be rebalanced
      */
     function rebalance(bytes32 poolId) external;
```
nventuro, line 83: I'd consider dropping this: it is equivalent to `rebalance(pId, true)`, and having overloads is a pain in client code.

pkg/asset-manager-utils/test/AaveATokenAssetManager.test.ts
```diff
@@ -377,83 +378,140 @@ describe('Aave Asset manager', function () {
...
+        });
+
+        it('returns the pool to its target allocation', async () => {
+          await assetManager['rebalance(bytes32,bool)'](poolId, forceRebalance);
+          const differenceFromTarget = await assetManager.maxInvestableBalance(poolId);
+          expect(differenceFromTarget.abs()).to.be.lte(1);
+        });
+
+        it("updates the pool's cash and managed balances correctly");
+      } else {
```
nventuro, line 412: Let's split this into two functions. Then, instead of 

```
context when the pool is not in the critical zone
  context('when forced', () => {
    itRebalancesCorrectly(true, true);
  });

  context('when not forced', () => {
    itRebalancesCorrectly(false, false);
  });
```

we will have

```
context when the pool is not in the critical zone
  context('when forced', () => {
    itRebalancesCorrectly()
  });

  context('when not forced', () => {
    itSkipsTheRebalance()
  });
```

pkg/asset-manager-utils/test/AaveATokenAssetManager.test.ts
```diff
@@ -377,83 +378,140 @@ describe('Aave Asset manager', function () {
...
+            .simulateATokenIncrease(tokens.DAI.address, amountReturned, assetManager.address);
+        });
+
+        context('when forced', () => {
+          itRebalancesCorrectly(true, true);
+        });
+
+        context('when not forced', () => {
+          itRebalancesCorrectly(false, false);
+        });
```
nventuro, line 457:
```suggestion
        context('when forced', () => {
          const forced = true;
          itRebalancesCorrectly(true, forced);
        });

        context('when not forced', () => {
          const forced = false;
          itRebalancesCorrectly(false, forced);
        });
```

nventuro: (this test is quite simple, but this pattern is very handy when the setting up of the test is a bit more involved and e.g. the `= true` would happen inside a `beforeEach`

pkg/asset-manager-utils/test/RewardsAssetManager.test.ts
```diff
@@ -342,76 +325,128 @@ describe('Rewards Asset manager', function () {
...
+      let poolConfig: PoolConfig;
 
       sharedBeforeEach(async () => {
-        const poolController = lp; // TODO
-        await assetManager.connect(poolController).setPoolConfig(poolId, poolConfig);
-
-        const { poolCash } = await assetManager.getPoolBalances(poolId);
-        await tokens.DAI.mint(assetManager.address, poolCash.mul(101).div(100));
+        poolConfig = await assetManager.getPoolConfig(poolId);
+      });
```
nventuro, line 333: This is only used in the 'transfers the expected tokens' test, I'd move it there :shrug:

nventuro (approved):
Fantastic work, thanks a lot!

pkg/asset-manager-utils/test/RewardsAssetManager.test.ts
TomAFrench: True, I was thinking that we'd be reusing it in other tests in that block but it's not really worth the `sharedBeforeEach`

nventuro (approved):

PR #642, opened 2021-06-16
Migration to SplitCode factories

nventuro:
This migrates the WeightedPoolFactory and LBPFactory to the split code one introduced in #629. I also added a few simple tests to check that the constructor is being encoded correctly (as we no longer have `new`'s type safety), as well as bubbed-up revert reasons in the factory, which was a missing feature.

This let us fully remove `allowUnlimitedContractSize: true` from the entire codebase!


pkg/solidity-utils/contracts/helpers/BaseSplitCodeFactory.sol
```diff
@@ -177,7 +177,15 @@ abstract contract BaseSplitCodeFactory {
...
 
-        _require(destination != address(0), Errors.FACTORY_CONTRACT_DEPLOYMENT_FAILED);
+        if (destination == address(0)) {
+            // Bubble up inner revert reason
+            // solhint-disable-next-line no-inline-assembly
+            assembly {
+                returndatacopy(0, 0, returndatasize())
+                revert(0, returndatasize())
+            }
+        }
```
facuspagnuolo, line 187: 🙌

facuspagnuolo (approved):

PR #644, opened 2021-06-16
Refactor AM's readAUM()

nventuro:
This renames `readAUM()` to `getAUM()`, for consistency with all other getters (`getPoolID()`, `getVault()`, etc.).

It also removes `balanceOf` by adding the `poolId` parameter to `getAUM()`. This IMO also makes the code more understandable: it was quite weird that we were using `readAUM()` directly, and it hinged on the fact that the AM only controls one pool.

### DIFF
--- a/pkg/asset-manager-utils/contracts/IAssetManager.sol
+++ b/pkg/asset-manager-utils/contracts/IAssetManager.sol
@@ -34,7 +34,7 @@ interface IAssetManager {
     /**
      * @return the current assets under management of this asset manager
      */
-    function readAUM() external view returns (uint256);
+    function getAUM() external view returns (uint256);
 
     /**
      * @return poolCash - The up-to-date cash balance of the pool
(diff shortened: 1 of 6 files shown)


pkg/asset-manager-utils/contracts/IAssetManager.sol
```diff
@@ -37,15 +37,10 @@ interface IAssetManager {
...
-    /**
-     * @notice Returns the invested balance
-     */
-    function balanceOf(bytes32 poolId) external view returns (uint256);
-
     /**
      * @return the current assets under management of this asset manager
      */
-    function readAUM() external view returns (uint256);
+    function getAUM(bytes32 poolid) external view returns (uint256);
```
TomAFrench, line 43:
```suggestion
    function getAUM(bytes32 poolId) external view returns (uint256);
```
nitpick

TomAFrench (approved):
Looks ok to me.

PR #645, opened 2021-06-17
refactor: remove _maxInvestableBalance

TomAFrench:
I've removed `maxInvestableBalance` and reworked the tests to no longer rely on it as we can calculate it offchain. The name `calcRebalanceAmount` is a little overloaded now so I'm open to suggestions for something less tied to rebalancing.


nventuro:
I think this can be a useful function to keep, not just for tests but also simple sanity checks. I'll open an alternative PR that performs a slight refactor on the internals of RewardsAM which naturally exposes this function.

TomAFrench:
Closing as this is well out of data.

PR #646, opened 2021-06-17
Make capitalin/out internal

nventuro:
This, uh, deletes quite a bit of code.

It basically makes `capitalIn` and `capitalOut` internal, as they don't make much sense with the existence of `rebalance`. It also deletes their tests, since they were essentially duplicates of `rebalance`'s tests, and removes their usage in test setup.

Additionally, I removed both capital and rebalance tests from the AaveAssetManager tests. While duplicates of the basic AM tests, these had _some_ usefulness, but I think it makes more sense to write more focused ones in a PR after this one than to make these improvements now, since this change is quite unrelated.

### DIFF
--- a/pkg/asset-manager-utils/contracts/RewardsAssetManager.sol
+++ b/pkg/asset-manager-utils/contracts/RewardsAssetManager.sol
@@ -21,7 +21,6 @@ import "@balancer-labs/v2-solidity-utils/contracts/math/FixedPoint.sol";
 
 import "./IAssetManager.sol";
 
-import "hardhat/console.sol";
 
 pragma solidity ^0.7.0;
 pragma experimental ABIEncoderV2;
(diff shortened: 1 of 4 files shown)


pkg/asset-manager-utils/contracts/RewardsAssetManager.sol
```diff
@@ -21,6 +21,8 @@ import "@balancer-labs/v2-solidity-utils/contracts/math/FixedPoint.sol";
 
 import "./IAssetManager.sol";
 
+import "hardhat/console.sol";
```
TomAFrench, line 24: Leftover import

TomAFrench (approved):
Looks ok to me. I'm happy with stripping most of the tests for capitalIn/Out although one sticking point is that testing `rebalance` doesn't ensure that any methods in a derived contract can't push the AM past its target investment level.

nventuro:
Yup I agree with your comment, with these two functions being internal we can now do the slight refactor I mentioned above: it e.g. now doesn't make sense for both `rebalance` and `_capitalIn` to compute the percentages - I just wanted to not alter the contract too much here due to the large-ish test changes.

PR #647, opened 2021-06-18
Multiple rewarders per rewards token

gtaschuk:
LP can now claim rewards from multiple rewarders for pools where they have staked bpt.

For instance, a pool with multiple tokens lent out to aave would earn stkAAVE from multiple asset managers (1 per token).

Whenever an LP stakes or unstakes, the rewards they have earned thus far are calculated, to account for the rewards they earned with their historical bpt staked.  Since this has to happen on the full set of rewards for a pool (the set of {reward token, rewarder} combinations), we need to limit (whitelist) who can issue rewards, to avoid griefing attacks (from unbounded iteration).

This has no hard-coded upper limit on `O(rewardTokens*rewarders)`, running out of gas could prevent you from receiving any rewards for a pool (and only that pool)

**Rewarders*
- Any address the pool adds to a whitelist, as they can flexibly work in the LPs best interest
- The owner of the contract can whitelist rewarders so balancer governance can add rewards, including BAL and approved airdrops
- Any of the pool's asset managers in the pool - thinking that the asset managers are acting in the pool's best interests.

**s/stakingToken/pool**

Assume that every staking token is a pool.  The asset manager implicit whitelist logic relies on the staked tokens being bpt (when it calls `getPoolId()`), I renamed `stakingToken`s to `pool`s, which I think makes this contract much easier to understand.

This was designed around a use case of instant rewards (rewardDuration = 1s)

### DIFF
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -108,7 +108,7 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
         IERC20 pool,
         IERC20 rewardsToken,
         uint256 rewardsDuration
-    ) public onlyWhitelistedRewarder(pool, rewardsToken) override {
+    ) public override onlyWhitelistedRewarder(pool, rewardsToken) {
         require(rewardsDuration > 0, "reward rate must be nonzero");
         require(rewardData[pool][msg.sender][rewardsToken].rewardsDuration == 0, "Duplicate rewards token");
         _rewardTokens[pool].add(address(rewardsToken));
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -58,11 +58,11 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
     mapping(IERC20 => mapping(address => mapping(IERC20 => Reward))) public rewardData;
     mapping(IERC20 => EnumerableSet.AddressSet) private _rewardTokens;
 
-    // stakingToken -> rewardToken -> rewarders
+    // pool -> rewardToken -> rewarders
     mapping(IERC20 => mapping(IERC20 => EnumerableSet.AddressSet)) private _rewarders;
     mapping(IERC20 => mapping(IERC20 => EnumerableSet.AddressSet)) private _whitelist;
 
-    // stakingToken -> rewarder ->  user -> reward token -> amount
+    // pool -> rewarder ->  user -> reward token -> amount
     mapping(IERC20 => mapping(address => mapping(address => mapping(IERC20 => uint256)))) public userRewardPerTokenPaid;
     mapping(IERC20 => mapping(address => mapping(address => mapping(IERC20 => uint256)))) public rewards;
 
@@ -75,18 +75,15 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
         vault = _vault;
     }
 
-    modifier onlyWhitelistedRewarder(IERC20 stakingToken, IERC20 rewardsToken) {
+    modifier onlyWhitelistedRewarder(IERC20 pool, IERC20 rewardsToken) {
         require(
-            isWhitelistedToReward(stakingToken, rewardsToken, msg.sender),
+            isWhitelistedToReward(pool, rewardsToken, msg.sender),
             "only accessible by whitelisted rewarders and asset managers"
         );
         _;
     }
-    modifier onlyOwnerOrStakingToken(IERC20 stakingToken) {
-        require(
-            msg.sender == owner() || msg.sender == address(stakingToken),
-            "only accessible by the owner or the staking token"
-        );
+    modifier onlyOwnerOrPool(IERC20 pool) {
+        require(msg.sender == owner() || msg.sender == address(pool), "only accessible by the owner or the pool");
         _;
     }
 
@@ -94,29 +91,29 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
      * @notice Allows a rewarder to be explicitly added to a whitelist of rewarders
      */
     function whitelistRewarder(
-        IERC20 stakingToken,
+        IERC20 pool,
         IERC20 rewardsToken,
         address rewarder
-    ) public onlyOwnerOrStakingToken(stakingToken) {
-        _whitelist[stakingToken][rewardsToken].add(rewarder);
+    ) public onlyOwnerOrPool(pool) {
+        _whitelist[pool][rewardsToken].add(rewarder);
     }
 
     /**
      * @notice Adds a new reward token to be distributed
-     * @param stakingToken - The bpt of the pool that will receive rewards
+     * @param pool - The bpt of the pool that will receive rewards
      * @param rewardsToken - The new token to be distributed to stakers
      * @param rewardsDuration - The duration over which each distribution is spread
      */
     function addReward(
-        IERC20 stakingToken,
+        IERC20 pool,
         IERC20 rewardsToken,
         uint256 rewardsDuration
-    ) public onlyWhitelistedRewarder(stakingToken, rewardsToken) {
+    ) public onlyWhitelistedRewarder(pool, rewardsToken) {
         require(rewardsDuration > 0, "reward rate must be nonzero");
-        require(rewardData[stakingToken][msg.sender][rewardsToken].rewardsDuration == 0, "Duplicate rewards token");
-        _rewardTokens[stakingToken].add(address(rewardsToken));
-        _rewarders[stakingToken][rewardsToken].add(msg.sender);
-        rewardData[stakingToken][msg.sender][rewardsToken].rewardsDuration = rewardsDuration;
+        require(rewardData[pool][msg.sender][rewardsToken].rewardsDuration == 0, "Duplicate rewards token");
+        _rewardTokens[pool].add(address(rewardsToken));
+        _rewarders[pool][rewardsToken].add(msg.sender);
+        rewardData[pool][msg.sender][rewardsToken].rewardsDuration = rewardsDuration;
         rewardsToken.approve(address(vault), type(uint256).max);
     }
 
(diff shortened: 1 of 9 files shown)


nventuro:
Thanks for the detailed description!

Could you also briefly comment on the expected order of operations? e.g. owner calls `whiteListRewarder(rewarder)`, rewarder calls `foo()`, etc.

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -70,191 +75,294 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
...
+    /**
+     * @notice Calculates the total amount of `rewardsToken` that `account` is able to claim
+     */
+    function totalEarned(
+        IERC20 pool,
+        address account,
+        IERC20 rewardsToken
+    ) public view returns (uint256 total) {
+        for (uint256 r; r < _rewarders[pool][rewardsToken].length(); r++) {
+            total = total.add(earned(pool, _rewarders[pool][rewardsToken].at(r), account, rewardsToken));
```
nventuro, line 230:
```suggestion
        uint256 length = _rewarders[pool][rewardsToken].length();
        for (uint256 r; r < length; r++) {
            total = total.add(earned(pool, _rewarders[pool][rewardsToken].unchecked_at(r), account, rewardsToken));
```

This will perform much fewer storage accesses.

nventuro: Doing 

```
EnumerableSet storage set = _rewarders[pool][rewardsToken];
for (...) {
  total = foo(set.at(r));
}
```

likely also saves gas - I'd be surprised if solc saved the result of the mapping computation.

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -70,191 +75,294 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
...
-        IERC20 rewardsToken,
-        address rewardsDistributor
-    ) external onlyOwner {
-        rewardData[stakingToken][rewardsToken].rewardsDistributor = rewardsDistributor;
-    }
-
-    function stake(IERC20 stakingToken, uint256 amount) external {
-        stake(stakingToken, amount, msg.sender);
+    function stake(IERC20 pool, uint256 amount) external {
+        stake(pool, amount, msg.sender);
```
nventuro, line 247: I'd consider removing this version and just leaving the 3 argument one - also, external overloads are a pain in client code.

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -70,191 +75,294 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
...
-        _getReward(stakingTokens, true);
+    function getRewardAsInternalBalance(IERC20[] calldata pools) public nonReentrant {
+        _getReward(pools, true);
     }
 
     /**
      * @notice Allows a user to claim any rewards to internal balance
      */
-    function _getReward(IERC20[] calldata stakingTokens, bool asInternalBalance) internal {
+    function _getReward(IERC20[] calldata pools, bool asInternalBalance) internal {
```
nventuro, line 309: Same comments about `lenght()`, `at()` vs `unchecked_at()` and storage pointers in the loops in this function.

pkg/vault/contracts/interfaces/IBasePool.sol
```diff
@@ -86,4 +86,7 @@ interface IBasePool is IPoolSwapStructs {
         uint256 protocolSwapFeePercentage,
         bytes memory userData
     ) external returns (uint256[] memory amountsOut, uint256[] memory dueProtocolFeeAmounts);
+
+
+    function getPoolId() external view returns (bytes32);
```
nventuro, line 91: I'm fine with adding this, but it likely requires also adding `override` in `pool-utils/BasePool` and `weighted/Weighted2Tokens`.

pkg/distributors/contracts/MultiRewards.sol
nventuro: (that is, assuming we care about arbitrary recipients)

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -70,191 +75,294 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
...
         uint256 amount,
         uint256 deadline,
         uint8 v,
         bytes32 r,
         bytes32 s
     ) public {
-        IERC20Permit(address(stakingToken)).permit(msg.sender, address(this), amount, deadline, v, r, s);
-        stake(stakingToken, amount, msg.sender);
+        IERC20Permit(address(pool)).permit(msg.sender, address(this), amount, deadline, v, r, s);
+        stake(pool, amount, msg.sender);
```
nventuro, line 286: If we care about arbitrary recipients, shouldn't we allow for one to be passed here?

nventuro (commented):
This seems quite focused and understandable, cheers and thanks for the effort and care!

I did a simple initial pass with things that immediately stood out, as I don't yet fully understand how this is supposed to work. Could you briefly comment on the expected order of operations? e.g. owner calls `whiteListRewarder(rewarder)`, rewarder calls `foo()`, etc.

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -70,191 +75,294 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
...
 
+    modifier onlyWhitelistedRewarder(IERC20 pool, IERC20 rewardsToken) {
+        require(
+            isWhitelistedToReward(pool, rewardsToken, msg.sender),
+            "only accessible by whitelisted rewarders and asset managers"
+        );
+        _;
+    }
+    modifier onlyOwnerOrPool(IERC20 pool) {
+        require(msg.sender == owner() || msg.sender == address(pool), "only accessible by the owner or the pool");
```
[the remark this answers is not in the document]
gtaschuk: Thanks.  Where else is that done?   Looking for a reference implementation

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -70,191 +75,294 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
...
+
+    /**
+     * @notice Allows a rewarder to be explicitly added to a whitelist of rewarders
+     */
+    function whitelistRewarder(
+        IERC20 pool,
+        IERC20 rewardsToken,
+        address rewarder
+    ) public onlyOwnerOrPool(pool) {
+        _whitelist[pool][rewardsToken].add(rewarder);
```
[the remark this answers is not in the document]
gtaschuk: I can imagine scenarios, sure.  Have been designing for near-instant rewards

if you deleted a reward without a snapshot, then users who hadn't claimed would not be able to claim - but maybe that's ok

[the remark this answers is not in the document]
gtaschuk: they wouldn't be able to claim for the portion of a reward duration that had elapsed

pkg/distributors/contracts/MultiRewards.sol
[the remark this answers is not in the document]
nventuro: This is slightly similar to how relayers work in the Vualt. You could inherit from `solidity-utils/helpers/Authentication` and provide an implementation for `_canPerform` (likely the same one in `VaultAuthorization`. Then you'd change this modifier (which is equivalent to `authenticateFor` in `VaultAuthorization) to call `_authenticateCaller()` if `msg.sender != pool`.

I'll grant that these names are not the best (as this isn't really about authentication so much as authorization) - but the tl;dr is you'll be assigning an action ID to each function with this modifier, and only accounts approved in the Authorizer to perform this action will be able to do so.

pkg/distributors/contracts/MultiRewards.sol
gtaschuk: A user `A` who stakes their tokens to a recipient `B` would also need to be able to unstake them back.  So this would require another level of accounting right?  I think it's better to keep stake as for the benefit of `msg.sender`.  Is there a use case you are imagining?

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -70,191 +75,300 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
...
-    function getReward(IERC20[] calldata stakingTokens) public nonReentrant {
-        _getReward(stakingTokens, false);
+    // todo accept array of claims [{pool, rewardToken}]
+    function getReward(IERC20[] calldata pools) public nonReentrant {
+        _getReward(pools, false);
     }
 
-    function getRewardAsInternalBalance(IERC20[] calldata stakingTokens) public nonReentrant {
-        _getReward(stakingTokens, true);
+    function getRewardAsInternalBalance(IERC20[] calldata pools) public nonReentrant {
```
gtaschuk, line 304: We should add a recipient here for implementing higher level reinvestment/swap/etc

pkg/distributors/contracts/MultiRewards.sol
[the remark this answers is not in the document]
gtaschuk: @nventuro I've done some work here to use the authorizer pattern, but I'm not sure it's a clear improvement - can that go in as a separate PR so we can look at it in isolation?

gtaschuk:
> as I don't yet fully understand how this is supposed to work. Could you briefly comment on the expected order of operations? e.g. owner calls `whiteListRewarder(rewarder)`, rewarder calls `foo()`, etc.

1. The pool or contract owner `whitelistRewarder` a rewarder so they can add rewards for a particular pool, reward token combination.  An asset manager for a pool can add themselves to the whitelist
2. A rewarder calls `addReward` to set up the config of a reward.
3. A rewarder calls `notifyReward` so the contract can requisition their tokens and start a reward

pkg/distributors/contracts/MultiRewards.sol
gtaschuk: Would like to address this as part of https://github.com/balancer-labs/balancer-v2-monorepo/issues/659

pkg/distributors/contracts/MultiRewards.sol
gtaschuk: Will cover this in #659

pkg/distributors/contracts/MultiRewards.sol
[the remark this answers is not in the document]
nventuro: Created https://github.com/balancer-labs/balancer-v2-monorepo/issues/662 for this

nventuro (approved):
Looks great, in it goes!

New topics spun out from this review have had issues created and are being tracked in [the milestone](https://github.com/balancer-labs/balancer-v2-monorepo/milestone/6).

PR #656, opened 2021-06-21
Turn on access control on asset manage config

TomAFrench:
Asset managers now only listen to the pool for setting their config.

The majority of this change is from swapping out the mocked pool implementations with a normal weighted pool and updating the test suite setup. Due to the order in which things need to be deployed I've updated the `WeightedPool` model to allow the vault address to be injected rather than deploying its own vault..

### DIFF
--- a/pkg/asset-manager-utils/contracts/RewardsAssetManager.sol
+++ b/pkg/asset-manager-utils/contracts/RewardsAssetManager.sol
@@ -64,9 +64,9 @@ abstract contract RewardsAssetManager is IAssetManager {
         token = _token;
     }
 
-    modifier onlyPoolController() {
+    modifier onlyPoolContract() {
         address poolAddress = address((uint256(poolId) >> (12 * 8)) & (2**(20 * 8) - 1));
-        require(msg.sender == poolAddress, "Only callable by pool controller");
+        require(msg.sender == poolAddress, "Only callable by pool");
         _;
     }
 
@@ -181,8 +181,7 @@ abstract contract RewardsAssetManager is IAssetManager {
 
     function _getAUM() internal view virtual returns (uint256);
 
-    // TODO restrict access with onlyPoolController
-    function setConfig(bytes32 pId, bytes memory rawConfig) external override withCorrectPool(pId) {
+    function setConfig(bytes32 pId, bytes memory rawConfig) external override withCorrectPool(pId) onlyPoolContract {
         InvestmentConfig memory config = abi.decode(rawConfig, (InvestmentConfig));
 
         require(
--- a/pkg/asset-manager-utils/test/RewardsAssetManager.test.ts
+++ b/pkg/asset-manager-utils/test/RewardsAssetManager.test.ts
@@ -27,7 +27,7 @@ const setup = async () => {
   const vault = await Vault.create();
 
   // Deploy Pool
-  const pool = await deploy('v2-vault/MockPool', { args: [vault.address, GeneralPool] });
+  const pool = await deploy('MockAssetManagedPool', { args: [vault.address, GeneralPool] });
   const poolId = await pool.getPoolId();
 
   // Deploy Asset manager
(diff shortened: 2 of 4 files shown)


pkg/asset-manager-utils/contracts/RewardsAssetManager.sol
```diff
@@ -177,8 +177,7 @@ abstract contract RewardsAssetManager is IAssetManager {
 
     function _getAUM() internal view virtual returns (uint256);
 
-    // TODO restrict access with onlyPoolController
-    function setConfig(bytes32 pId, bytes memory rawConfig) external override withCorrectPool(pId) {
+    function setConfig(bytes32 pId, bytes memory rawConfig) external override withCorrectPool(pId) onlyPoolController {
```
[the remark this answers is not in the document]
nventuro: e.g. on `registerPool`:

```
* The caller is expected to be a smart contract that implements either `IGeneralPool` or `IMinimalSwapInfoPool`,
* depending on the chosen specialization setting. This contract is known as the Pool's contract.
*
* Note that the same contract may register itself as multiple Pools with unique Pool IDs, or in other words,
* multiple Pools may share the same contract.
```

pkg/asset-manager-utils/test/RewardsAssetManager.test.ts
```diff
@@ -105,7 +106,9 @@ describe('Rewards Asset manager', function () {
         upperCriticalPercentage: 4,
         lowerCriticalPercentage: 2,
       };
-      await assetManager.connect(poolController).setConfig(poolId, encodeInvestmentConfig(updatedConfig));
+      await pool.instance
+        .connect(poolController)
```
nventuro, line 110: This seems wrong, in line 100 `poolController` is assigned to `lp`?

nventuro (commented):
Ah yes, I've been having issues with the fact that the Pool model creates its own Vault. Regardless, it should be noted that we don't _need_ an actual Pool contract for these tests: it could make sense to have an EOA as a Pool (as long as we don't join or exit), or a `MockPool`, which is easier to manage than the full-on WeightedPool.

Not saying we shouldn't have these sort of more end-to-end tests, but it seems like we might be mixing concerns a bit here (such as what is WeightedPool's mechanism for authorizing someone to call the 'set AM config' function).

pkg/asset-manager-utils/contracts/RewardsAssetManager.sol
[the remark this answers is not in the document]
TomAFrench: This makes sense and I was trying to think of a better name for this recently. I've updated this to `onlyPoolContract`.

pkg/asset-manager-utils/test/RewardsAssetManager.test.ts
TomAFrench: I've removed this as `MockAssetManagedPool` allows anyone to update the config so this is unnecessary.

pkg/asset-manager-utils/test/RewardsAssetManager.test.ts
```diff
@@ -94,10 +95,29 @@ describe('Rewards Asset manager', function () {
...
+      assetManager = await deploy('MockRewardsAssetManager', {
+        args: [vault.address, ZERO_BYTES32, tokens.DAI.address],
+      });
+
+      // Assign assetManager to the DAI token
+      const assetManagers = [assetManager.address, ZERO_ADDRESS];
+
+      await vault.connect(pool).registerTokens(poolId, tokens.addresses, assetManagers);
+
+      await assetManager.initialise(poolId);
```
nventuro, line 121: Why are we setting up a Pool again here from an EOA? Can't the one created in the `setup()` function be used?

TomAFrench: Purely because I did this and then realised I needed to create a MockPool for the rebalancing tests and haven't gotten round to updating this again. Will carve this out now,

nventuro (approved):

nventuro:
Extremely clean, thanks!

PR #657, opened 2021-06-21
Pull aToken + reward asset addresses from Aave contracts

TomAFrench:
To simplify the factory logic for the Aave weighted pools factory, we now pull addresses where possible from the Aave contracts.


gtaschuk (approved):
Great - this will make this easier to deploy

found the `function REWARD_TOKEN()` method a bit unexpected, but should be fine

TomAFrench:
> found the `function REWARD_TOKEN()` method a bit unexpected, but should be fine

Yeah, I'm not sure if Aave is planning on deprecating this at some point although it's a really useful function to have.

Issue #570, opened 2021-05-19
Create Pool and Factory that supports AssetManagers


nventuro:
The Asset Manager feature introduced in #568 is not yet used: both Weighted and Stable Pools set the asset managers to the zero address.

We need to discuss which kind of Pool we want to attach this to (a 2 token Weighted one?), whether we want to limit the assets this Pool can use, and how many of the Pool tokens can have an Asset Manager.

nventuro:
We decided that initial Asset Manager support will exist for a Weighted 2 token Pool, with no price oracle, and the Asset Managers and assets under management limited by the Factory to the ones we developed and know are sound.

nventuro:
Closing as this is not currently under development.

PR #661, opened 2021-06-21
Add Aave WeightedPool metafactory

TomAFrench:
WIP as it's not ready for review + this PR builds on #657, I'll rebase once that is merged to give a cleaner diff.

This PR adds `AaveWeightedPoolFactory` which sits on top of `WeightedPoolFactory` and performs any necessary deployments and initialisation of AMs.

~~There's also decisions to be made around whether we want to lock down `WeightedPoolFactory` so that only vetted metafactories can call into it (allowing us to use it as a single source of "safe" weighted pools) or leave it open and whitelist individual metafactories for the frontend, etc.~~ As discussed, the `WeightedPoolFactory` will remain open so no whitelisting is necessary.

Closes https://github.com/balancer-labs/balancer-v2-monorepo/issues/570

### DIFF
--- a/pkg/pool-weighted/contracts/managed/AaveWeightedPoolFactory.sol
+++ b/pkg/pool-weighted/contracts/managed/AaveWeightedPoolFactory.sol
@@ -16,6 +16,7 @@ pragma solidity ^0.7.0;
 pragma experimental ABIEncoderV2;
 
 import "@balancer-labs/v2-pool-utils/contracts/BasePool.sol";
+import "@balancer-labs/v2-pool-utils/contracts/factories/BasePoolFactory.sol";
 
 import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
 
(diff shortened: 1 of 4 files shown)


nventuro:
Seems this is ready for rebasing now!

nventuro:
Seems this is ready for rebasing now!

TomAFrench:
This PR still requires some tests to be added but I would like to have a discussion on how we want to handle setting the reward distributor. Both in terms of this initial asset manager before multimultirewards and into the future.

Currently I've allowed this to be provided at deployment time which is nice in terms of flexibility however this gives scope for pool deployers to siphon off any rewards to their own address by providing their own rewards contract so there's an argument for hardcoding this to multimultirewards - although this bumps up against the fact that multimultirewards will not be ready in time.

One solution is to store `rewardsDistributor = address(0)` in the factory and allow the multisig to update this value once to set it to the canonical rewards contract, at that point all future asset managers will have the correct value on deployment. We'd then need something similar to go in the asset managers so we can update the ones which have already been deployed.

pkg/pool-weighted/contracts/managed/AaveWeightedPoolFactory.sol
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * @dev Deploys a new `WeightedPool` which invests specified tokens into Aave.
+     */
+    function create(
+        string memory name,
+        string memory symbol,
+        IERC20[] memory tokens,
+        uint256[] memory weights,
+        uint256[] memory managedTokens,
```
nventuro, line 46: I found usage if this array slightly unintuitive. What if we simply had an array with the same length as `tokens`, holding `bool managed`? We'd do a few extra iterations, but that seems fine.

nventuro:
Re. the distributor: I believe the plan is to initially set it to a MerkleRedeem contract, and once MultiMultiRewards is ready do a one-off merkle claim, and switch to the new one. If so, we'd need to have capabilities to switch distributors.

TomAFrench:
Closing this as it's pretty stale at this point.

Issue #655, opened 2021-06-18
Add MetaStable support


nventuro:
This would be a derived StablePool that has contracts it can call to ask for conversion rates between tokens. A sensible interface for these would be to call them with the original value, and have they return the converted value. Given that this is very similar to scaling factors, we should look into including this in the entire scaling flow.

nventuro:
Implemented in #716

PR #663, opened 2021-06-22
Make Pool scaling factors fixed-point

nventuro:
Part of https://github.com/balancer-labs/balancer-v2-monorepo/issues/655

### DIFF
--- a/pkg/pool-utils/contracts/BasePool.sol
+++ b/pkg/pool-utils/contracts/BasePool.sol
@@ -509,10 +509,10 @@ abstract contract BasePool is IBasePool, BasePoolAuthorization, BalancerPoolToke
      * @dev Returns the scaling factor for one of the Pool's tokens. Reverts if `token` is not a token registered by the
      * Pool.
      *
-     * All scaling factors are fixed-point values with 18 decimals, to allow for this function to be overriden by
+     * All scaling factors are fixed-point values with 18 decimals, to allow for this function to be overridden by
      * derived contracts that need to apply further scaling, making these factors potentially non-integer.
      *
-     * The largest 'base' scaling factor (i.e. in tokens with less than 18 decimals) is 10**18, which in fixed-poin is
+     * The largest 'base' scaling factor (i.e. in tokens with less than 18 decimals) is 10**18, which in fixed-point is
      * 10**36. This value can be multiplied with a 112 bit Vault balance with no overflow by a factor of ~1e7, making
      * even relatively 'large' factors safe to use.
      *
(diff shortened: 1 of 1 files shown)


nventuro:
According to the benchmarks this increases gas costs by a bit, but not a large amount. ~200 gas in the worst case (4 token Stable Pool), typically less..

pkg/pool-utils/contracts/BasePool.sol
```diff
@@ -502,14 +502,23 @@ abstract contract BasePool is IBasePool, BasePoolAuthorization, BalancerPoolToke
...
     }
 
     /**
      * @dev Returns the scaling factor for one of the Pool's tokens. Reverts if `token` is not a token registered by the
      * Pool.
+     *
+     * All scaling factors are fixed-point values with 18 decimals, to allow for this function to be overriden by
+     * derived contracts that need to apply further scaling, making these factors potentially non-integer.
+     *
+     * The largest 'base' scaling factor (i.e. in tokens with less than 18 decimals) is 10**18, which in fixed-poin is
```
TomAFrench, line 515:
```suggestion
     * The largest 'base' scaling factor (i.e. in tokens with less than 18 decimals) is 10**18, which in fixed-point is
```

pkg/pool-utils/contracts/BasePool.sol
```diff
@@ -502,14 +502,23 @@ abstract contract BasePool is IBasePool, BasePoolAuthorization, BalancerPoolToke
...
         uint256 decimalsDifference = Math.sub(18, tokenDecimals);
-        return 10**decimalsDifference;
+        return FixedPoint.ONE * 10**decimalsDifference;
     }
 
     /**
      * @dev Returns the scaling factor for one of the Pool's tokens. Reverts if `token` is not a token registered by the
      * Pool.
+     *
+     * All scaling factors are fixed-point values with 18 decimals, to allow for this function to be overriden by
```
TomAFrench, line 512:
```suggestion
     * All scaling factors are fixed-point values with 18 decimals, to allow for this function to be overridden by
```

facuspagnuolo (approved):
LGTM!

pkg/pool-utils/contracts/BasePool.sol
```diff
@@ -561,7 +574,7 @@ abstract contract BasePool is IBasePool, BasePoolAuthorization, BalancerPoolToke
      */
     function _upscaleArray(uint256[] memory amounts, uint256[] memory scalingFactors) internal view {
         for (uint256 i = 0; i < _getTotalTokens(); ++i) {
-            amounts[i] = Math.mul(amounts[i], scalingFactors[i]);
+            amounts[i] = FixedPoint.mulDown(amounts[i], scalingFactors[i]);
```
dmf7z, line 577: I would add a note here of why we round down in the meantime. Later we can do a bit of more analysis if we need upscaleArrayUp/Down.

dmf7z (approved):

pkg/pool-utils/contracts/BasePool.sol
nventuro: I left a note in `_upscale`, since this function is the same except on an array.

PR #664, opened 2021-06-22
Compute scaling factors only once

nventuro:
While the base scaling factors (decimals) are cheap to compute, the MetaStablePool will perform external calls to retrieve them, making this very costly. This PR keeps those in memory and passes them around as needed. 

Benchmark analysis doesn't seem to show any difference in gas costs of joins nor exits.

Part of https://github.com/balancer-labs/balancer-v2-monorepo/issues/655


facuspagnuolo (approved):
👌

dmf7z (approved):
Looks good

Issue #654, opened 2021-06-18
Add support for TWO_TOKEN specialization


nventuro:
The best approach is likely to extend `BaseGeneralPool` (or maybe add a mixin) that uses this alternative specialization if the number of tokens is two, and implements the associated swap hooks.

PR #665, opened 2021-06-23
StablePool dual interface

nventuro:
Since both `BaseMinimalSwapInfoPool` and `BaseGeneralPool` call `BasePool`'s constructor, they are incompatible in their current version and cannot be mixed this way. I therefore changed them to _not_ call the base constructor, and left that task to the concrete Pool contracts (`WeightedPool` and `StablePool`). 

I also added tests for two token pools that use both interfaces and check that the results match.

### Benchmark results
Before
```
 # Stable Pools with 2 tokens

1 pools: 121.8k (simple swap)
1 pools: 133.5k (133.5k per pool)
2 pools: 187.5k (93.7k per pool)
3 pools: 247.1k (82.4k per pool)

## Using Internal Balance
1 pools: 96.2k (simple swap)
1 pools: 107.9k (107.9k per pool)
2 pools: 161.9k (81.0k per pool)
3 pools: 221.5k (73.8k per pool)
```

After
```
# Stable Pools with 2 tokens

1 pools: 111.7k (simple swap)
1 pools: 123.5k (123.5k per pool)
2 pools: 167.4k (83.7k per pool)
3 pools: 217.0k (72.3k per pool)

## Using Internal Balance
1 pools: 86.2k (simple swap)
1 pools: 97.9k (97.9k per pool)
2 pools: 141.9k (70.9k per pool)
3 pools: 191.4k (63.8k per pool)
```

Fixes https://github.com/balancer-labs/balancer-v2-monorepo/issues/654


nventuro:
I also tried a different approach where I changed `BaseGeneralPool` to make it also inherit from `BaseMinimalSwapInfo` and implement both interfaces (reusing the mechanism I added to `StablePool`, as there's no point in not using the Two Token setting in _all_ General pools if possible), but it got a bit messy because of how constructors must be called. 

It can be made better, it just needs a bit more work.

facuspagnuolo (approved):

PR #666, opened 2021-06-23
Expose a function to calculate whether AMs should rebalance after a join/exit

TomAFrench:
@facuspagnuolo here's a quick mockup of what the `shouldRebalance` function would roughly look like. Happy to update the interface if there's something else you'd prefer.

### DIFF
--- a/pkg/asset-manager-utils/contracts/IAssetManager.sol
+++ b/pkg/asset-manager-utils/contracts/IAssetManager.sol
@@ -56,9 +56,9 @@ interface IAssetManager {
     function updateBalanceOfPool(bytes32 poolId) external;
 
     /**
-     * @notice Checks whether the pool should rebalance upon adding or removing `delta` cash.
+     * @notice Determines whether the pool should rebalance given the provided balances
      */
-    function shouldRebalance(uint256 delta, bool positiveDelta) external view returns (bool);
+    function shouldRebalance(uint256 cash, uint256 managed) external view returns (bool);
 
     /**
      * @notice Rebalances funds between the pool and the asset manager to maintain target investment percentage.
--- a/pkg/asset-manager-utils/contracts/IAssetManager.sol
+++ b/pkg/asset-manager-utils/contracts/IAssetManager.sol
@@ -56,6 +56,11 @@ interface IAssetManager {
     function updateBalanceOfPool(bytes32 poolId) external;
 
     /**
+     * @notice Checks whether the pool should rebalance upon adding or removing `delta` cash.
+     */
+    function shouldRebalance(uint256 delta, bool positiveDelta) external view returns (bool);
+
+    /**
      * @notice Rebalances funds between the pool and the asset manager to maintain target investment percentage.
      * @param poolId - the poolId of the pool to be rebalanced
      * @param force - a boolean representing whether a rebalance should be forced even when the pool is near balance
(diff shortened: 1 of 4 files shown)


pkg/asset-manager-utils/contracts/IAssetManager.sol
```diff
@@ -48,6 +48,11 @@ interface IAssetManager {
      */
     function updateBalanceOfPool(bytes32 poolId) external;
 
+    /**
+     * @notice Checks whether the pool should rebalance upon adding or removing `delta` cash.
+     */
+    function shouldRebalance(uint256 delta, bool positiveDelta) external view returns (bool);
```
nventuro, line 54: I'm thinking about how this will be called by the Pool. We either a) store the list of asset managers in the Pool itself (a bit cumbersome as we'll need to do the whole if-else chain for `immutable` pseudo array lookup), or b) query the Vault for the AM.

Interestingly, if we were to query the AM, we'd be making the exact same call the AM would make: `vault.getPoolTokenInfo`, which returns `cash`, `managed` and `assetManager`. So I'm thinking we could do the following:
 - call `getPoolTokenInfo` from the Pool to query the AM
 - if the AM is non-zero, call `shouldRebalance(cash +- delta, managed)`, meaning the AM would skip the Vault query entirely and just go 'given this cash and managed, should I rebalance?'

This would also avoid the not-so-great boolean in the interface. What do you think?

TomAFrench: I think that's much cleaner and could be used to simplify our tests too.

pkg/asset-manager-utils/contracts/RewardsAssetManager.sol
```diff
@@ -226,21 +226,10 @@ abstract contract RewardsAssetManager is IAssetManager {
...
-        uint256 updatedCash = positiveDelta ? poolCash.add(delta) : poolCash.sub(delta);
-        return _shouldRebalance(updatedCash, poolManaged);
-    }
-
     /**
      * @notice Determines whether the pool should rebalance given the provided balances
      */
-    function _shouldRebalance(uint256 cash, uint256 managed) private view returns (bool) {
+    function shouldRebalance(uint256 cash, uint256 managed) public view override returns (bool) {
         uint256 investedPercentage = cash.mul(FixedPoint.ONE).divDown(cash + managed);
```
nventuro, line 233: Given that this is now `public` and the amounts are caller-supplied, we shouldn't make assumptions on them and use SafeMath in the `cash + managed` addition.

nventuro (commented):
Looks fantastic :+1: What do you think about adding a simple test in each scenario we have setup (when the AM is balanced, when it is above the threshold, etc) where we call this function with the current balances and check the result?

pkg/asset-manager-utils/contracts/RewardsAssetManager.sol
TomAFrench: Fair enough. My thoughts are that a user could provide wildly incorrect values while still not causing an overflow and even if they do then it doesn't affect AM logic. That said it doesn't really hurt.

nventuro: Agreed, but it feels a bit weird that we'd knowlingly process incorrect data and return an invalid result :shrug:

nventuro (approved):
The tests themselves are perfect, I'd just consider putting them in the `context` blocks we already have for `rebalance` (i.e. just add one more `it` to those). Instead of having that `describe` block be _exclusively_ about the `rebalance()` function, it'd be about rebalancing in general. This would mean we'd be fully reusing the setup of the different scenarios (under threshold, over threshold, etc.).

nventuro (approved):

PR #669, opened 2021-06-25
Deployments: Implement contracts verifier

facuspagnuolo:
Since we don't follow the expected layout for the `artifacts` directory in the `deployments` subrepo, the Hardhat's verifier plugin fails if try to use it as is. But, since we do have the required files to perform verifications,  we could point out hardhat where to fetch them from. Unfortunately, this is not possible with the current plugin interface. We have already reached out the Hardhat team to see if they can expose a few parameters to allow us parameterize these paths. In the meantime, I just grabbed a few lines of code form the hardhat plugin to build our own verifier.

The changing part here is that we should now use always `build-info` files instead of the `artifacts` ones.

### DIFF
--- a/pkg/deployments/src/task.ts
+++ b/pkg/deployments/src/task.ts
@@ -1,11 +1,13 @@
 import fs from 'fs';
 import path from 'path';
+import { BuildInfo } from 'hardhat/types';
 import { BigNumber, Contract } from 'ethers';
 import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/signers';
 
 import logger from './logger';
+import Verifier from './verifier';
 import { deploy, instanceAt } from './contracts';
-import { Artifact, Input, Output, RawInput, RawOutput, RawInputKeyValue, Network, NETWORKS, Param } from './types';
+import { Artifact, Input, Network, NETWORKS, Output, Param, RawInput, RawInputKeyValue, RawOutput } from './types';
 
 const TASKS_DIRECTORY = path.resolve(__dirname, '../tasks');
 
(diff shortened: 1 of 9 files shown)


pkg/deployments/src/task.ts
```diff
@@ -41,27 +49,42 @@ export default class Task {
     return instance;
   }
 
+  async verify(name: string, address: string, constructorArguments: unknown): Promise<void> {
+    if (!this._verifier) return logger.warn('Avoiding contract verification, no verifier defined');
```
nventuro, line 53:
```suggestion
    if (!this._verifier) return logger.warn('Skipping contract verification, no verifier defined');
```

pkg/deployments/src/task.ts
```diff
@@ -41,27 +49,42 @@ export default class Task {
...
   }
 
   dir(): string {
     if (!this.id) throw Error('Please provide a task deployment ID to run');
     return this._dirAt(TASKS_DIRECTORY, this.id);
   }
 
-  artifact(name: string): Artifact {
+  buildInfo(fileName: string): BuildInfo {
     const abiDir = this._dirAt(this.dir(), 'abi');
```
nventuro, line 74: This directory should likely not be called 'abi'

pkg/deployments/src/verifier.ts
```diff
@@ -0,0 +1,174 @@
...
+
+import EtherscanResponse, {
+  delay,
+  getVerificationStatus,
+} from '@nomiclabs/hardhat-etherscan/dist/src/etherscan/EtherscanService';
+
+import Task from './task';
+import logger from './logger';
+
+const MAX_VERIFICATION_INTENTS = 3;
```
nventuro, line 32: s/intent/attempt :stuck_out_tongue:

nventuro (approved):
Awesome! I imagine the Verifier code is mostly the same, except we're using `task.buildInfo`?

pkg/deployments/src/task.ts
facuspagnuolo: let's use `build-info`, I'll rename it in a further PR

PR #671, opened 2021-06-25
feat: add a BaseRelayer for easy approvals of relayers

### LINKED ISSUE #670: Add signature approval to BatchRelayer
Currently we require a separate call to approve the `RebalancingRelayer`, we should explore a `approveRelayerAndCall` type flow which we can mix into relayers such as this.
https://github.com/balancer-labs/balancer-v2-monorepo/blob/e29c13a6236bc7ea9c38f73bbb23920ae391f9eb/pkg/asset-manager-utils/contracts/RebalancingRelayer.sol#L63-L74

TomAFrench:
This PR adds a base class for relayers to inherit from which allows them to atomically set their approval and perform actions as standard.

~Draft PR as I'd like to move a few tests over.~

fixes #670

### DIFF
--- a/pkg/standalone-utils/contracts/BaseRelayer.sol
+++ b/pkg/standalone-utils/contracts/BaseRelayer.sol
@@ -26,21 +26,25 @@ import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
 contract BaseRelayer {
     using Address for address payable;
 
-    IVault public immutable vault;
+    IVault private immutable _vault;
 
-    constructor(IVault _vault) {
-        vault = _vault;
+    constructor(IVault vault) {
+        _vault = vault;
     }
 
     receive() external payable {
         // Accept ETH transfers only coming from the Vault. This is only expected to happen when joining a pool,
         // any remaining ETH value will be transferred back to this contract and forwarded back to the original sender.
-        _require(msg.sender == address(vault), Errors.ETH_TRANSFER);
+        _require(msg.sender == address(_vault), Errors.ETH_TRANSFER);
+    }
+
+    function getVault() public view returns (IVault){
+        return _vault;
     }
 
     function setRelayerApproval(bool approved, bytes calldata authorisation) external payable {
         bytes memory data = abi.encodePacked(
-            abi.encodeWithSelector(vault.setRelayerApproval.selector, msg.sender, address(this), approved),
+            abi.encodeWithSelector(_vault.setRelayerApproval.selector, msg.sender, address(this), approved),
             authorisation
         );
         _vaultAction(0, data);
@@ -52,7 +56,7 @@ contract BaseRelayer {
      */
     function _vaultAction(uint256 value, bytes memory data) private returns (bytes memory) {
         // solhint-disable-next-line avoid-low-level-calls
-        (bool success, bytes memory result) = address(vault).call{ value: value }(data);
+        (bool success, bytes memory result) = address(_vault).call{ value: value }(data);
 
         // Pass up revert if the call failed
         if (!success) {
--- a/pkg/standalone-utils/contracts/BaseRelayer.sol
+++ b/pkg/standalone-utils/contracts/BaseRelayer.sol
@@ -82,9 +82,9 @@ contract BaseRelayer {
                 if (result.length < 68) revert("MULTICALL_FAILED");
                 // solhint-disable-next-line no-inline-assembly
                 assembly {
-                    result := add(result, 0x04)
+                    returndatacopy(0, 0, returndatasize())
+                    revert(0, returndatasize())
                 }
-                revert(abi.decode(result, (string)));
             }
 
             results[i] = result;
--- a/pkg/standalone-utils/contracts/BaseRelayer.sol
+++ b/pkg/standalone-utils/contracts/BaseRelayer.sol
@@ -1,0 +1,97 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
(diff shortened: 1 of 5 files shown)


pkg/standalone-utils/contracts/BaseRelayer.sol
```diff
@@ -0,0 +1,97 @@
...
+        vault = _vault;
+    }
+
+    receive() external payable {
+        // Accept ETH transfers only coming from the Vault. This is only expected to happen when joining a pool,
+        // any remaining ETH value will be transferred back to this contract and forwarded back to the original sender.
+        _require(msg.sender == address(vault), Errors.ETH_TRANSFER);
+    }
+
+    function setRelayerApproval(
```
TomAFrench, line 41: Change this to hardcode to just this relayer

pkg/standalone-utils/test/BaseRelayer.test.ts
```diff
@@ -0,0 +1,168 @@
...
+    context('when relayer is allowed to set approval', () => {
+      sharedBeforeEach('authorise relayer', async () => {
+        const authorizer = vault.authorizer as Contract;
+        const setRelayerApproval = actionId(vault.instance, 'setRelayerApproval');
+        await authorizer.connect(admin).grantRole(setRelayerApproval, relayer.address);
+      });
+
+      context('when not approved by sender', () => {
+        context('when the first call gives permanent approval', () => {
+          it("doesn't require signatures on further calls", async () => {
```
nventuro, line 150: The line between Vault and BaseRelayer tests is a bit blurred here, but these tests are simple enough that I think it can be useful to keep them.

TomAFrench: Agreed, although considering this is a core feature of this relayer it's nice to have direct confirmation that this works.

pkg/standalone-utils/contracts/BaseRelayer.sol
```diff
@@ -0,0 +1,93 @@
...
+            authorisation
+        );
+        _vaultAction(0, data);
+    }
+
+    /**
+     * @notice Allows calling an arbitrary function on the Vault
+     * @dev To be used only to set relayer approval - other actions should be called with a permanent approval set
+     */
+    function _vaultAction(uint256 value, bytes memory data) private returns (bytes memory) {
```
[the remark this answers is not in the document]
TomAFrench: I don't see any benefit to doing this. The only reason to have `_vaultAction` is if you need to add additional calldata onto the end of the call.

I don't see much harm either though.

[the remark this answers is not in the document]
TomAFrench: We could roll `setRelayerApproval` and `_vaultAction` together into a single function if we wanted to but I felt the readability improvements from a well defined separation were worth it.

pkg/standalone-utils/test/BaseRelayer.test.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  let approvalAuthorisation: string;
+  const EMPTY_AUTHORISATION = '0x';
+
+  before('deploy base contracts', async () => {
+    [, admin, signer] = await ethers.getSigners();
+  });
+
+  sharedBeforeEach('set up relayer', async () => {
+    const { contracts } = await setup();
```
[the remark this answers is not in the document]
TomAFrench: Agreed, I didn't notice how small it had gotten while cutting things out.

pkg/standalone-utils/contracts/BaseRelayer.sol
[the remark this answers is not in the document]
nventuro: Agreed, if we expect for the first transaction to grant approval to the relayer via signature. Just wasn't sure if this was intentional.

pkg/standalone-utils/contracts/BaseRelayer.sol
```diff
@@ -0,0 +1,98 @@
...
+            (bool success, bytes memory result) = address(this).delegatecall(data[i]);
+
+            if (!success) {
+                // Next 6 lines from https://ethereum.stackexchange.com/a/83577
+                if (result.length < 68) revert("MULTICALL_FAILED");
+                // solhint-disable-next-line no-inline-assembly
+                assembly {
+                    returndatacopy(0, 0, returndatasize())
+                    revert(0, returndatasize())
+                }
```
nventuro, line 87:
```suggestion
                // If there's no revert reason, provide our own
                if (result.length < 68) revert("MULTICALL_FAILED");
                
                // Otherwise, bubble-up the original one
                // solhint-disable-next-line no-inline-assembly
                assembly {
                    returndatacopy(0, 0, returndatasize())
                    revert(0, returndatasize())
                }
```

nventuro:
(closed by mistake)

pkg/standalone-utils/contracts/BaseRelayer.sol
TomAFrench: :man_facepalming: Yeah, I should have spotted that comment needed to be changed.

nventuro:
We realized this is slightly more complicated, as `joinPool` forwards all of `msg.value`: those functions should likely have a new argument with the amount of ETH to forward (but they should all be `payable` since `multicall` is `payable`).

The issue we then run into is one of reentrancy: `multicall` should not be reentrant as it sends all ETH to the caller at the end, which could be used to steal funds by a malicious Pool or token, and something similar happens with `joinPool` and friends. 

We have a few conflicting requirements (being able to call `joinPool` directly and have it return ETH, being able to use `multicall`safely, being able to merge swaps and joins, and being able to merge relayer approval and other operations): I think we should consider all of them while discussing how to move forward.

nventuro:
@TomAFrench are you planning on tackling #689 here?

TomAFrench:
I was. Alternatively we can merge the original implementation along with #604 and then handle the re-entrancy issue in a separate PR.

No I think about it, this would probably be preferable as we'd need to make changes to the rebalancing relayer to account for the fact we're delegatecalling into a separate contract (use of unstructured storage etc).

nventuro:
I think you might be right in that splitting would be desirable, yes. If `BaseRelayer` becomes complicated due to #689, I'd suggest perhaps postponing it and leaving `RebalancingRelayer` in a secure state so that the Simple Asset Manager milestone is finished, and then improve it while tackling the BatchRelayer.

TomAFrench:
Gotcha, I've rolled back the last commit.

TomAFrench:
@nventuro considering we're postponing the use of the `RebalancingRelayer`, I think it's worth moving forward on merging this so we can build it into the `BatchRelayer` and the `JoinAndStakeRelayer`. What do you think?

PR #673, opened 2021-06-25
Pools: Improve rebalanced base pool

facuspagnuolo:
This PR introduces the concept of allowing users to join/exit pools w/ asset managers manually (instead of going through a relayer) in case the asset manager doesn't require a rebalance as a result.


pkg/pool-utils/contracts/test/MockRelayedBasePool.sol
```diff
@@ -16,9 +16,9 @@ pragma solidity ^0.7.0;
 pragma experimental ABIEncoderV2;
 
 import "../BasePool.sol";
-import "../RelayedBasePool.sol";
+import "../RebalancedBasePool.sol";
 
-contract MockRelayedBasePool is BasePool, RelayedBasePool {
+contract MockRelayedBasePool is BasePool, RebalancedBasePool {
```
nventuro, line 21: We should rename this contract

pkg/pool-utils/contracts/RebalancedBasePool.sol
```diff
@@ -0,0 +1,106 @@
...
+    ) internal view {
+        uint256 totalTokens = _getTotalTokens();
+        if (totalTokens > 0) _ensureAssetManagerBalanced(poolId, _token0, amounts[0], protocolFeeAmounts[0], positive);
+        if (totalTokens > 1) _ensureAssetManagerBalanced(poolId, _token1, amounts[1], protocolFeeAmounts[1], positive);
+        if (totalTokens > 2) _ensureAssetManagerBalanced(poolId, _token2, amounts[2], protocolFeeAmounts[2], positive);
+        if (totalTokens > 3) _ensureAssetManagerBalanced(poolId, _token3, amounts[3], protocolFeeAmounts[3], positive);
+        if (totalTokens > 4) _ensureAssetManagerBalanced(poolId, _token4, amounts[4], protocolFeeAmounts[4], positive);
+        if (totalTokens > 5) _ensureAssetManagerBalanced(poolId, _token5, amounts[5], protocolFeeAmounts[5], positive);
+        if (totalTokens > 6) _ensureAssetManagerBalanced(poolId, _token6, amounts[6], protocolFeeAmounts[6], positive);
+        if (totalTokens > 7) _ensureAssetManagerBalanced(poolId, _token7, amounts[7], protocolFeeAmounts[7], positive);
```
nventuro, line 89: :facepalm:  I hate not having immutable arrays

pkg/pool-utils/contracts/RebalancedBasePool.sol
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    function _ensureAssetManagerBalanced(
+        bytes32 poolId,
+        uint256[] memory amounts,
+        uint256[] memory protocolFeeAmounts,
+        bool positive
+    ) internal view {
+        uint256 totalTokens = _getTotalTokens();
+        if (totalTokens > 0) _ensureAssetManagerBalanced(poolId, _token0, amounts[0], protocolFeeAmounts[0], positive);
```
nventuro, line 82:
```suggestion
        if (totalTokens > 0) { _ensureAssetManagerBalanced(poolId, _token0, amounts[0], protocolFeeAmounts[0], positive); } else { return; }
```
Minor optimisation :shrug:

nventuro: (to be applied to all `if`s)

pkg/pool-utils/contracts/RebalancedBasePool.sol
```diff
@@ -0,0 +1,106 @@
...
+
+import "./BasePool.sol";
+import "./interfaces/IRebalancedBasePoolRelayer.sol";
+
+/**
+ * @dev Base Pool associated with a relayer that guarantees it can only be joined/exited from the relayer itself.
+ * This contract as a simple mixin for pools. Implementing pools must make sure to call the BasePool's constructor
+ * properly.
+ */
+abstract contract RebalancedBasePool is BasePool {
```
nventuro, line 30: Extremely nitpicky, but I noted we could just inherit from `vault/interfaces/IBasePool`, as we're only overriding interface functions and not relying on `BasePool` at all. 

I don't mind keeping things as-is though, just wanted to point this out.

nventuro (commented):
Love the implementation, thanks a lot! Should be good to go once we have tests

pkg/pool-utils/contracts/RebalancedBasePool.sol
TomAFrench: Agreed from a readability perspective but isn't this preferable for gas? Granted this is very minor savings.

facuspagnuolo: I'd propose doing it in another way to make it look nicer, pls see recent changes

pkg/pool-utils/contracts/RebalancedBasePool.sol
facuspagnuolo: I don't think it is possible cause we need to access `BasePool`'s storage variables here

pkg/pool-utils/contracts/RebalancedBasePool.sol
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    function _ensureAssetManagerBalanced(
+        bytes32 poolId,
+        uint256[] memory amounts,
+        uint256[] memory protocolFeeAmounts,
+        bool positive
+    ) internal view {
+        uint256 totalTokens = _getTotalTokens();
+        if (totalTokens == 0) return;
```
TomAFrench, line 96: Do we want to remove these first couple of `if` statements? We're not going to have less than 2 tokens in the current `BasePool`

pkg/pool-utils/test/RebalancedBasePool.test.ts
```diff
@@ -0,0 +1,198 @@
...
+        it('does not revert', async () => {
+          await expect(join()).not.to.be.reverted;
+        });
+      });
+    });
+
+    context('when some asset managers require a rebalance', () => {
+      sharedBeforeEach('mock asset manager', async () => {
+        await assetManagers[1].mockShouldRebalance(true);
+      });
```
nventuro, line 102: This suggestion might be a bit pedantic and overkill, so take it with a grain of salt.

It'd be _great_ if we had not just this scenario, but actually programmatically iterated over the asset managers for `tokens[0..7]`, and tested that when _any_ of all 8 are unbalanced, the checks occur. We're otherwise leaving a large portion of the error-prone `if-else` chain in `_ensureAssetManagerBalanced` untested.

Similarly, it'd be good to check that each asset manager receives the expected cash balances, but that might be a tad complicated as we can't e.g. emit events in the `view` call.

pkg/pool-utils/contracts/RebalancedBasePool.sol
```diff
@@ -0,0 +1,128 @@
...
+        IERC20 token,
+        uint256 amount,
+        uint256 protocolFeeAmount,
+        bool positive
+    ) internal view {
+        (uint256 cash, uint256 managed, , address assetManager) = getVault().getPoolTokenInfo(poolId, token);
+        if (assetManager != address(0)) {
+            uint256 finalCash = (positive ? cash.add(amount) : cash.sub(amount)).sub(protocolFeeAmount);
+            bool shouldRebalance = IAssetManager(assetManager).shouldRebalance(finalCash, managed);
+            _require(!shouldRebalance || _relayer.hasCalledPool(poolId), Errors.BASE_POOL_RELAYER_NOT_CALLED);
```
nventuro, line 125: While this is correct, it is currently lacking short-circuiting behavior: even after we find an asset manager requires rebalance and we check that the relayer has been called, we continue looking at all others. 

Perhaps this function could return `true` if the AM needs rebalancing and the relayer has been called? And then the version that calls this would immediately exit and skip all other AMs.

nventuro:
We're rolling back this change due to https://github.com/balancer-labs/balancer-v2-monorepo/issues/695 - relayers will not be optional any more.

PR #683, opened 2021-06-29
Certora fixes related to asset managers

TomAFrench:
Addressed:
- Optimization - excess calculations
- Unnecessary checked arithmetics
- Missing function from interface
- Unused return variable names
- Documentation errors, grammatical mistakes, and typos

Not addressed in this PR:
- Unsafe casting to int256
- Edge case reverts due to unsafe addition
- Overflow
- Wrong documentation
- Incomplete documentation

### DIFF
--- a/pkg/asset-manager-utils/contracts/RewardsAssetManager.sol
+++ b/pkg/asset-manager-utils/contracts/RewardsAssetManager.sol
@@ -65,7 +65,7 @@ abstract contract RewardsAssetManager is IAssetManager {
     }
 
     modifier onlyPoolContract() {
-        address poolAddress = address((uint256(poolId) >> (12 * 8)) & (2**(20 * 8) - 1));
+        address poolAddress = address(uint256(poolId) >> (12 * 8));
         require(msg.sender == poolAddress, "Only callable by pool");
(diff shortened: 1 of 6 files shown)


nventuro (approved):

pkg/asset-manager-utils/contracts/RewardsAssetManager.sol
```diff
@@ -213,13 +213,7 @@ abstract contract RewardsAssetManager is IAssetManager {
         return _config;
     }
 
-    function getPoolBalances(bytes32 pId)
-        public
-        view
-        override
-        withCorrectPool(pId)
-        returns (uint256 poolCash, uint256 poolManaged)
```
nventuro, line 221: I'm not sure how Solidity handles the ABI here: while IAssetManager may call these `poolCash` and `poolManaged`, if we interact with this contract with `RewardsAssetManager` ABI then this may override the interface declaration and drop the names. If so, we should keep them.

nventuro: I think the failing tests may be due to this.

PR #691, opened 2021-07-01
Asset Manager consistent getters

nventuro:
This is just a minor change that moves in the direction of having `getX()` type getters, like the rest of our contracts have. There are more improvements to be made related to these parts of the code (such as caching the value of `_poolId` to avoid repeated storage reads) - this PR just focuses on naming.


TomAFrench (approved):
This looks good to me. Very happy to get rid of those underscores.

I added another commit which just rearranges a couple of functions to group them a little better although it doesn't seem to be showing up in this PR.

nventuro:
@TomAFrench maybe you pushed to a different branch?

PR #696, opened 2021-07-01
Pools: Move oracle contracts to utils

facuspagnuolo:


nventuro (approved):
Seems like we have no tests for these contracts, only for the full oracle + pool, no?

facuspagnuolo:
There are some tests in `WeightedPool2Tokens.test.ts` but they are coupled with the weighted pool logic due to how misc data works, I'll try to decoupled it after the refactor

PR #701, opened 2021-07-02
Fix linter

facuspagnuolo:


gtaschuk:
Suggest we use `    "lint:typescript": "eslint --max-warnings 0 . --ext .ts",` so that lint fails on any warnings if we don't intend to tolerate them

PR #704, opened 2021-07-02
Add deployment ABIs, fix build infos, improve readmes

nventuro:
This renames the `abi` dir to `build-info` (and fixes a small bug introduced by #699 in the process), adds missing `build-info`s (e.g. of the weighted pools - these were generated in the correct commit), and adds all relevant abis (factories _and_ pools) in the `abi` directory, as well as make our readmes slightly less minimal.


markusbkoch:
Looks great. I noticed there's a `polygon.json` in every `output` directory but the `weighted-pool`'s

facuspagnuolo (approved):

PR #705, opened 2021-07-02
Add weighted pool factory task

TomAFrench:
This PR adds a task definition which I used to verify the Polygon deployments of the weighted pool factories - also included are the deployed addresses on this network.


nventuro (approved):
Brilliant, thanks! Hopefully soon we'll be able to abstract away most of the `force` pattern.

Issue #695, opened 2021-07-01
Update Asset Manager earnings before all joins and exits


nventuro:
The rebalancing relayer must have the Asset Managers update the Pool's total balances _before_ joins and exits to ensure that they are correctly assigned to the current LPs. Otherwise, a large (e.g. flash-loaned) LP could join the Pool and rapidly exit it, stealing all AM earnings since the last update.

This makes relayers mandatory.

TomAFrench:
Fixed in #706

TomAFrench:
(closed by mistake)

PR #706, opened 2021-07-05
Allow exits which requires dipping into pool's managed balance.

TomAFrench:
This PR adds behaviour to the `RebalancingRelayer` to allow it to process exits which are larger than the current cash balance of the pool.

The guards on investing/divesting past the target investment level have been removed as they prevented freeing up enough cash to perform the exits. This is fine as we've moved to a situation where these functions are not exposed by default.

Once this PR is merged we can
- Implement #695 by adding a call to `updatePoolBalance` to the before hook in the relayer.
- revert #651 as the risk of a malicious pool owner locking funds is removed.


pkg/asset-manager-utils/contracts/RebalancingRelayer.sol
```diff
@@ -34,8 +34,13 @@ contract RebalancingRelayer is IBasePoolRelayer, AssetHelpers {
...
     );
 
-    modifier rebalance(bytes32 poolId, IAsset[] memory assets) {
+    modifier rebalance(
+        bytes32 poolId,
+        IAsset[] memory assets,
+        uint256[] memory minCashBalances
+    ) {
         _require(_calledPool == _EMPTY_CALLED_POOL, Errors.REBALANCING_RELAYER_REENTERED);
+        _ensureCashBalance(poolId, _translateToIERC20(assets), minCashBalances);
```
nventuro, line 43: Let's store `tokens = _translate(assets)` as a variable, to avoid looping over this array twice.

pkg/asset-manager-utils/contracts/RebalancingRelayer.sol
```diff
@@ -76,11 +81,27 @@ contract RebalancingRelayer is IBasePoolRelayer, AssetHelpers {
...
     }
 
+    function _ensureCashBalance(
+        bytes32 poolId,
+        IERC20[] memory tokens,
+        uint256[] memory minCashBalances
+    ) internal {
+        for (uint256 i = 0; i < tokens.length; i++) {
+            (uint256 cash, , , address assetManager) = vault.getPoolTokenInfo(poolId, tokens[i]);
+            uint256 cashNeeded = minCashBalances[i];
```
nventuro, line 97:
```suggestion
            uint256 cashNeeded = minCashBalances[i];
            if (cashNeeded == 0) continue;
            
            (uint256 cash, , , address assetManager) = vault.getPoolTokenInfo(poolId, tokens[i]);
```

`joinPool` currently passes 0 values, but I imagine we may also have e.g. a single asset exit where we don't really care about the balances of other tokens.

pkg/asset-manager-utils/contracts/RewardsAssetManager.sol
```diff
@@ -63,8 +64,15 @@ abstract contract RewardsAssetManager is IAssetManager {
...
     modifier onlyPoolContract() {
-        address poolAddress = address(uint256(_poolId) >> (12 * 8));
-        require(msg.sender == poolAddress, "Only callable by pool");
+        require(msg.sender == getPoolAddress(), "Only callable by pool");
+        _;
+    }
+
+    modifier onlyPoolRebalancer() {
+        require(
+            address(IRelayedBasePool(getPoolAddress()).getRelayer()) == msg.sender,
```
nventuro, line 73:
```suggestion
            msg.sender == address(IRelayedBasePool(getPoolAddress()).getRelayer()),
```

Use Yoda conditions, we probably should not

pkg/asset-manager-utils/test/RebalancingRelayer.test.ts
```diff
@@ -120,18 +122,12 @@ describe('RebalancingRelayer', function () {
           it('rebalances the pool', async () => {
             const receipt = await relayer.connect(sender).joinPool(poolId, recipient.address, request);
 
-            expectEvent.inIndirectReceipt(await receipt.wait(), assetManagers[0].interface, 'Rebalanced', {
+            expectEvent.inIndirectReceipt(await receipt.wait(), assetManagers[0].interface, 'Rebalance', {
               poolId,
-              assetManager: assetManagers[0].address,
-              token: tokens.first.address,
-              force: false,
```
nventuro, line 127: Why remove these?

pkg/asset-manager-utils/test/RebalancingRelayer.test.ts
```diff
@@ -236,41 +226,69 @@ describe('RebalancingRelayer', function () {
...
+              const investmentConfig = {
+                targetPercentage: fp(0.95),
+                upperCriticalPercentage: fp(0.95),
+                lowerCriticalPercentage: fp(0),
+              };
+              await pool
+                .connect(admin)
+                .setAssetManagerPoolConfig(tokens.first.address, encodeInvestmentConfig(investmentConfig));
+
+              await assetManagers[0].rebalance(poolId, true);
```
nventuro, line 288: Given that this setup is somewhat complex, let's add an `expect` here checking that `cash` is indeed lower that required

nventuro (approved):
Great work!

pkg/asset-manager-utils/contracts/RebalancingRelayer.sol
TomAFrench: This will just be undone when we address #695 (I've got this on a local branch).

pkg/asset-manager-utils/test/RebalancingRelayer.test.ts
TomAFrench: I've replaced the MockAssetManager with a version which allows pulling cash into the managed balance and the events in these two contracts differ.

PR #707, opened 2021-07-05
MultiRewards distributor: rewarder allowlist

gtaschuk:
Uses a mapping rather than an `EnumerableSet` to more efficiently manage list of rewarders allowed to distribute in the staking contract

Renames whitelist -> allowlist re: @FernandoMartinelli 's point about outdated language https://balancerlabs.slack.com/archives/C0146NRBJEP/p1624636747191400

### DIFF
--- a/pkg/asset-manager-utils/contracts/AaveATokenAssetManager.sol
+++ b/pkg/asset-manager-utils/contracts/AaveATokenAssetManager.sol
@@ -60,7 +60,7 @@ contract AaveATokenAssetManager is RewardsAssetManager {
 
         distributor = IMultiRewards(rewardsDistributor);
         IERC20 poolAddress = IERC20(uint256(poolId) >> (12 * 8));
-        distributor.whitelistRewarder(poolAddress, stkAave, address(this));
+        distributor.allowlistRewarder(poolAddress, stkAave, address(this));
         distributor.addReward(poolAddress, stkAave, 1);
 
         stkAave.approve(rewardsDistributor, type(uint256).max);
--- a/pkg/asset-manager-utils/test/AaveATokenAssetManager.test.ts
+++ b/pkg/asset-manager-utils/test/AaveATokenAssetManager.test.ts
@@ -65,7 +65,7 @@ const setup = async () => {
 
   await assetManager.initialize(poolId, distributor.address);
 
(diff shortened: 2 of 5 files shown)


pkg/asset-manager-utils/contracts/AaveATokenAssetManager.sol
```diff
@@ -60,7 +60,7 @@ contract AaveATokenAssetManager is RewardsAssetManager {
 
         distributor = IMultiRewards(rewardsDistributor);
         IERC20 poolAddress = IERC20(uint256(poolId) >> (12 * 8));
-        distributor.whitelistRewarder(poolAddress, stkAave, address(this));
+        distributor.allowlistRewarder(poolAddress, stkAave, address(this));
```
nventuro, line 63: Doesn't 'allow' make more  sense as a verb? Also e.g. 'only allowed rewarder', 'is allowed rewarder', etc.

gtaschuk: :thinking: 
_tl;dr - I agree the term feels unfamiliar but I would suggest we use allowlist as a 1:1 standin for whitelist_

**Whitelist** is commonly understood as both a verb or a noun, which is why `whitelistRewarder` feels familiar.  **Allowlist** as a word is just not common in spoken english (yet?) but is being used as a direct stand-in in code.  I agree that the verb usage, especially, feels a little cumbersome, but I think that's just because it's uncommon today.

It is however becoming much more common, especially among larger companies.  Google, Microsoft, Youtube, VMWare etc are all replacing "whitelist" with "allowlist" so I suspect it will feel more familiar over time.

ie here's a vmware glossary definition about the practice of allowlisting (verb usage)
https://www.vmware.com/topics/glossary/content/allowlisting

I don't really like `allowRewarder` because it feels vague ("allow Rewarders to do what?").  The purpose of the function is to add the rewarder to the allowlist, and `allowlistRewarder` feels better than `addAddressToRewarderAllowlist`

nventuro: I was going to say that it might feel too jarring, but in the end I think you're right. 'allow' doesn't convey the same meaning of 'add to the set of allowed actors'. Thanks for the writeup!

nventuro (approved):

PR #703, opened 2021-07-02
MultiRewards: Allow LPs to claim rewards to a contract with a callback function

gtaschuk:
This implements #659 - allowing an LP to claim rewards to a callback contract

As an example of such a callback contract, `Reinvestor` reinvests rewards tokens in a specified `WeightedPool`

For example, a LP may want to claim all their BAL reward into a BAL-ETH pool

PR #708, opened 2021-07-05
MerkleRedeem distributor: allow user to claim merkle rewards to a callback contract

gtaschuk:
Up for discussion: Adds callbacks to the merkle distributor (as #703 does for the staking contract) to allow LPs to send rewards to a callback contract (ie for reinvestment).

I think this is most useful in the generalized use of this distributor - rather than our initially specific use case of distributing stkAAVE before the `MultiRewards` staking contract is adoptied.  Wanted to confirm my suspicion that it would be an easy extension of this contract


pkg/distributors/contracts/MerkleRedeem.sol
```diff
@@ -120,6 +114,19 @@ contract MerkleRedeem is IDistributor, Ownable {
             totalBalance = totalBalance.add(claim.balance);
             claimed[claim.week][liquidityProvider] = true;
         }
+    }
+
+    /**
+     * @notice Allows a user to claim a particular week's worth of rewards
```
TomAFrench, line 120:
```suggestion
     * @notice Allows a user to claim a particular set of weeks' rewards
```

pkg/distributors/contracts/MerkleRedeem.sol
```diff
@@ -128,6 +135,25 @@ contract MerkleRedeem is IDistributor, Ownable {
         }
     }
 
+    /**
+     * @notice Allows a user to claim several weeks of rewards to a callback
```
TomAFrench, line 139:
```suggestion
     * @notice Allows a user to claim a particular set of weeks' rewards to a callback
```

pkg/distributors/test/MerkleRedeem.test.ts
```diff
@@ -277,8 +277,56 @@ describe('MerkleRedeem', () => {
...
+        const proof1: BytesLike[] = merkleTree1.getHexProof(elements1[0]);
+        const proof2: BytesLike[] = merkleTree2.getHexProof(elements2[0]);
+
+        claims = [
+          { week: bn(1), balance: claimBalance1, merkleProof: proof1 },
+          { week: bn(2), balance: claimBalance2, merkleProof: proof2 },
+        ];
+      });
+
+      it('allows a user to claim the reward to a callback contract', async () => {
```
nventuro, line 302: Let's add a test that checks that callback has been granted the tokens as internal balance

nventuro (commented):
Looks great, thanks!

On an unrelated note, we've been using `.call` and reimplementing revert reason bubbling that I took the liberty of porting [`functionCall`](https://docs.openzeppelin.com/contracts/4.x/api/utils#Address-functionCall-address-bytes-) and using it here, which already does this for us. What do you think?

pkg/distributors/contracts/MerkleRedeem.sol
```diff
@@ -128,6 +135,25 @@ contract MerkleRedeem is IDistributor, Ownable {
...
+     */
+    function claimWeeksWithCallback(
+        address payable liquidityProvider,
+        address payable callbackContract,
```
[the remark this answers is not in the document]
gtaschuk: Correct, though unfortunately addresses need to be marked payable to receive `IAssets`

gtaschuk:
> Looks great, thanks!
> 
> On an unrelated note, we've been using `.call` and reimplementing revert reason bubbling that I took the liberty of porting [`functionCall`](https://docs.openzeppelin.com/contracts/4.x/api/utils#Address-functionCall-address-bytes-) and using it here, which already does this for us. What do you think?

Looks good.  Should also use for the callbacks on `MultiReward` then too

pkg/distributors/test/MerkleRedeem.test.ts
gtaschuk: I believe that's what this test does!

nventuro (approved):

PR #709, opened 2021-07-05
Pools: Store swap fee pct in misc data

facuspagnuolo:


pkg/pool-utils/contracts/BasePool.sol
```diff
@@ -61,7 +61,11 @@ abstract contract BasePool is IBasePool, BasePoolAuthorization, BalancerPoolToke
 
     uint256 private constant _MINIMUM_BPT = 1e6;
 
-    uint256 internal _swapFeePercentage;
+    // Storage slot that can be used to store unrelated pieces of information. In particular, by default is used
+    // to store only the swap fee percentage of a pool. But it can be extended to store some more pieces of information.
+    // The swap fee percentage can be stored using 64 bits, therefore the remaining 192 bits can be used to store any
+    // other piece of information.
```
nventuro, line 67:
```suggestion
    // Storage slot that can be used to store miscelaneous pieces of information, including the swap fee percentage.
    // Because this value is read in all swaps, joins and exits, the storage slot is 'warm' in these contexts, making further
    // reads inexpensive due to EIP 2929: https://eips.ethereum.org/EIPS/eip-2929.
    // The swap fee percentage itself is stored in the least significant 64 bits: the remaining 192 bits can be used for any
    // other purpose.
```

I'd make `getSwapFeePercentage()` and `_storeSwapFeePercentage()` perform the required masking to only use the relevant bits, making it easier to derive from this contract (as they'll only need to add behavior, and not change the original one).

nventuro (approved):
_nice_

pkg/pool-utils/contracts/BasePool.sol
facuspagnuolo: I like the idea of allowing each pool where to store/read the swap fee from

facuspagnuolo: I propose a new approach for this, you may want to update this comment to follow what's new

PR #710, opened 2021-07-05
Remove upper limit on AM investment level

TomAFrench:
Now LPs can force a divestment due to #706, we no longer need to hardcode an upper limit to the investment percentage.

### DIFF
--- a/pkg/asset-manager-utils/test/RewardsAssetManager.test.ts
+++ b/pkg/asset-manager-utils/test/RewardsAssetManager.test.ts
@@ -147,17 +147,6 @@ describe('Rewards Asset manager', function () {
       ).to.be.revertedWith('Target must be less than or equal to upper critical level');
     });
 
-    it('reverts when setting target above 95%', async () => {
-      const badConfig = {
-        targetPercentage: fp(0.95).add(1),
-        upperCriticalPercentage: fp(1),
-        lowerCriticalPercentage: 0,
-      };
-      await expect(
-        pool.setAssetManagerPoolConfig(assetManager.address, encodeInvestmentConfig(badConfig))
-      ).to.be.revertedWith('Target must be less than or equal to 95%');
-    });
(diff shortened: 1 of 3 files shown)


TomAFrench:
Closed as I was concerned that we didn't have a check to ensure that investment level remained <=100%. We do so everything is fine.

pkg/asset-manager-utils/test/RewardsAssetManager.test.ts
```diff
@@ -147,17 +147,6 @@ describe('Rewards Asset manager', function () {
       ).to.be.revertedWith('Target must be less than or equal to upper critical level');
     });
 
-    it('reverts when setting target above 95%', async () => {
```
nventuro, line 150: Let's replace this with a test that sets the target to 100%, and make sure we have a test involving the rebalancing relayer that exits a pool with 0 cash.

nventuro (commented):
Great catch!

pkg/asset-manager-utils/test/RewardsAssetManager.test.ts
TomAFrench: Would you not say that this is appropriately covered by the tests for checking `upperCriticalPercentage <=100%` and `targetPercentage <= upperCriticalPercentage`?

The rebalancing relayer already tests for a zero cash case we we've set the investment percentage to 100%

nventuro: If we already have such a test that's fine then, my point was we should at least have _a_ test.

Though how could we have a rebalancing relayer test with 100% target before this commit?

TomAFrench: We didn't, by "already" I mean we bump the percentage from 95% to 100% in this commit.

PR #711, opened 2021-07-05
feat: update pool balance prior to join/exit

### LINKED ISSUE #695: Update Asset Manager earnings before all joins and exits

TomAFrench:
We now call `updateBalanceOfPool` on all asset managers prior to joins/exits.

Fixes #695

### DIFF
--- a/pkg/asset-manager-utils/contracts/RebalancingRelayer.sol
+++ b/pkg/asset-manager-utils/contracts/RebalancingRelayer.sol
@@ -95,10 +95,19 @@ contract RebalancingRelayer is IBasePoolRelayer, AssetHelpers {
     ) internal {
         for (uint256 i = 0; i < tokens.length; i++) {
             (uint256 cash, , , address assetManager) = vault.getPoolTokenInfo(poolId, tokens[i]);
-            uint256 cashNeeded = minCashBalances[i];
-            if (assetManager != address(0) && cash < cashNeeded) {
-                // Withdraw the managed balance back to the pool to ensure that the cash covers the withdrawal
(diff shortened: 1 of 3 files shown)


pkg/asset-manager-utils/contracts/RebalancingRelayer.sol
```diff
@@ -95,10 +95,19 @@ contract RebalancingRelayer is IBasePoolRelayer, AssetHelpers {
...
+                uint256 cashNeeded = minCashBalances[i];
+                if (cash < cashNeeded) {
+                    // Withdraw the managed balance back to the pool to ensure that the cash covers the withdrawal
+                    // This will automatically update the vault with the most recent managed balance
+                    IAssetManager(assetManager).capitalOut(poolId, cashNeeded - cash);
+                } else {
+                    // We want to ensure that the pool knows about all asset manager returns
+                    // to avoid a new LP getting a share of returns earned before they joined.
+                    // We then wpdate the vault with the current managed balance manually.
+                    IAssetManager(assetManager).updateBalanceOfPool(poolId);
```
nventuro, line 109: Let's have at least a basic test for this. We could e.g. use the mock pool that emits events with the received balances, and check that, in all cases, the total balance matches the AUM and not the pre join/exit balance.

PR #714, opened 2021-07-07
Stable: Implement meta stable pool math

facuspagnuolo:
We still need to polish the math rounding

### DIFF
--- a/pkg/pool-stable/contracts/meta/StableOracleMath.sol
+++ b/pkg/pool-stable/contracts/meta/StableOracleMath.sol
@@ -55,7 +55,7 @@ contract StableOracleMath is StableMath {
         // a = amp param * n                                                                                         //
         // b = D + a.(S - D)                                                                                         //
         // D = invariant                                                                                             //
-        // S = sum of balances but x,y = 0                                                                           //
+        // S = sum of balances but x,y = 0 since x  and y are the only tokens                                        //
         **************************************************************************************************************/
 
         uint256 invariant = _calculateInvariant(amplificationParameter, _balances(balanceX, balanceY), true);
@@ -63,7 +63,7 @@ contract StableOracleMath is StableMath {
         uint256 a = (amplificationParameter * 2) / _AMP_PRECISION;
         uint256 b = Math.mul(invariant, a).sub(invariant);
 
-        uint256 axy2 = Math.mul(a * 2, balanceX).mulUp(balanceY);
+        uint256 axy2 = Math.mul(a * 2, balanceX).mulUp(balanceY); // n = 2
 
         // dx = a.x.y.2 + a.y^2 - b.y
         uint256 derivativeX = axy2.add(Math.mul(a, balanceY).mulUp(balanceY)).sub(b.mulUp(balanceY));
--- a/pkg/pool-stable/contracts/meta/StableOracleMath.sol
+++ b/pkg/pool-stable/contracts/meta/StableOracleMath.sol
@@ -31,7 +31,7 @@ contract StableOracleMath is StableMath {
         uint256 balanceX,
         uint256 balanceY,
         int256 logBptTotalSupply
-    ) internal view returns (int256 logSpotPrice, int256 logBptPrice) {
+    ) internal pure returns (int256 logSpotPrice, int256 logBptPrice) {
         uint256 spotPrice = _calcSpotPrice(amplificationParameter, balanceX, balanceY);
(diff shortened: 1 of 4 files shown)


pkg/pool-stable/contracts/meta/StableOracleMath.sol
```diff
@@ -0,0 +1,118 @@
...
+        //                             2.a.x.y + a.y^2 + b.y                                                         //
+        // spot price Y/X = - dx/dy = -----------------------                                                        //
+        //                             2.a.x.y + a.x^2 + b.x                                                         //
+        //                                                                                                           //
+        // n = 2                                                                                                     //
+        // a = amp param * n                                                                                         //
+        // b = D + a.(S - D)                                                                                         //
+        // D = invariant                                                                                             //
+        // S = sum of balances but x,y = 0                                                                           //
```
nventuro, line 58:
```suggestion
        // S = sum of balances but x,y = 0 since x  and y are the only tokens                                        //
```

pkg/pool-stable/contracts/meta/StableOracleMath.sol
```diff
@@ -0,0 +1,118 @@
...
+        **************************************************************************************************************/
+
+        uint256 invariant = _calculateInvariant(amplificationParameter, _balances(balanceX, balanceY), true);
+
+        uint256 a = (amplificationParameter * 2) / _AMP_PRECISION;
+        uint256 b = Math.mul(invariant, a).sub(invariant);
+
+        uint256 axy2 = Math.mul(a * 2, balanceX).mulUp(balanceY);
```
nventuro, line 66:
```suggestion
        uint256 axy2 = Math.mul(a * 2, balanceX).mulUp(balanceY); // n = 2
```

pkg/pool-stable/contracts/meta/StableOracleMath.sol
```diff
@@ -0,0 +1,118 @@
...
+
+        // dx = a.x.y.2 + a.y^2 - b.y
+        uint256 derivativeX = axy2.add(Math.mul(a, balanceY).mulUp(balanceY)).sub(b.mulUp(balanceY));
```
nventuro, line 69: `mulDown` is slightly less expensive than `mulUp`, and we don't care about rounding direction (since we introduce a lot of error later on).

pkg/pool-stable/contracts/meta/StableOracleMath.sol
```diff
@@ -0,0 +1,118 @@
...
+        uint256 spotPrice,
+        uint256 balanceX,
+        uint256 balanceY,
+        int256 logBptTotalSupply
+    ) internal pure returns (int256) {
+        /**************************************************************************************************************
+        //                                                                                                           //
+        //              balance X + (spot price Y/X * balance Y)                                                     //
+        // BPT price = ------------------------------------------                                                    //
+        //                          total supply                                                                     //
```
nventuro, line 97: I don't follow this formula, where does the derivation come from?

facuspagnuolo: It's simply transforming the balance of token Y into X using the spot price and then dividing the supply, that basically tells you how much your supply means in X

nventuro: Ah I see, the formula the price confused me. So the spot price is the price of Y in terms of X, meaning it has units [X/Y], and when multiplied by a Y amount you get an equivalent X amount. Thanks!

pkg/pool-stable/contracts/meta/StableOracleMath.sol
```diff
@@ -0,0 +1,118 @@
...
+
+        // The rounding direction is irrelevant as we're about to introduce a much larger error when converting to log
+        // space. We use `divUp` as it prevents the result from being zero, which would make the logarithm revert. A
+        // result of zero is therefore only possible with zero balances, which are prevented via other means.
+        return derivativeX.divUp(derivativeY);
+    }
+
+    /**
+     * @dev Calculates the price of BPT in token. A `logBptTotalSupply` should be the result of calling
+     * `LogCompression.toLowResLog` with the current BPT supply.
```
nventuro, line 82:
```suggestion
     * @dev Calculates the price of BPT in token X. `logBptTotalSupply` should be the result of calling
     * `LogCompression.toLowResLog` with the current BPT supply, and `spotPrice` the price of token
     * Y in token X (obtainable via `_calcSpotPrice()`.
```

nventuro (approved):

PR #715, opened 2021-07-07
Distributors: Fix tests and linter

facuspagnuolo:


gtaschuk (approved):
Thanks - it seems that master moved underneath #703 so the s/whitelist/allowlist and warning intolerances changes that were merged created problems.

`delta` was not deterministic due to the required sorting step on generated asset addresses that create the pools.  I amended your commit to make delta line up with the pool assets.

PR #716, opened 2021-07-07
Pools: Meta stable pool

### LINKED ISSUE #688: Add a MetaStablePool price oracle
MetaStablePools often benefit greatly from having price oracles, so it'd be very desirable to support these. To simplify the design, we'd be constraining them to just have two tokens (due to #665, two-token Stable Pools are cheaper to swap with).

It'd be very desirable to not have to copy-paste the entire StablePool code like we did for WeightedPool2Tokens and just augment the base one with an oracle - we'd likely need to introduce hooks in the right places in the base contracts for this however.

facuspagnuolo:
Fixes #688

### DIFF
--- a/pkg/pool-stable/contracts/meta/MetaStablePool.sol
+++ b/pkg/pool-stable/contracts/meta/MetaStablePool.sol
@@ -60,22 +60,43 @@ contract MetaStablePool is StablePool, StableOracleMath, PoolPriceOracle, IPrice
     // Swap
 
     /**
+     * Override to make sure sender is vault
+     */
+    function onSwap(
+        SwapRequest memory request,
+        uint256[] memory balances,
+        uint256 indexIn,
+        uint256 indexOut
+    ) public virtual override onlyVault(request.poolId) returns (uint256) {
+        return super.onSwap(request, balances, indexIn, indexOut);
+    }
+
+    /**
+     * Override to make sure sender is vault
+     */
+    function onSwap(
+        SwapRequest memory request,
+        uint256 balanceTokenIn,
+        uint256 balanceTokenOut
+    ) public virtual override onlyVault(request.poolId) returns (uint256) {
+        return super.onSwap(request, balanceTokenIn, balanceTokenOut);
+    }
+
+    /**
      * Update price oracle with the pre-swap balances
-     * Note this function does not perform any safety checks about swaps, it relies on upper implementations for that.
      */
     function _onSwapGivenIn(
         SwapRequest memory request,
         uint256[] memory balances,
         uint256 indexIn,
         uint256 indexOut
-    ) internal virtual override whenNotPaused returns (uint256) {
+    ) internal virtual override returns (uint256) {
         _updateOracle(request.lastChangeBlock, balances[0], balances[1]);
         return super._onSwapGivenIn(request, balances, indexIn, indexOut);
     }
 
     /**
      * Update price oracle with the pre-swap balances
-     * Note this function does not perform any safety checks about swaps, it relies on upper implementations for that.
      */
     function _onSwapGivenOut(
         SwapRequest memory request,
@@ -179,7 +200,7 @@ contract MetaStablePool is StablePool, StableOracleMath, PoolPriceOracle, IPrice
             userData
         );
 
-        // If the contract is paused, avoid updating the oracle to minimize code paths taken.
+        // If the contract is paused, the oracle is not updated to avoid extra calculations and reduce potential errors.
         if (_isNotPaused()) {
             _cacheInvariantAndSupply();
         }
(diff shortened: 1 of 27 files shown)


pkg/pool-stable/contracts/meta/MetaStablePool.sol
```diff
@@ -0,0 +1,372 @@
...
+     * Update price oracle with the pre-swap balances
+     * Note this function does not perform any safety checks about swaps, it relies on upper implementations for that.
+     */
+    function _onSwapGivenIn(
+        SwapRequest memory request,
+        uint256[] memory balances,
+        uint256 indexIn,
+        uint256 indexOut
+    ) internal virtual override whenNotPaused returns (uint256) {
+        _updateOracle(request.lastChangeBlock, balances[0], balances[1]);
```
nventuro, line 72: The base classes don't perform any safety checks, in particular, they don't check that `msg.sender == vault` (as WeightedPool2Tokens does). We should add this check either here or in BaseMinimalSwapInfoPool and BaseGeneralPool (I think this might be safer).

pkg/pool-stable/contracts/meta/MetaStablePool.sol
```diff
@@ -0,0 +1,372 @@
...
+    {
+        _require(tokens.length == 2, Errors.NOT_TWO_TOKENS);
+        _setOracleEnabled(oracleEnabled);
+    }
+
+    // Swap
+
+    /**
+     * Update price oracle with the pre-swap balances
+     * Note this function does not perform any safety checks about swaps, it relies on upper implementations for that.
```
nventuro, line 64: I'd delete this line from the comments, it doesn't seem to add much

pkg/pool-stable/contracts/meta/MetaStablePool.sol
```diff
@@ -0,0 +1,372 @@
...
+        virtual
+        override
+        returns (
+            uint256 bptAmountIn,
+            uint256[] memory amountsOut,
+            uint256[] memory dueProtocolFeeAmounts
+        )
+    {
+        // If the contract is paused, swap protocol fee amounts are not charged and the oracle is not updated
+        // to avoid extra calculations and reduce the potential for errors.
```
nventuro, line 211: I'd remove the comment about protocol fees as those don't apply to this `if`

pkg/pool-stable/contracts/meta/MetaStablePool.sol
```diff
@@ -0,0 +1,372 @@
...
+        int256 instantValue = _getInstantValue(variable, _getMiscData().oracleIndex());
+        return LogCompression.fromLowResLog(instantValue);
+    }
+
+    function getTimeWeightedAverage(OracleAverageQuery[] memory queries)
+        external
+        view
+        override
+        returns (uint256[] memory results)
+    {
```
nventuro, line 284: Does it make sense to try to extract these functions into a common module?

pkg/pool-stable/contracts/meta/OracleMiscData.sol
```diff
@@ -0,0 +1,112 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
+
+/**
+ * @dev This module provides an interface to store different pieces of information used by pools with a price oracle.
```
nventuro, line 20: Should we take it out of the `meta` directory?

nventuro (commented):
The tests are _a m a z i n g_

pkg/pool-stable/contracts/meta/MetaStablePool.sol
facuspagnuolo: Let's add it here for now, as it does not apply necessarily for the current stable and weighted pools

pkg/pool-stable/contracts/meta/OracleMiscData.sol
facuspagnuolo: I thought about placing it in the `pool-utils/oracle` dir but it is actually custom to each storage layout, and this one applies only to the stable pool for now. I think we can do the same learning process we did with the `2TokensWeightedPool` if we need it in the future

pkg/pool-stable/contracts/meta/MetaStablePool.sol
facuspagnuolo: Let's tackle it in a separate PR

nventuro (approved):

pkg/pool-stable/contracts/meta/MetaStablePool.sol
```diff
@@ -59,23 +59,44 @@ contract MetaStablePool is StablePool, StableOracleMath, PoolPriceOracle, IPrice
...
     /**
      * Update price oracle with the pre-swap balances
-     * Note this function does not perform any safety checks about swaps, it relies on upper implementations for that.
      */
     function _onSwapGivenIn(
         SwapRequest memory request,
         uint256[] memory balances,
         uint256 indexIn,
         uint256 indexOut
-    ) internal virtual override whenNotPaused returns (uint256) {
```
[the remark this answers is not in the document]
facuspagnuolo: Yes, the stable pool already does that

PR #718, opened 2021-07-09
Initial version of "balancerjs" SDK

TomAFrench:
This PR adds an initial version of a `balancerjs` sdk (I'm open to different names) which aims to package up the utils which we use which may be helpful to outside developers as well as exposing a number of functions which can help with avoiding common pitfalls (token sorting, etc.)

Apologies that this hits a number almost every package we have in the monorepo although most of the changes are purely switching an import so hopefully this shouldn't be overly involved to review.


pkg/solidity-utils/test/ERC20Permit.test.ts
```diff
@@ -86,7 +87,14 @@ describe('ERC20Permit', () => {
           const currentNonce = await token.nonces(holder.address);
           const otherToken = await deploy('ERC20PermitMock', { args: ['Token', 'TKN'] });
 
-          ({ v, r, s, deadline } = await signPermit(otherToken, holder, spender, amount, currentNonce));
+          ({ v, r, s, deadline } = await signPermit(
+            otherToken,
+            holder,
+            spender.address,
+            amount,
+            MAX_DEADLINE,
```
nventuro, line 95: Can't we use the implicit deadline and nonce?

nventuro: The same question applies to usage of the other `sign` methods

pvt/common/setupTests.ts
```diff
@@ -101,28 +101,28 @@ chai.use(function (chai, utils) {
...
-          let actualReason = reasons[actualErrorCode];
-          if (!actualReason) actualReason = 'Could not match a Balancer error message';
+          const actualErrorCode: string = matches[4];
+
+          let actualReason: string;
+          try {
+            actualReason = parseBalancerErrorCode(actualErrorCode);
+          } catch {
+            // If it's not a Balancer error then rethrow
+            if (!error.includes('BAL#')) return error;
```
nventuro, line 111: I'm not sure about this line - could you explain? Shouldn't it be before the `try`, and `throw` instead of `return`?

pvt/common/setupTests.ts
```diff
@@ -101,28 +101,28 @@ chai.use(function (chai, utils) {
...
 
-          // If there is no balancer error matching the expected revert reason re-throw the error
-          // @ts-ignore
-          const expectedError = Object.entries(reasons).find(([, value]) => value == expectedReason);
-          if (!expectedError) {
+          try {
+            // Assert the error code matched the actual reason
+            const expectedErrorCode = encodeBalancerErrorCode(expectedReason);
+            const message = `Expected transaction to be reverted with BAL#${expectedErrorCode} (${expectedReason}), but other exception was thrown: Error: VM Exception while processing transaction: revert BAL#${actualErrorCode} (${actualReason})`;
+            expect(actualErrorCode).to.be.equal(expectedErrorCode, message);
```
nventuro, line 120: I think this might also be wrong - we don't want to catch the exception that `except` might throw: I think we don't want to put this inside the `try` block.

nventuro (commented):
Love the initiative, and also how usage of this looks like with proper namespacing!

pkg/deployments/tasks/20210624-stable-pool/test/task.fork.ts
```diff
@@ -2,17 +2,16 @@ import hre from 'hardhat';
 import { expect } from 'chai';
 import { Contract } from 'ethers';
 
+import { encodeJoinStablePool, StablePoolJoinKind, SwapKind } from '@balancer-labs/balancer-js';
```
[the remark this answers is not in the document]
TomAFrench: Unfortunately yes, changes need to be built before they take effect from now on.

pkg/balancer-js/src/pool-weighted/exitEncoding.ts
```diff
@@ -0,0 +1,49 @@
+import { defaultAbiCoder } from '@ethersproject/abi';
+import { BigNumberish } from '@ethersproject/bignumber';
+
+export enum WeightedPoolExitKind {
```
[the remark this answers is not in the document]
TomAFrench: Just replacing some string literals for the different join/exit kinds with enums.

pkg/balancer-js/src/pool-utils/poolId.ts
```diff
@@ -0,0 +1,53 @@
+import { BigNumber } from '@ethersproject/bignumber';
+import { PoolSpecialization } from '../types';
+import invariant from 'tiny-invariant';
+
+/**
+ * Splits a poolId into its components, i.e. pool address, pool specialization and its nonce
+ * @param poolId - a bytes32 string of the pool's ID
+ * @returns an object with the decomposed poolId
+ */
+export const splitPoolId = (
```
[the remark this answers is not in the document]
TomAFrench: Yes, we're these are new but simple enough that I thought I'd chuck them in anyway.

pkg/balancer-js/rollup.config.ts
```diff
@@ -0,0 +1,31 @@
...
+import pkg from './package.json';
+
+const external = [
+  '@ethersproject/abi',
+  '@ethersproject/abstract-signer',
+  '@ethersproject/address',
+  '@ethersproject/bignumber',
+  '@ethersproject/bytes',
+  '@ethersproject/constants',
+  '@ethersproject/contracts',
```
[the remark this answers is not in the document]
TomAFrench: Correct. The reasoning behind this is that we want to use the copy of ethers that the user provides (to ensure consistency). We don't have the same situation with `tiny-invariant` as they likely wouldn't have it installed anyway. Better to bundle it in and make them not have to think about it.

pkg/solidity-utils/test/ERC20Permit.test.ts
TomAFrench: In this case yes, we shouldn't enforce their use though.

pkg/solidity-utils/test/ERC20Permit.test.ts
```diff
@@ -36,7 +37,7 @@ describe('ERC20Permit', () => {
 
     it('accepts holder signature', async function () {
       const previousNonce = await token.nonces(holder.address);
-      const { v, r, s } = await signPermit(token, holder, spender, amount);
+      const { v, r, s } = await signPermit(token, holder, spender.address, amount);
```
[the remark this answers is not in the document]
TomAFrench: We'd then need to include the `SignerWithAddress` type in the npm package (and so `hardhat-deploy`) afaik. I looked into adding `Signer` but I then ran into the issue that it doesn't include a way for getting the corresponding address (which makes sense considering the existence of `SignerWithAddress`)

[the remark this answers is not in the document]
TomAFrench: We only have to do this in a couple of instances so I think it's a small price to pay.

pkg/balancer-js/rollup.config.ts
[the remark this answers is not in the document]
nventuro: Shouldn't we remove `tiny-invariant` from `dependencies` then? (and add it to dev-dependencies)

pkg/solidity-utils/test/ERC20Permit.test.ts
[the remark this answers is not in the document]
nventuro: `SignerWithAddress` comes from hardhat-ethers, not hardhat-deploy, but I see your point. I'd still use `Signer`, which is a basic ethers type. We should be able to do `address = await signer.getAddress()`.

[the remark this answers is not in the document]
TomAFrench: Fair enough. I was also a little concerned about bloating the argument type on this function as I doubt that anyone will use a `Signer` except us for testing.

pvt/common/setupTests.ts
TomAFrench: Yeah I think this was relying on some throwing behaviour I had in `parseBalancerErrorCode` which I then stripped out. I've changed this up to make it more explicit. (`return` instead of `throw` is just a brainfart)

pvt/common/setupTests.ts
TomAFrench: Yeah, I wasn't thinking about the `expect`'s error. I've pulled that out to ensure that throws correctly.

pkg/balancer-js/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@balancer-labs/balancer-js",
```
nventuro, line 2: Is the name final?

pkg/balancer-js/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "version": "1.0.0",
+  "description": "",
```
nventuro, line 4:
```suggestion
  "description": "JavaScript utilities for interacting with the Balancer Protocol V2.",
```

pkg/balancer-js/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "homepage": "https://github.com/balancer-labs/balancer-v2-monorepo/tree/master/pkg/balancer-js#readme",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/balancer-labs/balancer-v2-monorepo.git",
+    "directory": "pkg/balancer-js"
+  },
+  "bugs": {
+    "url": "https://github.com/balancer-labs/balancer-v2-monorepo/issues"
+  },
+  "main": "dist/index.js",
```
nventuro, line 15:
```suggestion
  "files": [
    "dist/**/*"
  ],
  "main": "dist/index.js",
```

nventuro: No point in also publishing the full source, configs, etc.

pkg/balancer-js/README.md
```diff
@@ -0,0 +1 @@
+# SDK
```
nventuro, line 1: Want to take a stab at a readme so we can publish this? I can otherwise write something up myself.

pkg/balancer-js/src/utils/signatures.ts
```diff
@@ -0,0 +1,143 @@
...
+  BatchSwap = 'BatchSwap',
+  SetRelayerApproval = 'SetRelayerApproval',
+}
+
+export type Account = string | Signer | Contract;
+
+export async function accountToAddress(account: Account): Promise<string> {
+  if (typeof account == 'string') return account;
+  if (Signer.isSigner(account)) return account.getAddress();
+  if (account.address) return account.address;
```
nventuro, line 20:
```suggestion
  if (account.address) return account.address;
  if (Signer.isSigner(account)) return account.getAddress();
```

Subtle, but this will save RPCs when using SignerWithAddress

TomAFrench: This would need some nasty type assertions to work and calling `getAddress` on a `SignerWithAddress` just returns `this.address` so the current code is equivalent.

pkg/balancer-js/README.md
TomAFrench: Sure, I've popped something in however I'm expecting us to add more to this package before we publicise it so it'll evolve over time.

pkg/balancer-js/package.json
TomAFrench: It feels like a fairly standard choice. I've asked in #integrations-internal if anyone has strong feelings but unless anyone objects then I think we can say it's final.

nventuro (approved):

PR #719, opened 2021-07-10
Distributor: MultiRewards Exiter

gtaschuk:
This
a) simplifies accounting for the reward rollup by removing unnecessary rewarder parameter in mapping
b) Implements an `Exiter` callback contract that exits a pool with all the bpt sent to it and passes tokens on to the LP

Nearly the same as this, except the rewards are sent to the LP directly from the staking contract (open for debate)
staking-exit-callback

This opens the door for a lot of interesting use cases, like being able to seamlessly transfer assets from one pool to another without relayers, or to exit a pool and liquidate to WETH or something (provided a list of swaps).

I propose we focus more on the `MultiRewards#exitWithCallback` function for the purpose of this PR and discuss what exactly we want out of the callback contract as we think through UX - we may end up with a lot of these generic callback contracts to fit different use cases...

### DIFF
--- a/pkg/distributors/contracts/Exiter.sol
+++ b/pkg/distributors/contracts/Exiter.sol
@@ -1,0 +1,77 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
+
+import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
+import "@balancer-labs/v2-vault/contracts/interfaces/IBasePool.sol";
+import "@balancer-labs/v2-pool-weighted/contracts/BaseWeightedPool.sol";
+
+import "./PoolTokenManipulator.sol";
+
+contract Exiter is PoolTokenManipulator {
+    constructor(IVault _vault) PoolTokenManipulator(_vault) {
+        // solhint-disable-previous-line no-empty-blocks
+    }
+
+    /**
+     * @notice Exits specified pool with all bpt
+     * @param recipient - the recipient of the pool tokens
+     * @param pools - The pools to exit from (addresses)
+     */
+    function callback(IERC20[] calldata pools, address payable recipient) external {
+        for (uint256 p; p < pools.length; p++) {
+            address poolAddress = address(pools[p]);
+
+            IBasePool poolContract = IBasePool(poolAddress);
+            bytes32 poolId = poolContract.getPoolId();
+            ensurePoolTokenSetSaved(poolId);
+
+            IERC20 pool = IERC20(poolAddress);
+            _exitPool(pool, poolId, recipient);
+        }
+    }
+
+    /**
+     * @notice Exits the pool
+     * Exiting to a single token would look like:
+     * bytes memory userData = abi.encode(
+     * BaseWeightedPool.ExitKind.EXACT_BPT_IN_FOR_ONE_TOKEN_OUT,
+     * bptBalance,
+     * tokenIndexOut
+     * );
+     */
+    function _exitPool(
+        IERC20 pool,
+        bytes32 poolId,
+        address payable recipient
+    ) internal {
+        IAsset[] memory assets = _getAssets(poolId);
+        uint256[] memory minAmountsOut = new uint256[](assets.length);
+
+        uint256 bptAmountIn = pool.balanceOf(address(this));
+
+        bytes memory userData = abi.encode(BaseWeightedPool.ExitKind.EXACT_BPT_IN_FOR_TOKENS_OUT, bptAmountIn);
+        bool toInternalBalance = false;
+
+        IVault.ExitPoolRequest memory request = IVault.ExitPoolRequest(
+            assets,
+            minAmountsOut,
+            userData,
+            toInternalBalance
+        );
+        vault.exitPool(poolId, address(this), recipient, request);
+    }
+}
--- a/pkg/distributors/contracts/Exiter.sol
+++ b/pkg/distributors/contracts/Exiter.sol
@@ -20,27 +20,36 @@ import "@balancer-labs/v2-vault/contracts/interfaces/IBasePool.sol";
 import "@balancer-labs/v2-pool-weighted/contracts/BaseWeightedPool.sol";
 
 import "./PoolTokenManipulator.sol";
+import "./interfaces/IRewardsCallback.sol";
 
-contract Exiter is PoolTokenManipulator {
+contract Exiter is PoolTokenManipulator, IRewardsCallback {
     constructor(IVault _vault) PoolTokenManipulator(_vault) {
         // solhint-disable-previous-line no-empty-blocks
     }
 
+    struct CallbackParams {
+        address[] pools;
+        address payable recipient;
+    }
+
     /**
      * @notice Exits specified pool with all bpt
-     * @param recipient - the recipient of the pool tokens
-     * @param pools - The pools to exit from (addresses)
+     * @param callbackData are the encoded function arguments:
+     * recipient - the recipient of the pool tokens
+     * pools - The pools to exit from (addresses)
      */
-    function callback(IERC20[] calldata pools, address payable recipient) external {
-        for (uint256 p; p < pools.length; p++) {
-            address poolAddress = address(pools[p]);
+    function callback(bytes calldata callbackData) external override {
+        CallbackParams memory params = abi.decode(callbackData, (CallbackParams));
+
+        for (uint256 p; p < params.pools.length; p++) {
+            address poolAddress = params.pools[p];
 
             IBasePool poolContract = IBasePool(poolAddress);
             bytes32 poolId = poolContract.getPoolId();
             ensurePoolTokenSetSaved(poolId);
 
             IERC20 pool = IERC20(poolAddress);
-            _exitPool(pool, poolId, recipient);
+            _exitPool(pool, poolId, params.recipient);
         }
     }
 
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -277,12 +277,16 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
         stake(pool, amount, recipient);
     }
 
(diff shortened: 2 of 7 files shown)


pkg/distributors/contracts/PoolTokenManipulator.sol
```diff
@@ -0,0 +1,68 @@
...
+    IVault public immutable vault;
+
+    constructor(IVault _vault) {
+        vault = _vault;
+    }
+
+    mapping(bytes32 => EnumerableSet.AddressSet) private _poolTokenSets;
+    mapping(bytes32 => bool) private __poolTokenSetSaved;
+
+    function ensurePoolTokenSetSaved(bytes32 poolId) public {
```
TomAFrench, line 33: This could run into trouble when we have pools which don't have a fixed set of tokens, e.g. investment pools.

pkg/distributors/contracts/Exiter.sol
```diff
@@ -0,0 +1,78 @@
...
+        bytes32 poolId,
+        address payable recipient
+    ) internal {
+        IAsset[] memory assets = _getAssets(poolId);
+        uint256[] memory minAmountsOut = new uint256[](assets.length);
+        minAmountsOut[0] = 1;
+        minAmountsOut[1] = 1;
+
+        uint256 bptAmountIn = pool.balanceOf(address(this));
+        console.log("HERERE", recipient, bptAmountIn);
```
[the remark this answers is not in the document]
gtaschuk: 🙈 Oops.  Removed them

pkg/distributors/contracts/PoolTokenManipulator.sol
gtaschuk: I would like to avoid querying them every time.  This exiter contract is just an example of what could be done with the bpt when a user unstakes, so I'm not too worried about future proofing it.  Perhaps we add a method "refreshPoolTokenSet" that can be called manually when tokens are added?

pkg/distributors/contracts/Exiter.sol
```diff
@@ -0,0 +1,77 @@
...
+     * );
+     */
+    function _exitPool(
+        IERC20 pool,
+        bytes32 poolId,
+        address payable recipient
+    ) internal {
+        IAsset[] memory assets = _getAssets(poolId);
+        uint256[] memory minAmountsOut = new uint256[](assets.length);
+        minAmountsOut[0] = 1;
```
dmf7z, line 61: This works only if the pool has two tokens? I do not think you need to sent them to 1, you can leave all the array with zero.

gtaschuk: @dmf7z Removed them.  Would there be a case where a user would improperly construct this txn and end up with 0 tokens?

pkg/distributors/contracts/PoolTokenManipulator.sol
TomAFrench: Ah yeah, this should be fine as the worst case is we deploy a new version and there's no lock in on this one.

pkg/distributors/test/Exiter.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  let poolTokens: TokenList;
+  let vault: Contract;
+  let stakingContract: Contract;
+  let callbackContract: Contract;
+  let rewardToken: Token;
+  let pool: Contract;
+
+  before('deploy base contracts', async () => {
+    [, , lp, mockAssetManager] = await ethers.getSigners();
```
TomAFrench, line 27:
```suggestion
    [, , lp, rewarder] = await ethers.getSigners();
```

afaict this isn't actually an asset manager (as it's explicitly allowlisted as a rewarder) plus this is more descriptive

pkg/distributors/contracts/PoolTokenManipulator.sol
```diff
@@ -0,0 +1,68 @@
...
+    using EnumerableSet for EnumerableSet.AddressSet;
+
+    IVault public immutable vault;
+
+    constructor(IVault _vault) {
+        vault = _vault;
+    }
+
+    mapping(bytes32 => EnumerableSet.AddressSet) private _poolTokenSets;
+    mapping(bytes32 => bool) private __poolTokenSetSaved;
```
TomAFrench, line 31:
```suggestion
    mapping(bytes32 => bool) private _poolTokenSetSaved;
```

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -381,11 +386,29 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
...
+     *         and transfers them all the rewards
+     */
+    function exitWithCallback(
+        IERC20[] calldata pools,
+        address callbackContract,
+        bytes calldata callbackData
+    ) public returns (bytes memory) {
+        for (uint256 p; p < pools.length; p++) {
+            IERC20 pool = pools[p];
+            unstake(pool, _balances[pool][msg.sender], callbackContract);
```
TomAFrench, line 405:
```suggestion
            unstake(pools[p], _balances[pool][msg.sender], callbackContract);
```

I would inline this as I don't think it really hampers readability at all. Feel free to ignore.

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -381,11 +386,29 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
     function exit(IERC20[] calldata pools) external {
         for (uint256 p; p < pools.length; p++) {
             IERC20 pool = pools[p];
-            unstake(pool, _balances[pool][msg.sender]);
+            unstake(pool, _balances[pool][msg.sender], msg.sender);
```
TomAFrench, line 389: Similarly here

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -367,12 +374,10 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
...
         address callbackContract,
         bytes calldata callbackData
-    ) public nonReentrant {
+    ) public nonReentrant returns (bytes memory) {
         _getReward(pools, payable(callbackContract), true);
 
-        (bool success, ) = callbackContract.call(callbackData);
-        // solhint-disable-previous-line avoid-low-level-calls
-        require(success, "callback failed");
+        return Address.functionCall(callbackContract, callbackData);
```
TomAFrench, line 380: This is a sidenote to the Exit contract but doesn't this allow anyone to drain MultiRewards? I can just encode a transfer of all of the contract's assets to my own address. We need to have some controls on here.

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -381,11 +386,29 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
...
+        address callbackContract,
+        bytes calldata callbackData
+    ) public returns (bytes memory) {
+        for (uint256 p; p < pools.length; p++) {
+            IERC20 pool = pools[p];
+            unstake(pool, _balances[pool][msg.sender], callbackContract);
+        }
+        return Address.functionCall(callbackContract, callbackData);
+
+        getReward(pools);
```
TomAFrench, line 409: This code is unreachable.

TomAFrench (changes requested):
I've made a couple of minor notes although I'm quite concerned about how open the callbacks are to abuse. Have there been any discussions on how to lock it down to protect `MultiRewards`' assets?

pkg/distributors/contracts/MultiRewards.sol
gtaschuk: So it would actually be  `unstake(pools[p], _balances[pools[p]][msg.sender], callbackContract);` if inlined, which I find a bit confusing

pkg/distributors/contracts/MerkleRedeem.sol
```diff
@@ -138,16 +138,16 @@ contract MerkleRedeem is IDistributor, Ownable {
...
+        IRewardsCallback callbackContract,
         bytes calldata callbackData,
         Claim[] memory claims
-    ) external returns (bytes memory) {
+    ) external {
         require(msg.sender == liquidityProvider, "user must claim own balance");
         uint256 totalBalance = _processClaims(liquidityProvider, claims);
 
-        _disburseToInternalBalance(callbackContract, totalBalance);
+        _disburseToInternalBalance(payable(address(callbackContract)), totalBalance);
```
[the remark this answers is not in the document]
gtaschuk: It has to be payable to receive IAssets as internal balance: https://github.com/balancer-labs/balancer-v2-monorepo/blob/cf6576db6cab7a7aa731d74dcdff1c4babb9a935/pkg/vault/contracts/UserBalance.sol#L72

pkg/distributors/test/Exiter.test.ts
TomAFrench: It looks like we've updated this in some of the other files but not here

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -158,6 +164,9 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
         return _balances[pool][account];
     }
 
+    /**
+     * @notice this time is used when determining up until what time a reward has been accounted for
```
TomAFrench, line 168:
```suggestion
     * @notice This time is used when determining up until what time a reward has been accounted for
```

We should capitalise all of these comments

pkg/distributors/test/Exiter.test.ts
```diff
@@ -0,0 +1,98 @@
...
+      await stakingContract
+        .connect(mockAssetManager)
+        .allowlistRewarder(pool.address, rewardToken.address, mockAssetManager.address);
+      await stakingContract.connect(mockAssetManager).addReward(pool.address, rewardToken.address, rewardsDuration);
+
+      const bptBalance = await pool.balanceOf(lp.address);
+
+      await pool.connect(lp).approve(stakingContract.address, bptBalance);
+
+      await stakingContract.connect(lp)['stake(address,uint256)'](pool.address, bptBalance);
```
TomAFrench, line 57: I know that Nico really isn't a fan of overloading function names because of this. Have you talked about possibly renaming the three arg version `stakeFor` or something?

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -380,9 +386,26 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
...
+     *         and transfers them all the rewards
+     */
+    function exitWithCallback(
+        IERC20[] calldata pools,
+        IRewardsCallback callbackContract,
+        bytes calldata callbackData
+    ) public {
+        for (uint256 p; p < pools.length; p++) {
+            IERC20 pool = pools[p];
+            unstake(pool, _balances[pool][msg.sender], payable(address(callbackContract)));
```
TomAFrench, line 405:
```suggestion
            unstake(pool, _balances[pool][msg.sender], address(callbackContract));
```

This doesn't need to payable afaict

pkg/distributors/test/Exiter.test.ts
gtaschuk: Seems reasonable, though fwiw I see this as a problem with ethers, not a problem with the choice to overload.  Made the change

TomAFrench (commented):
LGTM.

TomAFrench (approved):

PR #720, opened 2021-07-12
Add initial implementation of BatchRelayer

TomAFrench:
This PR adds a relayer with the ability for `joinAndSwap`s, `joinAndStake`s, and `swapAndExit`s to handle metastable pools

~Draft PR in the meantime as I build out tests~


pkg/standalone-utils/contracts/BatchRelayer.sol
```diff
@@ -0,0 +1,168 @@
...
+        getVault().exitPool(poolId, msg.sender, recipient, request);
+        _sweepETH();
+    }
+
+    function _sweepETH() private {
+        uint256 remainingEth = address(this).balance;
+        if (remainingEth > 0) {
+            msg.sender.sendValue(remainingEth);
+        }
+    }
```
nventuro, line 167: I'm worried this and other reliance on global state might open the door to reentrancy vulns - we might be fine since the Vault is non-reentrant, but I'd still consider adding the `nonReentrant` modifier to external fucntions.

pkg/standalone-utils/contracts/BatchRelayer.sol
```diff
@@ -0,0 +1,168 @@
...
+        IERC20 bpt = IERC20(_getPoolAddress(poolId));
+        uint256 bptAmount = bpt.balanceOf(address(this));
+
+        // If necessary, give staking contract allowance to take BPT
+        if (bpt.allowance(address(this), address(getStakingContract())) < bptAmount) {
+            bpt.approve(address(getStakingContract()), type(uint256).max);
+        }
+
+        getStakingContract().stake(bpt, bptAmount, recipient);
+        _sweepETH();
```
nventuro, line 106: This looks very straightforward. I'm happy.

pkg/standalone-utils/contracts/BatchRelayer.sol
```diff
@@ -0,0 +1,168 @@
...
+        IVault.BatchSwapStep[] calldata swaps,
+        IAsset[] calldata assets,
+        int256[] calldata limits,
+        uint256 deadline
+    ) external payable {
+        // We can't output tokens to the user's internal balance
+        // as they need to have BPT on their address for the exit
+        IVault.FundManagement memory funds = IVault.FundManagement({
+            sender: msg.sender,
+            fromInternalBalance: false,
```
nventuro, line 123: Maybe add a comment mentioning this has to be false as it is incompatible with sending ETH

pkg/standalone-utils/contracts/BatchRelayer.sol
```diff
@@ -0,0 +1,168 @@
...
+            recipient: msg.sender,
```
nventuro, line 124: I'd use `address(this)` - both will work, but this seems more direct and less cryptic by not relying on relayer powers for the exit.

pkg/standalone-utils/contracts/BatchRelayer.sol
```diff
@@ -0,0 +1,168 @@
...
+                }
+            }
+
+            // Here we overwrite the bptAmountIn field of an `exactBptInForTokenOut` exit with the output of the swap
+            // We are mutating the userData field within request so no explicit assignment occurs
+            bytes memory userData = request.userData;
+            // solhint-disable-next-line no-inline-assembly
+            assembly {
+                mstore(add(userData, _BPT_AMOUNT_IN_OFFSET), bptAmount)
+            }
```
nventuro, line 155: This seems quite fragile - wouldn't it be better to do the `abi.encode` for `exactBPTIn` directly here, instead of assuming that's what we received and modifying it?

pkg/standalone-utils/contracts/BatchRelayer.sol
TomAFrench: The reasoning behind this is that we can ensure that should the user provide a swap which results in an output of other tokens in addition to the BPT then this will not be locked on the relayer.

pkg/standalone-utils/contracts/BatchRelayer.sol
TomAFrench: I'm not sure that this adds any additional fragility over what's existing in the pool for decoding the userData. `abi.decode` makes the same assumptions about what the format of this field is as we are doing here.

That said, I've replaced this with a more explicit abi.decode/abi.encode format for readability.

TomAFrench:
Closing this as superseded by #756

PR #721, opened 2021-07-14
Distributor: MultiRewards rewards scheduler

gtaschuk:
Allows a rewarder to schedule a reward for the MultiRewards contract

Once the reward startTime has passed, anyone can poke the contract to notify the MultiRewards contract of the reward

[OpenZeppelin Defender](https://openzeppelin.com/defender/) willl interact with this contract to start scheduled rewards

reward-scheduler

### DIFF
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -75,6 +75,8 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
     // pool -> user -> bpt balance staked
     mapping(IERC20 => mapping(address => uint256)) private _balances;
 
+    address public rewardsScheduler;
+
     /* ========== CONSTRUCTOR ========== */
 
     constructor(IVault _vault)
@@ -83,6 +85,10 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
         MultiRewardsAuthorization(_vault)
     {
         // solhint-disable-previous-line no-empty-blocks
+    }
+
+    function setRewardsScheduler(address rs) public onlyOwner {
+        rewardsScheduler = rs;
     }
 
     /**
@@ -382,6 +388,24 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
      * @param rewardsToken - the token to deposit into staking contract for distribution
      * @param reward - the amount of tokens to deposit
      */
+    function startScheduledReward(
+        IERC20 pool,
+        IERC20 rewardsToken,
+        uint256 reward,
+        address rewarder
+    ) external override updateReward(pool, address(0)) {
+        require(
+            rewarder == msg.sender || msg.sender == rewardsScheduler,
+            "Rewarder must be sender, or rewards scheduler"
+        );
+        _startReward(pool, rewardsToken, reward, rewarder);
+    }
+
+    /**
+     * @notice Allows a rewards distributor to deposit more tokens to be distributed as rewards
+     * @param rewardsToken - the token to deposit into staking contract for distribution
+     * @param reward - the amount of tokens to deposit
+     */
     function notifyRewardAmount(
         IERC20 pool,
         IERC20 rewardsToken,
@@ -389,9 +413,19 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
     ) external override updateReward(pool, address(0)) {
         require(_rewarders[pool][rewardsToken].contains(msg.sender), "Reward must be configured with addReward");
 
+        _startReward(pool, rewardsToken, reward, msg.sender);
+    }
+
+    function _startReward(
+        IERC20 pool,
+        IERC20 rewardsToken,
+        uint256 reward,
+        address rewarder
+    ) internal {
         // handle the transfer of reward tokens via `safeTransferFrom` to reduce the number
         // of transactions required and ensure correctness of the reward amount
-        rewardsToken.safeTransferFrom(msg.sender, address(this), reward);
+        address funder = msg.sender == rewardsScheduler ? rewardsScheduler : rewarder;
+        rewardsToken.safeTransferFrom(funder, address(this), reward);
 
         IVault.UserBalanceOp[] memory ops = new IVault.UserBalanceOp[](1);
 
@@ -405,23 +439,23 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
 
         getVault().manageUserBalance(ops);
 
-        if (block.timestamp >= rewardData[pool][msg.sender][rewardsToken].periodFinish) {
-            rewardData[pool][msg.sender][rewardsToken].rewardRate = Math.divDown(
+        if (block.timestamp >= rewardData[pool][rewarder][rewardsToken].periodFinish) {
+            rewardData[pool][rewarder][rewardsToken].rewardRate = Math.divDown(
                 reward,
-                rewardData[pool][msg.sender][rewardsToken].rewardsDuration
+                rewardData[pool][rewarder][rewardsToken].rewardsDuration
             );
         } else {
-            uint256 remaining = rewardData[pool][msg.sender][rewardsToken].periodFinish.sub(block.timestamp);
-            uint256 leftover = remaining.mulDown(rewardData[pool][msg.sender][rewardsToken].rewardRate);
-            rewardData[pool][msg.sender][rewardsToken].rewardRate = Math.divDown(
+            uint256 remaining = rewardData[pool][rewarder][rewardsToken].periodFinish.sub(block.timestamp);
+            uint256 leftover = remaining.mulDown(rewardData[pool][rewarder][rewardsToken].rewardRate);
+            rewardData[pool][rewarder][rewardsToken].rewardRate = Math.divDown(
                 reward.add(leftover),
-                rewardData[pool][msg.sender][rewardsToken].rewardsDuration
+                rewardData[pool][rewarder][rewardsToken].rewardsDuration
             );
         }
 
-        rewardData[pool][msg.sender][rewardsToken].lastUpdateTime = block.timestamp;
-        rewardData[pool][msg.sender][rewardsToken].periodFinish = block.timestamp.add(
-            rewardData[pool][msg.sender][rewardsToken].rewardsDuration
+        rewardData[pool][rewarder][rewardsToken].lastUpdateTime = block.timestamp;
+        rewardData[pool][rewarder][rewardsToken].periodFinish = block.timestamp.add(
+            rewardData[pool][rewarder][rewardsToken].rewardsDuration
         );
         emit RewardAdded(address(rewardsToken), reward);
     }
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -101,6 +101,14 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
         _allowlistRewarder(pool, rewardsToken, rewarder);
     }
 
+    function isAllowlistedRewarder(
+        IERC20 pool,
+        IERC20 rewardsToken,
+        address rewarder
+    ) public override view returns (bool) {
+        return _isAllowlistedRewarder(pool, rewardsToken, rewarder);
+    }
+
     /**
      * @notice Adds a new reward token to be distributed
      * @param pool - The bpt of the pool that will receive rewards
@@ -401,7 +409,7 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
             "Rewarder must be sender, or rewards scheduler"
         );
 
-        require(_rewarders[pool][rewardsToken].contains(msg.sender), "Reward must be configured with addReward");
+        require(_rewarders[pool][rewardsToken].contains(rewarder), "Reward must be configured with addReward");
 
         // handle the transfer of reward tokens via `safeTransferFrom` to reduce the number
         // of transactions required and ensure correctness of the reward amount
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -413,6 +413,8 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
 
         // handle the transfer of reward tokens via `safeTransferFrom` to reduce the number
         // of transactions required and ensure correctness of the reward amount
+        // Tokens always come from msg.sender because either `msg.sender == rewarder`
+        // or the`rewardsScheduler` is holding tokens on behalf of the `rewarder`
         rewardsToken.safeTransferFrom(msg.sender, address(this), reward);
 
         IVault.UserBalanceOp[] memory ops = new IVault.UserBalanceOp[](1);
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -105,7 +105,7 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
         IERC20 pool,
         IERC20 rewardsToken,
         address rewarder
-    ) public override view returns (bool) {
+    ) public view override returns (bool) {
         return _isAllowlistedRewarder(pool, rewardsToken, rewarder);
     }
 
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -397,7 +397,7 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
         address rewarder
     ) external override updateReward(pool, address(0)) {
         require(
-            rewarder == msg.sender || msg.sender == address(rewardsScheduler),
+            msg.sender == rewarder || msg.sender == address(rewardsScheduler),
             "Rewarder must be sender, or rewards scheduler"
         );
 
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -26,6 +26,8 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/IERC20Permit.sol
 import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
 import "@balancer-labs/v2-vault/contracts/interfaces/IAsset.sol";
 
+import "./RewardsScheduler.sol";
+
 import "./interfaces/IMultiRewards.sol";
 import "./interfaces/IDistributorCallback.sol";
 import "./interfaces/IDistributor.sol";
@@ -75,7 +77,7 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
     // pool -> user -> bpt balance staked
     mapping(IERC20 => mapping(address => uint256)) private _balances;
 
-    address public rewardsScheduler;
+    RewardsScheduler public immutable rewardsScheduler;
 
     /* ========== CONSTRUCTOR ========== */
 
@@ -85,10 +87,7 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
         MultiRewardsAuthorization(_vault)
     {
         // solhint-disable-previous-line no-empty-blocks
-    }
-
-    function setRewardsScheduler(address rs) public onlyOwner {
-        rewardsScheduler = rs;
+        rewardsScheduler = new RewardsScheduler();
     }
 
     /**
@@ -397,7 +396,7 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
         address rewarder
     ) external override updateReward(pool, address(0)) {
         require(
-            rewarder == msg.sender || msg.sender == rewardsScheduler,
+            rewarder == msg.sender || msg.sender == address(rewardsScheduler),
             "Rewarder must be sender, or rewards scheduler"
         );
 
@@ -405,8 +404,7 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
 
         // handle the transfer of reward tokens via `safeTransferFrom` to reduce the number
         // of transactions required and ensure correctness of the reward amount
-        address funder = (msg.sender == rewardsScheduler) ? rewardsScheduler : msg.sender;
-        rewardsToken.safeTransferFrom(funder, address(this), reward);
+        rewardsToken.safeTransferFrom(msg.sender, address(this), reward);
 
         IVault.UserBalanceOp[] memory ops = new IVault.UserBalanceOp[](1);
 
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -123,14 +123,14 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
     /* ========== VIEWS ========== */
 
     /**
-     * @notice Total supply of a staking token being added
+     * @notice Total supply of a pools bpt that has been added
      */
     function totalSupply(IERC20 pool) external view returns (uint256) {
         return _totalSupply[pool];
     }
 
     /**
-     * @notice The balance of a staking token than `account` has staked
+     * @notice The balance of a pools bpt that `account` has staked
      */
     function balanceOf(IERC20 pool, address account) external view returns (uint256) {
         return _balances[pool][account];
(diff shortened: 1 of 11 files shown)


pkg/distributors/contracts/RewardsScheduler.sol
```diff
@@ -0,0 +1,118 @@
...
+    IMultiRewards public multirewards;
+
+    constructor(IMultiRewards _multirewards) {
+        multirewards = _multirewards;
+    }
+
+    struct ScheduledReward {
+        IERC20 pool;
+        IERC20 rewardsToken;
+        uint256 startTime;
```
nventuro, line 38:
```suggestion
        uint256 startTime; // A start time of 0 is used as a sentinel value of the reward not existing
```

nventuro: (we could also add a `bool`, which would be more explicit and packed in the same storage slot anyway since we're using a struct. also we don't really care about gas costs that much here)

pkg/distributors/contracts/RewardsScheduler.sol
```diff
@@ -0,0 +1,118 @@
...
+            rewardsToken: rewardsToken,
+            rewarder: msg.sender,
+            amount: amount,
+            startTime: startTime
+        });
+
+        emit RewardScheduled(rewardId, msg.sender, address(rewardsToken), startTime);
+    }
+
+    function unscheduleReward(bytes32 rewardId) external {
```
nventuro, line 104: I'd get rid of this function, mostly to reduce complexity and the exploitable surface that arises from the fact that we're transferring tokens out of the scheduler. Yes, it can be useful to unschedule a scheduled reward, but it feels like a weird exception in the irreversible blockchain context, and I don't think it's worth the added risk.

pkg/distributors/contracts/RewardsScheduler.sol
```diff
@@ -0,0 +1,118 @@
...
+            require(scheduledReward.startTime != 0, "reward has not been created");
+            require(scheduledReward.startTime <= block.timestamp, "reward cannot be started");
+
+            multirewards.startScheduledReward(
+                scheduledReward.pool,
+                scheduledReward.rewardsToken,
+                scheduledReward.amount,
+                scheduledReward.rewarder
+            );
+            delete _rewards[rewardId];
```
nventuro, line 62: I'd not delete this entry, and instead a) mark it as completed (maybe have an enum with `uninit = 0, pending, completed` instead of using start time == 0), and b) add a getter `getRewardInfo(rewardId)` so that past rewards can be more easily queried.

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -70,6 +70,8 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
     mapping(IERC20 => uint256) private _totalSupply;
     mapping(IERC20 => mapping(address => uint256)) private _balances;
 
+    address public rewardsScheduler;
```
nventuro, line 73:
```suggestion
    address public immutable rewardsScheduler;
```

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -81,6 +83,10 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
         _;
     }
 
+    function setRewardsScheduler(address rs) public onlyOwner {
+        rewardsScheduler = rs;
+    }
```
nventuro, line 88: If the scheduler is immutable (and I think it should be immutable - I can't imagine much of a reason why we'd want ot upgrade and I'd avoid this extra complexity), what we could do instead is the following, which will be simpler and cheaper:

```
contract MultRewards
  constructor() {
    scheduler = new Scheduler();
  }
  ...
}

contract Scheduler {
  constructor() {
    multirewards = msg.sender;
  }
}
```

nventuro: (by the way, did you mean to make this mutable on purpose? If so, I'd be interested in hearing the rationale. And in that case we should at least emit an event).

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -388,6 +397,24 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
...
+        IERC20 rewardsToken,
+        uint256 reward,
+        address rewarder
+    ) external override updateReward(pool, address(0)) {
+        require(
+            rewarder == msg.sender || msg.sender == rewardsScheduler,
+            "Rewarder must be sender, or rewards scheduler"
+        );
+        _startReward(pool, rewardsToken, reward, rewarder);
+    }
```
nventuro, line 416: Do actually need this function at all? It seems like we should just call `notifyRewardAmount` directly from the scheduler.

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -90,7 +96,10 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
         address rewarder
     ) external override {
         require(
-            msg.sender == owner() || msg.sender == address(pool) || isAssetManager(pool, msg.sender),
+            msg.sender == owner() ||
+                msg.sender == address(pool) ||
+                msg.sender == address(rewardsScheduler) ||
+                isAssetManager(pool, msg.sender),
```
nventuro, line 102: Is this maybe the result of an incomplete refactor from an alternative approach? The scheduler is not adding any rewarders to any whitelist (nor does it need to), I'm not sure why we're letting it do it.

pkg/distributors/contracts/RewardsScheduler.sol
```diff
@@ -0,0 +1,118 @@
...
+
+    function poke(bytes32[] calldata rewardIds) external {
+        for (uint256 r; r < rewardIds.length; r++) {
+            bytes32 rewardId = rewardIds[r];
+            ScheduledReward memory scheduledReward = _rewards[rewardId];
+
+            require(scheduledReward.startTime != 0, "reward has not been created");
+            require(scheduledReward.startTime <= block.timestamp, "reward cannot be started");
+
+            multirewards.startScheduledReward(
```
nventuro, line 56: We should also emit an event here, to not make following the schedule lifecycle depend on looking at multiple contracts.

nventuro (commented):
Thanks @gtaschuk! The diagram was very usful in understanding what you set out to do.

I left a few minor comments regarding events, data structures and so on, but overall I'm not sure about the approach taken. I thought the underlying idea was to leave `MultiRewards` as close to the original as possible, with the only change being that the scheduler would be able to call `notifyReward`, since `MultiRewards` trusts that it'll perform proper access control checks.

Is there a need to distinguish between the scheduled reward start and the regular one?

gtaschuk:
> I thought the underlying idea was to leave MultiRewards as close to the original as possible, with the only change being that the scheduler would be able to call notifyReward, since MultiRewards trusts that it'll perform proper access control checks.

The changes to multirewards are to allow the rewardScheduler to start a reward *on behalf of* a rewarder.

The staking contract had no `rewarder` parameter for `notifyReward`, instead started a reward on behalf of msg.sender.  So I added the function `startScheduledReward` for the `rewardScheduler`.  Perhaps we should always require the `rewarder` parameter (which would be msg.sender, unless called by the `rewardScheduler`)?

pkg/distributors/contracts/MultiRewards.sol
gtaschuk: I thought it was more clear to have 2 functions - `startScheduledReward` for the rewardScheduler, that takes a rewarder parameter, and another function `notifyRewardAmount' for self-scheduled rewards (which omits the rewarder parameter, and uses msg.sender).

If we want to require rewarders to pass their own address for that parameter, we can combine into 1 function.

pkg/distributors/contracts/RewardsScheduler.sol
```diff
@@ -0,0 +1,118 @@
...
+    using SafeERC20 for IERC20;
+
+    IMultiRewards public multirewards;
+
+    constructor(IMultiRewards _multirewards) {
+        multirewards = _multirewards;
+    }
+
+    struct ScheduledReward {
+        IERC20 pool;
```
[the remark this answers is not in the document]
gtaschuk: `pool` is used as the parameter for the staking tokens everywhere in `multirewards`.  While there is no explicit check, since we for example, use the pool address to get the asset managers for the purpose of the allowlist, I think it is better to scope the use of this contract, and multirewards, to just staking pool bpt.

pkg/distributors/contracts/MultiRewards.sol
gtaschuk: Yes, I figured we should leave ourselves the flexibility to upgrade the scheduler (to better handle recurring rewards, for example).  I guess, as part of that, we would want to allow people to cancel scheduled rewards under any conditions as part of upgradeability, as (scheduled rewards currently can't be cancelled once they are scheduled to start).

If we opt for immutable reward schedulers then the deployment pattern above makes sense

pkg/distributors/contracts/RewardsScheduler.sol
```diff
@@ -0,0 +1,118 @@
...
+
+// solhint-disable not-rely-on-time
+
+/**
+ * Scheduler for MultiRewards contract
+ */
+contract RewardsScheduler {
+    using SafeERC20 for IERC20;
+
+    IMultiRewards public multirewards;
```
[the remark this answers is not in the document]
gtaschuk: To clarify, you want it to be a private, immutable variable, with a `getMultiRewards()` getter function?

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -389,18 +395,26 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
...
     ) external override updateReward(pool, address(0)) {
+        require(
+            rewarder == msg.sender || msg.sender == rewardsScheduler,
+            "Rewarder must be sender, or rewards scheduler"
+        );
+
         // handle the transfer of reward tokens via `safeTransferFrom` to reduce the number
         // of transactions required and ensure correctness of the reward amount
-        rewardsToken.safeTransferFrom(msg.sender, address(this), reward);
+        address funder = (msg.sender == rewardsScheduler) ? rewardsScheduler : msg.sender;
```
rabmarut, line 416: Doesn't this always resolve to `msg.sender`?

pkg/distributors/contracts/RewardsScheduler.sol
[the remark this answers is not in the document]
nventuro: Yes - though the renaming isn't strictly necessary, just suggesting going with the same style we already have.

pkg/distributors/contracts/RewardsScheduler.sol
```diff
@@ -0,0 +1,118 @@
...
+
+    struct ScheduledReward {
+        IERC20 pool;
+        IERC20 rewardsToken;
+        uint256 startTime;
+        address rewarder;
+        uint256 amount;
+    }
+
+    event RewardScheduled(bytes32 rewardId, address scheduler, address rewardsToken, uint256 startTime);
```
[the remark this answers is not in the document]
nventuro: The pool should also be added here.

pkg/distributors/contracts/RewardsScheduler.sol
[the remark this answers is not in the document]
nventuro: It seems like `notifyReward` has `stakingToken` in its docs? That's where my confusion came from.

I personally don't mind either name (thought I don't think it's obvious that the Pool's contract is also the token contract), but we should at least use the same one everywhere.

pkg/distributors/contracts/MultiRewards.sol
nventuro: We _could_ have such a thing, but if so I'd make it so `startSchedulerReward` was only callable by the scheduler - I found the fact that rewarders could use either very confusing.

I guess there might be benefits in keeping `notify` as-is in that we'd remain interface-compatible with the original implementation? But I'm not sure.

pkg/distributors/contracts/RewardsScheduler.sol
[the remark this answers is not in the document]
gtaschuk: Don't think we actually need a getter for this variable so I just made it private

pkg/distributors/contracts/RewardsScheduler.sol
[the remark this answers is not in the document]
gtaschuk: > It seems like notifyReward has stakingToken in its docs? That's where my confusion came from.

Sorry, where do you see `stakingToken`?  I see 2 references to staked token.  Also note multireward docs have been somewhat updated by the other PR for `Exiter`s that I opened last week.

Recommend we continue using `pool` everywhere

pkg/distributors/contracts/RewardsScheduler.sol
gtaschuk: Hmm.  I think it would be useful if someone for example wanted to split a larger reward into multiple smaller ones, or otherwise change a reward, more confidently schedule a reward far in advance etc.  Can remove, just, well, that's why I thought it was worth including.  I don't see an attack where it could be used to steal funds from another user.

pkg/distributors/contracts/RewardsScheduler.sol
gtaschuk: What's the use case of querying past reward schedules on-chain?  I didn't opt to save them for simplicity/cleanup

By the same logic should we be saving past records of rewards added to the staking contract itself?  Those seem like arguably the more important ones to save.

pkg/distributors/contracts/RewardsScheduler.sol
nventuro: I actually think the commitment to the schedule is an important property, and don't see being able to reschedule or cancel as that useful since all other related actions are fully final - it feels strange to have one that isn't.

pkg/distributors/contracts/RewardsScheduler.sol
nventuro: Well, we didn't introdce many changes in MultiRewards precisely to avoid changes, so I wouldn't use it as a basis for comparison.

My point about data availability was more that it is strange for some piece of data to be available and then deleted after it is 'used' - I've seen projects be bitten by this, and in general try to avoid `delete` altogether.

gtaschuk: I disagree, but I've made the change

pkg/distributors/contracts/RewardsScheduler.sol
```diff
@@ -0,0 +1,135 @@
...
+contract RewardsScheduler {
+    using SafeERC20 for IERC20;
+
+    IMultiRewards private immutable _multirewards;
+
+    constructor() {
+        _multirewards = IMultiRewards(msg.sender);
+    }
+
+    enum RewardStatus { PENDING, STARTED }
```
TomAFrench, line 35:
```suggestion
    enum RewardStatus { UNINITIALIZED, PENDING, STARTED }
```

I think it makes more sense to have the "empty state" flag on the status enum rather than use a sentinel value on the start time.

pkg/distributors/contracts/RewardsScheduler.sol
```diff
@@ -0,0 +1,135 @@
...
+    function getScheduledRewardInfo(bytes32 rewardId) external view returns (ScheduledReward memory reward) {
+        return _rewards[rewardId];
+    }
+
+    function startRewards(bytes32[] calldata rewardIds) external {
+        for (uint256 r; r < rewardIds.length; r++) {
+            bytes32 rewardId = rewardIds[r];
+            ScheduledReward memory scheduledReward = _rewards[rewardId];
+
+            require(scheduledReward.startTime != 0, "reward has not been created");
```
TomAFrench, line 74: This check is redundant due to L76.

pkg/distributors/test/MultiRewards.test.ts
```diff
@@ -69,7 +69,9 @@ describe('Staking contract', () => {
       .allowlistRewarder(pool.address, rewardToken.address, mockAssetManager.address);
 
     await expect(
-      stakingContract.connect(mockAssetManager).notifyRewardAmount(pool.address, rewardToken.address, fp(100))
+      stakingContract
+        .connect(mockAssetManager)
```
TomAFrench, line 73:
```suggestion
        .connect(rewarder)
```

I would change the tests to use an explicitly allowlisted rewarder and change the variable name as this is more descriptive (otherwise we need to know that asset managers are privileged rewarders). We can then have a separate test for just checking that asset managers can add rewards.

pkg/distributors/contracts/RewardsScheduler.sol
gtaschuk: Nice, I like this

pkg/distributors/test/MultiRewards.test.ts
gtaschuk: I made this rename in the other exiter PR, ok if I leave the rename out of this PR, and explicitly allowlist the user in a separate PR?

(lint inlines a lot of functions when changing `mockAssetManger` to `rewarder` and I don't want to introduce the noise to this PR, which should change `MultiRewards` minimally)

TomAFrench: Sure thing, this is tangential to the rest of this PR so makes sense to tackle elsewhere.

pkg/distributors/contracts/RewardsScheduler.sol
```diff
@@ -0,0 +1,135 @@
...
+            require(scheduledReward.startTime <= block.timestamp, "reward cannot be started");
+            require(scheduledReward.status == RewardStatus.PENDING, "scheduled reward already activated");
+
+            scheduledReward.rewardsToken.approve(address(_multirewards), scheduledReward.amount);
```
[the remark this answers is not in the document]
gtaschuk: Agreed that there is no risk to granting infinite approval to the MultiRewards contract as it should only be able to deposit tokens to it's vault balance as part of starting a scheduled reward.

We _do_ hold token balances on the `RewardScheduler` - there is a safeTransferFrom the rewarder when you schedule a reward.

I'm considering it... find it a bit harder to read but yes, it would save gas for those poking this contract to start the reward, so it's probably worth it.   Would it make sense to do that approval when the reward is scheduled to shift the gas burden from the contract pokers to the reward schedulers?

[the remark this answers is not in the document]
TomAFrench: Ah yes, I got a bit mixed up on it holding balances. Besides I think the fact that it's approving a single known contract which we trust means it's an acceptable risk.

re moving approval from poker to scheduler: It's not a particularly gas intensive step and I'd prefer the readability improvements from keeping the approval and the call which requires it close to each other.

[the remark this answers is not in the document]
nventuro: Is this the only pending item on this PR?

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -389,16 +394,23 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
...
     ) external override updateReward(pool, address(0)) {
-        require(_rewarders[pool][rewardsToken].contains(msg.sender), "Reward must be configured with addReward");
+        require(
+            msg.sender == rewarder || msg.sender == address(rewardsScheduler),
+            "Rewarder must be sender, or rewards scheduler"
+        );
+        require(_rewarders[pool][rewardsToken].contains(rewarder), "Reward must be configured with addReward");
 
         // handle the transfer of reward tokens via `safeTransferFrom` to reduce the number
         // of transactions required and ensure correctness of the reward amount
```
nventuro, line 416: I'd add a comment here mentioning than the reward scheduler holds the tokens for `rewarder`, so we always use `msg.sender` as the sender address.

pkg/distributors/contracts/RewardsScheduler.sol
```diff
@@ -0,0 +1,134 @@
...
+            emit RewardStarted(
+                rewardId,
+                msg.sender,
+                scheduledReward.pool,
+                scheduledReward.rewardsToken,
+                scheduledReward.startTime,
+                scheduledReward.amount
+            );
+
+            _rewards[rewardId].status = RewardStatus.STARTED;
```
nventuro, line 93: Shouldn't we move this up a few lines, right after the `require status == pending` check, to avoid reentrancy issues?

pkg/distributors/contracts/RewardsScheduler.sol
```diff
@@ -0,0 +1,134 @@
...
+        bytes32 rewardId,
+        address indexed scheduler,
+        IERC20 indexed pool,
+        IERC20 indexed rewardsToken,
+        uint256 startTime,
+        uint256 amount
+    );
+    event RewardStarted(
+        bytes32 rewardId,
+        address indexed scheduler,
```
nventuro, line 56: It seems like this 'scheduler' is called 'rewarder' everywhere else.

pkg/distributors/contracts/RewardsScheduler.sol
```diff
@@ -0,0 +1,134 @@
...
+        rewardsToken.safeTransferFrom(msg.sender, address(this), amount);
+
+        _rewards[rewardId] = ScheduledReward({
+            pool: pool,
+            rewardsToken: rewardsToken,
+            rewarder: msg.sender,
+            amount: amount,
+            startTime: startTime,
+            status: RewardStatus.PENDING
+        });
```
nventuro, line 130: As in the other comment, I'd initialize the mapping before calling `safeTransferFrom` to avoid reentrancy issues.

pkg/distributors/test/RewardsScheduler.test.ts
```diff
@@ -0,0 +1,143 @@
...
+  let rewardsScheduler: Contract;
+  let rewardsToken: Token;
+  let pool: Contract;
+
+  before('deploy base contracts', async () => {
+    [admin, , lp, , rewarder] = await ethers.getSigners();
+  });
+
+  sharedBeforeEach('set up asset manager and mock callback', async () => {
+    const { contracts } = await setup();
```
nventuro, line 32: If we're going to rely so heavily on `setup`, it should export the signers as well instead of having us keep both calls in sync.

pkg/distributors/test/RewardsScheduler.test.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    pool = contracts.pool;
+    vault = contracts.vault;
+    stakingContract = contracts.stakingContract;
+    rewardsToken = contracts.rewardTokens.DAI;
+    rewardTokens = contracts.rewardTokens;
+
+    //rewardsScheduler = await deploy('RewardsScheduler', { args: [stakingContract.address] });
```
nventuro, line 40:
```suggestion
```

pkg/distributors/test/RewardsScheduler.test.ts
```diff
@@ -0,0 +1,143 @@
...
+
+        await expectBalanceChange(
+          () => rewardsScheduler.connect(lp).startRewards([rewardId]),
+          rewardTokens,
+          [{ account: stakingContract.address, changes: { DAI: ['very-near', expectedReward] } }],
+          vault
+        );
+      });
+
+      it('emits RewardAdded in MultiRewards', async () => {
```
nventuro, line 133: We seem to be missing:
 a) a test for the event emitted by the scheduler
 b) a test for what happens once the scheduled reward is started (the status changes to done, it cannot be started again)

pkg/distributors/contracts/RewardsScheduler.sol
gtaschuk: moving it, but, how would reentrancy happen?  This function only calls `multiRewards.notifyRewardAmount` and `multirewards` is an immutable contract which does not call this contract as part of `notifyRewardAmount`

pkg/distributors/contracts/RewardsScheduler.sol
gtaschuk: So actually in this case the param is referring to the user that actually started the reward (the person that poked the contract) but I think on second thought it does make more sense to report the `rewarder` here, so I have changed it

pkg/distributors/test/RewardsScheduler.test.ts
gtaschuk: Makes a lot of sense - I updated the RewardScheduler test to use values exported from setup and will update the other tests to use them too in a separate PR

pkg/distributors/contracts/RewardsScheduler.sol
nventuro: `notifyRewardAmounts` calls token functions, both directly and indirectly via the Vault, so there's a lot of potential for this.

gtaschuk: Ah, yes, I suppose a malicious token would be able to drain the balance with a reentrancy attack -I think that would only allow them to drain the token's own balance (still bad)

either way, the change is in

pkg/distributors/test/MultiRewards.test.ts
```diff
@@ -362,9 +358,13 @@ describe('Staking contract', () => {
     });
 
     it.skip('emits RewardPaid for each pool', async () => {
```
nventuro, line 360: Can we unskip this test?

nventuro (approved):
Looking good, this seems ready! Thanks a lot @gtaschuk 

Moving forward, please avoid rebasing PRs after a review unless it's necessary for some reason. It makes following up on the review that much harder, as a) it is not possible to identify new commits, and b) it is not possible to look at the code that led to the original review comments, as those commits no longer exist.

PR #722, opened 2021-07-14
MultiRewards Distributor: check reward exists in `notifyRewardAmount`

gtaschuk:
Fix: Explicitly check reward exists in `notifyRewardAmount` before starting reward period (previously this function would have `reverted` if called before `addReward` because of division by zero (with an uninitialized reward duration).

### DIFF
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -398,10 +398,7 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
         IERC20 rewardsToken,
         uint256 reward
     ) external override updateReward(pool, address(0)) {
-        require(
-            rewardData[pool][msg.sender][rewardsToken].rewardsDuration != 0,
-            "Reward must be configured with addReward"
-        );
+        require(_rewarders[pool][rewardsToken].contains(msg.sender), "Reward must be configured with addReward");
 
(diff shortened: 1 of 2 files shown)


pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -398,6 +398,11 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
         IERC20 rewardsToken,
         uint256 reward
     ) external override updateReward(pool, address(0)) {
+        require(
+            rewardData[pool][msg.sender][rewardsToken].rewardsDuration != 0,
```
nventuro, line 402: This seems a bit indirect, shouldn't we do `_rewarders[pool][rewardsToken].contains(msg.sender)`?

gtaschuk: Sure.  fwiw I used the `rewardDuration` check as that's what snx/curve use to check the existance of a reward.  `_rewarders` is not in their implementation since they don't allow multiple rewarders...

nventuro: I see what you mean - I was referrencing this line: https://github.com/curvefi/multi-rewards/blob/master/contracts/MultiRewards.sol#L502

gtaschuk: @nventuro so we're all set now?

nventuro (approved):

PR #723, opened 2021-07-14
Mutable Tokens

EndymionJkb:
Move fixed token/scaling factors from BasePool down to most derived contracts, to allow for mutable tokens. Expand WeightedPool to 20 tokens


pkg/pool-utils/contracts/BasePool.sol
```diff
@@ -30,15 +30,11 @@ import "@balancer-labs/v2-asset-manager-utils/contracts/IAssetManager.sol";
...
-// reads. Because immutable arrays are not supported, we instead declare a fixed set of state variables plus a total
-// count, resulting in a large number of state variables.
-
 // solhint-disable max-states-count
 
 /**
- * @dev Reference implementation for the base layer of a Pool contract that manages a single Pool with an immutable set
- * of registered tokens, no Asset Managers, an admin-controlled swap fee percentage, and an emergency pause mechanism.
+ * @dev Reference implementation for the base layer of a Pool contract that manages a single Pool with no
+ * Asset Managers, an admin-controlled swap fee percentage, and an emergency pause mechanism.
```
TomAFrench, line 37:
```suggestion
 * @dev Reference implementation for the base layer of a Pool contract that manages a single Pool with optional
 * Asset Managers, an admin-controlled swap fee percentage, and an emergency pause mechanism.
```

This comment is stale as we allow these now.

EndymionJkb: Missed that; thanks. And as discussed, I'll add a factory that hard-codes zeroes so that we can whitelist a factory without having to worry about people sneaking in rogue asset managers

pvt/benchmarks/joinExit.ts
```diff
@@ -14,7 +14,7 @@ import {
 import { deployedAt } from '@balancer-labs/v2-helpers/src/contract';
 
 // setup environment
-const BPTAmount = bn(5e18);
+const BPTAmount = bn(1e18);
```
rabmarut, line 17: This seems very minor but just wondering what's the reason for this change?

EndymionJkb: Otherwise it runs out of balance when you run it with 20 tokens. I didn't try other values; maybe 2 or 3 would work

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -23,18 +23,80 @@ import "./BaseWeightedPool.sol";
 contract WeightedPool is BaseWeightedPool {
     using FixedPoint for uint256;
 
+    uint256 private constant _MAX_TOKENS = 20;
```
rabmarut, line 26: This is private but everything else is internal. What is the intended behavior here? Is this to allow for a future contract to derive from this one add more tokens, like your original design of the 8- and 16-token weighted pools?

EndymionJkb: Yes - I do want them to be able to derive from WeightedPool, but it might be confusing to have multiple _MAX_TOKENS defined if this were internal. You'd have _MAX_TOKENS and then _MAX_NEW_POOL_TOKENS.

If they want to keep the same max token count in the new pool, they would just not override _getMaxTokens()

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -23,18 +23,80 @@ import "./BaseWeightedPool.sol";
...
+
+    uint256 internal immutable _totalTokens;
```
rabmarut, line 28: Similar question: This has an `internal` getter, and it's `immutable`, so it can actually be made `private`. Too nitpicky? Same goes for the scaling factors and normalized weights.

EndymionJkb: _totalTokens definitely can be private. The only reason to make the others internal would be if you needed to reference token3 or scalingFactor1 directly from a derived class, which is hard to imagine. You can call the internal methods to access them. So yes, I can make all those private.

rabmarut: In BasePool, where the `_scalingFactor`s were originally defined, they were actually private.

rabmarut: Anyway, I think the `private` model from `BasePool` makes the most sense for all three of these (`_totalTokens`, `_scalingFactor`s, and `_normalizedWeight`s). And it should be changed not just here in `WeightedPool` but also in `StablePool` and `LiquidityBootstrappingPool`. Let me know if you disagree.

EndymionJkb: There's a case for WeightedPool and LBP (consulted Nico to see if there was a convention). Gist of that was:

Generally you want to keep things as restricted as possible - but, these are base classes that we intend to be easily extensible. So that's the tradeoff. There's no "danger" in making them internal, since they're immutable.

In the case of the WeightedPools, Nico points out that there is a BaseWeightedPool, and anybody doing something that would require special handling of tokens would probably derive from that instead of WeightedPool. So that would argue for making them private.

However... what if you wanted to make a 30- or 40-token pool that had identical behavior to WeightedPool, just more tokens - we talked about this exact case today in the retro. In that case, it would be inefficient to implement _getNormalizedWeights() and _getScalingFactors() if you couldn't see the "base" tokens. You'd have to reserve memory for the full size, call the base class, fill in the first 20 slots from the base class, then continue for the rest.

So that would argue for making everything internal (except perhaps the actual _tokenX variables). If you want the tokens, you get them from the Vault. What do you think of that? private _tokens (and _totalTokens), and scalingFactors/normalizedWeights internal for extensibility?

Probably the same applies to StablePool- and there is no base stable pool above it to derive from

EndymionJkb: Stable Pool does look at _token values in swap callbacks. So maybe those should stay internal

rabmarut: Interesting... the `_token`s have actually always been `internal`, even in `BasePool`, so making them `private` instead would constitute an actual change rather than just a return to norm like I was suggesting with the others. I guess part of this discussion hinges on the design decisions that went into the original `BasePool`, and I'm not privy to those (but Nico probably is).

> In that case, it would be inefficient to implement _getNormalizedWeights() and _getScalingFactors() if you couldn't see the "base" values.

I think this argument is quite strong, so perhaps leaving most/all of these `internal` is indeed the best path forward. Maybe we should just revert the most recent commit, or else make `_totalTokens` private and leave everything else alone. Sorry for the unnecessary tangent.

The rest of the PR is good to go from my perspective, so if we can just make one more patch to arrive at a consistent answer to the `private`/`internal` problem across all relevant pool types then I can take one last glance and approve.

EndymionJkb: Ok; I'm leaving _tokenX internal because I don't want to change StablePool, and though I'm skeptical derived classes would need to use them, it's "safe" to have them internal if they're immutable. Latest push has those changes.

pkg/pool-utils/contracts/test/MockBasePool.sol
```diff
@@ -97,4 +97,20 @@ contract MockBasePool is BasePool {
...
+    function _getTotalTokens() internal pure override returns (uint256) {
+        return 2;
+    }
+
+    function _scalingFactor(IERC20) internal pure override returns (uint256) {
+        return 1e18;
+    }
+
+    function _scalingFactors() internal pure override returns (uint256[] memory) {
+        return new uint256[](2);
```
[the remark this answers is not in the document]
EndymionJkb: Yeah I thought about that :) The relayer one does - I can make it more like that

pkg/pool-weighted/test/WeightedPool.test.ts
```diff
@@ -9,13 +9,13 @@ describe('WeightedPool', function () {
   let allTokens: TokenList;
 
   const POOL_SWAP_FEE_PERCENTAGE = fp(0.01);
-  const WEIGHTS = [fp(30), fp(20), fp(15), fp(10), fp(8), fp(6), fp(4), fp(2)];
+  const WEIGHTS = Array(20).fill(fp(0.05));
```
nventuro, line 12: It is important that we use different weights, otherwise we're not testing that each token is associated with the correct weight (i.e. that there are no mistakes in the if-else chain).

We should also test different scaling facotrs (i.e. different number of decimals). Alternatively we could have 20 programatically created tests (`i = 0..19`), where we have token `i` have a different number of decimals and check that the scaling factor is correct.

pkg/pool-weighted/contracts/WeightedPoolNoAMFactory.sol
```diff
@@ -0,0 +1,59 @@
...
+pragma experimental ABIEncoderV2;
+
+import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
+
+import "@balancer-labs/v2-pool-utils/contracts/factories/BasePoolSplitCodeFactory.sol";
+import "@balancer-labs/v2-pool-utils/contracts/factories/FactoryWidePauseWindow.sol";
+
+import "./WeightedPool.sol";
+
+contract WeightedPoolNoAMFactory is BasePoolSplitCodeFactory, FactoryWidePauseWindow {
```
nventuro, line 25: Do we need this contract?

EndymionJkb: This came out of a discussion at the Integrations retro. We have customers who would use a WeightedPool with 20 tokens, so might want to deploy this soonish. This would allow us to list a new WeightedPool type on the UI without worrying about somebody sneaking in a rogue asset manager. Otherwise we might need to approve each individual pool for safety

pkg/pool-weighted/test/WeightedPool.test.ts
EndymionJkb: Ok; will expand on this

EndymionJkb: It uses different weights now and, if you pass in a varyDecimals parameter, it will set each token to a different decimal. For instance, a 3-token pool would have 18,17,16; a 5-token pool would have 18,17,16,14 - and a 20 token pool represents all 18 + 0 decimal tokens.

Scaling factors are not exposed, so I had to create a MockWeightedPool to expose them and deploy that instead of the regular WeightedPool. The test now checks that the scaling factors match the predicted ones

rabmarut (approved):

nventuro:
I added weight and scaling factor tests for the StablePool and LBP, since we also changed those files and they were relying on BasePool's correctness.

nventuro (approved):

PR #725, opened 2021-07-15
Stable: Cache token rates

facuspagnuolo:


pkg/pool-stable/test/meta/MetaStablePool.test.ts
```diff
@@ -539,6 +540,7 @@ describe('MetaStablePool', function () {
 
   describe('price rates', () => {
     let rateProviders: Contract[];
+    const expirations = [MINUTE, MINUTE * 2];
```
nventuro, line 543: These are the durations, right?

pkg/pool-stable/contracts/meta/MetaStablePool.sol
```diff
@@ -27,45 +27,77 @@ import "./OracleMiscData.sol";
...
     using OracleMiscData for bytes32;
 
     IRateProvider private immutable _rateProvider0;
     IRateProvider private immutable _rateProvider1;
 
+    // Price rate caches are used to avoid querying the price rate for a token every time we need to work with it.
+    // Data is stored with the following structure:
+    //
+    // [ last update | cache expires | price rate value ]
+    // [    uint64   |     uint64    |      uint128     ]
```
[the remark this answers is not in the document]
facuspagnuolo: I fully agree, but why not using the whole slot if we have it?

pkg/pool-stable/test/meta/MetaStablePool.test.ts
facuspagnuolo: Yes

pkg/pool-stable/contracts/meta/MetaStablePool.sol
```diff
@@ -79,6 +111,7 @@ contract MetaStablePool is StablePool, StableOracleMath, PoolPriceOracle, IPrice
         uint256 indexIn,
         uint256 indexOut
     ) public virtual override onlyVault(request.poolId) returns (uint256) {
+        _cachePriceRateIfNecessary();
```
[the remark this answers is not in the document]
facuspagnuolo: Note that making `_priceRate()` non-view will make `_scalingFactors()` non-view as well. Initially I thought that will force us to change `onSwap` as well, but it is actually non-view in the interface, so we can change this.

pkg/pool-stable/contracts/meta/MetaStablePool.sol
```diff
@@ -27,45 +27,78 @@ import "./OracleMiscData.sol";
...
     using FixedPoint for uint256;
     using OracleMiscData for bytes32;
 
     IRateProvider private immutable _rateProvider0;
     IRateProvider private immutable _rateProvider1;
 
+    // Price rate caches are used to avoid querying the price rate for a token every time we need to work with it.
+    // Data is stored with the following structure:
+    //
+    // [ last update | duration | price rate value ]
```
nventuro, line 40:
```suggestion
    // [ expires | duration | price rate value ]
```

pkg/pool-stable/contracts/meta/MetaStablePool.sol
```diff
@@ -420,17 +454,53 @@ contract MetaStablePool is StablePool, StableOracleMath, PoolPriceOracle, IPrice
...
+     * @dev Overrides scaling factor getter to introduce the tokens' price rate.
+     * Note that it may update the price rate cache if necessary.
      */
     function _scalingFactors() internal view virtual override returns (uint256[] memory scalingFactors) {
+        // There is no need to check the arrays length since both are based on `_getTotalTokens`
+        // Given there is no generic direction for this rounding, it simply follows the same strategy as the BasePool.
         scalingFactors = super._scalingFactors();
         uint256[] memory priceRates = _priceRates();
+        scalingFactors[0] = scalingFactors[0].mulDown(priceRates[0]);
+        scalingFactors[1] = scalingFactors[1].mulDown(priceRates[1]);
```
nventuro, line 466: Extremely minor, but given we know it's only two rates, I think we can remove `_priceRates` and just call `_priceRate(token0)` here, removing the temporary arrays.

nventuro (approved):
Fantastic! I'd merge this as-is, and do cache eviction and duration update in a separate PR.

PR #726, opened 2021-07-16
Group related functions in the SDK into classes

TomAFrench:
The way we're exporting functions from the SDK currently doesn't help discoverability for outside devs and is a bit messy. Some functions will near always be imported together (relayer auth) and on others we're overloading arguments leading to having to pass large objects causing unnecessary line breaks (join/exit encoding).

To mitigate this I've grouped related functions into classes so they can be imported together in a namespaced way.


nventuro (approved):
I love everything about this PR.

Re. discoverability, while TS should help, I imagine at some point we'll want to have proper docs, at least listing the different functions.

PR #727, opened 2021-07-16
Generate typechain bindings for all contracts in deployments package into sdk

TomAFrench:
I think it would be a good idea to distribute typechain bindings for any pools that are in production as it's relatively painless on our side.

Arguably the place it makes sense to store these would be in Balancer.js so that they always get installed alongside it (plus I'm planning on making use of it in there in future anyway) although we may want to consider distributing it separately as its own package.


nventuro:
Good idea!

Seems you forgot to add those files to .gitignore.

nventuro:
This kind of usage means we should avoid having contracts with the same name, right? E.g. once we deploy a new WeightedPoolFactory, we should name it something different.

TomAFrench:
Yes, if the interface has changed we would need to rename the contract so that previous typechain bindings don't get overwritten. I think this would be fairly desirable anyway to avoid confusion though (numbering should be fine I think).

nventuro (approved):

PR #729, opened 2021-07-16
MultiRewards: joinAndStake with batch relayer

gtaschuk:
Joins a pool and stakes the bpt to the LP in the `MultiRewards` contract
staking-relayer

### DIFF
--- a/pkg/standalone-utils/contracts/BatchRelayer.sol
+++ b/pkg/standalone-utils/contracts/BatchRelayer.sol
@@ -86,7 +86,11 @@ contract BatchRelayer {
         IERC20 bpt = IERC20(_getPoolAddress(poolId));
         uint256 bptAmount = bpt.balanceOf(address(this));
 
-        bpt.approve(address(stakingContract), bptAmount);
+        // If necessary, give Vault allowance to take BPT
+        if (bpt.allowance(address(this), address(getVault())) < bptAmount) {
+            bpt.approve(address(getVault()), type(uint256).max);
+        }
+        
         stakingContract.stake(bpt, bptAmount, recipient);
     }
 
--- a/pkg/standalone-utils/contracts/BatchRelayer.sol
+++ b/pkg/standalone-utils/contracts/BatchRelayer.sol
@@ -16,6 +16,7 @@ pragma solidity ^0.7.0;
 pragma experimental ABIEncoderV2;
 
 import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
+import "@balancer-labs/v2-distributors/contracts/interfaces/IMultiRewards.sol";
 
 /**
  * @title Batch Relayer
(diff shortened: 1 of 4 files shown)


pkg/standalone-utils/contracts/BatchRelayer.sol
```diff
@@ -74,6 +75,21 @@ contract BatchRelayer {
...
+        address payable recipient,
+        IVault.JoinPoolRequest calldata joinPoolRequest,
+        IMultiRewards stakingContract
+    ) external {
+        getVault().joinPool(poolId, msg.sender, address(this), joinPoolRequest);
+
+        IERC20 bpt = IERC20(_getPoolAddress(poolId));
+        uint256 bptAmount = bpt.balanceOf(address(this));
+
+        bpt.approve(address(stakingContract), bptAmount);
```
TomAFrench, line 89:
```suggestion
        // If necessary, give Vault allowance to take BPT
        if (bpt.allowance(address(this), address(getVault())) < bptAmount) {
            bpt.approve(address(getVault()), type(uint256).max);
        }
        
```

TomAFrench: This will reduce gas costs by about 20k after the first usage for each pool.

An infinite approval is safe here as the relayer will never hold a balance of BPT as it is instantly staked.

TomAFrench (commented):
Looks good to me. I wasn't really expecting this to be added to the `BatchRelayer` but it would only be an issue if the staking contract isn't ready before we deploy this for metastable pools.

That said, it seems like we could just deploy this with the staking contract set to zero and redeploy when the staking contract is deployed.

gtaschuk:
> That said, it seems like we could just deploy this with the staking contract set to zero and redeploy when the staking contract is deployed.

We're currently passing in the staking contract's address as a param.  Are you suggesting we add a method `onlyOwner` that can `setStakingAddress`?  (seems reasonable)

Fwiw I made it a param because I was reluctant to add ownership to this contract, and it would revert if the provided `stakingContract` didn't respond to that method

Thanks for fixing the few little lint errors

TomAFrench:
I was thinking more along the lines of we set an immutable staking contract upon deployment of the relayer, no need for ownership. This avoids the issue of users accidentally staking into a malicious contract if we allow them to provide their own address.

Redeploying the relayer once we have this address isn't an issue as users should only need to sign a new authorisation message and we can disallow further interactions from the old version.

pkg/standalone-utils/test/BatchRelayer.test.ts
```diff
@@ -221,6 +221,52 @@ describe('BatchRelayer', function () {
...
+        ).wait();
+
+        expectEvent.inIndirectReceipt(receipt, vault.instance.interface, 'PoolBalanceChanged', {
+          poolId: basePoolId,
+          liquidityProvider: sender.address,
+        });
+
+        expectEvent.inIndirectReceipt(receipt, stakingContract.interface, 'Staked', {
+          pool: basePool.address,
+          account: sender.address,
```
TomAFrench, line 264:
```suggestion
          account: recipient.address,
```
I've just noticed that this line is wrong, yet the test passes on CI - running it locally fails however.

I've fixed this locally but I think we should find out why CI isn't giving the right result.

TomAFrench: Looks like CI is running tests for `solidity-utils` instead

TomAFrench: fixed in #731. In the meantime I've verified that tests pass locally.

PR #730, opened 2021-07-17
Update deployment benchmark

nventuro:
This updates the bytecodesize measurement script to use the artifacts directly instead of deploying, and targets the appropiate contracts (i.e. WeightedPool instead of the factory since it is now split code, and MetaStable).

This is the current status, after setting MetaStable to 200 runs:

```
== Deployment measurements ==

# v2-vault/Vault
Deployed bytecode size is 23.938 kB

# v2-pool-weighted/WeightedPool
Deployed bytecode size is 23.094 kB

# v2-pool-weighted/WeightedPool2TokensFactory
Deployed bytecode size is 23.965 kB

# v2-pool-weighted/LiquidityBootstrappingPool
Deployed bytecode size is 19.718 kB

# v2-pool-stable/StablePoolFactory
Deployed bytecode size is 22.730 kB

# v2-pool-stable/meta/MetaStablePool
Deployed bytecode size is 25.069 kB
```


facuspagnuolo (approved):
:clap:

PR #731, opened 2021-07-19
ci: run correct test suite for standalone-utils

TomAFrench:
It looks like we're running the tests for `solidity-utils` in the `standalone-utils` task rather than the `standalone-utils` tests.


nventuro (approved):
Great catch!

PR #732, opened 2021-07-19
Meta: Add cache duration setter

facuspagnuolo:


pkg/pool-stable/test/meta/MetaStablePool.test.ts
```diff
@@ -583,104 +596,129 @@ describe('MetaStablePool', function () {
...
+            updatePriceRateCache();
+            itAdaptsTheScalingFactorsCorrectly();
+          });
 
-            expect(newPriceRate0).to.be.gt(oldPriceRate0);
-            expect(newPriceRate1).to.be.gt(oldPriceRate1);
+          context('when it passed the previous duration', () => {
+            mockNewRatesAndAdvanceTime(MINUTE * 2);
+            updatePriceRateCache();
+            itAdaptsTheScalingFactorsCorrectly();
```
nventuro, line 707: It seems like these tests don't really test the new duration is in effect?

pkg/pool-stable/test/meta/MetaStablePool.test.ts
```diff
@@ -583,104 +596,129 @@ describe('MetaStablePool', function () {
...
+          });
+        });
```
nventuro, line 709: It seems like these tests don't really test the new duration is in effect?

nventuro (commented):
Fantastic, very clean :ok_hand:

pkg/pool-stable/test/meta/MetaStablePool.test.ts
facuspagnuolo: It does, note that `itAdaptsTheScalingFactorsCorrectly` is querying the provider directly and these where updated right before the assertion, then the current scaling factors are being tested against the latest values coming from the providers

nventuro (approved):

PR #737, opened 2021-07-20
Move typechain bindings to their own package

TomAFrench:
I've moved typechain bindings to be distributed under the @balancer-labs/typechain package as they weren't surviving the bundling process intact. Balancer.js can now just import these types (allowing us to rollup its own types) and others can use them without being required to install Balancer.js.


pkg/typechain/package.json
```diff
@@ -0,0 +1,37 @@
...
+  },
+  "main": "dist/index.ts",
+  "typings": "dist/index.d.ts",
+  "files": [
+    "dist/"
+  ],
+  "scripts": {
+    "build": "rm -rf dist/ && yarn typechain && tsc && cp temp/*.d.ts dist/ && yarn postbuild",
+    "postbuild": "rm -rf temp/",
+    "typechain": "typechain --target ethers-v5 --out-dir temp/ '../deployments/tasks/*/abi/*.json' 'abi/*.json'"
```
nventuro, line 23: This means that we should not use these typings in our tests, correct? As the code in master may have diverged from the last deployment.

nventuro (approved):
Looks great!

pkg/typechain/package.json
TomAFrench: Yes, I think we shouldn't do that. I'm seeing this as something for Balancer.js to use along with outside developers, trying to keep these up to date for in-development contracts along with maintain backwards compatibility with deployed ones might be tricky.

Issue #685, opened 2021-06-29
Add tests for WeightCompression


nventuro:
#630 introduced this library, it'd be good to have some basic tests for it.

nventuro:
See e.g. the tests for `WordCodec`, in `pool-utils/test/oracle/Samples.test.ts`.

PR #738, opened 2021-07-20
Investment pool v1

EndymionJkb:
Replaces the boolean flags for pool types with an enum. Introduces the InvestmentPool: mapping storage for tokens (up to 100). Since gradual updates and pausing will follow soon (but isn't needed on the first iteration), some "time traveling" code is present but commented out, and the structure (e.g., initializing the weights) reflects how it will be when those features (similar to LBP) are added


pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -0,0 +1,271 @@
...
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
+
+import "../BaseWeightedPool.sol";
+import "./WeightCompression.sol";
+
+/**
+ * @dev Weighted Pool with mutable weights, designed to support V2 Liquidity Bootstrapping
+ */
+contract InvestmentPool is BaseWeightedPool, ReentrancyGuard {
+    // The Pause Window and Buffer Period are timestamp-based: they should not be relied upon for sub-minute accuracy.
```
rabmarut, line 28: Why is this comment here? Seems like it was probably copied from another contract but it's not relevant to this one.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -0,0 +1,271 @@
...
+    // solhint-disable not-rely-on-time
+
+    using FixedPoint for uint256;
+    using WordCodec for bytes32;
+    using WeightCompression for uint256;
+
+    // State variables
+
+    // The current number of tokens in the pool
+    // Technically redundant; cached here to avoid calling getTokens on the pool,
```
rabmarut, line 38: Why do you say this is redundant? This seems perfectly fine to me.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -0,0 +1,271 @@
...
+    //   which would be very gas-intensive for large numbers of tokens
+    // Can only change if tokens are added/removed
+    uint256 private _totalTokens;
+
+    // Store scaling factor and start/end weights for each token
+    // Mapping should be more efficient than trying to compress it further
+    // into a fixed array of bytes32 or something like that, especially
+    // since tokens can be added/removed - and re-ordered in the process
+    // For each token, we store:
+    // [ 155 bits|   5 bits |  32 bits   |   64 bits    |
```
rabmarut, line 48: Why is the start weight at double precision compared to the end weight? In the LBP, this was necessary because there was basically just enough room to pack everything this way. But here, it seems like the end weight could just as easily be 64 bits, or each weight could even be 100+ bits; there is plenty of room. Are we saving it for something else?

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -0,0 +1,271 @@
...
+    function _getMaxTokens() internal pure virtual override returns (uint256) {
+        return _MAX_WEIGHTED_TOKENS;
+    }
+
+    function _getTotalTokens() internal view virtual override returns (uint256) {
+        return _totalTokens;
+    }
+
+    function _scalingFactor(IERC20 token) internal view virtual override returns (uint256) {
+        bytes32 tokenData = _poolState[token];
```
rabmarut, line 108: I could see this snippet being used quite a bit, so I'd consider wrapping it into a function that returns the bytes32 tokenData:

```
bytes32 tokenData = _poolState[token];

// A valid token can't be zero (must have non-zero weights)
if (tokenData == 0) {
    _revert(Errors.INVALID_TOKEN);
}
```

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -0,0 +1,271 @@
...
+
+    function _getNormalizedWeight(IERC20 token) internal view override returns (uint256) {
+        bytes32 tokenData = _poolState[token];
+
+        // A valid token can't be zero (must have non-zero weights)
+        if (tokenData == 0) {
+            _revert(Errors.INVALID_TOKEN);
+        }
+
+        uint256 startWeight = tokenData.decodeUint64(_START_WEIGHT_OFFSET).uncompress64();
```
rabmarut, line 137: Same here, I think this bit should be broken out into a function (similar to how it's done by index in the LBP, it can be done by passing tokenData and pctProgress here):

```
uint256 startWeight = tokenData.decodeUint64(_START_WEIGHT_OFFSET).uncompress64();
uint256 endWeight = tokenData.decodeUint32(_END_WEIGHT_OFFSET).uncompress32();

return _interpolateWeight(startWeight, endWeight, pctProgress);
```

I just don't like having so many lines duplicated across functions because it makes me want to double-check that they are actually the same, which is a waste of time/energy. If there's a good reason not to do this please let me know.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -0,0 +1,271 @@
...
+            bytes32 tokenData = _poolState[tokens[i]];
+
+            uint256 startWeight = tokenData.decodeUint64(_START_WEIGHT_OFFSET).uncompress64();
+            uint256 endWeight = tokenData.decodeUint32(_END_WEIGHT_OFFSET).uncompress32();
+
+            normalizedWeights[i] = _interpolateWeight(startWeight, endWeight, pctProgress);
+        }
+    }
+
+    function _getNormalizedWeightsAndMaxWeightIndex() internal view override returns (uint256[] memory, uint256) {
```
rabmarut, line 163: Why is this function ever so slightly different from the one in the LBP? It has the same outcome, but I just don't understand why the conventions are different. I think we should try to be consistent about this stuff.

pkg/pool-weighted/contracts/smart/WeightCompression.sol
```diff
@@ -23,6 +23,7 @@ import "@balancer-labs/v2-solidity-utils/contracts/math/FixedPoint.sol";
  */
 library WeightCompression {
     uint256 private constant _UINT31_MAX = 2**(31) - 1;
+    uint256 private constant _UINT32_MAX = 2**(32) - 1;
```
rabmarut, line 26: Use `type(uint32).max` instead?

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
EndymionJkb: I mean the ground truth is the Vault, and it could be slightly confusing to read from a cache (that could technically get out of sync).

Maybe I'll remove the word "redundant" and explicitly call it a cache; that should improve the clarity.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
EndymionJkb: Yes; stale cut/paste. Case of staring at it too long and not seeing things :)

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
EndymionJkb: Yes - there is a lot more to store later (circuit breaker stuff, mainly). If necessary, the start weight precision could be reduced later (32 is probably still ok).

It is still desirable to have start weights higher precision than end weights. End weights are set by the user, presumably to "round numbers" - while start weights can be anything (e.g., if they restart the weight update in the middle). Higher resolution start weights help avoid (potentially game-able) weight "jumps" due to lack of precision.

pkg/pool-weighted/contracts/smart/WeightCompression.sol
EndymionJkb: We seem to be going back and forth on this. That's what it was, and then somebody on SC wanted it this way. I guess for consistency with the rest of the codebase - mainly WordCodec - that does it this way

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
EndymionJkb: Good idea - should save some bytecode. Added _getValidTokenData(token)

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
EndymionJkb: Since this is only called before interpolateWeight, I'll just move it inside there.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
EndymionJkb: No reason - just wrote it independently. I've made it identical now.

pkg/pool-weighted/contracts/smart/WeightCompression.sol
rabmarut: But if that's the case then why use type(uint64).max below?

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
rabmarut: But we have `_totalTokens` in all pools, and the Vault is nonetheless the source of ground truth in those pools. Does the fact that it's mutable here actually make a difference to that design pattern?

pkg/pool-weighted/contracts/smart/WeightCompression.sol
EndymionJkb: Oversight (fixed in a later branch) - will backport here to resolve confusion

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -0,0 +1,260 @@
...
+        }
+    }
+
+    function _getValidTokenData(IERC20 token) private view returns (bytes32 tokenData) {
+        tokenData = _poolState[token];
+
+        // A valid token can't be zero (must have non-zero weights)
+        if (tokenData == 0) {
+            _revert(Errors.INVALID_TOKEN);
+        }
```
TomAFrench, line 258:
```suggestion
        _require(tokenData != 0, Errors.INVALID_TOKEN);
```
Is there a reason to prefer an explicit if statement?

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -0,0 +1,260 @@
...
+        if (startWeight > endWeight) {
+            uint256 weightDelta = pctProgress.mulDown(startWeight - endWeight);
+            return startWeight.sub(weightDelta);
+        } else {
+            uint256 weightDelta = pctProgress.mulDown(endWeight - startWeight);
+            return startWeight.add(weightDelta);
+        }
+    }
+
+    function _getValidTokenData(IERC20 token) private view returns (bytes32 tokenData) {
```
TomAFrench, line 252:
```suggestion
    function _getTokenData(IERC20 token) private view returns (bytes32 tokenData) {
```

It feels weird to me to specify "valid" here. I understand this function enforces that we actually have data for `token` but I think that we can assume that functions return valid data (and revert otherwise) unless explicitly stated.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
EndymionJkb: Well it's going away in the next PR anyway (moved into packed storage), so it doesn't really matter.

pkg/balancer-js/src/pool-weighted/normalizedWeights.ts
```diff
@@ -1,12 +1,22 @@
...
  * @param weights - an array of token weights to be normalized
  * @returns an equivalent set of normalized weights
  */
 export function toNormalizedWeights(weights: BigNumber[]): BigNumber[] {
+  // When the number is exactly equal to the max, normalizing with common inputs
+  // leads to a value < 0.01, which reverts. In this case fill in the weights exactly.
+  if (weights.length == MaxWeightedTokens) {
+    return Array(MaxWeightedTokens).fill(ONE.div(MaxWeightedTokens));
+  }
+
```
nventuro, line 19: This feels quite hacky. If we honestly think a pool with e.g. 90 tokens might exist, I'd consider allowing for smaller weights (maybe up to 0.1%?). It should be easy to review our arithmetic analysis to understand how this might change the limits (max join/exit ratio in join/exitswaps). cc @dmf7z

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
nventuro: Seems like it wasn't removed?

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
nventuro: See my comment above re. allowing for smaller weights - they will essentially be fixed for pools with a large amount of tokens unless we reduce the minimum one.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -0,0 +1,259 @@
...
+    uint256 private constant _DECIMAL_DIFF_OFFSET = 96;
+
+    // Time travel comment
+    // [ 192 bits | 32 bits  |  32 bits   |
+    // [  unused  | end time | start time |
+    // |MSB                            LSB|
+    //bytes32 private _gradualUpdateTimestamps;
+
+    //uint256 private constant _START_TIME_OFFSET = 0;
+    //uint256 private constant _END_TIME_OFFSET = 32;
```
nventuro, line 63: Is this from another branch?

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -0,0 +1,259 @@
...
+            owner
+        )
+    {
+        uint256 numTokens = tokens.length;
+        InputHelpers.ensureInputLengthMatch(numTokens, normalizedWeights.length, assetManagers.length);
+
+        _tokenCountCache = numTokens;
+
+        // I'm time-traveling a bit here - storing the weights in a form where they can be changed
+        uint256 currentTime = block.timestamp;
```
nventuro, line 95: Not sure I follow this comment, what do you mean by time-traveling? This seems consistent with what we do in LBPs: an instant weight change (start and end time are the same).

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -0,0 +1,259 @@
...
+        _require(normalizedSum == FixedPoint.ONE, Errors.NORMALIZED_WEIGHT_INVARIANT);
+
+        //_gradualUpdateTimestamps = newTimestamps
+        //    .insertUint32(startTime, _START_TIME_OFFSET)
+        //    .insertUint32(endTime, _END_TIME_OFFSET);
+
+        //emit GradualWeightUpdateScheduled(startTime, endTime, startWeights, endWeights);
+    }
+
+    function _computeScalingFactor(bytes32 tokenState) private view returns (uint256) {
```
nventuro, line 204: Let's chose a different name: `_computeScalingFactor` already exists (for an argument of type `IERC20`) and does something completely different.

pkg/pool-weighted/contracts/smart/WeightCompression.sol
nventuro: `type(uint32).max` is the better choice - but `uintX` is only defined for multiples of 8, so `uint31` does not exist unfortunately.

pkg/pool-weighted/contracts/smart/WeightCompression.sol
```diff
@@ -23,6 +23,8 @@ import "@balancer-labs/v2-solidity-utils/contracts/math/FixedPoint.sol";
  */
 library WeightCompression {
```
nventuro, line 24: Let's make sure we tackle #685 reasonably soon

pkg/pool-weighted/test/BaseWeightedPool.behavior.ts
```diff
@@ -26,9 +26,12 @@ export function itBehavesAsWeightedPool(numberOfTokens: number, useCustomTwoToke
   async function deployPool(params: RawWeightedPoolDeployment = {}): Promise<void> {
     const assetManagers = Array(numberOfTokens).fill(assetManager.address);
 
+    const poolTypeParam = useCustomTwoTokenPool
+      ? WeightedPoolType.WEIGHTED_POOL_2TOKENS
+      : WeightedPoolType.WEIGHTED_POOL;
```
nventuro, line 31: I'd drop the `bool` in `itBehavesAsAWeightedPool` and use this enum directly.

pkg/balancer-js/src/pool-weighted/normalizedWeights.ts
EndymionJkb: Well, I mean the typescript convenience function that creates slightly different but similar weights by calling toNormalizedWeights(range(10000, 100100)) will have some < 0.01.

You mean change the Math to allow < 0.01? We did an analysis on that (see the limits Notion doc), and it seemed like 100 was a good limit, and I don't think more than 50ish will be practical.

I asked the main client waiting for this - they said 20 was the flagship, and they couldn't imagine doing more than 40. (Higher effective counts should just use composition.)

pkg/pool-weighted/contracts/smart/WeightCompression.sol
EndymionJkb: Ok... so put it back to using max, except for 31

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
EndymionJkb: I just mean I'm putting code in that anticipates the next PR

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
EndymionJkb: Well now I'm really confused, because it looks like this comment is from a later branch somehow? Panicked that I checked into the wrong branch somehow (there are three branched off each other), but it looks right otherwise. Well, the code will catch up to it soon.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
EndymionJkb: Commented out stuff that says time travel anticipates the next PR (included for context, since otherwise it would be really weird to call functions like startGradualWeightUpdate to initialize weights, when the weight update functions aren't there yet

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
EndymionJkb: computeScalingMultiplier?

EndymionJkb: Well, maybe _readScalingFactor - it does the exact same thing as BasePool, except it reads from the mapping rather than calling the token.decimals()

pkg/pool-weighted/contracts/smart/WeightCompression.sol
EndymionJkb: I was going to use WordCodec as an example, but we don't have that one either

pkg/balancer-js/src/pool-weighted/normalizedWeights.ts
rabmarut: I hear Nico's point but honestly I'm pretty content with this. I concur that it is very unlikely we ever see close to 100 tokens.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
rabmarut: @EndymionJkb I don't understand what you mean; the code will never catch up to this comment because the Pause Window and Buffer Period are implemented in BaseWeightedPool, no? The comment should be removed.

EndymionJkb: Ok - now I finally understand. You're talking about the Pause window comment, not the dev comment on top! The highlighted section included both, and I was looking at the top one :) Yes, will remove.

pkg/pool-weighted/test/InvestmentPoolFactory.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+      const pool = await createPool();
+      const { pauseWindowEndTime, bufferPeriodEndTime } = await pool.getPausedState();
+      const now = await currentTimestamp();
+
+      expect(pauseWindowEndTime).to.equal(now);
+      expect(bufferPeriodEndTime).to.equal(now);
+    });
+
+    it('does not have asset managers', async () => {
```
rabmarut, line 89: Not necessarily saying this is wrong, but the WeightedPoolFactory test has an asset manager and an owner, whereas here they are both zeroed. I'm just wondering what the decision process is in constructing these tests.

EndymionJkb:
I think this is left over from when it didn't allow asset managers (that
was a very recent addition). Will add.

pvt/helpers/src/models/types/TypesConverter.ts
```diff
@@ -20,6 +20,11 @@ import {
 } from '../tokens/types';
 import { ZERO_ADDRESS } from '../../constants';
 
+export function computeDecimalsFromIndex(i: number): number {
+  // Produces sequential series (18,17,16,...,0,17,16,15,...,1,18,17,16,15,...)
+  return i > 18 ? 18 - ((i - 18) % 18) : 18 - i;
```
rabmarut, line 25:
```suggestion
  return 18 - (i % 18);
```

pvt/helpers/src/models/types/TypesConverter.ts
```diff
@@ -20,6 +20,11 @@ import {
 } from '../tokens/types';
 import { ZERO_ADDRESS } from '../../constants';
 
+export function computeDecimalsFromIndex(i: number): number {
+  // Produces sequential series (18,17,16,...,0,17,16,15,...,1,18,17,16,15,...)
```
rabmarut, line 24: I'm confused by this comment... the line below produces the series 18,17,16,...,3,2,1 over and over, right? But the comment shows it going to zero, then starting at 17 the next time, then going to 1 and restarting at 18.

rabmarut: Oh, sorry for the confusion. Your line actually does produce 18..0, followed by 17..1, and thereafter 18..1 repeatedly. Is that actually the desired behavior? If so, my reduction below is incorrect. But why is that desirable?

rabmarut: Perhaps you want the infinitely repeating series 18..0?

```
return 18 - (i % 19);
```

EndymionJkb: Yes; that's much cleaner. This was one of those late night things I just didn't spend time on making better (partly because the exact sequence really doesn't matter; just should contain all possible decimal values). Originally it was developed for the 20-token pool, so it went 18-0 and stayed at 0. I didn't want that pattern for the 100-token pools (though I guess it would have been fine).

Anyway that's much more readable, so I'll change it.

rabmarut (approved):

PR #740, opened 2021-07-20
LBP: Merge swaps enabled with pool state

facuspagnuolo:


EndymionJkb (approved):
Looks good! Will incorporate these in pools going forward

PR #743, opened 2021-07-21
Investment Pool v2

EndymionJkb:
Add "LBP" features to the Investment Pool: enable/disable swaps + gradual weight change


pkg/solidity-utils/contracts/helpers/BalancerErrors.sol
```diff
@@ -144,6 +144,7 @@ library Errors {
     uint256 internal constant SWAPS_DISABLED = 327;
     uint256 internal constant CALLER_IS_NOT_LBP_OWNER = 328;
     uint256 internal constant PRICE_RATE_OVERFLOW = 329;
+    uint256 internal constant DISPROPORTIONATE_JOIN_OR_EXIT = 330;
```
rabmarut, line 147: This is a bit confusing as an error code because Balancer pools typically allow for disproportionate joins/exits. Can we somehow clarify "while paused" without stretching the length of the name beyond what's reasonable?

pvt/benchmarks/joinExit.ts
```diff
@@ -48,7 +48,7 @@ async function main() {
   }
   console.log('\n');
 
-  for (let numTokens = 40; numTokens <= 90; numTokens += 10) {
+  for (let numTokens = 40; numTokens <= 80; numTokens += 10) {
```
rabmarut, line 51: Consider defining this as a constant at the top, since you're going to have to keep changing it.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -85,22 +103,99 @@ contract InvestmentPool is BaseWeightedPool, ReentrancyGuard {
...
+        )
+    {
+        // Load current pool state from storage
+        bytes32 poolState = _poolState;
+
+        startTime = poolState.decodeUint32(_START_TIME_OFFSET);
+        endTime = poolState.decodeUint32(_END_TIME_OFFSET);
+
+        (IERC20[] memory tokens, , ) = getVault().getPoolTokens(getPoolId());
+        uint256 totalTokens = tokens.length;
```
rabmarut, line 147: I'm assuming there's a good reason, but wondering what it is. Why not use the cached `_getTotalTokens()` here?

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -193,29 +287,156 @@ contract InvestmentPool is BaseWeightedPool, ReentrancyGuard {
...
+        uint256 currentBalanceTokenIn,
+        uint256 currentBalanceTokenOut
+    ) internal view override returns (uint256) {
+        _require(getSwapEnabled(), Errors.SWAPS_DISABLED);
+
+        return super._onSwapGivenOut(swapRequest, currentBalanceTokenIn, currentBalanceTokenOut);
+    }
+
+    /**
+     * @dev Extend ownerOnly functions to include the LBP control functions
```
rabmarut, line 324:
```suggestion
     * @dev Extend ownerOnly functions to include the InvestmentPool control functions
```

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -193,29 +287,156 @@ contract InvestmentPool is BaseWeightedPool, ReentrancyGuard {
...
+
+        // If the pool is paused, only allow proportional join
+        if (!getSwapEnabled()) {
+            // We know the bptOut from the join - so proportional exit with that bptOut should reproduce amountsIn,
+            // if the join was proportional
+            uint256[] memory amountsOut = WeightedMath._calcTokensOutGivenExactBptIn(
+                balances,
+                bptAmountOut,
+                totalSupply().add(bptAmountOut)
+            );
```
rabmarut, line 376: I don't love the fact that this block implicitly depends on `balances` having been properly mutated during `super._onJoinPool()`. As far as I can tell, the underlying call to `BaseWeightedPool._invariantAfterJoin()` should update `balances` accordingly, but are we willing to make that assumption at this level? Even in `BasePool.sol`, you'll find this comment after the `_onJoinPool()` call:

```
// Note we no longer use `balances` after calling `_onJoinPool`, which may mutate it.
```

The assumption is that `_onJoinPool` _may_ mutate `balances` while performing computations (for gas savings), not that `balances` _must_ correctly reflect the post-join state at the end of the call. I'd be really curious to get @nventuro's thoughts on this. Even if this currently works, are we willing to bet that it always will? And that if we make changes such that it no longer works, we will remember this conversation and fix this (and/or the tests will catch it)?

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -193,29 +287,156 @@ contract InvestmentPool is BaseWeightedPool, ReentrancyGuard {
...
+            uint256 diff;
+
+            for (uint256 i = 0; i < amountsOut.length; i++) {
+                diff = amountsOut[i] > amountsIn[i] ? amountsOut[i].sub(amountsIn[i]) : amountsIn[i].sub(amountsOut[i]);
+
+                _require(
+                    amountsIn[i] > 0 && diff.divUp(amountsIn[i]) <= _PROPORTIONAL_TOLERANCE_PCT,
+                    Errors.DISPROPORTIONATE_JOIN_OR_EXIT
+                );
+            }
```
rabmarut, line 386: While I think this is just about the best we can do with what we currently have, I can't help but wonder if a better solution might be to add a new `JoinKind` to `BaseWeightedPool` that is strictly proportional. Something like `TOKENS_IN_FOR_EXACT_BPT_OUT` (note the 's' in 'tokens'). But I'm lacking some context on why this was originally omitted (whereas the equivalent `ExitKind` exists). Maybe @nventuro could shed some light?

The way I see it, no matter how low we set this percentage threshold, it could always be gamed for a sufficiently large pool. As long as the arbitrageur has reserves of all tokens, they could continuously join/exit to make swaps on one pair. And that would completely ruin the value proposition of the paused state. If we could instead strictly enforce proportionality, then the integrity of the paused state is maintained without exceptions.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -46,20 +58,25 @@ contract InvestmentPool is BaseWeightedPool, ReentrancyGuard {
...
     uint256 private constant _DECIMAL_DIFF_OFFSET = 96;
 
-    // Time travel comment
-    // [ 192 bits | 32 bits  |  32 bits   |
-    // [  unused  | end time | start time |
-    // |MSB                            LSB|
-    //bytes32 private _gradualUpdateTimestamps;
+    // Adding or removing liquidity while paused must be proportional
+    // Proportional join is defined as one where the BPT prices are all similar
+    uint256 private constant _PROPORTIONAL_TOLERANCE_PCT = 1e16; // 1%
```
rabmarut, line 69: After reviewing the mechanics of how this threshold is imposed, I'd recommend shrinking it to 0.1% or lower.

pkg/solidity-utils/contracts/helpers/BalancerErrors.sol
EndymionJkb: Yes; I'll think of something. INVALID_JOIN_EXIT_WHILE_PAUSED?

pvt/benchmarks/joinExit.ts
EndymionJkb: I have 3: minRange, maxRange, and overall max

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
EndymionJkb: Yes - I thought about that too, but hesitated to add a whole new join kind. Makes sense; let's ask.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
EndymionJkb: Sounds good. Made the test tighter so it still works.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
EndymionJkb: Yes, but I'm not sure what could be done about it other than repeat all the calculations, or extract it from the new lastInvariant (which makes the same assumptions). Could at least add documentation

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
rabmarut: After chatting with @nventuro, we both concur that adding the new `JoinKind` makes the most sense here. It sounds like it was only omitted to begin with because it seemed redundant at the time - in the absence of a pause, there is actually no good UX reason to support this type of join over an uneven join (which can incidentally be proportional). But now that we have the pause capability, I think adding this strictly enforced proportional join to `BaseWeightedPool` is a necessity, and there shouldn't be any blockers to doing this. @EndymionJkb, are you content to insert that change into this PR in lieu of the functionality above?

EndymionJkb: Yes - that's much cleaner. Will do.

pkg/balancer-js/src/pool-weighted/encoder.ts
```diff
@@ -4,7 +4,8 @@ import { BigNumberish } from '@ethersproject/bignumber';
 export enum WeightedPoolJoinKind {
   INIT = 0,
   EXACT_TOKENS_IN_FOR_BPT_OUT,
-  TOKEN_IN_FOR_EXACT_BPT_OUT,
+  ONE_TOKEN_IN_FOR_EXACT_BPT_OUT,
+  TOKENS_IN_FOR_EXACT_BPT_OUT,
```
nventuro, line 8: I'm not a fan of the rename - ideally this would be as backwards compatible as possible. I'd just name the new one something different, like `ALL_TOKENS_IN...` or maybe even `PROPORTIONAL_TOKENS_IN_...`

pkg/pool-weighted/contracts/WeightedMath.sol
```diff
@@ -214,6 +214,33 @@ contract WeightedMath {
...
+        amountsIn = new uint256[](balances.length);
+
+        // Calculate the factor by which the invariant will increase after minting BPTAmountOut
+        uint256 balanceMultiplier = bptTotalSupply.add(bptAmountOut).divUp(bptTotalSupply).sub(FixedPoint.ONE);
+
+        // Balances need to increase by (invariantRatio - 1)
+        for (uint256 i = 0; i < balances.length; i++) {
+            amountsIn[i] = balanceMultiplier.mulUp(balances[i]);
+        }
+    }
```
nventuro, line 242: This is needlessly complex:

`(total + out) / total - 1` equals 
`total / total + out / total - 1`, which equals 
`1 + out / total - 1`, which of course equals
`out / total`

nventuro: In order words, the amounts in are just the bptout/total ratio of the current balances. I'd copy the exit implementation:

```
        uint256 bptRatio = bptAmountIn.divDown(totalBPT);

        uint256[] memory amountsOut = new uint256[](balances.length);
        for (uint256 i = 0; i < balances.length; i++) {
            amountsOut[i] = balances[i].mulDown(bptRatio);
        }

        return amountsOut;
```

nventuro (commented):
I think this PR is a bit too unwieldy. I'd have a separate, simpler PR that only adds the new join kind, along with proper tests. Then we can look into how Investment Pools might use that feature.

pkg/balancer-js/src/pool-weighted/encoder.ts
EndymionJkb: Done - didn't notice you'd already answered this question here.

pkg/pool-weighted/contracts/WeightedMath.sol
EndymionJkb: Too late at night :) Now in the proportional PR

pkg/balancer-js/src/utils/errors.ts
```diff
@@ -55,6 +55,7 @@ const balancerErrorCodes: Record<string, string> = {
   '327': 'SWAPS_DISABLED',
   '328': 'CALLER_IS_NOT_LBP_OWNER',
   '329': 'PRICE_RATE_OVERFLOW',
+  '330': 'INVALID_JOIN_EXIT_WHILE_PAUSED',
```
nventuro, line 58:
```suggestion
  '330': 'INVALID_JOIN_EXIT_KIND_WHILE_SWAPS_DISABLED',
```
It is important to not get 'paused' (an emergency mechanism that can only be used the first 3 months) and 'swaps disabled' (a manager action).

pkg/pool-weighted/contracts/WeightedPoolUserDataHelpers.sol
```diff
@@ -41,7 +41,11 @@ library WeightedPoolUserDataHelpers {
         (, amountsIn, minBPTAmountOut) = abi.decode(self, (BaseWeightedPool.JoinKind, uint256[], uint256));
     }
 
-    function tokenInForExactBptOut(bytes memory self) internal pure returns (uint256 bptAmountOut, uint256 tokenIndex) {
+    function oneTokenInForExactBptOut(bytes memory self)
+        internal
+        pure
+        returns (uint256 bptAmountOut, uint256 tokenIndex)
```
nventuro, line 47: Why did this change? I'd keep the original name

pkg/pool-weighted/test/InvestmentPool.test.ts
```diff
@@ -124,4 +75,384 @@ describe('InvestmentPool', function () {
...
+            });
+
+            it('disallows disproportionate joins (multi-token, 95%)', async () => {
+              const amountsIn = [...initialBalances];
+              amountsIn[0] = fp(0.95);
+
+              await expect(pool.joinGivenIn({ from: sender, amountsIn })).to.be.revertedWith(
+                'INVALID_JOIN_EXIT_WHILE_PAUSED'
+              );
+            });;*/
```
nventuro, line 264: Let's not have commented code checked in - we'll add it once it's uncommented.

EndymionJkb: Some of this is obsolete; rest will be uncommented (since we know have that join type) or deleted if unused

nventuro (approved):

pkg/pool-weighted/test/InvestmentPool.test.ts
```diff
@@ -124,4 +75,265 @@ describe('InvestmentPool', function () {
...
+            const bptBeforeJoin = await pool.balanceOf(owner.address);
+            await expect(pool.joinGivenIn({ from: owner, amountsIn: initialBalances })).to.not.be.reverted;
+
+            const bptAfterJoin = await pool.balanceOf(owner.address);
+            expect(bptAfterJoin).to.gt(bptBeforeJoin);
+
+            await expect(pool.exitGivenOut({ from: owner, amountsOut: initialBalances })).to.not.be.reverted;
+            const bptAfterExit = await pool.balanceOf(owner.address);
+            expect(bptAfterExit).to.lt(bptAfterJoin);
+          });
```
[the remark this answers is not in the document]
EndymionJkb: It might be left over from LBPs, where join/exit is permissioned? Do you want me to remove it?

PR #746, opened 2021-07-22
Investment Pool (circuit breakers)

### LINKED ISSUE #680: Implement Circuit Breakers
Circuit breakers are an automatic stopgap mechanism that prevents a Pool from buying too much of a given token as its price drops, mitigating the risk of participating in Pools with tokens for which there is less confidence.

There are many ways to define such a mechanism, including e.g. minimum relative price of a token respective to other tokens. The issue is that since a Pool may have a large of number of tokens (#674), specifying all of these min price relationships is a very error-prone endeavor.

Alternatively, we could look into defining this mechanism as a maximum balance of a token relative to some measure (the invariant?), such that the Pool would halt swaps for it if it ever got too much (due to falling prices). It'd be good to explore this avenue some more and try to arrive at an intuitive understanding of what such a circuit breaker would actually implement.

Another idea proposed by @rabmarut is that of latching: what if the Pool stopped swaps _until_ some sort of action was taken, instead of reenabling them once (if) the situation goes back to normal. After all, such an unforeseen scenario might require special action (such as token deregistration - #677). An obvious issue here is griefing: we might need a mechanism such as requiring for a swap to be the first in the block to be confident the price has not been manipulated.

Finally, whatever path we go forward with or analyse, we should be mindful of the fact that Investment Pools will have changing weights (#679), which play a major role in asset price and balances.

EndymionJkb:
PoC circuit breaker implementation (for reference/discussion)

Closes #680


EndymionJkb:
Superseded by #1307

PR #749, opened 2021-07-23
Oracle: Move query methods to linked library

nventuro:
This reduces the bytecode size of MetaStablePools enough that they can be deployed. With 200 runs, they end up at 23.919kB.

The way bytecode is reduced is by removing all oracle query functionality outside of the pool contract itself, and into a linked library. This adds a slight overhead of ~3k gas (2600 from the `delegatecall`), but gets rid of the binary search code and exponentiation (from `fromLowResLog`).

Ideally, we'd be able to deploy the library at construction time in either the Pool or the Factory. Unfortunately, Solidity does not allow for this (though I'll create a language proposal soon with this idea). Some available options are:

1) use linked libraries, which requires external (i.e. outside of Solidity itself) linking (bytecode modification)
2) use delegatecall directly, which can only be done in assembly (as `delegatecall` is not considered `view`) - essentially implementing a kind of proxy pattern. This could work, but it requires a way to keep the storage slots of the base and library contracts in sync
3) drop usage of delegate calls and make `view` calls back into the main contract for storage reads. Even if these repeated calls only cost 100 gas, the overhead of call encoding, decoding, function dispatch, return value encoding and decoding quickly adds up.

In the end, I went with 1), as  this is the most 'standard' approach. I modified the `deploy` function to allow for libraries, which are manually linked. Unfortunately, Hardhat does not support linking an artifact directly (see https://github.com/nomiclabs/hardhat/issues/1716), so this will have to make do for now.

This PR is split into two commits. The first moves functionality from the pool and oracle into the `QueryProcessor` library, while the second moves the query tests from the pools into the pool oracle, which is where they are implemented. No funtionality should change as a result of this.

### DIFF
--- a/pkg/pool-utils/contracts/oracle/PoolPriceOracle.sol
+++ b/pkg/pool-utils/contracts/oracle/PoolPriceOracle.sol
@@ -13,6 +13,7 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
 
 import "@balancer-labs/v2-solidity-utils/contracts/helpers/BalancerErrors.sol";
 
@@ -21,6 +22,7 @@ import "../interfaces/IPoolPriceOracle.sol";
 
 import "./Buffer.sol";
 import "./Samples.sol";
+import "./QueryProcessor.sol";
 
 /**
  * @dev This module allows Pools to access historical pricing information.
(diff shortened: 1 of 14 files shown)


pkg/pool-utils/contracts/oracle/PoolPriceOracle.sol
```diff
@@ -31,9 +33,12 @@ import "./Samples.sol";
...
- * timestamp when the index last changed.
+ * timestamp when the index last changed. Aditionally, access to the latest circular buffer index must be exposed by
+ * implementing `_getOracleIndex`.
+ *
+ * This contract relies on the `QueryProcessor` linked library to reduce bytecode size.
  */
-contract PoolPriceOracle is IPoolPriceOracle {
+abstract contract PoolPriceOracle is IPoolPriceOracle, IPriceOracle {
     using Buffer for uint256;
     using Samples for bytes32;
```
facuspagnuolo, line 43: Do we still need these?

facuspagnuolo (approved):

PR #752, opened 2021-07-26
MultiRewards: use authorizer pattern for access control

gtaschuk:
This delegates control of the `MultiRewards` contract to the `Vault`'s authorizer

Namely, this gives it the power to allowlist rewarders.  Pool controllers and asset managers are allowed to allowlist themselves for their own pools

I opted to move all authentication functions to the `MultiRewardsAuthorization` contract for a cleaner separation of concerns

_The end result of using the authorizer here is a little confusing because_
_- the allowlist requires {pool, rewarder} specific control which the authorizer cannot provide with it's actionId based approach (afaik)_
_- pool controller and asset managers need implicit access to add allowlisted rewarders, and the mix of implicit access and the authorizers explicit access is confusing_

### DIFF
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -55,9 +55,6 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
         uint256 rewardPerTokenStored;
     }
 
-    // delegate ownership to the vault
-    address private constant _DELEGATE_OWNER = 0xBA1BA1ba1BA1bA1bA1Ba1BA1ba1BA1bA1ba1ba1B;
-
     mapping(IERC20 => mapping(address => mapping(IERC20 => Reward))) public rewardData;
     mapping(IERC20 => EnumerableSet.AddressSet) private _rewardTokens;
 
@@ -76,7 +73,7 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
     constructor(IVault _vault)
         // Multirewards is a singleton, so it simply uses its own address to disambiguate action identifiers.
         Authentication(bytes32(uint256(address(this))))
-        MultiRewardsAuthorization(_DELEGATE_OWNER, _vault)
+        MultiRewardsAuthorization(_vault)
         TemporarilyPausable(3600, 3600)
     {
         // solhint-disable-previous-line no-empty-blocks
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -74,8 +74,8 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
     /* ========== CONSTRUCTOR ========== */
 
     constructor(IVault _vault)
-        /* disambiguate based on who is deploying staking contract */
-        Authentication(bytes32(uint256(msg.sender)))
+        // Multirewards is a singleton, so it simply uses its own address to disambiguate action identifiers.
+        Authentication(bytes32(uint256(address(this))))
         MultiRewardsAuthorization(_DELEGATE_OWNER, _vault)
         TemporarilyPausable(3600, 3600)
     {
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -79,12 +79,6 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
         // solhint-disable-previous-line no-empty-blocks
     }
 
-    function _getAuthorizer() internal view override returns (IAuthorizer) {
-        // Access control management is delegated to the Vault's Authorizer. This lets Balancer Governance manage which
-        // accounts can call permissioned functions: for example, to perform emergency pauses.
-        return getVault().getAuthorizer();
-    }
-
     /**
      * @notice Allows a rewarder to be explicitly added to an allowlist of rewarders
      */
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -17,7 +17,6 @@ pragma experimental ABIEncoderV2;
 
 import "@balancer-labs/v2-solidity-utils/contracts/math/Math.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/math/FixedPoint.sol";
-import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/Ownable.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/ReentrancyGuard.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/EnumerableSet.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/helpers/TemporarilyPausable.sol";
@@ -27,10 +26,11 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/IERC20Permit.sol
 
 import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
 import "@balancer-labs/v2-vault/contracts/interfaces/IAsset.sol";
-import "@balancer-labs/v2-vault/contracts/interfaces/IBasePool.sol";
 
 import "./interfaces/IMultiRewards.sol";
 import "./interfaces/IDistributor.sol";
+
+import "./MultiRewardsAuthorization.sol";
 
 // solhint-disable not-rely-on-time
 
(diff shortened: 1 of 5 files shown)


pkg/distributors/contracts/MultiRewardsAuthorization.sol
```diff
@@ -0,0 +1,101 @@
...
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/Authentication.sol";
+import "@balancer-labs/v2-vault/contracts/interfaces/IAuthorizer.sol";
+import "@balancer-labs/v2-vault/contracts/interfaces/IBasePool.sol";
+import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
+
+/**
+ * @dev Base authorization layer implementation for MultiRewards
+ */
+abstract contract MultiRewardsAuthorization is Authentication {
+    address private immutable _owner;
```
TomAFrench, line 27: We don't seem to use this variable anywhere that I can see. Is it needed?

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -72,36 +73,30 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
 
     /* ========== CONSTRUCTOR ========== */
 
-    constructor(IVault _vault) Ownable() TemporarilyPausable(3600, 3600) {
-        vault = _vault;
+    constructor(IVault _vault)
+        /* disambiguate based on who is deploying staking contract */
+        Authentication(bytes32(uint256(msg.sender)))
```
TomAFrench, line 78:
```suggestion
        // Multirewards is a singleton, so it simply uses its own address to disambiguate action identifiers.
        Authentication(bytes32(uint256(address(this))))
```

For singletons we tend to use their own address as the disambiguator.

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -72,36 +73,30 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
...
+        MultiRewardsAuthorization(_DELEGATE_OWNER, _vault)
+        TemporarilyPausable(3600, 3600)
```
TomAFrench, line 80: 1 hour seems very short, is there much benefit to this?

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -72,36 +73,30 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
...
+    {
+        // solhint-disable-previous-line no-empty-blocks
     }
 
-    modifier onlyAllowlistedRewarder(IERC20 pool, IERC20 rewardsToken) {
-        require(isAllowlistedRewarder(pool, rewardsToken, msg.sender), "only accessible by allowlisted rewarders");
-        _;
+    function _getAuthorizer() internal view override returns (IAuthorizer) {
```
TomAFrench, line 85: Move to `MultiRewardsAuthorization`?

pkg/distributors/test/MultiRewardsSharedSetup.ts
```diff
@@ -56,6 +60,10 @@ export const setup = async (): Promise<{ data: SetupData; contracts: SetupContra
     args: [vault.address],
   });
 
+  // authorize vault authorizer to allowlistRewarders
```
TomAFrench, line 63:
```suggestion
  // authorize admin to allowlistRewarders
```

pkg/distributors/test/MultiRewards.test.ts
```diff
@@ -61,6 +76,29 @@ describe('Staking contract', () => {
...
+      await vault.connect(admin).setAuthorizer(newAuthorizer.address);
+
+      action = await actionId(stakingContract, 'allowlistRewarder');
+      await newAuthorizer.connect(other).grantRole(action, other.address);
+
+      // that authorizer allowlists a rewarder
+      await stakingContract.connect(other).allowlistRewarder(pool.address, rewardToken.address, lp.address);
+
+      expect(await stakingContract.isAllowlistedRewarder(pool.address, rewardToken.address, lp.address)).to.equal(true);
+    });
```
TomAFrench, line 101: I'm not sure how useful this test is. At the end of the day you're just testing that `MultiRewards` maintains sync with the Vault in terms of which Authorizer is uses so this is a duplicate of one of the tests above.

pkg/distributors/test/MultiRewards.test.ts
```diff
@@ -33,13 +32,29 @@ describe('Staking contract', () => {
...
     rewardTokens = contracts.rewardTokens;
   });
 
+  describe('authorizer', () => {
+    it('uses the authorizer of the vault', async () => {
+      expect(await stakingContract.getAuthorizer()).to.equal(authorizer.address);
+    });
+
+    it('tracks authorizer changes in the vault', async () => {
+      const action = await actionId(vault, 'setAuthorizer');
```
TomAFrench, line 49:
```suggestion
      const action = actionId(vault, 'setAuthorizer');
```

This function is synchronous so we don't need an await.

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -55,13 +55,14 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, Temporari
         uint256 rewardPerTokenStored;
     }
 
-    IVault public immutable vault;
+    // delegate ownership to the vault
+    address private constant _DELEGATE_OWNER = 0xBA1BA1ba1BA1bA1bA1Ba1BA1ba1BA1bA1ba1ba1B;
```
TomAFrench, line 59: Seems like this can be removed.

TomAFrench (commented):
Looks good. There are a couple of what looks like leftovers from the migration away from `Ownable` to clean up.

I also noticed that we've split the events across the `IMultiRewards` and `MultiRewards` previously, what's the reasoning behind this?

pkg/distributors/contracts/MultiRewards.sol
gtaschuk: good question (didn't change the pause window in this PR) what do you think it should be?  24 hours?

pkg/distributors/test/MultiRewards.test.ts
gtaschuk: That's fair. A bit lengthy, but intended as a sanity check.  Do you think I should just remove it?

pkg/distributors/contracts/MultiRewards.sol
TomAFrench: I think this should be at least a couple of rounds of rewards so maybe a month or so?

pkg/distributors/test/MultiRewards.test.ts
TomAFrench: Yeah, I'd just nix it.

pkg/distributors/contracts/MultiRewardsAuthorization.sol
```diff
@@ -0,0 +1,103 @@
...
+
+    modifier onlyAllowlistedRewarder(IERC20 pool, IERC20 rewardsToken) {
+        require(isAllowlistedRewarder(pool, rewardsToken, msg.sender), "only accessible by allowlisted rewarders");
+        _;
+    }
+
+    modifier onlyAllowlisters(IERC20 pool) {
+        require(
+            _canPerform(getActionId(msg.sig), msg.sender) ||
+                msg.sender == address(pool) ||
```
TomAFrench, line 43:
```suggestion
            _canPerform(getActionId(msg.sig), msg.sender) ||
                msg.sender == address(pool) ||
```

Depending on how often we expect the pool to be the one allowlisting rewarders we could swap these two lines for some gas savings. Reasonably minor however.

pkg/distributors/contracts/MultiRewardsAuthorization.sol
```diff
@@ -0,0 +1,103 @@
...
+    IVault private immutable _vault;
+
+    mapping(IERC20 => mapping(IERC20 => mapping(address => bool))) private _allowlist;
+
+    constructor(IVault vault) {
+        _vault = vault;
+    }
+
+    modifier onlyAllowlistedRewarder(IERC20 pool, IERC20 rewardsToken) {
+        require(isAllowlistedRewarder(pool, rewardsToken, msg.sender), "only accessible by allowlisted rewarders");
```
TomAFrench, line 36:
```suggestion
        require(isAllowlistedRewarder(pool, rewardsToken, msg.sender), "Only accessible by allowlisted rewarders");
```

We should capitalise revert strings for consistency.

TomAFrench:
I think this is good to go once the current comments are addressed.

pkg/distributors/contracts/MultiRewardsAuthorization.sol
gtaschuk: I suspect that the most frequent case will be balancer multisig allowlisting itself, asset managers allowlisting themselves, balancer governance allowlisting airdrops, then pool-governance allowlisting (which I don't think would be for a while since there is currently no mechanism to do so in the deployed pools.

Since this is a one-off, I'm not sure I see much savings in letting the pool contract skip the line to save gas :thinking:

TomAFrench: Agreed that this is very small, let's leave it as is.

TomAFrench (approved):

Issue #689, opened 2021-06-30
Relayer mutlicall reentrancy protection


nventuro:
We're planning on having relayers that use Multicall (chiefly, the Batch Relayer) rely on 'global' state (e.g. relayer token balances) for some actions, such as returning all owned tokens (dust) to the caller at the end. This requires reentrancy protection to avoid having an attacker re-enter the relayer and sweep these amounts.

@TomAFrench proposed a very interesting pattern: make the relayer contract _only_ have the `multicall` function in its external interface, and have a separate secondary contract with the `joinPool`, `approveRelayer`, etc. functions that the relayer will `delegatecall` into. This means that:
 - the secondary contract is by itself unusable because it is not a relayer: all calls to the Vault will fail
 - there is a single entry point in the relayer (`multicall`), which can be reentrancy-protected, while it can still 're-enter' the join, swap, etc. functions that are delegatecall-ed into in the secondary contract

PR #756, opened 2021-07-27
feat: implement lido functionality in BatchRelayer

TomAFrench:
This PR allows the BatchRelayer to automatically wrap/unwrap Lido's stETH tokens, avoiding users having to wrap/unwrap stETH before/after interacting with Balancer.

~~Pros:~~
- ~~Avoids having to track multiple addresses in the frontend based on which asset we're using~~
- ~~Better gas efficiency due to performing all actions on the same contract.~~

~~Cons:~~
- ~~Rather ugly.~~
- ~~Doesn't handle all wstETH situations.~~

~~As mentioned in Slack, my preference from a code readability standpoint would be to have the wrapping/unwrapping happen in a separate (non-relayer) contract which would perform wrapping/unwrapping and then call into the standard BatchRelayer. We would also need something similar to this for trades on the wstETH-ETH pool which wouldn't need the BatchRelayer.~~

~~@facuspagnuolo I'd be interested to hear your views on if you have any strong preference here.~~

While a zap contract would work for swaps, it would increase complexity and gas costs of joins/exits greatly, we're then continuing down this route.

Draft PR as tests need to be added for the new Lido flavoured functions.

### DIFF
--- a/pkg/distributors/contracts/interfaces/IMultiRewards.sol
+++ b/pkg/distributors/contracts/interfaces/IMultiRewards.sol
@@ -34,4 +34,10 @@ interface IMultiRewards {
         IERC20 rewardsToken,
         address rewarder
     ) external;
+
+    function stakeFor(
+        IERC20 pool,
+        uint256 amount,
+        address receiver
+    ) external;
 }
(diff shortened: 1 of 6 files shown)


TomAFrench:
I'm going to park this PR as now this functionality is no longer critical for Lido launch, I'd like to tackle #689 first.

PR #765, opened 2021-07-30
Meta: Redeploy testnet instances

facuspagnuolo:


pkg/deployments/tasks/20210727-meta-stable-pool/output/kovan.json
```diff
@@ -0,0 +1,5 @@
+{
+  "query": "0x88919A6C36359d449322Cd40451749553C6E33A2",
```
nventuro, line 2: Seems these are not verified?

nventuro (approved):

pkg/deployments/tasks/20210727-meta-stable-pool/output/kovan.json
facuspagnuolo: Yup we need to tweak the verification process to support libraries 😬

PR #766, opened 2021-07-30
Reward callbacks update

TomAFrench:
I've made a couple of minor changes to #761 so that we only make addresses payable once we're feeding them into a `userBalanceOp` which improves readability imo. I've also removed some shadowing of a function name in `PoolTokenManipulator`.

If you're happy with these changes then we can merge into #761 and immediately merge that.


gtaschuk (approved):
Looks good

imho regrettable that we made the UserBalanceOp require a payable argument because ability to send/receive ETH isn't quite the same as ability to send/receive vault assets

PR #767, opened 2021-07-31
de-overload stake function to stakeFor a recipient

gtaschuk:
As suggested on the https://github.com/balancer-labs/balancer-v2-monorepo/pull/719 review

This syntax is easier to read


TomAFrench (approved):
LGTM

PR #768, opened 2021-08-02
Distributor: MerkleRedeem redeploy

gtaschuk:
setup task for redeployment of LDO rewards contract

I dated it to today but can easily change to deploy it tomorrow or whenever

Can be deployed with:
`npx hardhat deploy --id 20210802-ldo-merkle-redeploy --key <KEY> --network mainnet`


nventuro:
Given a) we never used the original one, b) we don't plan to use it, and more importantly c) it contains a severe vulnerability, I'd delete the previous deployment and replace it with this one.

Keep in mind we'll need to update the build info and ABIs to account for #776.

nventuro:
Can you merge the changes from https://github.com/balancer-labs/balancer-v2-monorepo/pull/782 here? tldr: the `distributor` key in the output artifact should be called `MerkleRedeem`, and there's a few helpers to reduce boilerplate

PR #769, opened 2021-08-03
Distributor: MultiRewards certora improvements

gtaschuk:
A collection of suggestions from Certora

Most notably we allow anyone to refresh (poke) the pool token list that the exiter and reinvestor use

### DIFF
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -78,7 +78,7 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
     /* ========== CONSTRUCTOR ========== */
 
     constructor(IVault _vault)
-        // Multirewards is a singleton, so it simply uses its own address to disambiguate action identifiers.
+        // MultiRewards is a singleton, so it simply uses its own address to disambiguate action identifiers
         Authentication(bytes32(uint256(address(this))))
         MultiRewardsAuthorization(_vault)
     {
@@ -143,7 +143,7 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
     }
 
     /**
-     * @notice Calculates the amount of reward per staked bpt for a rewardToken
+     * @notice Calculates the amount of reward token per staked bpt
      */
     function rewardPerToken(
         IERC20 pool,
@@ -287,7 +287,7 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
     }
 
     /**
-     * @notice Allows a user to claim any rewards to internal balance
+     * @notice Allows a user to claim any rewards to an internal balance or EOA
      */
     function _getReward(
         IERC20[] calldata pools,
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -153,9 +153,10 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
         if (_totalSupply[pool] == 0) {
             return rewardData[pool][rewarder][rewardsToken].rewardPerTokenStored;
         }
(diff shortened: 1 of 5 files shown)


pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -153,9 +153,8 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
         if (_totalSupply[pool] == 0) {
             return rewardData[pool][rewarder][rewardsToken].rewardPerTokenStored;
         }
-        uint256 unrewardedDuration = lastTimeRewardApplicable(pool, rewarder, rewardsToken).sub(
-            rewardData[pool][rewarder][rewardsToken].lastUpdateTime
-        );
+        uint256 unrewardedDuration = lastTimeRewardApplicable(pool, rewarder, rewardsToken) -
```
nventuro, line 156: Can we add a comment explaining why this is fine?

pkg/distributors/contracts/PoolTokenManipulator.sol
```diff
@@ -30,13 +30,20 @@ contract PoolTokenManipulator {
     mapping(bytes32 => EnumerableSet.AddressSet) private _poolTokenSets;
     mapping(bytes32 => bool) private _poolTokenSetSaved;
 
+    function savePoolTokenSet(bytes32 poolId) public {
+        (IERC20[] memory poolTokens, , ) = vault.getPoolTokens(poolId);
+        if (_poolTokenSetSaved[poolId]) {
+            delete _poolTokenSets[poolId];
```
nventuro, line 36: Have you tested this? I'm not sure it does what you want, since `delete` leaves mappings untouched: https://docs.soliditylang.org/en/v0.7.4/types.html#delete (a set is composed of a mapping and an array)

See [this issue on the OZ repo](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2607) for more discussion on clearing sets, particularly [this comment](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2607#issuecomment-829340362).

gtaschuk: Great catch.  Fixed to loop through and remove the tokens, and added a test using a `MockPool` with mutable tokens.

pkg/distributors/contracts/PoolTokenCache.sol
```diff
@@ -30,13 +30,25 @@ contract PoolTokenCache {
     mapping(bytes32 => EnumerableSet.AddressSet) private _poolTokenSets;
     mapping(bytes32 => bool) private _poolTokenSetSaved;
 
+    function savePoolTokenSet(bytes32 poolId) public {
+        (IERC20[] memory poolTokens, , ) = vault.getPoolTokens(poolId);
+        if (_poolTokenSetSaved[poolId]) {
+            uint256 numTokens = _poolTokenSets[poolId].length();
```
nventuro, line 36:
```suggestion
        if (_poolTokenSetSaved[poolId]) {
            // Purge potentially stale cached data
            uint256 numTokens = _poolTokenSets[poolId].length();
```

pkg/distributors/contracts/PoolTokenCache.sol
```diff
@@ -30,13 +30,25 @@ contract PoolTokenCache {
...
+            for (uint256 t; t < numTokens; t++) {
+                // always the 0 index since we're removing all elements
```
nventuro, line 38:
```suggestion
            // Clear the set by removing the zeroeth element n times
            for (uint256 t; t < numTokens; t++) {
```

pkg/distributors/contracts/PoolTokenCache.sol
```diff
@@ -30,13 +30,25 @@ contract PoolTokenCache {
...
+                address tokenAddress = _poolTokenSets[poolId].unchecked_at(0);
+                _poolTokenSets[poolId].remove(tokenAddress);
```
nventuro, line 40: This works, but it is inefficient for no good reason due a quirk on the OZ original implementation. Will follow up with an improvement.

nventuro (approved):

PR #770, opened 2021-08-03
Oracle: Allow querying past accum for incomplete buffers

facuspagnuolo:


EndymionJkb (commented):
I wasn't worried about the InvestmentPool warnings, because this is placeholder code that gets replaced in the next PR (and has to be a view function there). Of course it doesn't hurt anything to change it here.

dmf7z (approved):
👍

pkg/pool-utils/contracts/oracle/QueryProcessor.sol
```diff
@@ -111,22 +111,30 @@ library QueryProcessor {
...
-                // The only remaining condition to check is for the look up time to be between the oldest and latest
-                // timestamps.
+                if (oldestTimestamp > 0) {
+                    // If the oldest timestamp is not zero, it means the buffer was fully initialized.
+                    bufferLength = Buffer.SIZE;
+                } else {
+                    // If the buffer was not fully initialized, we can assume the oldest index is the first item in
+                    // the buffer. We can overwrite this variable since it is no longer used ahead.
+                    bufferLength = oldestIndex;
+                    oldestIndex = 0;
```
nventuro, line 128:
```suggestion
                    // If the buffer was not fully initialized, we haven't wrapped around it yet and can treat it as a regular array
                    // where the oldest index is the first one, and the length the number of samples.
                    bufferLength = oldestIndex; // Equal to latestIndex.next()
                    oldestIndex = 0;
```

nventuro (approved):
Looks great!

PR #772, opened 2021-08-03
Initialize oracle to save gas on swaps

EndymionJkb:
Add method to initialize (a range of) oracle sample slots, to save gas on swaps.


pkg/pool-utils/contracts/oracle/PoolPriceOracle.sol
```diff
@@ -77,6 +77,24 @@ abstract contract PoolPriceOracle is IPoolPriceOracle, IPriceOracle {
...
+     * @dev Manually fill oracle slots with dummy data, so that subsequent swaps filling them with real data
+     * cost less gas. Initialize by range, since filling a slot costs roughly 20k gas, so trying to fill the
+     * entire buffer at once would run out of gas.
+     */
+    function initializeOracle(uint256 startSlot, uint256 endSlot) external {
+        _require(startSlot < endSlot && endSlot <= Buffer.SIZE, Errors.OUT_OF_BOUNDS);
+
+        // Cannot use zero, or it will not actually do anything (uninitialized entries are already 0)
+        // Cannot use 0x01, since the timestamp is in the LSB position, and needs to be zero for invalid entries
+        bytes32 initSample = bytes32(0xba10000000000000000000000000000000000000000000000000000000000000);
```
nventuro, line 90:
```suggestion
        // Uninitialized samples are identified by a zero timestamp -- all other fields are ignored, so any non-zero value
        // with a zero timestamp suffices.
        bytes32 initSample = Samples.pack(1, 0, 0, 0, 0, 0, 0);
```

pkg/pool-utils/contracts/oracle/PoolPriceOracle.sol
```diff
@@ -77,6 +77,24 @@ abstract contract PoolPriceOracle is IPoolPriceOracle, IPriceOracle {
...
+        for (uint256 i = startSlot; i < endSlot; i++) {
+            if (_samples[i] == 0) {
```
nventuro, line 92:
```suggestion
            if (_samples[i].timestamp() == 0) {
```

pkg/pool-weighted/test/WeightedPool2Tokens.test.ts
```diff
@@ -119,6 +130,38 @@ describe('WeightedPool2Tokens', function () {
...
+    const itUpdatesTheOracleDataAndMeasuresGas = (action: PoolHook, lastChangeBlockOffset = 0) => {
+      context('without updated oracle', () => {
+        it('updates the oracle data (and measures gas)', async () => {
+          const previousData = await pool.getMiscData();
+
+          await advanceTime(MINUTE * 10); // force index update
+          const { receipt } = await action(await calcLastChangeBlock(lastChangeBlockOffset));
+          const uninitializedGasUsed = receipt.gasUsed;
+          let initializedGasUsed;
+          console.log(`Gas used (uninitialized oracle): ${printGas(uninitializedGasUsed)}`);
```
nventuro, line 142: We should not be printing logs in tests: they should just pass or fail. 

Regarding the tests themselves, I would not bother at all looking at receipts and the gas spent. As seen in this PR, doing so is quite convoluted and hard to specify, as there are many factors that affect a transaction's gas cost, and we cannot isolate the buffer update itself. Also, we don't really care about testing the EVM gas semantics.

I'd remove this and just have simple tests for the `iniitalizeOracle` function that check:
 - start < end
 - end <= 1024
 - already init samples are not affected
 - uninit samples are init with a non-zero value

pkg/pool-utils/contracts/oracle/PoolPriceOracle.sol
```diff
@@ -77,6 +77,24 @@ abstract contract PoolPriceOracle is IPoolPriceOracle, IPriceOracle {
         return Buffer.SIZE;
     }
 
+    /**
+     * @dev Manually fill oracle slots with dummy data, so that subsequent swaps filling them with real data
+     * cost less gas. Initialize by range, since filling a slot costs roughly 20k gas, so trying to fill the
+     * entire buffer at once would run out of gas.
+     */
+    function initializeOracle(uint256 startSlot, uint256 endSlot) external {
```
nventuro, line 85: 'initialize' is not a good name here: it leads people into thinking that this, well, initializes the oracle, which it doesn't.

I'd use highly-specific vocabulary here, specially considering this is a low-level utility that the vast, vast majority of people don't care about. Maybe something like `dirtyUninitializedOracleSamples`.

pkg/pool-utils/contracts/oracle/PoolPriceOracle.sol
nventuro: I don't like that constant being coupled to the layout of the samples - it's too error prone. We should also have tests here to prevent such issues.

pkg/pool-utils/contracts/oracle/PoolPriceOracle.sol
EndymionJkb: I had it like that so it didn't overwrite slots it already initialized. If someone initialized 100-200, and then somebody else did 150-250, the second one would only write to 201-250. I suppose it doesn't hurt (or cost as much gas) to overwrite something. Is it cheaper to check timestamp() == 0 vs the whole thing 0?

pkg/pool-utils/contracts/oracle/PoolPriceOracle.sol
EndymionJkb: initializeOracleStorage?

pkg/pool-weighted/test/WeightedPool2Tokens.test.ts
EndymionJkb: Ok; I wanted to at least see it to verify that it's working, since the whole point is to save gas. I can replace them

pkg/pool-utils/contracts/oracle/PoolPriceOracle.sol
EndymionJkb: I added tests for it

pkg/pool-utils/contracts/oracle/PoolPriceOracle.sol
nventuro: Sorry, that's not the point of my change. The point is that we're currently relying on 'timestamp != 0' to mean 'slot is uninitialized' - we should use that throughout (maybe even include a method that returns a bool).

This doesn't change the behavior of the contract.

pkg/pool-utils/contracts/oracle/PoolPriceOracle.sol
nventuro: No, my gripe is with the 'initialize' word: we're not initializing anything on the oracle, it is as 'empty' as it used to, there's no new data. 

What we're doing is an extremely niche and low level thing, which is dirtying the storage slot to make the next write cheaper due to it being already allocated in the node non volatile storage. The name should reflect that.

PR #773, opened 2021-08-04
Dynamic oracle parameters

EndymionJkb:
Default preserves present behavior. Can extend (but not shrink) the buffer size, up to a maximum 128k (to prevent an "infinite" buffer). Can change the sample duration between 0 and the max of 2 minutes.

Could also create an immutable sample duration and pass it in, but this extra argument would break compatibility with the past factory.


nventuro:
The changes themselves look good, but in the end we decided there'd be too much added complexitiy by allowing this level of granular configuration - we think the default oracle parameters are fine.

PR #778, opened 2021-08-10
Lido relayer

TomAFrench:
This PR strips out all of the `joinAndSwap` functionality from #756 to give a relayer which focuses on allowing performing the 4 basic vault operations using stETH.


pkg/standalone-utils/contracts/LidoRelayer.sol
```diff
@@ -0,0 +1,210 @@
...
+        address recipient = funds.recipient;
+
+        // Find the index of wstETH in the assets array
+        uint256 wstETHIndex;
+        for (uint256 i; i < assets.length; i++) {
+            if (assets[i] == IAsset(address(_wstETH))) {
+                wstETHIndex = i;
+                break;
+            }
+            require(i < assets.length, "Does not require wstETH");
```
nventuro, line 117: Same comment as in `join`

pkg/standalone-utils/contracts/LidoRelayer.sol
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        int256 wstETHLimit = limits[wstETHIndex];
+        if (wstETHLimit > 0) {
+            // If wstETH is an input then we want to send it from the relayer
+            // as we wrap it there.
+            funds.sender = address(this);
```
nventuro, line 124: Might be obvious, but this assumes that the _only_ input token is wstETH

pkg/standalone-utils/contracts/LidoRelayer.sol
```diff
@@ -0,0 +1,210 @@
...
+            require(!funds.fromInternalBalance, "Cannot send from internal balance");
+
+            _pullStETHAndWrap(msg.sender, uint256(wstETHLimit));
+            _approveToken(IERC20(address(_wstETH)), address(getVault()), uint256(wstETHLimit));
+        } else {
+            // If wstETH is an output then we want to receive it on the relayer
+            // so we can unwrap it before forwarding stETH to the user
+            funds.recipient = payable(address(this));
+            require(!funds.toInternalBalance, "Cannot send to internal balance");
```
nventuro, line 133: Similarly, this assumes the only output token is wstETH. We should be explicit about this in the documentation.

pkg/standalone-utils/contracts/LidoRelayer.sol
```diff
@@ -0,0 +1,210 @@
...
+        swapAmounts = getVault().batchSwap{ value: msg.value }(kind, swaps, assets, funds, limits, deadline);
+
+        if (swapAmounts[wstETHIndex] < 0) {
+            // Unwrap any received wstETH for the user automatically
+            uint256 wstETHAmount = uint256(-swapAmounts[wstETHIndex]);
+            _unwrapAndPushStETH(recipient, wstETHAmount);
+        } else if (kind == IVault.SwapKind.GIVEN_OUT) {
+            // GIVEN_OUT swaps with wstETH input may leave some dust on the relayer
+            // This should be forwarded on to the user
+            _unwrapAndPushStETH(recipient, IERC20(address(_wstETH)).balanceOf(address(this)));
```
nventuro, line 145: Can't we also have dust on 'given in'? Since we pulled the full limit.

pkg/standalone-utils/contracts/LidoRelayer.sol
```diff
@@ -0,0 +1,210 @@
...
+        require(sender == msg.sender, "Invalid sender");
+
+        // Pull in wstETH, wrap and return to user
+        uint256 wstETHAmount;
+        for (uint256 i; i < request.assets.length; i++) {
+            if (request.assets[i] == IAsset(address(_wstETH))) {
+                wstETHAmount = request.maxAmountsIn[i];
+                break;
+            }
+            require(i < request.assets.length, "Does not require wstETH");
```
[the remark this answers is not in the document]
TomAFrench: Ah, good old off by two errors. This is meant to be

```
require(i < request.assets.length - 1, "Does not require wstETH");
```

pkg/standalone-utils/contracts/LidoRelayer.sol
TomAFrench: Technically true but this was written assuming that the user would set the limit equal to the amount of wstETH they want to swap.

pkg/standalone-utils/contracts/LidoRelayer.sol
```diff
@@ -0,0 +1,210 @@
...
+        require(sender == msg.sender, "Invalid sender");
+
+        uint256 wstETHBalanceBefore = IERC20(address(_wstETH)).balanceOf(recipient);
+
+        getVault().exitPool(poolId, sender, recipient, request);
+
+        uint256 wstETHBalanceAfter = IERC20(address(_wstETH)).balanceOf(recipient);
+
+        // Pull in wstETH, unwrap and return to user
+        uint256 wstETHAmount = wstETHBalanceAfter - wstETHBalanceBefore;
```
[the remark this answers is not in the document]
TomAFrench: Happy to change but the user would have to be interacting with pretty weird tokens/pools for this to fail.

pkg/standalone-utils/contracts/LidoRelayer.sol
TomAFrench: I've added a couple of lines on the risks of multiple inputs/outputs.

pkg/standalone-utils/contracts/LidoRelayer.sol
```diff
@@ -71,21 +71,26 @@ contract LidoRelayer is RelayerAssetHelpers, ReentrancyGuard {
             uint256 wstETHAmount = singleSwap.kind == IVault.SwapKind.GIVEN_IN ? singleSwap.amount : limit;
             _pullStETHAndWrap(msg.sender, wstETHAmount);
             _approveToken(IERC20(address(_wstETH)), address(getVault()), wstETHAmount);
+
+            swapAmount = getVault().swap{ value: msg.value }(singleSwap, funds, limit, deadline);
+
+            // GIVEN_OUT trades can leave an unknown amount of wstETH
+            // We then must refund the full relayer balance
+            _unwrapAndPushStETH(msg.sender, IERC20(address(_wstETH)).balanceOf(address(this)));
```
nventuro, line 79: We could put this behind an `if kind == given out` and skip the `balanceOf` call, no?

pkg/deployments/tasks/20210812-lido-relayer/index.ts
```diff
@@ -0,0 +1,25 @@
...
+  goerli: '0x6320cd32aa674d2898a68ec82e869385fc5f7e2f',
+  kovan: '0xA387B91e393cFB9356A460370842BC8dBB2F29aF',
+};
+
+export default async (task: Task, { force, from }: TaskRunOptions = {}): Promise<void> => {
+  const output = task.output({ ensure: false });
+  const input = task.input() as LidoRelayerDeployment;
+  const args = [input.vault, wstETHMap[task.network]];
+
+  if (force || !output.relayer) {
```
nventuro, line 17: There's now a `deployAndVerify` helper that does all of this

nventuro:
I've verified locally that the build-info artifacts from commit 7724a38 of the rate provider repo correspond to compiling WstETHRateProvider in said commit.

PR #779, opened 2021-08-10
Deployments: Polish publish task

nventuro:
An issue the new structure has is that importing a contract now requires knowing not only the task id, contract name and network, but also the internal id used in the `output/<network>.json` file, which so far has been roughly arbitrary (e.g. `factory` for the multiple factories, `vault` for the vault, etc.). We might want to use the contract name instead of these, as I don't expect a task will include multiple deployments of the same contract.

### DIFF
--- a/pkg/deployments/README.md
+++ b/pkg/deployments/README.md
@@ -3,17 +3,60 @@
 # Balancer V2 Deployments
 
 [![NPM Package](https://img.shields.io/npm/v/@balancer-labs/v2-deployments.svg)](https://www.npmjs.org/package/@balancer-labs/v2-deployments)
+[![GitHub Repository](https://img.shields.io/badge/github-master-lightgrey?logo=github)](https://github.com/balancer-labs/balancer-v2-monorepo/tree/master/pkg/deployments)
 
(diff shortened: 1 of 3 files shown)


pkg/deployments/README.md
```diff
@@ -3,17 +3,60 @@
...
+| Description                                     | Task ID                                                                                  |
+| ----------------------------------------------- | ---------------------------------------------------------------------------------------- |
+| Authorizer, governance contract                 | [`20210418-authorizer`](./tasks/20210418-authorizer)                                     |
+| Vault, main protocol contract                   | [`20210418-vault`](./tasks/20210418-vault)                                               |
+| Weighted Pools of up to 8 tokens                | [`20210418-weighted-pool`](./tasks/20210418-weighted-pool)                               |
+| Weighted Pools with two tokens and price oracle | [`20210418-weighted-pool`](./tasks/20210418-weighted-pool)                               |
+| Liquidity Bootstrapping Pools of up to 4 tokens | [`20210721-liquidity-bootstrapping-pool`](./tasks/20210721-liquidity-bootstrapping-pool) |
+| Stable Pools of up to 5 tokens                  | [`20210624-stable-pool`](./tasks/20210624-stable-pool)                                   |
```
facuspagnuolo, line 62: I'd invert the order of the columns

facuspagnuolo (approved):

PR #780, opened 2021-08-11
Linear pool math rounding

dmf7z:
Add initial rounding to linear pool math. It is possible that we may also want to round up and down nominal functions. We will know it after stress testing this.


facuspagnuolo (approved):
LGTM!

PR #783, opened 2021-08-11
Deployments: Provide basic helpers

facuspagnuolo:


pkg/deployments/index.ts
```diff
@@ -0,0 +1,72 @@
+import { Contract } from 'ethers';
+
+/**
+ * @dev Creates an ethers Contract object for a canonical contract deployed on a specific network
+ * @param task ID of the task to fetch the deployed contract
+ * @param contract Name of the contract to be fetched
+ * @param network Name of the network looking the deployment for (e.g. mainnet, rinkeby, ropsten, etc)
+ */
+export async function getBalancerDeployedContract(task: string, contract: string, network: string): Promise<Contract> {
```
nventuro, line 9: What about `getBalancerCanonicalContract`? 'deployed' might be confusing, since all contracts are 'deployed'

pkg/deployments/index.ts
```diff
@@ -0,0 +1,72 @@
...
+  return `@balancer-labs/v2-deployments/tasks/${task}/abi/${contract}.json`;
+}
+
+/**
+ * @dev Returns the deployment path for a specific task on a network
+ * @param task ID of the task to look the deployment path for the required network
+ * @param network Name of the network looking the deployment path for (e.g. mainnet, rinkeby, ropsten, etc)
+ */
+export function getBalancerDeploymentPath(task: string, network: string): string {
+  return `@balancer-labs/v2-deployments/tasks/${task}/output/${network}.json`;
```
nventuro, line 71: I wouldn't export these to avoid having people depend on them, coupling external code too much to our directory structure.

pkg/deployments/index.ts
facuspagnuolo: True, but I'd try to simplify it instead and use `getBalancerContract` directly

PR #784, opened 2021-08-11
Deployments: Fix ABI exports

facuspagnuolo:


nventuro (approved):
Looks great, but let's wait until https://github.com/balancer-labs/balancer-v2-monorepo/pull/768 is merged to avoid further conflicts.

PR #785, opened 2021-08-11
Deployments: Fix published files

facuspagnuolo:


pkg/deployments/package.json
```diff
@@ -12,17 +12,19 @@
...
-    "!test.json"
+    "dist/"
   ],
   "scripts": {
     "lint": "eslint . --ext .ts --max-warnings 0",
     "test": "hardhat test ./**/test/*.ts",
     "test:deployed": "yarn test --network mainnet",
-    "test:fork": "yarn test --fork mainnet --block-number 12731159"
+    "test:fork": "yarn test --fork mainnet --block-number 12731159",
+    "prepublishOnly": "tsc"
```
nventuro, line 27: Let's move this to `build` and run it on CI. Also, it seems to be currently failing:

```
error TS2339: Property 'ethers' does not exist on type '{ default: HardhatRuntimeEnvironment; config: HardhatConfig; hardhatArguments: HardhatArguments; tasks: TasksMap; run: RunTaskFunction; network: Network; artifacts: Artifacts; }'.

17   const { ethers } = await import('hardhat');
```

pkg/deployments/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "moduleResolution": "node",
+    "declaration": true,
+    "outDir": "dist",
+    "resolveJsonModule": true
+  },
+  "include": [
+    "src",
```
nventuro, line 10:
```suggestion
    "index.ts",
    "src",
```

:stuck_out_tongue:

pkg/deployments/package.json
```diff
@@ -12,17 +12,19 @@
...
   },
   "devDependencies": {
```
nventuro, line 29: We might consider making hardhat, ethers and hardhat-ethers peer-dependencies.

PR #789, opened 2021-08-12
Add test for wstETH rate provider deployment

TomAFrench:
Very simple test just to check that we're matching the response from wstETH.


nventuro (approved):
Just great! :raised_hands:

PR #791, opened 2021-08-17
Cache purge

nventuro:
This follows up https://github.com/balancer-labs/balancer-v2-monorepo/pull/769 with a few improvements and comments, mostly related to cache purging. I then realized that a Pool changing its token set is a very uncommon event that doesn't make sense to optimize for, but oh well.

I also extracted the few cache tests we had into its own file, adding a few basic ones.

### DIFF
--- a/pkg/distributors/contracts/PoolTokenCache.sol
+++ b/pkg/distributors/contracts/PoolTokenCache.sol
@@ -37,10 +37,13 @@ contract PoolTokenCache {
             // Purge potentially stale cached data
             uint256 numTokens = _poolTokenSets[poolId].length();
 
-            // Clear the set by removing the zeroeth element n times
-            for (uint256 t; t < numTokens; t++) {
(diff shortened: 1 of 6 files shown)


gtaschuk (commented):
So this is a tradeoff right?  It increases gas costs for `EnumerableSet/Map` delete operations (especially for large sets) across the codebase to slightly reduce gas costs of callback contracts when used with mutable token sets.

If that's a tradeoff we are willing to make, then the implementation here looks fine.

nventuro:
It increases gas costs of some forms of `delete` _very slightly_ (i.e. there's one extra `if` comparison that uses stack variables), in exchange of greatly reducing the cost of clearing the set.

nventuro:
Also turns out the implementation is wrong :stuck_out_tongue:

gtaschuk (approved):

PR #794, opened 2021-08-19
Add BPT Allowlist

### LINKED ISSUE #793: Allowlist BPT Holders
Some institutional use cases require that only allowed parties participate in the pool, i.e. hold the BPT. A very simple way to achieve this is at the BPT contract level, without altering Pool logic.

Note that swaps are not affected by this.

rabmarut:
This draft is a proof of concept for an alternative version of BalancerPoolToken with allowlist functionality, namely:
- Only addresses on a list can ever hold the token.
- Only the pool owner can add to or remove from this list.
- The enforcement of the allowlist happens on all internal transfers, including mint and burn. Both sender and recipient must be verified, or the transfer reverts.

There are a few different ways to do this, so if we want to go a different way that's fine. I implemented this at the `_beforeTokenTransfer` level, which means that it automatically gets checked on every `_transfer`, `_mint`, and `_burn`. It could alternatively be done by overriding both `_transfer` and `_mint` separately, saving one needless check in `_mint` and two in `_burn`.

Authentication is a bit goofy because this allowlist pattern is typically implemented with a single owner, but I wanted to use the Authorizer instead. The goofy part is that the Vault's Authorizer is introduced in `BasePool`, and the BPT concept lives above that. So I opted to have the new BPT derive from `BasePoolAuthorization`, which introduces the concept of the `_owner` and basic authentication. As a result, the new BPT must be `abstract`, which is a bit odd. One alternative might be to move the completed authentication logic from `BasePool` to `BalancerPoolToken`.

Closes #793.


rabmarut:
Looks like inheritance is not working how I expected it to...

```
contracts/smart/InvestmentPool.sol:30:1: TypeError: Linearization of inheritance graph impossible
contract InvestmentPool is RestrictedBalancerPoolToken, BaseWeightedPool, ReentrancyGuard {
^ (Relevant source part starts here and spans across multiple lines).
```

It may be that this method is untenable and I need to try another way. Will revisit tomorrow and consult with the experts.

EndymionJkb:
I had been trying to do this, with the idea being a project wanting more complex validation could replace the content of LiquidityProviderController, but it wouldn't change anything else. This doesn't quite work either: I got (different) inheritance errors:

```
abstract contract LiquidityProviderController is BalancerPoolToken {
    mapping(address => bool) internal _transferAllowlist;

    event AddressAddedToAllowList(address indexed addr);
    event AddressRemovedFromAllowList(address indexed addr);

    function transfer(address recipient, uint256 amount) public override(ERC20) returns (bool) {
        _require(_transferAllowlist[recipient], Errors.ERC20_TRANSFER_NOT_ALLOWED);

        return super.transfer(recipient, amount);
    }

    function transferFrom(address sender, address recipient, uint256 amount)
        public
        override(BalancerPoolToken)
        returns (bool)
    {
        _require(_transferAllowlist[sender], Errors.ERC20_TRANSFER_NOT_ALLOWED);
        _require(_transferAllowlist[recipient], Errors.ERC20_TRANSFER_NOT_ALLOWED);

        return super.transferFrom(sender, recipient, amount);
    }

    function _addAddressToTransferAllowlist(address addr) internal {
        if (!_transferAllowlist[addr]) {
            _transferAllowlist[addr] = true;

            emit AddressAddedToAllowList(addr);
        }
    }

    function _removeAddressFromTransferAllowlist(address addr) internal {
        if (_transferAllowlist[addr]) {
            _transferAllowlist[addr] = false;

            emit AddressRemovedFromAllowList(addr);
        }
    }
}

contract InvestmentPool is BaseWeightedPool, ReentrancyGuard, LiquidityProviderController {
   ...

   function addAddressToTransferAllowlist(address addr) external authenticate {
        _addAddressToTransferAllowlist(addr);
    }

    function removeAddressFromTransferAllowlist(address addr) external authenticate {
        _removeAddressFromTransferAllowlist(addr);
    }

    function _isOwnerOnlyAction(bytes32 actionId) internal view override returns (bool) {
        return
            (actionId == getActionId(InvestmentPool.addAddressToTransferAllowlist.selector)) ||
            (actionId == getActionId(InvestmentPool.removeAddressFromTransferAllowlist.selector)) ||
            super._isOwnerOnlyAction(actionId);
    }
}

```

pkg/pool-utils/contracts/RestrictedBalancerPoolToken.sol
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * @dev Override to enforce address allowlist.
+     */
+    function _beforeTokenTransfer(
+        address from,
+        address to,
+        uint256
+    ) internal virtual override {
+        _require(isAllowedAddress(from), Errors.ERC20_TRANSFER_FROM_PROHIBITED_ADDRESS);
```
nventuro, line 53: Do we need this? I imagined restricting the recipients would be enough - the only way a sender would not be allowed would be if at some point they were allowed, but that was later revoked.

pkg/pool-utils/contracts/RestrictedBalancerPoolToken.sol
```diff
@@ -0,0 +1,92 @@
...
+            (actionId == getActionId(this.removeAllowedAddress.selector));
+    }
+
+    // Public functions
+
+    /**
+     * @dev Verifies that a given address is allowed to hold tokens.
+     */
+    function isAllowedAddress(address member) public view returns (bool) {
+        return _allowedAddresses[member];
```
nventuro, line 72:
```suggestion
        return member == address(0) && _allowedAddresses[member];
```

`_burn` sends tokens to the zero address, so the current implementation prevents burns unless we add this. Similarly, `_mint` sends from the zero address, so if we want to restrict `from` we'd also need it. This additionally removes these 'unnecessary' storage loads. (see https://docs.openzeppelin.com/contracts/3.x/api/token/erc20#ERC20-_beforeTokenTransfer-address-address-uint256-)

Alternatively, we could just override `_transfer` and `_mint`.

nventuro: Actually, I'd rather leave `isAllowedAddress` untouched, and have the `if` on the `_beforeTokenTransfer` hook, where we have context on the semantics of `from` and `sender`.

pkg/pool-utils/contracts/RestrictedBalancerPoolToken.sol
```diff
@@ -0,0 +1,92 @@
...
+
+    event AddressAdded(address member);
+    event AddressRemoved(address member);
+
+    constructor(string memory tokenName, string memory tokenSymbol, address owner)
+        BalancerPoolToken(tokenName, tokenSymbol)
+        BasePoolAuthorization(owner)
+    {
+        // Add zero address to allowlist to avoid blocking mints and burns
+        _allowedAddresses[address(0)] = true;
```
nventuro, line 40: With the other changes this would be unnecessary, resulting in overall less gas usage (and I'd argue more robust code, as e.g. the zero address could never be removed from the list of allowed addresses).

pkg/pool-utils/contracts/RestrictedBalancerPoolToken.sol
```diff
@@ -0,0 +1,92 @@
...
+        _require(isAllowedAddress(to), Errors.ERC20_TRANSFER_TO_PROHIBITED_ADDRESS);
+    }
+
+    /**
+     * @dev Override to enforce address allowlist
+     */
+    function _isOwnerOnlyAction(bytes32 actionId) internal view virtual override returns (bool) {
+        return
+            (actionId == getActionId(this.addAllowedAddress.selector)) ||
+            (actionId == getActionId(this.removeAllowedAddress.selector));
```
nventuro, line 63:
```suggestion
            (actionId == getActionId(this.removeAllowedAddress.selector)) ||
            super._isOwnerOnlyAction(actionId);
```

We shouldn't replace the original implementation, just add new action ids.

nventuro (commented):
Looks great!

Re. the linearization errors:
>contract InvestmentPool is BaseWeightedPool, RestrictedBalancerPoolToken

The order matters: you can try swapping the order of `RestrictedBPT` and `BaseWeightedPool` - I believe the new BPT should come before the BasePool, since BasePool uses that. 

This may not work though, if so we can try alternative approaches.

pkg/pool-utils/contracts/RestrictedBalancerPoolToken.sol
rabmarut: I think the situation you described is actually be a valid use case: someone is on the allowlist and then later removed because of bad behavior.

pkg/pool-utils/contracts/RestrictedBalancerPoolToken.sol
rabmarut: > Alternatively, we could just override _transfer and _mint.

Did you read my text in the body of the PR?

pkg/pool-utils/contracts/RestrictedBalancerPoolToken.sol
rabmarut: Fair point about removing the zero address!

pkg/pool-utils/contracts/RestrictedBalancerPoolToken.sol
rabmarut: Are you recommending this as a matter of convention? In this particular case, this is the highest level at which `_isOwnerOnlyAction` is implemented. There is no implementation in `super`.

rabmarut: I implemented it the same way it's implemented in `BasePool`.

rabmarut:
> The order matters

@nventuro I actually did try every possible order, and none worked - same linearization error on all of them. Does Solidity not properly handle two derivations off the same base? In this case, both `BasePool` and `RestrictedBalancerPoolToken` derive from both `BalancerPoolToken` and `BasePoolAuthorization`. There are definitely other ways to map the inheritance tree if this doesn't work, but I'm wondering if it's tenable at all because it is the easiest solution given what we currently have in place.

pkg/pool-utils/contracts/RestrictedBalancerPoolToken.sol
nventuro: Yup, I was referrencing that (I should've added a 'as you pointed out' at the end). The point I tried to get accross is that my suggestion is simply an alternative to the mint and transfer override - both are equally good solutions.

rabmarut: Sorry, I misunderstood your suggestion. I believe I've now implemented it as described.

nventuro:
I'm closing this for tidyness' sake since the implementation never got very far and the code is now a bit stale, even though we may want to reconsider this in the future.

PR #795, opened 2021-08-20
Investment Pool : enable/disable swaps

EndymionJkb:
Separated V2 into two PRs (V2 is now just weight change; this one covers halting trading). Permissioned tests still need work.

### DIFF
--- a/pkg/pool-weighted/contracts/smart/InvestmentPool.sol
+++ b/pkg/pool-weighted/contracts/smart/InvestmentPool.sol
@@ -19,11 +19,11 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/ReentrancyGuard.
 import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
 
 import "../BaseWeightedPool.sol";
+import "../WeightedPoolUserDataHelpers.sol";
 import "./WeightCompression.sol";
 
 /**
- * @dev Weighted Pool with mutable weights, designed to support investment use cases: large token counts,
- * rebalancing through gradual weight updates, and enabling/disabling trading.
+ * @dev Weighted Pool with mutable weights, designed to support investment use cases: enabling/disabling trading.
  */
 contract InvestmentPool is BaseWeightedPool, ReentrancyGuard {
     // solhint-disable not-rely-on-time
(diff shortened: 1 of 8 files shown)


pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -19,47 +19,47 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/ReentrancyGuard.
 import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
 
 import "../BaseWeightedPool.sol";
+import "../WeightedPoolUserDataHelpers.sol";
 import "./WeightCompression.sol";
 
 /**
- * @dev Weighted Pool with mutable weights, designed to support investment use cases: large token counts,
- * rebalancing through gradual weight updates, and enabling/disabling trading.
+ * @dev Weighted Pool with mutable weights, designed to support investment use cases: enabling/disabling trading.
```
nventuro, line 26: Why this change? The original comment was accurate.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -19,47 +19,47 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/ReentrancyGuard.
...
+    // Use the _miscData slot in BasePool
+    // First 64 bits are reserved for the swap fee
+    //
+    // Store non-token-based values:
+    //
+    // [ 64 bits  | 184 bits |    7 bits   |   1 bit   ]
+    // [ reserved | unused   | token count | swap flag ]
+    // |MSB                                         LSB|
+    uint256 private constant _SWAP_ENABLED_OFFSET = 0;
+    uint256 private constant _TOTAL_TOKENS_OFFSET = 1;
```
nventuro, line 47:
```suggestion
    uint256 private constant _TOTAL_TOKENS_OFFSET = 1; // 7 bits is enough since MAX_WEIGHTED_TOKENS is 100
```

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -88,19 +89,44 @@ contract InvestmentPool is BaseWeightedPool, ReentrancyGuard {
         uint256 numTokens = tokens.length;
         InputHelpers.ensureInputLengthMatch(numTokens, normalizedWeights.length, assetManagers.length);
 
-        _tokenCountCache = numTokens;
+        _setMiscData(_getMiscData().insertUint7(numTokens, _TOTAL_TOKENS_OFFSET));
```
nventuro, line 92: We don't _need_ it, but it won't hurt to have a `require(numTokens < 128)` (due to the 7 bits), just in case.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -159,6 +185,110 @@ contract InvestmentPool is BaseWeightedPool, ReentrancyGuard {
...
+    )
+        internal
+        override
+        returns (
+            uint256,
+            uint256[] memory,
+            uint256[] memory
+        )
+    {
+        // If swaps are disabled, require proportional join
```
nventuro, line 228:
```suggestion
        // If swaps are disabled, the only join kind that is allowed is the proportional one, as all others involve 
        // implicit swaps and alter token prices.
```

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -159,6 +185,110 @@ contract InvestmentPool is BaseWeightedPool, ReentrancyGuard {
...
+        internal
+        virtual
+        override
+        returns (
+            uint256,
+            uint256[] memory,
+            uint256[] memory
+        )
+    {
+        // If swaps are disabled, require proportional exit
```
nventuro, line 266:
```suggestion
        // If swaps are disabled, the only exit kind that is allowed is the proportional one, as all others involve 
        // implicit swaps and alter token prices.
```

pkg/pool-weighted/test/InvestmentPool.test.ts
```diff
@@ -124,4 +74,214 @@ describe('InvestmentPool', function () {
...
+      });
+
+      it('sets token weights', async () => {
+        const normalizedWeights = await pool.getNormalizedWeights();
+
+        // Not exactly equal due to weight compression
+        expect(normalizedWeights).to.equalWithError(pool.normalizedWeights, 0.0001);
+      });
+
+      describe.skip('permissioned actions', () => {
```
nventuro, line 182: Are these skipped on purpose?

nventuro (commented):
Thanks for making the effort of splitting the v2 PR!

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
EndymionJkb: Maybe: _require(_getMaxTokens() == numTokens, Errors.MAX_TOKENS);
101-127 would still be an overflow. If you get the same value out that you put in, you know it didn't overflow

pkg/pool-weighted/test/InvestmentPool.test.ts
EndymionJkb: No - they didn't work and I was too tired to see the problem. Maybe I can fix them now

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
nventuro: A bit indirect, but I like it. Go for it!

pkg/pool-weighted/test/InvestmentPool.test.ts
```diff
@@ -124,4 +74,214 @@ describe('InvestmentPool', function () {
...
+              const startingBpt = await pool.balanceOf(sender);
+
+              const { amountsIn } = await pool.joinAllGivenOut({ from: sender, bptOut: startingBpt });
+
+              const endingBpt = await pool.balanceOf(sender);
+              expect(endingBpt).to.be.gt(startingBpt);
+              expect(amountsIn).to.deep.equal(initialBalances);
+            });
+
+            it('disallows disproportionate exits (single token)', async () => {
```
nventuro, line 254: The first and third tests should not be called the same, as they are testing different join kinds. I'd consider having two describe blocks, one for proportional join kinds and one for disproportional ones, to make reading these easier.

PR #796, opened 2021-08-20
Add deployment config for Arbitrum mainnet

rabmarut:
The multisig which acts as the Authorizer admin lives here: https://explorer.arbitrum.io/address/0x6207ed574152496c9B072C24FD87cE9cd9E17320

And the WETH address (`0x82aF49447D8a07e3bd95BD0d56f35241523fBab1`) is taken from the official Arbitrum mainnet token list: https://bridge.arbitrum.io/token-list-42161.json

Please let me know if I missed anything.

### DIFF
--- a/pkg/deployments/src/types.ts
+++ b/pkg/deployments/src/types.ts
@@ -4,7 +4,7 @@ import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/dist/src/signer-wit
 
 import Task from './task';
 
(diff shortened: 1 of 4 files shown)


facuspagnuolo (approved):
It looks fine, maybe we could try testing the vault's deployment on an arbitrum-mainnet fork. We should be able to use alchemy to for that. Let me know if you need a hand setting it up

rabmarut:
Thanks! I actually test deployed to mainnet already so we should be good to go.

PR #800, opened 2021-08-25
Linear pool

dmf7z:
This PR adds the linear pool smart contract with the swap hook and phantom bpt.

Pending tasks:

1) Inheriting from BaseGeneralPool and BasePool forces to override parent functions because fees are managed differently. It also makes it impossible to do the initial join (bpt mint) in the constructor because it lacks the poolId. It is done in a function called initialize that is public because it needs to be removed. BasePool needs to be refactored or linear pool must have its own LinearBasePool. 

2) ~~Increase or decrease targets~~

3) ~~Add external rate to wrapped token (cache, etct).~~ ([#806](https://github.com/balancer-labs/balancer-v2-monorepo/pull/806))

4) Integration tests for swaps, ~~scaling factor tests ([#806](https://github.com/balancer-labs/balancer-v2-monorepo/pull/806))~~ and more tests to math contract.


pkg/balancer-js/src/utils/errors.ts
```diff
@@ -56,6 +56,10 @@ const balancerErrorCodes: Record<string, string> = {
   '328': 'CALLER_IS_NOT_LBP_OWNER',
   '329': 'PRICE_RATE_OVERFLOW',
   '330': 'INVALID_JOIN_EXIT_KIND_WHILE_SWAPS_DISABLED',
+  '331': 'LOWER_GREATER_THAN_UPPER_TARGET',
+  '332': 'UPPER_TARGET_TOO_HIGH',
+  '333': 'UNHANDLED',
```
facuspagnuolo, line 61: Let's be a bit more descriptive on this one

pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -0,0 +1,541 @@
...
+
+    function _getWrappedTokenCachedRate() internal view virtual returns (uint256) {
+        return _wrappedTokenRateCache.getValue();
+    }
+
+    function getTargets() external view returns (uint256 lowerTarget, uint256 upperTarget) {
+        return (_lowerTarget, _upperTarget);
+    }
+
+    function setTargets(uint256 lowerTarget, uint256 upperTarget) external authenticate {
```
facuspagnuolo, line 520: Shouldn't we emit an event here?

pkg/balancer-js/src/utils/errors.ts
dmf7z: what are you thinking for example? because it is used onJoin, onExit and onInitialized that are not handled

pkg/pool-linear/contracts/LinearPool.sol
dmf7z: yeap! 👍

pkg/balancer-js/src/utils/errors.ts
dmf7z: renaming to `UNHANDLED_BY_LINEAR_POOL`

facuspagnuolo (approved):

PR #801, opened 2021-08-28
MultiRewards: allows stakeWithPermit to stake on behalf of another user

gtaschuk:
Previously `stakeWithPermit` only allowed a user to use their own permit signature to stake.  This allows anyone to use the permit signature

### DIFF
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -227,8 +227,8 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
     }
 
     /* ========== MUTATIVE FUNCTIONS ========== */
-    function stake(IERC20 pool, uint256 amount) external {
-        stakeFor(pool, amount, msg.sender);
+    function stake(IERC20 pool, uint256 amount) external nonReentrant {
+        _stakeFor(pool, amount, msg.sender, msg.sender);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -241,11 +241,20 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
...
+    ) public nonReentrant {
+        _stakeFor(pool, amount, msg.sender, receiver);
+    }
+
+    function _stakeFor(
+        IERC20 pool,
+        uint256 amount,
+        address account,
+        address receiver
+    ) internal updateReward(pool, receiver) {
```
TomAFrench, line 253:
```suggestion
    ) internal nonReentrant updateReward(pool, receiver) {
```

I'd move the reentrancy guard to here to ensure that we don't forget to add another way to stake and forget to add the guard there.

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -262,13 +271,14 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
...
+        address account,
         address recipient,
         uint8 v,
         bytes32 r,
         bytes32 s
-    ) external {
-        IERC20Permit(address(pool)).permit(msg.sender, address(this), amount, deadline, v, r, s);
-        stakeFor(pool, amount, recipient);
+    ) external nonReentrant {
+        IERC20Permit(address(pool)).permit(account, address(this), amount, deadline, v, r, s);
```
TomAFrench, line 280: This is dangerous. I can pull this signature out of a transaction in the mempool and then include it in another transaction which sets me as the recipient - stealing the user's BPT.

We'd need to check another signature which includes the recipient field as well to make this secure.

TomAFrench (changes requested):
The current implementation is vulnerable to having the signature extracted and used maliciously.

pkg/distributors/contracts/MultiRewards.sol
TomAFrench: Either that or enforce that `account == recipient`

PR #803, opened 2021-08-30
Investment Pool (management fees)

EndymionJkb:
Merging "disable swap" and "management fees" was one too many arguments, so had to use a struct for the constructor: which broke all the deploys in InvestmentPool tests. So that needs to be fixed (they were working before).


pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -71,44 +82,76 @@ contract InvestmentPool is BaseWeightedPool, ReentrancyGuard {
         uint256[] endWeights
     );
     event SwapEnabledSet(bool swapEnabled);
+    event ManagementFeePercentageSet(uint256 managementFee);
```
TomAFrench, line 85:
```suggestion
    event ManagementFeePercentageChanged(uint256 managementFee);
```

We have a `SwapFeePercentageChanged` event so it might make sense to use the same format here (although we do have a split between "set" and "changed" already)

pkg/balancer-js/src/pool-weighted/encoder.ts
```diff
@@ -89,4 +90,11 @@ export class WeightedPoolEncoder {
       ['uint256', 'uint256[]', 'uint256'],
       [WeightedPoolExitKind.BPT_IN_FOR_EXACT_TOKENS_OUT, amountsOut, maxBPTAmountIn]
     );
+
+  /**
+   * Encodes the userData parameter for exiting an InvestmentPool for withdrawing management fees.
+   * This can only be done by the pool owner.
+   */
+  static exitForManagementFees = (): string =>
```
TomAFrench, line 98: I'm not entirely sure if this should live here seeing as 99% of people aren't going to use it. At the very least I think it should probably be in an `InvestmentPoolEncoder`?

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
EndymionJkb: I made it "Set" because it's immutable, and "Changed" implies otherwise

pkg/balancer-js/src/pool-weighted/encoder.ts
EndymionJkb: Yes - this was just easier. If this design (method of extracting fees) survives, we can refactor

TomAFrench: Agreed for this initial release.

EndymionJkb:
Superceded by #814

PR #804, opened 2021-08-30
Investment Pools: minimum weight change duration

EndymionJkb:
This was a thing we talked about that dropped. LBPs (V1 and V2) let you effectively do immediate weight changes, but that's less dangerous because there are no LPs (well, there could be on V1 if they configure it that way).

That would be dangerous on a big public pool, though. This is a simple hard-coded limit. Could pass it in if we wanted to make it configurable, but we'd either still need a hard-coded minimum, or rely on managers to not make it 0, or users to notice that it's 0 and the pool requires more trust.

Keeping it dirt simple for now. It at least prevents instantaneous weight changes.


nventuro (approved):
The changes themselves are fine, though I'd prefer we had a slightly more conservative minimum duration, such as 48 hours.

EndymionJkb:
> The changes themselves are fine, though I'd prefer we had a slightly more conservative minimum duration, such as 48 hours.

Agreed. Hopefully they will answer, and we will set this longer.

PR #805, opened 2021-08-30
MultiRewards: use Math.mul when dealing with time

gtaschuk:
Fix for issue [C01] of the OZ audit.

### DIFF
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -429,25 +429,33 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
 
         getVault().manageUserBalance(ops);
 
-        if (block.timestamp >= rewardData[pool][rewarder][rewardsToken].periodFinish) {
-            rewardData[pool][rewarder][rewardsToken].rewardRate = Math.divDown(
-                reward,
-                rewardData[pool][rewarder][rewardsToken].rewardsDuration
-            );
-        } else {
-            uint256 remaining = rewardData[pool][rewarder][rewardsToken].periodFinish.sub(block.timestamp);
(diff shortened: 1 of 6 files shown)


pkg/distributors/test/Reinvestor.test.ts
```diff
@@ -114,7 +114,7 @@ describe('Reinvestor', () => {
         ).wait();
 
         const deltas = [bn(0), bn(0)];
-        deltas[assets.indexOf(rewardToken.address)] = bn('999999999999999898');
+        deltas[assets.indexOf(rewardToken.address)] = bn('999999999999999498');
```
nventuro, line 117: Is this value different because we're advancing a different amount of time? Can this be computed instead of hardcoded?

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -429,25 +429,33 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
...
+            rewardData[pool][rewarder][rewardsToken].rewardsDuration
+        );
+        emit RewardAdded(address(rewardsToken), reward);
+    }
+
+    function _updateRewardRate(
+        IERC20 pool,
+        address rewarder,
+        IERC20 rewardsToken,
+        uint256 newReward
```
nventuro, line 444: I find the addition of this function slightly odd, as we're altering the same data structure both inside it and on its callsite.

I can it it being potentially motivated by how verbose all of this is due to repeated access to `rewardData[pool][rewarder][rewardsToken]`, and I propose we do the following instead: keep the original code as-is (just changing the fixed point mul for regular mul), and then in a separate PR use storage pointers to the struct to a) reduce verbosity and b) save gas by computing the storage slot only once.

nventuro: Went ahead and implemented this myself in the interest of time.

PR #806, opened 2021-08-31
Linear: Implement wrapped token rate cache

facuspagnuolo:


pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -433,6 +460,59 @@ contract LinearPool is BasePool, IGeneralPool, LinearMath, IRateProvider {
...
+            uint256 rate,
+            uint256 duration,
+            uint256 expires
+        )
+    {
+        rate = _wrappedTokenRateCache.getValue();
+        (duration, expires) = _wrappedTokenRateCache.getTimestamps();
+    }
+
+    function setWrappedTokenRateCacheDuration(uint256 duration) external authenticate {
```
dmf7z, line 480: Do we want the owner to manage all these rate authenticate functions? In that case we should add it to `_isOwnerOnlyAction`

dmf7z (approved):

pkg/pool-linear/contracts/LinearPool.sol
facuspagnuolo: I followed the same strategy we did for the meta stable pool, but let's discuss who should be able to access it

PR #812, opened 2021-09-01
MultiRewards: Made exitWithCallback nonReentrant

gtaschuk:
I don't exactly see an attack vector here but to be on the safe side...


TomAFrench:
This would result in us being unable to restake as part of a callback which is something that I think we'd want to be able to do.

gtaschuk:
> This would result in us being unable to restake as part of a callback which is something that I think we'd want to be able to do.

Great point.  That use case does seem worth preserving

Issue #676, opened 2021-06-28
Add Management Pool fees


nventuro:
These are fees that are charged by Pool administrators, independent of protocol fees. The mechanism we'll use for collecting these fees will likely be a new exit kind, where the sender (or recipient? both?) are permissioned accounts, and tokens are sent for no BPT, or a special join where new BPT is minted for no tokens.

A naive model would mean that LPs will be better off joining immediately after fees are collected, and exiting immediately before, but if fees are expected to be very low this might be fine. Alternatively, we could perform a type of accounting similar to the one we do with protocol fees, avoiding these issues.

Something else to discuss is what sort of fees we could support. At least two have been considered:
 - a 'streaming' fee, which would grant a percentage of the Pool's total holdings (e.g. a yearly 2% of total managed funds)
 - a 'performance' fee, in the form of a percentage of swap fees (the only 'performance' metric)

If we're interested in both, for simplicity's sake I believe we should handle both at once (instead of having two flavors).

The 'performance' fee is somewhat straightforward, as it is not too different from regular protocol fees. The 'streaming' case is more interesting, as we'd likely need to track BPT changes over time (in joins and exits), e.g. in an accumulator.

nventuro:
#814  impelemented the 'performance fee', albeit in a less-than-optimal manner for pools with multiple tokens: we'll need to do some more research on the math to find a more gas efficient implementation (e.g. one that mints BPT to the manager).

EndymionJkb:
As of PRs #1163, #1164, #1214 (after some infrastructure work in #1111), managed pools pay management fees (both swap and AUM kinds) in BPT.

PR #814, opened 2021-09-03
Investment pool management fee

nventuro:
This PR implements part of #676, in particular, the so-called 'performance fees'. A percentage of the swap fees charged on every swap, join and exit are allocated to the owner, who can collect them at any moment.

**This is a very large PR**, but it's been designed to be easily reviewable by looking at each individual commit in sequence. **Don't try to review the entire PR at once**. Each commit builds on top of the previous one in a logical way, adding tests where appropriate. A few odd commits just reorder some bits around (mostly the test suite).

Because weights in InvestmentPools are not fixed, the standard method of computing swap fees by measuring invariant growth over time is not suitable. Instead, we keep track of how many tokens have been collected as management swap fees, and subtract those from all incoming balances (as if they didn't exist in the Pool in the first place). Collected management fees can be claimed with a function that triggers a special exit on the Vault.

In order to provide as much confidence in this change as possible, we decided to simplify operation by making this Pool skip protocol fee computation: those are hardcoded to zero. This is acceptable as protocol fees are currently turned off, and this is expected to be a short-lived Pool type (replaced in the future by a full-fledged InvestmentPool).

Roughly speaking, these are the changes this PR introduces:
 - protocol fees are harcoded to zero
 - the notion of a management fee percentage is introduced. This value is immutable
 - each swap, joinswap and exitswap accumulates collected management fees
 - management fees are deducted from all incoming Vault balances (i.e. in swap, join and exit hooks)
 - the owner can withdraw collected fees by calling a permissioned function on the Pool, which triggers a special exit on the Vault

Supercedes #803


pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -307,56 +397,171 @@ contract InvestmentPool is BaseWeightedPool, ReentrancyGuard {
...
+
+        // If swaps are disabled, the only exit kind that is allowed is the proportional one (as all others involve
+        // implicit swaps and alter token prices) and management fee collection (as there's no point in restricting
+        // that).
+        ExitKind kind = userData.exitKind();
         _require(
-            getSwapEnabled() || userData.exitKind() == ExitKind.EXACT_BPT_IN_FOR_TOKENS_OUT,
+            getSwapEnabled() ||
+                kind == ExitKind.EXACT_BPT_IN_FOR_TOKENS_OUT ||
+                kind == ExitKind.MANAGEMENT_FEE_TOKENS_OUT,
```
TomAFrench, line 433: `_exitManagerFeeTokensOut` has `whenNotPaused` modifer so this will still fail.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -72,56 +80,88 @@ contract InvestmentPool is BaseWeightedPool, ReentrancyGuard {
         uint256[] startWeights,
         uint256[] endWeights
     );
-    event SwapEnabledSet(bool swapEnabled);
+    event SwapEnabledChanged(bool swapEnabled);
```
TomAFrench, line 83:
```suggestion
    event SwapEnabledSet(bool swapEnabled);
```

I'd like to maintain this as `SwapEnabledSet` for consistency with LBPs if possible. Having two events which are semantically the same could be annoying for subgraph purposes in future when I'm deduplicating some of the data source templates.

pkg/solidity-utils/test/EnumerableMap.behavior.ts
```diff
@@ -98,6 +99,30 @@ export function shouldBehaveLikeMap(
...
+      await store.map.set(keyA, valueA);
+      await store.map.set(keyB, valueB);
+
+      await store.map.remove(keyA);
+
+      // B is now the only element; its index must be 0
+      expect(await store.map.indexOf(keyB, indexOfErrorCode)).to.equal(0);
+    });
+
+    it('reverts if they key is not in the map', async () => {
```
TomAFrench, line 121:
```suggestion
    it('reverts if the key is not in the map', async () => {
```

pkg/pool-weighted/contracts/BaseWeightedPool.sol
```diff
@@ -280,21 +290,28 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool, WeightedMath {
...
             getSwapFeePercentage()
         );
 
+        // Note that swapFee is already upscaled
+        _processSwapFeeAmount(tokenIndex, swapFee);
+
+        // We join in a single token, so we initialize amountsIn with zeros
+        uint256[] memory amountsIn = new uint256[](_getTotalTokens());
+        // And then assign the result to the selected token
+        amountsIn[tokenIndex] = amountIn;
```
TomAFrench, line 313: Any reason for the difference here and in WeightedPool2Tokens? https://github.com/balancer-labs/balancer-v2-monorepo/blob/36c3847a0dd0cd9f3e98fa7130414153c4a58fe0/pkg/pool-weighted/contracts/WeightedPool2Tokens.sol#L574-L581

TomAFrench: Similarly for exits

pkg/pool-weighted/contracts/BaseWeightedPool.sol
```diff
@@ -377,43 +394,49 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool, WeightedMath {
...
+        (uint256 amountOut, uint256 swapFee) = WeightedMath._calcTokenOutGivenExactBptIn(
             balances[tokenIndex],
             normalizedWeights[tokenIndex],
             bptAmountIn,
             totalSupply(),
             getSwapFeePercentage()
         );
 
+        // This in exceptional situation in which the fee is charged on a token out instead of a token in. Note that
+        // swapFee is already upscaled.
```
TomAFrench, line 432:
```suggestion
        // This is an exceptional situation in which the fee is charged on a token out instead of a token in.
        // Note that swapFee is already upscaled.
```

pkg/pool-weighted/contracts/BaseWeightedPool.sol
```diff
@@ -455,6 +478,10 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool, WeightedMath {
         );
         _require(bptAmountIn <= maxBPTAmountIn, Errors.BPT_IN_MAX_AMOUNT);
 
+        // This in exceptional situation in which the fees are charged on tokens out instead of a token in. Note that
+        // swapFees is already upscaled.
```
TomAFrench, line 482:
```suggestion
        // This is an exceptional situation in which the fee is charged on a token out instead of a token in.
        // Note that swapFee is already upscaled.
```

TomAFrench (commented):
Everything looks reasonable from an initial look through of each commit. I'd need to have another go through in a bit more detail to be totally happy though.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -159,7 +199,9 @@ contract InvestmentPool is BaseWeightedPool, ReentrancyGuard {
     }
 
     function _getMaxTokens() internal pure virtual override returns (uint256) {
-        return _MAX_WEIGHTED_TOKENS;
+        // The upper bound is WeightedMath.MAX_WEIGHTED_TOKENS, but this is constrained by other factors, such as Pool
+        // creation gas consumption (which is linear).
+        return 50;
```
TomAFrench, line 204:
```suggestion
        return _INVESTMENT_MAX_TOKENS;
```

Can we pull this out to a constant at the top of the contract? It'll make it easier to tell the maximum number of tokens at a glance.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
nventuro: Ah fair point, I didn't remember the LBPs also had this event - I went with 'changed' as that is what we use in other events. I'll revert this.

pkg/pool-weighted/contracts/BaseWeightedPool.sol
nventuro: WeightedPool2Tokens ignores the swap fee, so I just kept the old code and assigned directly into the array. We could change those to make them as similar as possible.

Part of the reason is that Solidity lets you either assign to variables in tuple deconstruction or declare new ones, but not both. For example, the following is invalid:

```
function baz() returns (uint256, uint256);

uint256 foo;
(foo, uint256 bar) = baz();
```

TomAFrench: Ah, I missed the difference in handling of swap fees. In that case then I agree we should leave this as be as we won't really gain much in return for the reduced readability.

pkg/pool-utils/contracts/BaseMinimalSwapInfoPool.sol
```diff
@@ -97,4 +109,41 @@ abstract contract BaseMinimalSwapInfoPool is IMinimalSwapInfoPool, BasePool {
         uint256 balanceTokenIn,
         uint256 balanceTokenOut
     ) internal virtual returns (uint256);
+
+    /**
+     * @dev Called whenever a swap fee is charged. Implementations should call their parent's via super, to ensure all
+     * implementations in the inheritance tree are called.
+     *
+     * Callers should make sure that any one of the three `_processSwapFeeAmount` functions is called when swap fees are
+     * computed, and to make `amount` upscaled.
```
EndymionJkb, line 118:
```suggestion
     * @dev Called whenever a swap fee is charged. Implementations should call their parents via super, to ensure all
     * implementations in the inheritance tree are called.
     *
     * Callers must call one of the three `_processSwapFeeAmount` functions when swap fees are computed,
     * and upscale `amount`.
```

pkg/pool-weighted/contracts/WeightedMath.sol
```diff
@@ -317,7 +374,10 @@ contract WeightedMath {
         uint256 taxableAmount = amountOutWithoutFee.mulUp(taxablePercentage);
         uint256 nonTaxableAmount = amountOutWithoutFee.sub(taxableAmount);
 
-        return nonTaxableAmount.add(taxableAmount.mulDown(FixedPoint.ONE.sub(swapFeePercentage)));
+        uint256 taxableAmountMinusFees = taxableAmount.mulDown(FixedPoint.ONE.sub(swapFeePercentage));
+
+        swapFee = taxableAmount - taxableAmountMinusFees;
+        amountOut = nonTaxableAmount.add(taxableAmountMinusFees);
```
rabmarut, line 380: Taking inspiration from lines +194 to +197, this should be a little more readable:

```suggestion
        swapFee = taxableAmount.mulUp(swapFeePercentage);
        amountOut = nonTaxableAmount.add(taxableAmount.sub(swapFee));
```

rabmarut (approved):
Looks great! I made one very minor suggestion which is totally optional.

pkg/pool-weighted/contracts/WeightedMath.sol
nventuro: You're right, thanks!

pkg/pool-utils/contracts/BaseMinimalSwapInfoPool.sol
```diff
@@ -38,7 +38,13 @@ abstract contract BaseMinimalSwapInfoPool is IMinimalSwapInfoPool, BasePool {
 
         if (request.kind == IVault.SwapKind.GIVEN_IN) {
             // Fees are subtracted before scaling, to reduce the complexity of the rounding direction analysis.
-            request.amount = _subtractSwapFeeAmount(request.amount);
+            uint256 amountInMinusSwapFees = _subtractSwapFeeAmount(request.amount);
+
+            // Process the (upscaled!) swap fee.
+            uint256 swapFee = request.amount - amountInMinusSwapFees;
```
facuspagnuolo, line 44: This sounds a bit awkward, isn't better to have another helper in the base pool that works as follows:

```solidity
uint256 swapFeeAmount = _getSwapFeeAmounts(request.amount);
uint256 amountInMinusSwapFees = request.amount.sub(swapFeeAmount);
```

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -462,7 +462,14 @@ contract InvestmentPool is BaseWeightedPool, ReentrancyGuard {
...
         bptAmountIn = 0;
-        amountsOut = new uint256[](_getTotalTokens()); // TODO: set the actual token amounts
+
+        amountsOut = new uint256[](_getTotalTokens());
+        for (uint256 i = 0; i < _getTotalTokens(); ++i) {
+            // We can use unchecked getters and setters as we know the map has the same size (and order!) as the Pool's
+            // tokens.
+            amountsOut[i] = _tokenCollectedManagementFees.unchecked_valueAt(i);
+            _tokenCollectedManagementFees.unchecked_setAt(i, 0);
+        }
```
facuspagnuolo, line 505: I find a bit weird we are recomputing the manager fees in the callback, should we find a way to re-use what we do in `collectManagementFees` or maybe try to decode it from `userData` here?

facuspagnuolo (commented):
Leaving here some initial thoughts

pkg/pool-stable/contracts/StablePool.sol
```diff
@@ -413,9 +413,10 @@ contract StablePool is BaseGeneralPool, BaseMinimalSwapInfoPool, StableMath, IRa
             return _exitExactBPTInForTokenOut(balances, userData);
         } else if (kind == ExitKind.EXACT_BPT_IN_FOR_TOKENS_OUT) {
             return _exitExactBPTInForTokensOut(balances, userData);
-        } else {
-            // ExitKind.BPT_IN_FOR_EXACT_TOKENS_OUT
+        } else if (kind == ExitKind.BPT_IN_FOR_EXACT_TOKENS_OUT) {
             return _exitBPTInForExactTokensOut(balances, scalingFactors, userData);
+        } else {
+            _revert(Errors.UNHANDLED_EXIT_KIND);
```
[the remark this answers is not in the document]
nventuro: Yes, absolutely. That's why I also changed it on StablePool.

pkg/pool-weighted/contracts/BaseWeightedPool.sol
```diff
@@ -384,9 +391,10 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool, WeightedMath {
             return _exitExactBPTInForTokenOut(balances, normalizedWeights, userData);
         } else if (kind == ExitKind.EXACT_BPT_IN_FOR_TOKENS_OUT) {
             return _exitExactBPTInForTokensOut(balances, userData);
-        } else {
-            // ExitKind.BPT_IN_FOR_EXACT_TOKENS_OUT
+        } else if (kind == ExitKind.BPT_IN_FOR_EXACT_TOKENS_OUT) {
             return _exitBPTInForExactTokensOut(balances, normalizedWeights, scalingFactors, userData);
+        } else {
+            _revert(Errors.UNHANDLED_EXIT_KIND);
```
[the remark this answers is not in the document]
nventuro: Maybe? The underlying issue is that in the new exit method we added (`collectFees`) we needed an extra argument (`sender`) which is not available in the base `_doJoin`. We did the change this way to minimize disruption.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -183,6 +183,29 @@ contract InvestmentPool is BaseWeightedPool, ReentrancyGuard {
         _startGradualWeightChange(startTime, endTime, _getNormalizedWeights(), endWeights, tokens);
     }
 
+    function collectManagementFees(address recipient) external authenticate whenNotPaused nonReentrant {
+        (IERC20[] memory tokens, , ) = getVault().getPoolTokens(getPoolId());
```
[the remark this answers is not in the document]
nventuro: Indeed - this is replaced in a latter commit with a lookup over the set.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -282,81 +367,184 @@ contract InvestmentPool is BaseWeightedPool, ReentrancyGuard {
         return super._onSwapGivenOut(swapRequest, currentBalanceTokenIn, currentBalanceTokenOut);
     }
 
+    // We override _onJoinPool and _onExitPool as we need to not compute the current invariant and calculate protocol
+    // fees, since that mechanism does not work for Pools in which the weights change over time. Instead, this Pool
+    // always pays zero protocol fees.
+    // Additionally, we also check that only non-swap join and exit kinds are allowed while swaps are disabled.
+
```
EndymionJkb, line 374: Move this down to on top of _onJoinPool?

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
```diff
@@ -324,56 +329,41 @@ contract InvestmentPool is BaseWeightedPool, ReentrancyGuard {
...
-                sender,
-                recipient,
-                balances,
-                lastChangeBlock,
-                protocolSwapFeePercentage,
-                scalingFactors,
-                userData
-            );
+        (bptAmountOut, amountsIn) = _doJoin(balances, _getNormalizedWeights(), scalingFactors, userData);
+        dueProtocolFeeAmounts = new uint256[](_getTotalTokens());
```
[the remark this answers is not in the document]
nventuro: Discussed this internally and came to the conclusion that this value is only really useful for getting insights in how swap fees are being computed and how many have accumulated, so it's actually better to revert in that query.

pkg/pool-weighted/contracts/smart/InvestmentPool.sol
nventuro: We have to go over the loop anyway, and I liked having the redudancy provided by both lookups. This is also an infrequent operation, so I'm fine with doing it this way.

pkg/pool-utils/contracts/BaseMinimalSwapInfoPool.sol
nventuro: Yes, but that would require e.g. also changing `GeneralBasePool`, and I wanted to minimize disruption, especially considering we're going to revert these changes soon.

PR #815, opened 2021-09-03
Linear: Improve swap internal functions

facuspagnuolo:


dmf7z (approved):
Great work, reads much better!

PR #816, opened 2021-09-03
MultiDistributor - Added a StakingRelayer that enables join and stake

gtaschuk:
Bringing this back as a separate relayer that inherits `LidoRelayer`

Could be difficult to keep track of relayers and their capabilities in the frontend when we have multiple versions...

ExitWithCallback + an Exiter contract enables unstake and exit

`joinAndStake` from the relayer uses 294k gas vs 140k gas just to join (as measured from the stable pool tests)


TomAFrench:
Closing this as it's using the old relayer architecture and staking interface is changing.

PR #819, opened 2021-09-06
Linear pool math optimization and fixes

dmf7z:
This PR fixes an error in toNominal function.

It also optimize calcs for `calcBptOutPerMainIn` and `calcBptInPerMainOut`: 

```
const newBptSupply = bptSupply.mul(decimal(1).sub(deltaNominalMain.div(invariant)));
const bptIn = bptSupply.sub(newBptSupply);
```

equals

`const bptIn = bptSupply.mul(deltaNominalMain).div(invariant);`


facuspagnuolo (approved):
Nice catch!

PR #821, opened 2021-09-06
Add InvestmentPool deployment task

nventuro:
Someone should independently verify that the build info and ABIs on this commit are accurate and correspond to building the factory in this state.


TomAFrench:
I've recompiled the `pool-weighted` contracts on commit e388fdb9 and when copying the build-info file with id `bc3070b9b4f78c6318b791cb8cd9aee9` into `InvestmentPoolFactory.json` which resulted in an empty diff. The committed build-info should then be accurate.

PR #823, opened 2021-09-06
Add a factory for Investment Pools that doesn't allow asset managers

EndymionJkb:


EndymionJkb:
Superceded by 821 - we're just changing the main factory to exclude them

PR #827, opened 2021-09-08
Use a single transfer per token in MerkleOrchard

TomAFrench:
Not as pretty as before but will most likely substantially reduce gas costs.

### DIFF
--- a/pkg/distributors/contracts/MerkleOrchard.sol
+++ b/pkg/distributors/contracts/MerkleOrchard.sol
@@ -60,10 +60,10 @@ contract MerkleOrchard is IDistributor, Ownable {
         Claim[] memory claims,
         bool asInternalBalance
     ) internal {
-        IVault.UserBalanceOpKind kind = asInternalBalance
-            ? IVault.UserBalanceOpKind.TRANSFER_INTERNAL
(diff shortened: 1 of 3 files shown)


nventuro:
I suspect this is way better than repeated transfers since the overhead is quite low, but it should be noted that repeated transfers are not necessarily the end of the world with [EIP-2929](https://eips.ethereum.org/EIPS/eip-2929) live: the repeated `call`, `sload` and `sstore`s are quite less expensive than the original ones. It'd be great to have at least a simple comparison of measured costs.

gtaschuk (approved):
Yes - this seems like a clear enough win for me especially in the case of LM where people would be claiming across many weeks rather than a 1-off airdrop.

Obviously the orchard is just the seed of an idea rn but I'm glad you're jumping in on this - I'm getting excited about the potential.

TomAFrench:
I threw together some quick benchmarks on this. They're very non-representative of real gas costs as it's a single leave merkle tree but it's fine for seeing the effects that this change has.

tl;dr at two claims we're making a saving of ~8.6k gas so it's worth the extra overhead.

Before:
image

After:
image

PR #828, opened 2021-09-08
Use bitmap for tracking claims in MerkleOrchard

TomAFrench:
We now use a bitmap to track whether a particular leaf in a merkle tree has been used or not. With this up to 256 recipients can share a single storage slot in order to mark their claimed status, avoiding many cold stores.

Note: this has resulted in us having to add an explicit index into the merkle leaves.

I've also flattened the `rewardToken > rewarder > distribution` mapping structure to avoid repeated hashing (ever so slightly reducing gas costs)

### DIFF
--- a/pkg/distributors/contracts/MerkleOrchard.sol
+++ b/pkg/distributors/contracts/MerkleOrchard.sol
@@ -1,0 +1,220 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


TomAFrench:
Quick benchmark on this to show how the bitmap affects the costs of verifying claims (I've removed the actual distribution of tokens as that adds an extra variable). The first claimer for a given distribution has to initialise the bitmap so has to pay an additional gas costs (all claimers have to pay this currently) but the second gets reduced costs.

First claimer (Writing to an empty slot):
image

Second claimer (Writing to a dirty slot):
image

gtaschuk:
This is a great idea.  I do find it a bit cumbersome that we have to start including indices in the leaf nodes and it could create confusion if indices are improperly set/provided

I want to take a quick stab at inferring indices based on the index of the tree element that the submitted proof implies.  If it's a minor gas cost to compute that index, I think it's worth considering

nventuro:

PR #829, opened 2021-09-08
Use global incrementing distribution nonce in MerkleOrchard

TomAFrench:
I'm not a huge fan of allowing rewarders to specify the value of `distribution` for a given merkledrop. We have several functions which rely on it being incremented by one each time (`merkleRoots` and `claimStatus`) but nothing enforces this so we can't rely on these functions being useful. These functions only really make sense to read offchain so we can always use a multicall contract anyway.

My preference is that we lean into the fact that this is just a nonce, i.e. make it global and increment it each time we add a distribution.

### DIFF
--- a/pkg/distributors/contracts/MerkleOrchard.sol
+++ b/pkg/distributors/contracts/MerkleOrchard.sol
@@ -33,6 +33,7 @@ contract MerkleOrchard is IDistributor, Ownable {
     using SafeERC20 for IERC20;
 
     // Recorded distributions
+    uint256 public nextDistributionNonce;
     // rewardToken > rewarder > distribution > root
     mapping(IERC20 => mapping(address => mapping(uint256 => bytes32))) public trees;
     // rewardToken > rewarder distribution > lp > root
@@ -47,7 +48,7 @@ contract MerkleOrchard is IDistributor, Ownable {
     }
 
     struct Claim {
-        uint256 distribution;
+        uint256 distributionNonce;
         uint256 balance;
         address rewarder;
         IERC20 rewardToken;
@@ -70,7 +71,7 @@ contract MerkleOrchard is IDistributor, Ownable {
             claim = claims[i];
 
             require(
-                !isClaimed(claim.rewardToken, claim.rewarder, claim.distribution, liquidityProvider),
+                !isClaimed(claim.rewardToken, claim.rewarder, claim.distributionNonce, liquidityProvider),
                 "cannot claim twice"
             );
             require(
@@ -78,7 +79,7 @@ contract MerkleOrchard is IDistributor, Ownable {
                     claim.rewardToken,
                     claim.rewarder,
                     liquidityProvider,
-                    claim.distribution,
+                    claim.distributionNonce,
                     claim.balance,
                     claim.merkleProof
                 ),
@@ -98,7 +99,7 @@ contract MerkleOrchard is IDistributor, Ownable {
                 kind: kind
             });
 
-            claimed[claim.rewardToken][claim.rewarder][claim.distribution][liquidityProvider] = true;
+            claimed[claim.rewardToken][claim.rewarder][claim.distributionNonce][liquidityProvider] = true;
 
             suppliedBalance[claim.rewardToken][claim.rewarder] =
                 suppliedBalance[claim.rewardToken][claim.rewarder] -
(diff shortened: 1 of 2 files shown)


TomAFrench:
> I think we should leave in the `merkleRoots` and `claimStatus` functions though as that will make it easier to adapt the frontend

The issue is that with non-continuous values for `distributionNonce`, these functions will stop working whenever a second rewarder publishes a merkle root. We could pass arrays of nonces but I think as multicall is an industry standard we shouldn't feel the need to work around it.

nventuro:
If we do make distributions increase by one (which I think is a good idea), we should do it per distribution channel, not globally. See #883

gtaschuk:
I'm actually in favor of leaving these as user supplied nonces.

Having set up merkle distributions for a few teams and initiated a number of distributions, I think it's useful to allow people to explicitly supply nonces that have real-world meaning (ie starting at week 27 of rewards) and as a safeguard against a transaction being submitted twice and creating duplicate distributions in error.

I think that it's reasonable to leave it up to the admin to ensure continuous ranges, especially since there's nothing that prevents a user from claiming if they don't

TomAFrench:
> If we do make distributions increase by one (which I think is a good idea), we should do it per distribution channel, not globally. See #883

Agreed, having these per channel is a requirement for bitmaps (unless we're using a global one which runs the risks of effectiveness being diluted by many distributions with non-overlapping recipients)

TomAFrench:
> Having set up merkle distributions for a few teams and initiated a number of distributions, I think it's useful to allow people to explicitly supply nonces that have real-world meaning (ie starting at week 27 of rewards) and as a safeguard against a transaction being submitted twice and creating duplicate distributions in error.

Having a user-provided nonce as an safeguard against accidentally repeating a distribution is a good point, however we can have this while still enforcing incrementing nonces by requiring that the provided nonce is the expected value.

> I think that it's reasonable to leave it up to the admin to ensure continuous ranges, especially since there's nothing that prevents a user from claiming if they don't.

Giving rewarders this flexibility creates a few headaches for a future "fully featured" `MerkleOrchard` frontend for not a ton of benefits (allowing an initial offset to create a meaningful nonce value).

I'm imagining something where given a rewarder+token you can see all the distributions in the channel (allowing you to claim a distribution or upload a merkle tree to be verified against the root and then stored for other users). With fixed nonces we can easily query how many distributions exist and then display them, while if we allow arbitrary nonces we'd need to use a subgraph-like service or there's always the chance that there's another distribution with a nonce that's higher than we've checked.

Even if we knew a upper bound on the last nonce we'd then have to perform filtering as any of the previous nonces could be an empty tree (granted there's no guarantees that any of the merkle hashes are valid trees).

> I think we should leave in the merkleRoots and claimStatus functions though as that will make it easier to adapt the frontend

We could emulate these functions in JS with a multicall very easily so I don't think that we need to worry about adapting the frontend in the short term.

gtaschuk:
> I'm imagining something where given a rewarder+token you can see all the distributions in the channel (allowing you to claim a distribution or upload a merkle tree to be verified against the root and then stored for other users). With fixed nonces we can easily query how many distributions exist and then display them, while if we allow arbitrary nonces we'd need to use a subgraph-like service or there's always the chance that there's another distribution with a nonce that's higher than we've checked.

Not to be pedantic but I think the use of `nonce` here is the source of our difference in perspective.  I think of this value as an id for the distribution that can only be used once (to ensure uniqueness).  An alternate interpretation is more as an incrementing value that identifies ordering in a channel - I don't think the order is particularly important to the distribution channel for the purposes of this contracts use even though claims will often be claimed in order.

If the goal is to query from the frontend, I propose we instead add merkle roots, and distribution (ids) to the `DistributionAdded` event and either omit `merkleRoots` and `claimStatus` or have them accept an array of distribution ids

nventuro:
>Having a user-provided nonce as an safeguard against accidentally repeating a distribution is a good point, however we can have this while still enforcing incrementing nonces by requiring that the provided nonce is the expected value.

Agreed, creating distributions should be idempotent.

I don't have strong feelings re. what the nonce should be, but lean towards @TomAFrench's take: having them be sequential is much more straightforward and makes many components that much simpler by avoiding the question 'what does the nonce mean'. 

> I think it's useful to allow people to explicitly supply nonces that have real-world meaning (ie starting at week 27 of rewards) 

Arbitrary nonces are arguably 'incorrect' usage, as non sequentional ids are strictly inferior in a core metric (claim gas cost), might makes e.g. timestamp-based ids seem to provide zero (or negative) benefit.

gtaschuk:
> Arbitrary nonces are arguably 'incorrect' usage, as non sequentional ids are strictly inferior in a core metric (claim gas cost), might makes e.g. timestamp-based ids seem to provide zero (or negative) benefit.

non-sequential ids do not increase claim gas cost.  It is only useful for them to be sequential for the purpose of the query functions, which this PR currently removes

I don't feel so strongly, if someone wants to update this PR to apply to the new channel structure, I won't object, I just find it unnecessary and a little less flexible

TomAFrench:
> non-sequential ids do not increase claim gas cost. 

This is not true when distribution channels use bitmaps to reduce gas costs. Non sequential IDs result in positions in the bitmap being wasted (the extreme example being a timestamp based id as unless distributions are with 256 seconds of each other we only use a single bit per slot.)

gtaschuk:
> This is not true when distribution channels use bitmaps to reduce gas costs. Non sequential IDs result in positions in the bitmap being wasted (the extreme example being a timestamp based id as unless distributions are with 256 seconds of each other we only use a single bit per slot.)

Right I guess I more specifically should have said that although sparse nonces waste spaces in the bitmap it doesn't matter the order that the first 256 come in

TomAFrench:
Yep, I think we're all in agreement that we don't really care which order the nonces are used. The main drives for me are ensuring dense usage of bitmaps and simplifications on the frontend.

PR #830, opened 2021-09-10
Correct error message

EndymionJkb:
Noticed this when updating docs - there is no test for either of the "didn't converge" messages (probably very hard to trigger).


dmf7z:
Not really that hard to trigger, just need some crazy balances. I added a test.

nventuro (approved):

PR #831, opened 2021-09-10
Save storage pointers to save gas

nventuro:
Based on the comments in https://github.com/balancer-labs/balancer-v2-monorepo/pull/805.

This saves storage pointers to avoid recomputing them (which involves hash and memory-intensive operations), saving gas (though I haven't measured the effect).

There's multiple places where this could be applied: I only performed the change in the most critical functions (those that are called during `stake` or `unstake`) to try to keep the diff small, but am not against doing it accross the entire contract.

### DIFF
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -429,24 +429,19 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
 
         getVault().manageUserBalance(ops);
 
-        if (block.timestamp >= rewardData[pool][rewarder][rewardsToken].periodFinish) {
-            rewardData[pool][rewarder][rewardsToken].rewardRate = Math.divDown(
-                reward,
-                rewardData[pool][rewarder][rewardsToken].rewardsDuration
-            );
+        // Save the storage pointer to compute the slot only once.
+        Reward storage data = rewardData[pool][rewarder][rewardsToken];
+
+        if (block.timestamp >= data.periodFinish) {
+            data.rewardRate = Math.divDown(reward, data.rewardsDuration);
         } else {
-            uint256 remaining = rewardData[pool][rewarder][rewardsToken].periodFinish.sub(block.timestamp);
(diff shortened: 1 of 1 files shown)


gtaschuk (approved):

gtaschuk:
approving pending fixing failing tests

nventuro:
Tests shouldn't have failed - the PR only contained contract changes, so something clearly broke :sweat_smile: Turns out I had accidentally removed a line.

Should be fixed now.

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -478,19 +477,28 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
...
+        // Save the storage pointer to compute the slot only once.
+        EnumerableSet.AddressSet storage rewarders = _rewarders[pool][token];
+        uint256 rewardersLength = rewarders.length();
+
+        for (uint256 r; r < rewardersLength; r++) {
+            address rewarder = rewarders.unchecked_at(r);
+            Reward storage data = rewardData[pool][rewarder][token];
+
+            // Cache storage variables to avoid repeated access.
+            uint256 perToken = rewardPerToken(pool, rewarder, token);
```
TomAFrench, line 490: We're only using `pool`, `rewarder` and `token` here to recompute `rewardData[pool][rewarder][token]`. We could make an internal `_rewardPerToken(Reward storage data)` to avoid this.

Similarly in `lastTimeRewardApplicable`.

TomAFrench: We're also just using a one or two fields from `data` and not writing anything back to storage in those functions so could just pass those fields in directly.

TomAFrench: Other than this small possible improvement I'm happy with it.

nventuro: Oh this is very good, I didn't notice those functions accessed the same data structure. Will include this.

TomAFrench (approved):
LGTM

PR #832, opened 2021-09-13
Phantom stable pool MVP

facuspagnuolo:
This PR provides an initial implementation of the "Phantom Stable Pool". 

The idea was to mix a few components from the Linear and the Stable pool:
- Stable pool math
- Phantom BPT strategy
- Join/Exit as swaps
- Cache rate providers


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -0,0 +1,466 @@
...
+        uint256 bufferPeriodDuration;
+        address owner;
+    }
+
+    constructor(NewPoolParams memory params)
+        StablePool(
+            params.vault,
+            params.name,
+            params.symbol,
+            _insertSorted(params.tokens, IERC20(this)),
```
dmf7z, line 75: Small note. Stable pool can handle 5 tokens + bpt because bpt are not part of invariant math. We are limiting it to 4 tokens + bpt. I think it is ok for our current use of this pool.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -0,0 +1,466 @@
...
+        }
+
+        uint256 bptIndex;
+        for (bptIndex = params.tokens.length; bptIndex > 0 && params.tokens[bptIndex - 1] > IERC20(this); bptIndex--) {
+            // solhint-disable-previous-line no-empty-blocks
+        }
+        _bptIndex = bptIndex;
+    }
+
+    function initialize() external {
```
dmf7z, line 104: Can you add a note that it is important that this function is called right after creation of the pool and that we recommend to use the factory. In case someone wants to deploy this pool without the factory.

dmf7z (approved):
good work!

PR #834, opened 2021-09-13
Fix conversion of WETH to ETH in sortTokens

TomAFrench:
All instances of WETH were being converted to ETH, rather than just when the user provided ETH as one of the inputs.


nventuro (approved):

nventuro:
I found it a bit odd that the docs for `sortTokens` begin with 'Converts an asset to the equivalent ERC20 address'

nventuro:
Woops closed by mistake

TomAFrench:
One test isn't passing but it's not related to this code and it passes locally so I'm going to merge.

PR #836, opened 2021-09-13
deploying BAL distribution contract on Arbitrum

gtaschuk:
We have opted to use a merkle redeem contract for liquidity mining rewards on arbitrum.  This deploys the merkle redeem contract for BAL rewards and will be run as
```
npx hardhat deploy --id 20210913-bal-arbitrum-merkle --key <KEY> --network arbitrum
```


rabmarut (approved):
Building pkg/distributors out of commit 206ce3a yields a build-info file called 521ef37390e7374683fe8f66688ff735.json which matches build-info/MerkleRedeem.json as verified by a diff.

PR #838, opened 2021-09-14
fix entrypoint to @balancer-labs/typechain package

TomAFrench:
`dist/index.js` rather than `dist/index.ts`


nventuro (approved):
Great catch :sweat_smile:

PR #839, opened 2021-09-14
Changelog and package update

nventuro:
I created the `stable` branch, which we can now direct people to when looking for the latest stable version of contract code. In it, I added changelogs, updated package versions, and re-published everything on npm.

By merging `stable` into `master`, we'll have these added changelogs, which we should populate as work is done. Eventually, `stable` will be updated with the latest commits from `master`, and new package versions published.


nventuro:
This is being successfully used in the template repo, merging.

PR #840, opened 2021-09-15
feat: add function to get a contract's artifact from its task id + name

TomAFrench:
Thoughts on adding a function which allows us to pull out the contract artifacts using their name + task id? Main usecase being allowing people to deploy the exact bytecode in tests, etc. without having to integrate tasks into their workflow / extract artifacts from the build-info themselves.

I've implemented this by using the `Task` class to avoid duplication. Although the other functions have avoided using `Tasks`, is there a reason I'm missing here?


nventuro (approved):
We ended up going with a different approach altogether to both not expose `Task` and its dependencies, but also not have to publish the full build infos (which are currently ~150MB).

PR #841, opened 2021-09-15
Distributors: Store bitmap claims with an implicit index from the merkle proof

gtaschuk:
Iterates on top of the awesome idea in https://github.com/balancer-labs/balancer-v2-monorepo/pull/828 (to realize gas savings by tightly packing claim status) but computes the implicit index of the element in the (balanced) merkle tree as part of verifying the merkle proof (based on if you go right or left in each level up from the root)

I favor this approach because it means we don't have to add indexes to the merkle tree, which makes it less error-prone (and is a smaller diff).  It also uses less gas based on the numbers in the PR above vs:

```
➜  benchmarks git:(merkle-orchard-bitmap-claims-implicit-index) ✗ yarn measure-merkle-claim

# Claiming 1 distributions

1 claims: 63.2k (63.2k per claim)

## Using Internal Balance
1 claims: 54.1k (54.1k per claim)

# Claiming 2 distributions

2 claims: 60.9k (30.4k per claim)

## Using Internal Balance
2 claims: 51.8k (25.9k per claim)

# Claiming 3 distributions

3 claims: 75.7k (25.2k per claim)

## Using Internal Balance
3 claims: 66.7k (22.2k per claim)

# Claiming 4 distributions

4 claims: 90.6k (22.6k per claim)

## Using Internal Balance
4 claims: 81.5k (20.4k per claim)

# Claiming 5 distributions

5 claims: 105.4k (21.1k per claim)

## Using Internal Balance
5 claims: 96.3k (19.3k per claim)
```

### DIFF
--- a/pkg/solidity-utils/contracts/openzeppelin/MerkleProof.sol
+++ b/pkg/solidity-utils/contracts/openzeppelin/MerkleProof.sol
@@ -12,7 +12,7 @@ library MerkleProof {
      * sibling hashes on the branch from the leaf to the root of the tree. Each
      * pair of leaves and each pair of pre-images are assumed to be sorted.
      */
-    function verify(bytes32[] memory proof, bytes32 root, bytes32 leaf) internal pure returns (bool) {
+    function verify(bytes32[] memory proof, bytes32 root, bytes32 leaf) internal pure returns (bool, uint index) {
         bytes32 computedHash = leaf;
 
         for (uint256 i = 0; i < proof.length; i++) {
@@ -21,6 +21,7 @@ library MerkleProof {
             if (computedHash <= proofElement) {
                 // Hash(current computed hash + current element of the proof)
                 computedHash = keccak256(abi.encodePacked(computedHash, proofElement));
+                index += 2**i;
             } else {
                 // Hash(current element of the proof + current computed hash)
                 computedHash = keccak256(abi.encodePacked(proofElement, computedHash));
@@ -28,6 +29,6 @@ library MerkleProof {
         }
 
         // Check if the computed hash (root) is equal to the provided root
-        return computedHash == root;
+        return (computedHash == root, index);
     }
 }
--- a/pkg/distributors/contracts/MerkleOrchard.sol
+++ b/pkg/distributors/contracts/MerkleOrchard.sol
@@ -1,0 +1,220 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma experimental ABIEncoderV2;
+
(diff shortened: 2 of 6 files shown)


gtaschuk:
The gas usage in the original implementation's PR is measured differently than the numbers in this PR's description. Corresponding measurements (this PRs implementation is only slightly better for gas):

```
➜  benchmarks git:(merkle-orchard-bitmap-claims) ✗ cd ../../pkg/distributors && yarn compile && popd && yarn measure-merkle-claim
Compiling 51 files with 0.7.1
Compilation finished successfully
~/core/pvt/benchmarks ~/core ~ ~/erc20-redeemable

# Claiming 1 distributions

1 claims: 63.3k (63.3k per claim)

## Using Internal Balance
1 claims: 54.3k (54.3k per claim)

# Claiming 2 distributions

2 claims: 61.2k (30.6k per claim)

## Using Internal Balance
2 claims: 52.2k (26.1k per claim)

# Claiming 3 distributions

3 claims: 76.2k (25.4k per claim)

## Using Internal Balance
3 claims: 67.2k (22.4k per claim)

# Claiming 4 distributions

4 claims: 91.2k (22.8k per claim)

## Using Internal Balance
4 claims: 82.1k (20.5k per claim)

# Claiming 5 distributions

5 claims: 106.2k (21.2k per claim)

## Using Internal Balance
5 claims: 97.1k (19.4k per claim)
```

pkg/solidity-utils/contracts/openzeppelin/MerkleProof.sol
```diff
@@ -21,13 +21,14 @@ library MerkleProof {
             if (computedHash <= proofElement) {
                 // Hash(current computed hash + current element of the proof)
                 computedHash = keccak256(abi.encodePacked(computedHash, proofElement));
+                index += 2**i;
```
TomAFrench, line 24: I think this breaks in the situation where there's any sorting of hashes in the upper layers. I'm pretty confident that if you tried this for larger trees then it would start marking incorrect indices as claimed in the bitmap as we'd need the hashes in each layer to coincidentally be in ascending order.

TomAFrench (commented):
I don't think this method of calculating the index works for larger tree sizes and if we started using trees with more than 2 layers then we'd start seeing incorrect claim indices.

pkg/solidity-utils/contracts/openzeppelin/MerkleProof.sol
TomAFrench: As discussed, this does break the idea of an "index" as used in #828 however it's unique per merkle leave so is acceptable to use as a method to determine whether a particular user has claimed their distribution.

TomAFrench: This does have downsides for gas in comparison to explicit indices.

>One downside I see is that it will make claims with a number of participants which is just a bit over a power of 2 less efficient.
Say we have 1025 people in the tree, to balance the tree we need to pad it to 2048 which needs a bitmap 8 slots long. With explicit indices we'll fill up the first 4 slots and then the last person spills over into the next slot. With implicit indices users get placed as a near random position in those 8 slots so we'll end up initialising all of them.

That said, at worst each slot will be half filled on average so there's still significant gains.

gtaschuk: Right - this is not an index per se but an identifier that spans up to ```2**(Math.ceiling(Math.log2(elements)))``` which could be at worst half full

Both are wayyy better than storing claim booleans to the point where the difference between true indices vs identifiers becomes less significant compared to the other gas costs

I find this solution elegant despite the inefficiency of storing identifiers vs indexes.  I think we should do some benchmarking to put numbers to our assumptions.

nventuro:

PR #842, opened 2021-09-16
Alternative orchard claim storage

nventuro:
After seeing all the fun everyone's been having on #828 and #841, I wanted in on the fun.

image

---

I challenge the notion that 'storage efficiency' is a useful metric and something to try to optimize for. Or rather, I challenge the notion that having 'half full storage bitmaps' is inherently a bad thing. I embrace this inefficiency: each LP gets their own bitmap for all token distributions: week 0 is bit 0, week 1 is bit 1, and so on. After 256 weeks (~5 years), a new storage slot is used.

The result of this simple change is that an LP claiming multiple weeks (whitin the same 5 year window) **reads from and writes to a single storage slot**. That's it. No need for indices, balanced trees, nothing.

---

This PR is intentionally simple and clear-cut to avoid introducing noise: many of the optimizations present in #828 and #841 can also be applied here. Indeed, _new_ optimizations can also be applied. Due to [EIP 2929](https://eips.ethereum.org/EIPS/eip-2929), repeatead storage accesses are cheaper, which is what we see here. But they are not zero: we can cut these down by reading from storage only once, setting the different bits in memory, and then doing a single storage write with the final bitmap. This would require for the claimer to send the claims sorted in a particular order, but I don't think that's too much to ask.

---

### Benchmark

Unmodified, including all token transfers, etc.

```

1 claims: 62.5k (62.5k per claim)
2 claims: 71.9k (36.0k per claim)
3 claims: 81.3k (27.1k per claim)
4 claims: 90.7k (22.7k per claim)
5 claims: 100.2k (20.0k per claim)

## Using Internal Balance

1 claims: 53.5k (53.5k per claim)
2 claims: 62.9k (31.4k per claim)
3 claims: 72.3k (24.1k per claim)
4 claims: 81.7k (20.4k per claim)
5 claims: 91.1k (18.2k per claim)
```

### DIFF
--- a/pkg/distributors/contracts/MerkleOrchard.sol
+++ b/pkg/distributors/contracts/MerkleOrchard.sol
@@ -35,8 +35,8 @@ contract MerkleOrchard is IDistributor, Ownable {
     // Recorded distributions
     // rewardToken > rewarder > distribution > root
     mapping(IERC20 => mapping(address => mapping(uint256 => bytes32))) public trees;
-    // rewardToken > rewarder distribution > lp > root
-    mapping(IERC20 => mapping(address => mapping(uint256 => mapping(address => bool)))) public claimed;
+    // rewardToken > rewarder > lp > distribution / 256 -> bitmap
+    mapping(IERC20 => mapping(address => mapping(address => mapping(uint256 => uint256)))) public claimedBitmap;
     // rewardToken > rewarder > balance
     mapping(IERC20 => mapping(address => uint256)) public suppliedBalance;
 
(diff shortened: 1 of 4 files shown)


gtaschuk:
Posted some thoughts on usage/tradeoffs on #843

I think this is well suited to our single rewarder/reward token `MerkleRedeem.sol`.  If we think we're deploying any of those anytime soon we should implement this for that and save everyone a lot of gas.

nventuro:
This was only meant as a proof of concept - closing to be properly implemented as per #885

PR #843, opened 2021-09-17
Yet Another Orchard Claim Storage Strategy

gtaschuk:
I reaffirm the notion (refuted in #842) that storage efficiency is a useful metric and extend my efforts in #841 but slightly more efficiently.

I don't think the typical usage of this contract will be to claim multiple distributions from the same reward token + rewarder as optimized for in #842 - that should be the role of the staking contract.

The role of the orchard should be airdrops, and more frequent claims across a number of different reward tokens/rewarder pairs, and possibly fro a number of different rewardees.

**Approach**
This breaks the tree into max 256 leaf (max 8 level) branches, and stores a bitmap for each branch - maximally efficient storage for the claim statuses of a balanced tree.

**Gas Costs**
Per User approach - 20k initial claim(s), ~5k for unlimited weeks of claims thereafter.  This is great if you are claiming 4+ weeks.  Propose if we go that route though, that we instead store on a per-rewarder basis, as it seems likely a single rewarder will send many rewards of different kinds to many different users
Per Distribution approach - 1/256 chance you pay 20k, 255/256 chance of 5k per claim (depending on the size of the tree) - so an average of ~5k per claim

```
➜  benchmarks git:(merkle-orchard-branch-index) ✗ yarn measure-merkle-claim          

# Claiming 1 distributions

1 claims: 63.2k (63.2k per claim)

## Using Internal Balance
1 claims: 54.2k (54.2k per claim)

# Claiming 2 distributions

2 claims: 61.0k (30.5k per claim)

## Using Internal Balance
2 claims: 52.0k (26.0k per claim)

# Claiming 3 distributions

3 claims: 75.9k (25.3k per claim)

## Using Internal Balance
3 claims: 66.9k (22.3k per claim)

# Claiming 4 distributions

4 claims: 90.8k (22.7k per claim)

## Using Internal Balance
4 claims: 81.8k (20.5k per claim)

# Claiming 5 distributions

5 claims: 105.7k (21.1k per claim)

## Using Internal Balance
5 claims: 96.7k (19.3k per claim)
```
Note: Branches, and thus `bytes32 branch` root hashes of the branches are unique for a tree with unique node entries

### DIFF
--- a/pkg/distributors/contracts/MerkleOrchard.sol
+++ b/pkg/distributors/contracts/MerkleOrchard.sol
@@ -36,7 +36,7 @@ contract MerkleOrchard is IDistributor, Ownable {
     // keccak(rewardToken|rewarder|distribution) > root
     mapping(bytes32 => bytes32) public trees;
     // keccak(rewardToken|rewarder|distribution) > bitmap
-    // bitmap is made up of a mapping(uint256 => uint256) to allow more than 256 recipients
+    // bitmap is made up of a mapping(bytes32 => uint256) to allow more than 256 recipients
     mapping(bytes32 => mapping(bytes32 => uint256)) public claimedBitMap;
     // rewardToken > rewarder > balance
     mapping(IERC20 => mapping(address => uint256)) public suppliedBalance;
--- a/pkg/distributors/contracts/MerkleOrchard.sol
+++ b/pkg/distributors/contracts/MerkleOrchard.sol
@@ -37,7 +37,7 @@ contract MerkleOrchard is IDistributor, Ownable {
     mapping(bytes32 => bytes32) public trees;
     // keccak(rewardToken|rewarder|distribution) > bitmap
     // bitmap is made up of a mapping(uint256 => uint256) to allow more than 256 recipients
-    mapping(bytes32 => mapping(uint256 => uint256)) public claimedBitMap;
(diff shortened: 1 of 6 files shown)


TomAFrench:
Can you go into more detail on the differences between this and #828? From an initial glance it seems like this is trying to address the issue of indices of "filled" leaves being distributed over the full set of bitmaps by enforcing that they leaves which are adjacent to each other in the bottom layer are represented as adjacent bits in the bitmaps?

gtaschuk:
> Can you go into more detail on the differences between this and #828? From an initial glance it seems like this is trying to address the issue of indices of "filled" leaves being distributed over the full set of bitmaps by enforcing that they leaves which are adjacent to each other in the bottom layer are represented as adjacent bits in the bitmaps?

Exactly. #841 calculates an index from the merkle leaf to the root, and then used used that index%256 in words of claim booleans.  In that, each word index is [roof -> leaf] >> 8.  This instead groups leaves by the intermediate hash - the "branch" - that they have in common which is at most 8 levels down from the leaf meaning (I think!) that words would be more full.  Actually, now that I think about it, perhaps we need to be more careful picking which hashes are used for the words for branches with mixed levels -.ie in a max 10 level tree, we might want to use the hashes 2 from the bottom rather than 8 from the top to properly account for mixed-level branches... will think about this over the weekend

nventuro:

PR #846, opened 2021-09-17
Distributors: Stake with permit (OZ-M01)

gtaschuk:
Fixes behavior of stake with permit to allow a user to stake tokens on another users behalf

As noted here https://github.com/balancer-labs/balancer-v2-monorepo/pull/801#discussion_r701839593 `msg.sender` must stake the user with the LP as the recipient to avoid mempool attacks

### DIFF
--- a/pkg/distributors/contracts/MultiRewards.sol
+++ b/pkg/distributors/contracts/MultiRewards.sol
@@ -277,6 +277,7 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
         bytes32 r,
         bytes32 s
     ) external nonReentrant {
+        require(account == recipient, "Cannot stake an accounts bpt to an alternate address");
         IERC20Permit(address(pool)).permit(account, address(this), amount, deadline, v, r, s);
         _stakeFor(pool, amount, account, recipient);
     }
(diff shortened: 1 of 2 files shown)


pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -262,13 +271,15 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
...
+        address account,
         address recipient,
         uint8 v,
         bytes32 r,
         bytes32 s
-    ) external {
-        IERC20Permit(address(pool)).permit(msg.sender, address(this), amount, deadline, v, r, s);
-        stakeFor(pool, amount, recipient);
+    ) external nonReentrant {
+        require(account == recipient, "Cannot stake an accounts bpt to an alternate address");
```
nventuro, line 280:
```suggestion
        // Force the  sender to the be the recipient to ensure the signature can only be used in a single way
        require(account == recipient, "Sender and recipient must equal for permit staking");
```

The original error message was not accurante and might be misleading, since it _is_ possible to stake to an alternate address - the restriction only shows up when permit staking.

nventuro: On a different note, we should also later review the fact that `amount` also has to be the same for both. This e.g. means that it is impossible to use `stakeWithPermit` while granting infinite allowance.

I imagine such a use case can be covered in some other contract by first doing a permit and then staking, but it still requires some analysis.

nventuro (approved):

PR #847, opened 2021-09-17
Distributors - Multirewards events (OZ-M02)

gtaschuk:
Added event for rewarder allowlisting, and relevant details to existi…ng events [OZ-M02]

### DIFF
--- a/pkg/distributors/contracts/MerkleRedeem.sol
+++ b/pkg/distributors/contracts/MerkleRedeem.sol
@@ -40,6 +40,8 @@ contract MerkleRedeem is IDistributor, Ownable {
 
     IVault public immutable vault;
 
+    event RewardAdded(address indexed token, uint256 amount);
+
     constructor(IVault _vault, IERC20 _rewardToken) {
         vault = _vault;
         rewardToken = _rewardToken;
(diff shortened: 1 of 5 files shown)


pkg/distributors/contracts/MerkleRedeem.sol
```diff
@@ -40,6 +40,8 @@ contract MerkleRedeem is IDistributor, Ownable {
 
     IVault public immutable vault;
 
+    event RewardAdded(address indexed token, uint256 amount);
```
nventuro, line 43: It seems this new event is not emitted anywhere?

gtaschuk: This event was moved from `IDistributor` to `MultiRewards` because the addition of the `token` parameter doesn't apply to MerkleRedeem and is used at the end of `notifyRewardAmount`

The parameter would likely be a good addition to the orchard's events, and even the 1-token `MerkleRedeem` so it might find itself back into the `IDistributor` but for now I left `MerkleRedeem.sol` as-is.

nventuro: Ahh I see, thanks

nventuro (approved):

PR #848, opened 2021-09-20
[OZ-M02] Remove start amp events on amp stop

nventuro:
From the OZ preliminary report:

>Owing to the fact that one version of _setAmplificationData calls the other version of _setAmplificationData, both events will be emitted in a single transaction when the constructor of StablePool calls _setAmplificationData. Additionally, if startAmplificationParameterUpdate is called followed by a call to stopAmplificationParameterUpdate, two instances of AmpUpdateStarted will be emitted followed by one instance of AmpUpdateStopped.

This makes it so stopping no longer emits a 'start' event.

### DIFF
--- a/pkg/pool-stable/contracts/StablePool.sol
+++ b/pkg/pool-stable/contracts/StablePool.sol
@@ -726,12 +726,21 @@ contract StablePool is BaseGeneralPool, BaseMinimalSwapInfoPool, StableMath, IRa
     }
 
     function _setAmplificationData(uint256 value) private {
-        _setAmplificationData(value, value, block.timestamp, block.timestamp);
-
+        _storeAmplificationData(value, value, block.timestamp, block.timestamp);
         emit AmpUpdateStopped(value);
     }
 
     function _setAmplificationData(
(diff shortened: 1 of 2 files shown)


pkg/pool-stable/contracts/StablePool.sol
```diff
@@ -726,8 +726,7 @@ contract StablePool is BaseGeneralPool, BaseMinimalSwapInfoPool, StableMath, IRa
     }
 
     function _setAmplificationData(uint256 value) private {
-        _setAmplificationData(value, value, block.timestamp, block.timestamp);
-
+        _storeAmplificationData(value, value, block.timestamp, block.timestamp);
         emit AmpUpdateStopped(value);
```
dmf7z, line 730: Not a big deal, but with this change we are emitting only `AmpUpdateStopped` when setting amp data and no start event.

nventuro: Yes, I think that is perfectly fine. This was noted in the audit report.

PR #851, opened 2021-09-20
Distributors: added check to setRewardsDuration that a reward exists OZ-H01

gtaschuk:
Prevents a rewarder from setting a reward duration for a reward before they add it, and thus lock themselves out of adding the reward

### DIFF
--- a/pkg/distributors/test/MultiRewards.test.ts
+++ b/pkg/distributors/test/MultiRewards.test.ts
@@ -96,6 +96,12 @@ describe('Staking contract', () => {
     ).to.be.revertedWith('Reward must be configured with addReward');
   });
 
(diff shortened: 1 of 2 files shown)


pkg/distributors/test/MultiRewards.test.ts
```diff
@@ -88,6 +88,14 @@ describe('Staking contract', () => {
     });
   });
 
+  it('reverts if a rewarder attempts to setRewardsDuration before adding a reward', async () => {
```
nventuro, line 91: Can we also add a test for the newly added allowlist modifier?

gtaschuk: Done - test failure appears to be a flicker on the pools package (unrelated to this test case being added here)

PR #856, opened 2021-09-21
Phantom: Filter BPT balance when calling stable math lib

facuspagnuolo:


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -469,4 +471,15 @@ contract StablePhantomPool is StablePool {
     function _getMinimumBpt() internal pure override returns (uint256) {
         return _MINIMUM_BPT;
     }
+
+    function _skipBptIndex(uint256 index) internal view returns (uint256) {
+        return index < _bptIndex ? index : index + 1;
```
dmf7z, line 476: When `_skipBptIndex`  is used by `_onSwapGivenIn` and `_onSwapGivenOut`,  shouldn't it be `index - 1`?

facuspagnuolo: If we use `index - 1`, we would be repeating a past value, I think we should skip the current value and replace it by the following ones

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -463,10 +470,14 @@ contract StablePhantomPool is StablePool {
...
-    function _getMinimumBpt() internal pure override returns (uint256) {
-        return _MINIMUM_BPT;
+    function _skipBptIndex(uint256 index) internal view returns (uint256) {
+        return index < _bptIndex ? index : index - 1;
+    }
+
+    function _dropBptItem(uint256[] memory _balances) internal view returns (uint256[] memory balances) {
+        balances = new uint256[](_balances.length - 1);
+        for (uint256 i = 0; i < balances.length; i++) {
+            balances[i] = _balances[_skipBptIndex(i)];
```
dmf7z, line 480: For the `_dropBptItem ` it was ok to have `index + 1` :) 

Maybe because it is used only here you can do it inline:
`balances[i] = _balances[i < _bptIndex ? i : i + 1];`

facuspagnuolo: I forgot to push that last part, see my last commit please :)

dmf7z (approved):

PR #858, opened 2021-09-21
Phantom: Support proportional initialize

facuspagnuolo:


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -244,20 +227,39 @@ contract StablePhantomPool is StablePool {
...
+
+        (uint256 currentAmp, ) = _getAmplificationParameter();
+        uint256 invariantAfterJoin = StableMath._calculateInvariant(currentAmp, _dropBptItem(amountsIn), true);
+
+        // Set the initial BPT to the value of the invariant
+        uint256 bptAmountOut = invariantAfterJoin;
+        _updateLastInvariant(invariantAfterJoin, currentAmp);
+
+        // Mint the total amount of BPT to the sender forcing the Vault to pull it
+        uint256 initialBpt = _MAX_TOKEN_BALANCE.sub(bptAmountOut);
```
dmf7z, line 258: Should we have to subtract the minimum bpt also? Because it is going to be minted after calling `_onInitializePool` in the `onJoinPool` hook.

dmf7z: Nevermind, I just checked and it has been done in another PR.

dmf7z (approved):

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
facuspagnuolo: Exactly

PR #860, opened 2021-09-23
Deployments: InvestmentPoolFactory on production networks

EndymionJkb:


nventuro (approved):
Verified all deployed contracts, and published v2.3 with the changes.

I also added the bytecode for the two most recent merkle deployments, which was missing.

PR #861, opened 2021-09-23
Linear pool fix rate math

dmf7z:
Fix wrong use of rate in linear math. In depth tests for this will be added in future PRs. 
This PR also simplifies `_calcWrappedOutPerMainIn` and `_calcWrappedInPerMainOut` math functions

### DIFF
--- a/pkg/pool-linear/contracts/LinearMath.sol
+++ b/pkg/pool-linear/contracts/LinearMath.sol
@@ -68,7 +68,6 @@ contract LinearMath {
     function _calcWrappedOutPerMainIn(
         uint256 mainIn,
         uint256 mainBalance,
-        uint256 wrappedBalance,
         Params memory params
     ) internal pure returns (uint256) {
         // Amount out, so we round down overall.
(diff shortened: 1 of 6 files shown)


facuspagnuolo (approved):
Nice catch!

pkg/pool-linear/contracts/LinearMath.sol
```diff
@@ -68,31 +68,27 @@ contract LinearMath {
...
         Params memory params
     ) internal pure returns (uint256) {
         // Amount in, so we round up overall.
 
         uint256 previousNominalMain = _toNominal(mainBalance, params);
         uint256 afterNominalMain = _toNominal(mainBalance.sub(mainOut), params);
         uint256 deltaNominalMain = previousNominalMain.sub(afterNominalMain);
-        uint256 newWrappedBalance = wrappedBalance.add(deltaNominalMain.mulUp(params.rate));
-        return newWrappedBalance.sub(wrappedBalance);
+        return deltaNominalMain.divDown(params.rate);
```
sergioyuhjtman, line 91: why divDown and not divUp?

dmf7z: true, we are rounding up. Fixing it now. Thx.

PR #863, opened 2021-09-23
Make WeightedMath and StableMath libraries

nventuro:
This is a bit overdue - these were contracts as at some point they held some state, but there's no point now, and I believe we all think of them as libraries already anyway.


dmf7z (approved):
👍

PR #868, opened 2021-09-24
Grant infinite BPT allowance to the Vault

nventuro:
Doing this is fine as the Vault already has authorization mechanisms in place to validate `sender`. In fact, `BasePool` already relies on this fact when in performs `_burnPoolTokens` without explicit `sender` approval.

We haven't needed this feature as since burn BPT and it is never swapped, but as we start creating nested Pools this will become relevant.

Best reviewed commit by commit since the very first commit improves documentation a bit to make intent clearer.


dmf7z (approved):
great work!

PR #869, opened 2021-09-28
Distributors: added deploy task for MCB merkle distributor

gtaschuk:


rabmarut (approved):
Building pkg/distributors out of commit 206ce3a (same as MerkleRedeem contracts for BAL Arbitrum and VITA mainnet) yields a build-info file called 521ef37390e7374683fe8f66688ff735.json which matches build-info/MerkleRedeem.json as verified by a diff.

PR #872, opened 2021-09-29
Avoid double invariant computation of invariant on Phantom Pools

nventuro:
This leads to gas savings when protocol fees are enabled.


dmf7z (approved):
Great work nico, it reads much better

PR #894, opened 2021-10-02
MerkleOrchard: User-supplied token list

### LINKED ISSUE #888: [orchard] Avoid excessive memory usage
A user might claim a very large number of rewards at once, but we have no way of knowing beforehand how many tokens are involved in those claims. Unfortunately, we need to keep per-token data (namely, the withdrawn amounts) while we loop over the claims array, so we need to have allocated space for this.

A naive solution would be to allocate arrays as large as the number of claims, but this results in unnecessarily large arrays for the vast majority of use cases. An alternative solution could be to have the caller supply the array of tokens along with the claims, and provide token indices in this array instead of addresses, which would also let us access the right slot directly instead of having to perform O(n) lookup.

```solidity
struct Claim {
  uint256 tokenIndex;
  address distributor:
  ...
}

function claim(Claim[] memory claims, IERC20[] memory tokens...) { ... }
```

Note that this is the same pattern we follow [in the Vault](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/vault/contracts/interfaces/IVault.sol#L567).

gtaschuk:
Closes #888

Reduces gas by offloading token indexing to the client

(tagging: @evgenyboxer for visibility as this changes the contract interface slightly)

### DIFF
--- a/pkg/distributors/contracts/MerkleOrchard.sol
+++ b/pkg/distributors/contracts/MerkleOrchard.sol
@@ -52,7 +52,7 @@ contract MerkleOrchard is IDistributor, Ownable {
         uint256 distribution;
         uint256 balance;
         address rewarder;
-        IERC20 rewardToken;
+        uint256 tokenIndex;
         bytes32[] merkleProof;
     }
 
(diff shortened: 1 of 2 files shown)


pkg/distributors/contracts/MerkleOrchard.sol
```diff
@@ -52,32 +52,28 @@ contract MerkleOrchard is IDistributor, Ownable {
...
-        // We want to keep track of the number of unique tokens so we can aggregate transfers
-        uint256 numRewardTokens;
-        IERC20[] memory rewardTokens = new IERC20[](claims.length);
-        uint256[] memory rewardAmounts = new uint256[](claims.length);
+        uint256[] memory amounts = new uint256[](tokens.length);
 
         Claim memory claim;
         for (uint256 i = 0; i < claims.length; i++) {
             claim = claims[i];
+            IERC20 token = tokens[claim.tokenIndex];
```
TomAFrench, line 71:
```suggestion
        Claim memory claim;
        IERC20 token;
        for (uint256 i = 0; i < claims.length; i++) {
            claim = claims[i];
            token = tokens[claim.tokenIndex];
```

For consistency with `claim`. (Although sometimes I've found that doing this increases gas costs instead of decreasing them.)

TomAFrench (commented):
Looks good to me.

TomAFrench (approved):
Lgtm

PR #895, opened 2021-10-03
Orchard Renaming

gtaschuk:
Renaming to reflect that these contracts distribute tokens on behalf of distributors

closes https://github.com/balancer-labs/balancer-v2-monorepo/issues/886

PR #900, opened 2021-10-04
Merkle orchard renaming

TomAFrench:
Following on from #895, I've named `seedAllocations` -> `createDistribution` to be more in keeping with the rest of the terminology.

I've also removed the underscore from `_merkleRoot`


gtaschuk:
too many cooks in the kitchen!

Would you mind rebasing this onto or committing on top of the much larger change in #901 to avoid merge conflicts?

gtaschuk (approved):

Issue #890, opened 2021-10-01
[orchard] Optimize for reduced storage accesses


nventuro:
While https://github.com/balancer-labs/balancer-v2-monorepo/issues/885 will greatly reduce gas cost by making each distribution channel use its own storage slot (for the most part), we could still achieve larger savings by only reading to and from storage when it is actually required.

For example, if processing two distribution for the same channel, we could read the claims bitmap once, update in memory twice, and then write it back once. The cost of accessing dirty storage is reduced (100 gas), but non-zero, and there's extra overhead in decoding data, computing storage pointers, etc.

A simple way to achieve this would be to bundle adjacent claims, only committing the results to storage when a claim for a different distribution comes up. This would require for the client code to send claims sorted by distribution channel in order to get savings, but that should be fairly straightfoward to do (and would not be required either way).

PR #901, opened 2021-10-04
Orchard per user storage

gtaschuk:
Closes #895 and #890

This PR recognizes that the predominant use case of merkle redeem distribution is for periodic distributions of the same token

It therefore optimizes for gas by storing a users claim status in a way that minimizes storage writes

### DIFF
--- a/pkg/distributors/contracts/MerkleOrchard.sol
+++ b/pkg/distributors/contracts/MerkleOrchard.sol
@@ -259,7 +259,7 @@ contract MerkleOrchard {
         uint256 distribution,
         uint256 claimedBalance,
         bytes32[] memory merkleProof
-    ) public view returns (bool) {
+    ) external view returns (bool) {
         bytes32 channelId = _getChannelId(token, distributor);
         return _verifyClaim(channelId, liquidityProvider, distribution, claimedBalance, merkleProof);
     }
--- a/pkg/distributors/contracts/MerkleOrchard.sol
+++ b/pkg/distributors/contracts/MerkleOrchard.sol
@@ -30,12 +30,12 @@ contract MerkleOrchard {
     using SafeERC20 for IERC20;
 
     // Recorded distributions
-    // token > distributor > distribution > root
-    mapping(IERC20 => mapping(address => mapping(uint256 => bytes32))) public trees;
-    // token > distributor distribution > lp > root
-    mapping(IERC20 => mapping(address => mapping(uint256 => mapping(address => bool)))) public claimed;
-    // token > distributor > balance
-    mapping(IERC20 => mapping(address => uint256)) public suppliedBalance;
+    // channelId > distributor > distribution > root
+    mapping(bytes32 => mapping(uint256 => bytes32)) public trees;
+    // channelId > lp > distribution / 256 -> bitmap
+    mapping(bytes32 => mapping(address => mapping(uint256 => uint256))) public claimedBitmap;
+    // channelId > balance
+    mapping(bytes32 => uint256) public suppliedBalance;
 
     event DistributionAdded(address indexed token, uint256 amount);
     event DistributionPaid(address indexed user, address indexed token, uint256 amount);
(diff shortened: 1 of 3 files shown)


pkg/distributors/contracts/MerkleOrchard.sol
```diff
@@ -63,16 +63,57 @@ contract MerkleOrchard {
     ) internal {
         uint256[] memory amounts = new uint256[](tokens.length);
 
+        // To save gas when setting claimed statuses in storange we group updates
```
TomAFrench, line 66:
```suggestion
        // To save gas when setting claimed statuses in storage we group updates
```

pkg/distributors/contracts/MerkleOrchard.sol
```diff
@@ -63,16 +63,57 @@ contract MerkleOrchard {
...
         for (uint256 i = 0; i < claims.length; i++) {
             claim = claims[i];
-            token = tokens[claim.tokenIndex];
 
-            require(!isClaimed(token, claim.distributor, claim.distribution, liquidityProvider), "cannot claim twice");
+            // When we process a new claim we either
+            // a) aggregate the new claim bit with previous claims of the same channel/claim bitmap
+            // b) set claim status and start aggregating a new set of currentBits for a new channel/word
+            if (currentChannelId == bytes32(0)) {
+                currentChannelId = keccak256(abi.encodePacked(address(tokens[claim.tokenIndex]), claim.distributor));
```
TomAFrench, line 84: I think we need something like

```
function _getChannelId(IERC20 token, address distributor) private returns (bytes32) {
 return keccak256(abi.encodePacked(address(token, distributor))
}
```
As this is popping up a bunch of times and causing line breaks

pkg/distributors/contracts/MerkleOrchard.sol
```diff
@@ -63,16 +63,57 @@ contract MerkleOrchard {
...
+                currentBits = 1 << claim.distribution % 256;
+            }
+
+            if (i == claims.length - 1) {
+                _setClaimedBits(liquidityProvider, currentChannelId, currentWordIndex, currentBits);
+                _deductClaimedBalance(currentChannelId, currentClaimAmount);
+            }
+
             require(
                 verifyClaim(
```
TomAFrench, line 115: We can pass in the channel ID directly here

pkg/distributors/contracts/MerkleOrchard.sol
gtaschuk: yeah, I was encountering stack too deep orders, adding that function would probably be a better way to avoid them

pkg/distributors/contracts/MerkleOrchard.sol
```diff
@@ -63,16 +63,57 @@ contract MerkleOrchard {
...
-        IERC20 token;
         for (uint256 i = 0; i < claims.length; i++) {
             claim = claims[i];
-            token = tokens[claim.tokenIndex];
 
-            require(!isClaimed(token, claim.distributor, claim.distribution, liquidityProvider), "cannot claim twice");
+            // When we process a new claim we either
+            // a) aggregate the new claim bit with previous claims of the same channel/claim bitmap
+            // b) set claim status and start aggregating a new set of currentBits for a new channel/word
+            if (currentChannelId == bytes32(0)) {
```
TomAFrench, line 83: I'd remove this if statement and change the else block to conditionally update the stored bitmap as so
```
if (channelId != bytes32(0)){
   _setClaimedBits(liquidityProvider, currentChannelId, currentWordIndex, currentBits);
   _deductClaimedBalance(currentChannelId, currentClaimAmount);
 }

This would remove us having to perform a check that's only relevant for the first claim on each iteration.

File: pkg/distributors/contracts/MerkleOrchard.sol [hunk 5]
```diff
@@ -63,16 +63,57 @@ contract MerkleOrchard {
...
+            ) {
+                if (currentWordIndex == claim.distribution / 256) {
+                    currentBits |= 1 << claim.distribution % 256;
+                    currentClaimAmount += claim.balance;
+                } else {
+                    _setClaimedBits(liquidityProvider, currentChannelId, currentWordIndex, currentBits);
+
+                    currentWordIndex = claim.distribution / 256;
+                    currentBits = 1 << claim.distribution % 256;
+                }
```
TomAFrench, line 99:
```suggestion
                currentClaimAmount += claim.balance;
                if (currentWordIndex == claim.distribution / 256) {
                    currentBits |= 1 << claim.distribution % 256;
                } else {
                    _setClaimedBits(liquidityProvider, currentChannelId, currentWordIndex, currentBits);

                    currentWordIndex = claim.distribution / 256;
                    currentBits = 1 << claim.distribution % 256;
                }
```

Currently if you go into the next word of the channel then it looks like you won't get any of the tokens.

pkg/distributors/contracts/MerkleOrchard.sol
```diff
@@ -201,23 +260,24 @@ contract MerkleOrchard {
         uint256 claimedBalance,
         bytes32[] memory merkleProof
     ) public view returns (bool) {
```
TomAFrench, line 262:
```suggestion
    ) external view returns (bool) {
```

TomAFrench (approved):
💯

PR #902, opened 2021-10-04
refactor: remove merkleRoots and claimStatus functions

TomAFrench:
I'm of the opinion that these functions are unnecessary due to us being able to call into the orchard through a multicall contract when it comes to querying this information and in the situation where we're calling one then we're likely calling the other (which would require a multicall contract to perform both queries in a single call).

### DIFF
--- a/pkg/distributors/contracts/MerkleOrchard.sol
+++ b/pkg/distributors/contracts/MerkleOrchard.sol
@@ -162,37 +162,6 @@ contract MerkleOrchard {
         return claimed[token][distributor][distribution][liquidityProvider];
     }
 
-    function claimStatus(
-        address liquidityProvider,
-        IERC20 token,
-        address distributor,
-        uint256 begin,
-        uint256 end
-    ) external view returns (bool[] memory) {
-        require(begin <= end, "distributions must be specified in ascending order");
-        uint256 size = 1 + end - begin;
-        bool[] memory arr = new bool[](size);
-        for (uint256 i = 0; i < size; i++) {
-            arr[i] = isClaimed(token, distributor, begin + i, liquidityProvider);
-        }
-        return arr;
-    }
-
-    function merkleRoots(
-        IERC20 token,
-        address distributor,
-        uint256 begin,
-        uint256 end
-    ) external view returns (bytes32[] memory) {
-        require(begin <= end, "distributions must be specified in ascending order");
-        uint256 size = 1 + end - begin;
-        bytes32[] memory arr = new bytes32[](size);
-        for (uint256 i = 0; i < size; i++) {
-            arr[i] = trees[token][distributor][begin + i];
-        }
-        return arr;
-    }
-
     function verifyClaim(
         IERC20 token,
         address distributor,
--- a/pkg/distributors/test/MerkleOrchard.test.ts
+++ b/pkg/distributors/test/MerkleOrchard.test.ts
@@ -327,15 +327,8 @@ describe('MerkleOrchard', () => {
     });
 
     it('reports distributions as unclaimed', async () => {
-      const expectedResult = [false, false];
-      const result = await merkleOrchard.claimStatus(lp1.address, token.address, distributor.address, 1, 2);
-      expect(result).to.eql(expectedResult);
-    });
-
-    it('returns an array of merkle roots', async () => {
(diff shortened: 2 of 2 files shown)


gtaschuk:
I think I agree (https://github.com/balancer-labs/balancer-v2-monorepo/pull/829#issuecomment-933049502).

The current frontend PR will need to be updated https://github.com/balancer-labs/frontend-v2/pull/906 to not use those query methods, but I think that's fine

I think we should consider doing this in tandem with adding merkle roots to `DistributionAdded` events

pkg/distributors/contracts/MerkleOrchard.sol
```diff
@@ -37,7 +37,7 @@ contract MerkleOrchard {
     // token > distributor > balance
     mapping(IERC20 => mapping(address => uint256)) public suppliedBalance;
 
-    event DistributionAdded(address indexed token, uint256 amount);
+    event DistributionAdded(address indexed token, bytes32 merkleRoot, uint256 amount);
     event DistributionPaid(address indexed user, address indexed token, uint256 amount);
```
nventuro, line 41: Shouldn't these have the channel id?

TomAFrench: in place of `token`?

TomAFrench: `DistributionPaid` can refer to multiple channels as it's currently emitted once for each token being claimed

nventuro: No, I meant as an extra argument. If there are multiple channels for `token` (e.g. multiple teams making distributions in the same token), `DistributionPaid` isn't enough to figure out which channel was claimed.

TomAFrench: I've split up where we emit the `DistributionPaid` (now `DistributionSent`) event to support this.

pkg/distributors/contracts/MerkleOrchard.sol
```diff
@@ -125,6 +125,7 @@ contract MerkleOrchard {
                 if (currentChannelId != bytes32(0)) {
                     _setClaimedBits(liquidityProvider, currentChannelId, currentWordIndex, currentBits);
                     _deductClaimedBalance(currentChannelId, currentClaimAmount);
+                    emit DistributionSent(currentChannelId, recipient, tokens[claim.tokenIndex], currentClaimAmount);
```
TomAFrench, line 128: I've noticed that we're currently emitting `recipient` as `user` in the `DistributionSent` event. This seems odd to me as when using a callback contract the contract's address will be listed here instead of the `claimer`'s

gtaschuk: yes, recipient is a better term

pkg/distributors/contracts/MerkleOrchard.sol
```diff
@@ -37,8 +37,8 @@ contract MerkleOrchard {
     // channelId > balance
     mapping(bytes32 => uint256) private _suppliedBalance;
 
-    event DistributionAdded(address indexed token, uint256 amount);
-    event DistributionSent(address indexed user, address indexed token, uint256 amount);
+    event DistributionAdded(bytes32 indexed channelId, IERC20 indexed token, bytes32 merkleRoot, uint256 amount);
```
gtaschuk, line 40: I think I prefer `event DistributionAdded(address indexed distributor, IERC20 indexed token, bytes32 merkleRoot, uint256 amount);` as `channelId` is just an implementation detail irrelevant to the claimant and this would facilitate better filtering based on the address adding rewards

gtaschuk (changes requested):

gtaschuk:
I am now thinking that we should leave the range methods in to simplify the frontend integration - we can always just not use them in the future if we move to a multicall.

TomAFrench:
> I agree, removing the `claimStatus` range function at this point will create additional work to the frontend team which I would prefer to avoid. The frontend doesn't use `merkleRoots` though, in case you think it makes sense to remove it.

tbh, I don't really understand this argument. We can create a equivalent function in javascript really easily (I'd be happy to do this) so there's not really any need to bloat the contract to save such a small amount of work.

PR #903, opened 2021-10-05
Orchard: use private state variables

### LINKED ISSUE #887: Follow code style for external functions
Among others, in all our contracts we prefix getters with `get` (e.g. `getVault()`), resulting in `private` state variables and explicit getters. This is often done to make it easier to understand in client code which calls send a transaction and which don't, but also to isolate state variables accross contracts in an inheritance tree. We should follow this practice even in contracts that are not (yet) part of complex inheritance schemes.

gtaschuk:
closes #887

### DIFF
--- a/pkg/distributors/contracts/MerkleOrchard.sol
+++ b/pkg/distributors/contracts/MerkleOrchard.sol
@@ -31,19 +31,19 @@ contract MerkleOrchard {
 
     // Recorded distributions
     // channelId > distributor > distribution > root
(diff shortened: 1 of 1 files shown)


TomAFrench (approved):
You could argue that we should add a getter for `_claimedBitmap` as well but I think `isClaimed` fills that niche better.

TomAFrench:
Ah didn't see that this had auto-merge on. pinging @gtaschuk to make you aware of the extra commits I added.

PR #904, opened 2021-10-05
Reorganise some MerkleOrchard functions

TomAFrench:
There's no new functionality in this PR, just changing how we lay out the contract.

- Getter functions have all been moved to the top of the file to match how we lay out the pool contracts
- The order of args in `claimDistributionsWithCallback` have been changed to match that in the other two `claimDistributions` functions.
- The order of args in other functions such as `_verifyClaim` have been updated so that we always provide them in the order `channelId (token+distributor) > distribution > recipient address` as we had some inconsistencies and this seems the most logical way to set it out.
- `liquidityProvider` has been replaced with `claimer` as there's no requirement to provide liquidity to be included in a distribution.

### DIFF
--- a/pkg/distributors/contracts/MerkleOrchard.sol
+++ b/pkg/distributors/contracts/MerkleOrchard.sol
@@ -32,7 +32,7 @@ contract MerkleOrchard {
     // Recorded distributions
     // channelId > distribution > root
     mapping(bytes32 => mapping(uint256 => bytes32)) private _distributionRoot;
-    // channelId > lp > distribution / 256 -> bitmap
+    // channelId > claimer > distribution / 256 -> bitmap
     mapping(bytes32 => mapping(address => mapping(uint256 => uint256))) private _claimedBitmap;
     // channelId > balance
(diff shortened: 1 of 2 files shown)


pkg/distributors/contracts/MerkleOrchard.sol
```diff
@@ -76,17 +78,144 @@ contract MerkleOrchard {
         IERC20 token,
         address distributor,
         uint256 distribution,
-        address liquidityProvider
+        address claimer
```
nventuro, line 81: Note that some comments list 'lp', for example on line 35 when describing the mapping layout.

TomAFrench: Thanks, I can't find any references other than this.

gtaschuk (approved):
This all looks good but to be honest I would have rather you communicated about these changes before you did them, or just created issues for me since I was under the impression I would be the only person working on these

I have some in progress stuff it's just not really a smooth workflow to have to keep rebasing.  I don't want to sound ungrateful, just more communication would be appreciated.

PR #926, opened 2021-10-12
Merkle Orchard Deployment

gtaschuk:
This is the deployment of the merkle orchard to all networks


nventuro:
This PR has been updated with the changes introduced in https://github.com/balancer-labs/balancer-v2-monorepo/pull/946, and all contracts redeployed.

@gtaschuk can you check that the new build-info and ABI are correct, and that all addresses point to the new version? Thanks!

PR #932, opened 2021-10-12
Initial implementation of safe multicall relayer

### LINKED ISSUE #689: Relayer mutlicall reentrancy protection

TomAFrench:
This PR has an implementation of multicall relayer with logic stored in a separate contract in order to resolve #689.

Contracts which implement the actual logic of a relayer inherit from `BaseRelayerImplementation` which has the bare necessities for signature approvals, etc. Any extra functionality is added through exposing `external payable` functions on this contract.

Within the `BaseRelayerImplementation` constructor we deploy a `RelayerEntrypoint` contract. This has a single external function of a reentrancy protected `multicall` which automatically sweeps any ETH back to `msg.sender` at the end of the tx.

### DIFF
--- a/pkg/standalone-utils/contracts/relayer/BalancerRelayer.sol
+++ b/pkg/standalone-utils/contracts/relayer/BalancerRelayer.sol
@@ -53,7 +53,7 @@ contract BalancerRelayer is IBalancerRelayer, ReentrancyGuard {
     /**
      * @dev This contract is not meant to be deployed directly by an EOA, but rather during construction of a child of
      * `BaseRelayerLibrary` which will provides its own address to be used as the relayer's library.
-     */ 
+     */
     constructor(IVault vault, address libraryAddress) {
         _vault = address(vault);
         _library = libraryAddress;


TomAFrench:
I was considering whether to add `Library` to the `VaultActions`, etc. but decided that it's a bit noisy and they're unable to be deployed except as part of a derived contract which would have the `Library` suffix.

I think this PR as a multicall relayer as described in #689 is complete. I'd like to tackle #934 in a separate PR by migrating the LidoRelayer's functionality over into this relayer (after some toy examples to settle on the syntax). We'd want to do this before launching the new relayer anyway and Lido is a more concrete example than trying to do a complex set of operations involving Linear pools.

Issue #907, opened 2021-10-05
MultiDistributor: Code styling issues


facuspagnuolo:
Some minor details we could address to improve the code styling here:

- [x] Rename "allowlist" to "whitelist" (https://github.com/balancer-labs/balancer-v2-monorepo/pull/940)
- [x] Rename "pool" to "staking token" (https://github.com/balancer-labs/balancer-v2-monorepo/pull/939)
- [x] Simplify naming strategy for indexes in loops (#952)
- [x] Merge "account" and "user" terminologies (#952)
- [x] Improve inline documentation (https://github.com/balancer-labs/balancer-v2-monorepo/pull/968)
- [x] Simplify and polish revert reasons, try using less than 32 characters (https://github.com/balancer-labs/balancer-v2-monorepo/pull/968)

Issue #922, opened 2021-10-07
MultiDistributor: Consider using a permissionless model


gtaschuk:
Up for discussion:

Currently we use a permissioned model to add distributors to a pool (either via governance, pools, or asset managers).  This is to prevent a DDOS attack in which looping through too many distributions would make staking/unstaking/claiming rewards expensive or potentially completely block unstaking from a pool.  And yet, if the permissioned entities were compromised, they could still force distributions on users, paralyzing pool stakes.

An alternate approach would be to make this contract completely permissionless - users would instead opt in to _channels_ {staked token, distributed token, distributor}, that they wish to receive distributions for and anyone would be able to become a  distributor.

I'm imagining opt-in being binary - a user would have all their staking tokens staked to the channels they want to claim.  The UX of opting in to channels could be mitigated with a permit scheme .  Best case we figure out how to pack 256 opt-in/opt-out into a storage slot and allow it to be updated en masse

It's worth noting that while reward rates and reward durations vary, it doesn't matter who issues a distribution, and users should opt for the channels that best suit their staking plans.  Perhaps this could translate into automatic opt-in for particular distributions or permit-based opt in (for instance, any channel with a sufficiently high reward rate, or added through a pool etc)

Moving to a permissionless model results in much less operational overhead for managing who is an allowed distributor and puts the users in control of which distributions are worth their gas to receive.  It also makes this contract useful beyond our ecosystem and allows people to set up staking schemes for any token in an environment where there is cheap liquidity for distributed tokens through balancer.

PR #940, opened 2021-10-15
Staking: Rename 'allowlist' to 'whitelist'

facuspagnuolo:
Related to https://github.com/balancer-labs/balancer-v2-monorepo/issues/907


facuspagnuolo:
> Are we not getting rid of this anyway due to #922?

Yes, but this was done before we decided to move forward with that one 😄

Issue #910, opened 2021-10-05
MultiDistributor: Improve interface


facuspagnuolo:
The `MultiRewards` interface is a bit vague and can be extended to provide further details about is functionality, an idea could be:

```solidity
interface IMultiRewards {
   event Staked(address indexed pool, address indexed user, uint256 amount);
   event Withdrawn(address indexed pool, address indexed user, uint256 amount);
   event Claimed(address indexed pool, address indexed token, address indexed user, uint256 amount);
   event RewardSet(...);

   function stake(address pool, uint256 amount) external;
   function stakeFor(address pool, address user, uint256 amount) external;
   function stakeWithPermit(address pool, address user, uint256 amount, ...) external;
   function withdraw(address pool, uint256 amount) external;
   function claim(address pool, address token) external;
   function exit(address pool, address token) external;

   function isRewarderWhitelisted(address pool, address rewarder);
   function isRewardTokenWhitelisted(address pool, address rewarder, address token);

   function getReward(address pool, address rewarder, address token) external view returns (Reward memory);
   function getRewardPerToken(address pool, address rewarder, address token) external view returns (uint256);
   function getLastTimeRewardApplicable(address pool, address rewarder, address token) external view returns (uint256);

   function getTotalStaked(address pool) external view returns (uint256);
   function getStakedBalance(address pool, address user) external view returns (uint256);
   function getRewardsEarned(address pool, address token, address user) external view returns (uint256);
   function getRewardsEarned(address pool, address rewarder, address token, address user) external view returns (uint256);
}
```

PR #967, opened 2021-10-27
Staking: Massive tests

facuspagnuolo:
**MERGE AGAINST MASTER AFTER #963**

Fixes https://github.com/balancer-labs/balancer-v2-monorepo/issues/910

A few functions were improved to provide a better and handier interface


nventuro:
Merging as is to unblock work on other PRs, further changes to the tests will come in standalone PRs.

PR #968, opened 2021-10-28
Staking: Code styling and general improvements

### LINKED ISSUE #905: MultiDistributor: Rename `getReward` to `claim`
The original function name refers more to a view function rather than sth that actually claims the rewards for a user.
Same with `getRewardAsInternalBalance`.

### LINKED ISSUE #906: MultiDistributor: Inconsistent naming strategy `unstake` vs `withdarwn`
It would be nice if we could align the naming strategy between "unstake" and "withdraw"

### LINKED ISSUE #907: MultiDistributor: Code styling issues

### LINKED ISSUE #908: MultiDistributor: Align `notifyRewardAmount` and `addReward`
I'd try to be consistent with the naming strategy between the method and the event name here.

facuspagnuolo:
**MERGE AGAINST MASTER AFTER #967**

It is heavily recommended to review this PR commit by commit.

Fixes #905
Fixes #906 
Fixes #907
Fixes #908


nventuro:
Force-pushed to rebase on top of #967

pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -494,134 +477,169 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
...
+                recipient: payable(recipient),
+                kind: kind
+            });
+
+            if (amounts[i] > 0){
+                emit RewardPaid(msg.sender, address(tokens[i]), amounts[i]);
+            }
+        }
+
+        getVault().manageUserBalance(ops);
```
TomAFrench, line 579: I've included the mechanism of batching transfers used in #827. This seems more suitable to me than the method used in #894 as a simplified interface is more important here as `MultiDistributor` is likely to have many contracts as users.

nventuro:
I temporarily undid the MultiRewards -> MultiDistributor rename in a87a636 (to be reverted later) so that GitHub wouldn't treat this as the old file being deleted and a new one created, making review harder.

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -25,319 +25,356 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/IERC20Permit.sol
...
+    function getDistributionId(
         IERC20 stakingToken,
-        IERC20 rewardsToken,
-        uint256 duration
-    ) external override returns (bytes32 distributionId) {
-        require(duration > 0, "reward rate must be nonzero");
-        require(address(stakingToken) != address(0), "STAKING_TOKEN_ZERO_ADDRESS");
-        require(address(rewardsToken) != address(0), "REWARDS_TOKEN_ZERO_ADDRESS");
+        IERC20 distributionToken,
+        address distributor
```
TomAFrench, line 106:
```suggestion
        address owner
```
I need to update this to match whats in `Distribution`

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -25,319 +25,356 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/IERC20Permit.sol
...
-        returns (uint256)
-    {
-        // If the user is not subscribed to the queried distribution, it should be handled as if the user has no stake.
-        // Then, it can be short cut to zero.
-        if (!userStaking.allowedDistributions.contains(distributionId)) {
-            return 0;
+        Distribution storage distribution = _getDistribution(distributionId);
+        require(distribution.duration > 0, "DISTRIBUTION_DOES_NOT_EXIST");
+        // TODO: add test for this
+        require(distribution.owner == msg.sender, "SENDER_NOT_DISTRIBUTOR");
```
TomAFrench, line 256:
```suggestion
        require(distribution.owner == msg.sender, "SENDER_NOT_OWNER");
```

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -390,245 +427,233 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
...
+
+    function _globalTokensPerStake(Distribution storage distribution) internal view returns (uint256) {
+        uint256 supply = distribution.totalSupply;
+        if (supply == 0) {
+            return distribution.globalTokensPerStake;
+        }
+
+        // Underflow is impossible here because _lastTimePaymentApplicable(...) is always greater than last update time
+        uint256 unpaidDuration = _lastTimePaymentApplicable(distribution) - distribution.lastUpdateTime;
+        uint256 unpaidAmountPerToken = Math.mul(unpaidDuration, distribution.paymentRate).divDown(supply);
```
nventuro, line 605: I found the fixed point division here bizarre. Created https://github.com/balancer-labs/balancer-v2-monorepo/issues/1010 to track this effort.

paymentRate is created as total / duration (non-fp). So what this does is:

`(elapsed * total / duration) * 1e18 / supply` (the 1e18 comes from fixed point division)

which seems incorrect?

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -390,245 +427,233 @@ contract MultiRewards is IMultiRewards, IDistributor, ReentrancyGuard, MultiRewa
...
-            _updateUserRewardRatePerToken(userStaking, distributionId);
+            _updateUserTokensPerStake(userStaking, distributionId);
         }
     }
 
-    /* ========== MODIFIERS ========
+    function _updateUserTokensPerStake(UserStaking storage userStaking, bytes32 distributionId) internal {
+        uint256 globalTokensPerStake = _updateDistributionRate(distributionId);
+        UserDistribution storage userDistribution = userStaking.distributions[distributionId];
+        userDistribution.unclaimedTokens = _totalUnclaimedTokens(userStaking, distributionId);
```
nventuro, line 579: This is not incorrect, but calling `_totalUnclaimedTokens` here is quite inefficient as most of what it does is recompute things we've already computed (e.g. check that the user is subscribed, compute globalTokensPerStake, etc.).

nventuro (commented):
Brilliant :raised_hands: It's way easier to follow what's going on now, thank you very much!

I created new issues for most of the things I pointed out in the interest of keeping this contained to naming and style

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -25,319 +25,356 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/IERC20Permit.sol
...
-        distribution.rewarder = msg.sender;
-        distribution.rewardsToken = rewardsToken;
-        distribution.stakingToken = stakingToken;
+    /**
+     * @dev Returns the information of a distribution
+     * @param distributionId ID of the distribution being queried
+     */
+    function getDistribution(bytes32 distributionId) external view override returns (Distribution memory) {
+        return _getDistribution(distributionId);
+    }
```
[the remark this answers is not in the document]
TomAFrench: One of these returns a storage pointer while the other returns memory so they fulfil different roles.

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -25,319 +25,356 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/IERC20Permit.sol
...
      * @param distributionId ID of the distribution being queried
-     * @param user The address of the user being queried
+     * @param user Address of the user being queried
      */
-    function isSubscribed(bytes32 distributionId, address user) external view returns (bool) {
+    function unaccountedUnclaimedTokens(bytes32 distributionId, address user) external view override returns (uint256) {
         IERC20 stakingToken = _getDistribution(distributionId).stakingToken;
-        return _userStakings[stakingToken][user].allowedDistributions.contains(distributionId);
+        UserStaking storage userStaking = _userStakings[stakingToken][user];
+        return _unaccountedUnclaimedTokens(userStaking, distributionId);
```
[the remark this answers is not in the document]
TomAFrench: This roughly lines up with my thoughts. I think they're accessible purely because they were in the original contract but for some of these I don't see a reason to expose them.

pkg/distributors/contracts/MultiRewards.sol
TomAFrench: Yes, I saw this and thought it was weird as it's the only usage of `FixedPoint` in the contract. Definitely going to dig into this more but I didn't want to muck with maths in this PR.

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -25,319 +25,356 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/IERC20Permit.sol
...
             uint256 amount = userStaking.balance;
-            if (amount == 0) userStaking.allowedDistributions.add(distributionId);
-            else {
-                userStaking.allowedDistributions.add(distributionId);
-                // The unpaid rewards remains the same because was not subscribed to the distribution
-                userStaking.distributions[distributionId].paidRatePerToken = _updateDistributionRate(distributionId);
+            if (amount > 0) {
+                subscribedDistributions.add(distributionId);
+                // The unclaimed tokens remains the same because the user was not subscribed to the distribution
+                userStaking.distributions[distributionId].userTokensPerStake = _updateDistributionRate(distributionId);
```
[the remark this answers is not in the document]
TomAFrench: I don't quite follow the concern here. The only situation where we don't update it is in the case where a user doesn't have a staked balance and so the user doesn't accrue rewards.

Either a user (1) stakes and subscribes or (2) subscribes and stakes.

1. `userTokensPerStake` is updated on both function calls.
2. The user has no staked tokens so accrues no rewards and `userTokensPerStake` is updated when the user stakes (just before their balance is updated)

I think this is acceptable as the places where we'd have to make a mistake in accounting for this to be an issue are few and well defined. (stake, unstake, unsubscribe.)

pkg/distributors/contracts/MultiRewards.sol
TomAFrench: Agreed, we can pass down some more storage pointers but in a later PR.

pkg/distributors/contracts/MultiRewards.sol
```diff
@@ -25,319 +25,356 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/IERC20Permit.sol
...
-     * @param distributionId ID of the distribution being queried
+     * @dev Creates a new distribution
+     * @param stakingToken The staking token that will be eligible for this distribution
+     * @param distributionToken The token to be distributed to users
+     * @param duration The duration over which each distribution is spread
      */
-    function lastTimeRewardApplicable(bytes32 distributionId) public view returns (uint256) {
-        return _lastTimeRewardApplicable(_getDistribution(distributionId));
-    }
+    function create(
```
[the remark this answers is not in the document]
TomAFrench: I think a happy medium is to have `Distribution` suffixes but not `Token` ones.

```
stake
unstake
claim
exit
createDistribution
fundDistribution
subscribeDistributions
unsubscribeDistributions
```

This keeps the main user functions simple and familiar while more unfamiliar ones are more explicit

pkg/distributors/contracts/MultiRewards.sol
nventuro: Agreed, we'll follow up in that issue!

pkg/distributors/contracts/MultiRewards.sol
nventuro: Created https://github.com/balancer-labs/balancer-v2-monorepo/issues/1013

pkg/distributors/contracts/MultiRewards.sol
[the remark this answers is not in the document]
nventuro: Agreed, created https://github.com/balancer-labs/balancer-v2-monorepo/issues/1016

pkg/distributors/contracts/MultiRewards.sol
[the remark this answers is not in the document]
nventuro: My concern was exactly that we'd have a mistake in the accounting :stuck_out_tongue: But I agree with your rationale. We have soft-stake and soft-subscribe, both of which don't do much (updated staked tokens and update set of subscriptions respectively): it is only the hard-stake (stake when subscribed or subscribe when staked) that matters, and there's limited places where it happens.

The same is true for unstake+unsubscribe.

nventuro (approved):
Reviewing your comments to my review was an extremely pleasant experience, thank you for taking the time to address each so cleanly! :raised_hands:

PR #971, opened 2021-10-28
Staking: Drop reinvestor

facuspagnuolo:
**MERGE AGAINST MASTER AFTER #968**

PR #974, opened 2021-10-28
Distributors: test reinvestor against orchard

gtaschuk:
Submitting as an alternative to #971 

This uses the merkle orchard to test the reinvestor instead of the staking contract.  Reinvestment is one of our advertised benefits of the callback pattern so seems worth preserving/deploying


TomAFrench:
Closing as we're going to be transitioning away from this with the gauge system.

PR #978, opened 2021-10-31
Remove extraneous base class

EndymionJkb:
Discovered when writing Fei documentation. Cleaner class diagram this way.


TomAFrench (dismissed):
LGTM

TomAFrench (approved):
nvm, my mistake. I was thinking we shouldn't call the ERC20 constructor here as it would be done in ERC20Permit but that's not the case.

PR #980, opened 2021-10-31
Feat/587 where field authorizer

Ramarti:
Modifies `AccessControl` and  `Authorizer` to add more fine grained permissions for actions.

Now permission to execute an action (role) can be granted / revoked globally (working for all contracts, like before) or for a specific set of addresses.

`canPerform` will return true if the address asking for an `action` has global or by contract permission. 
This changes are transparent for the rest of the contracts

// TODO: migration script for Authorizer


pkg/solidity-utils/contracts/openzeppelin/AccessControl.sol
```diff
@@ -122,38 +167,73 @@ abstract contract AccessControl {
...
      * Requirements:
      *
      * - the caller must have ``role``'s admin role.
+     * - list of ``where``'s can't be empty
      */
-    function grantRole(bytes32 role, address account) public virtual {
-        _require(hasRole(_roles[role].adminRole, msg.sender), Errors.GRANT_SENDER_NOT_ADMIN);
+    function grantRole(bytes32 role, address account, address[] calldata where) public virtual {
+        _require(where.length > 0, Errors.INPUT_LENGTH_MISMATCH);
+        _require(hasRole(_roles[role].adminRole, msg.sender, GLOBAL_ROLE_ADMIN), Errors.GRANT_SENDER_NOT_ADMIN);
```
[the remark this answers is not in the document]
nventuro: This is however entirely optional. If we go with https://github.com/balancer-labs/balancer-v2-monorepo/issues/985, then I'd not change anything now and just use this PR as a stepping stone to get there.

pkg/vault/contracts/Authorizer.sol
```diff
@@ -35,47 +35,84 @@ contract Authorizer is AccessControl, IAuthorizer {
...
+        InputHelpers.ensureInputLengthMatch(roles.length, accounts.length);
+        for (uint256 i = 0; i < roles.length; i++) {
+            grantRole(roles[i], accounts[i], where);
+        }
+    }
+
+    /**
+     * @dev Grants roles to a list of accounts for all contracts.
+     */
+    function grantRolesGloballyToMany(bytes32[] memory roles, address[] memory accounts) external {
```
nventuro, line 74: These functions are starting to grow a bit out of control: they're fine for now, but I created https://github.com/balancer-labs/balancer-v2-monorepo/issues/986 to propose reducing how many flavors we support.

pkg/vault/test/Authorizer.test.ts
```diff
@@ -20,6 +20,8 @@ describe('Authorizer', () => {
   const ROLE_2 = '0x0000000000000000000000000000000000000000000000000000000000000002';
 
   const ROLES = [ROLE_1, ROLE_2];
+  const WHERE = [ethers.Wallet.createRandom().address, ethers.Wallet.createRandom().address];
+  const NOT_WHERE = ethers.Wallet.createRandom().address;
```
nventuro, line 24: Huh I didn't know about `createRandom()`, that's cool! We often use the `other` signer to represent this, but this is a nifty tool.

pkg/vault/test/Authorizer.test.ts
```diff
@@ -87,30 +131,62 @@ describe('Authorizer', () => {
         authorizer = authorizer.connect(admin);
       });
 
-      context('when the roles where granted', () => {
+      context('when the roles ANYWHERE granted to a set of contracts', () => {
```
nventuro, line 134:
```suggestion
      context('when the roles were granted to a set of contracts', () => {
```

We had a typo in the original description ('where' instead of 'were'), and it seems that fell prey to your find+replace :)

pkg/vault/test/Authorizer.test.ts
```diff
@@ -133,32 +217,66 @@ describe('Authorizer', () => {
         authorizer = authorizer.connect(admin);
       });
 
-      context('when the roles where granted', () => {
+      context('when the roles ANYWHERE granted globally', () => {
```
nventuro, line 220:
```suggestion
      context('when the roles were granted globally', () => {
```

Same here

pkg/vault/test/Authorizer.test.ts
```diff
@@ -133,32 +217,66 @@ describe('Authorizer', () => {
...
-            expect(await authorizer.canPerform(role, grantee.address, WHERE)).to.be.false;
-            expect(await authorizer.canPerform(role, other.address, WHERE)).to.be.false;
+            expect(await authorizer.canPerform(role, grantee.address, ANYWHERE)).to.be.false;
+            expect(await authorizer.canPerform(role, other.address, ANYWHERE)).to.be.false;
           }
         });
       });
 
-      context('when one of the roles was not granted', () => {
+      context('when the roles ANYWHERE granted to a set of contracts', () => {
```
nventuro, line 235:
```suggestion
      context('when the roles were granted to a set of contracts', () => {
```

nventuro (commented):
Looks good, thanks a lot @Ramarti!

We've already had a lot of back and forth in discussing how to best tackle this, but after seeing the full extent of the changes I created two more issues to guide this effort in a direction that I think will be more sustainable (https://github.com/balancer-labs/balancer-v2-monorepo/issues/985 and https://github.com/balancer-labs/balancer-v2-monorepo/issues/986). 

We don't need to worry about those two now (they can be addressed in the future after this PR is merged), but I'd appreciate any feedback you might have on them.

Thanks again!

pkg/solidity-utils/contracts/openzeppelin/AccessControl.sol
```diff
@@ -45,14 +45,16 @@ abstract contract AccessControl {
...
+        EnumerableSet.AddressSet globalMembers;
+        mapping(address => EnumerableSet.AddressSet) membersByContract;
         bytes32 adminRole;
     }
 
     mapping(bytes32 => RoleData) private _roles;
 
     bytes32 public constant DEFAULT_ADMIN_ROLE = 0x00;
-
+    address public constant GLOBAL_ROLE_ADMIN = address(0);
```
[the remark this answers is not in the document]
Ramarti: Yes, kind of.

When you check for has role:

```
function hasRole(bytes32 role, address account, address where) public view virtual returns (bool) {
        return _roles[role].globalMembers.contains(account) || 
            _roles[role].membersByContract[where].contains(account);    
}
```
If it has the global role, it will ignore the `where` address, and in case it does not have the global role, you should not set address(0) as a `where` so it will return false. It's a way to make more evident you are checking for a global role without another method.

Actually, just realized we should require not giving a non-global role to address(0) just in case. Will make a commit

pkg/solidity-utils/contracts/openzeppelin/AccessControl.sol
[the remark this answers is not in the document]
Ramarti: Yeah if we make a more specific AccessControl, maybe it makes sense to have 2 hasRole versions, one for globals

pkg/vault/contracts/Authorizer.sol
Ramarti: I was not sure about them either but wanted to keep consistent. Also adds some pretty trivial tests and a bit of complexity

pkg/vault/test/Authorizer.test.ts
Ramarti: I was spending too much time context-switching between the test setup files and the test files to really get which address what was, and I think at some point a test was failing because I misused one, so I thought 'there must be a better way'

pkg/vault/contracts/Authorizer.sol
nventuro: Makes sense! We can just drop them later as part of #986

pkg/solidity-utils/contracts/openzeppelin/AccessControl.sol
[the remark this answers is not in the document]
nventuro: Ah that's a good realization, yes. I think it is another piece of evidence pointing towards #985 being desirable (particularly, hvaing the notion of adminship be more explicit).

nventuro:
@Ramarti I didn't notice I had to approve the PR for the CI tests to run, so I only now saw that some of them failed. Could you look into that? Thanks!

nventuro (approved):
Looks great, thanks a lot @Ramarti!

PR #982, opened 2021-11-02
Merge Lido and Batch relayers

TomAFrench:
I've written a relayer module for interacting with Lido so that users can perform wrap/unwrap actions within a multicall. Tests have been updated to check functionality is maintained.

`VaultActions` now allows the relayer to be the source of funds in order to avoid having to unnecessarily transfer wstETH back and forth from the user.

As a bonus we now allow users to stake ETH as part of a relayer interaction.


pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -72,383 +65,543 @@ describe('LidoRelayer', function () {
...
+      sender,
+      recipient,
+      amount,
+      outputReference ?? 0,
+    ]);
+  }
 
-          sharedBeforeEach('build swap request', async () => {
-            singleSwap = {
+  describe('swap', () => {
```
nventuro, line 118: These tests were good back when the primitives of the relayer were swap, join, etc. I'd change this up now for more focused tests the use the different functionalities of the wrapper library (chained references, recipient redirection, etc.)

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -72,383 +65,543 @@ describe('LidoRelayer', function () {
...
+        const tokenIn = tokens.findBySymbol('wstETH');
+        const tokenOut = tokens.WETH;
+        const amount = fp(1);
+
+        const receipt = await (
+          await relayer.connect(sender).multicall([
+            encodeWrap(sender.address, relayer.address, amount, toChainedReference(0)),
+            encodeApprove(tokenIn, MAX_UINT256),
+            encodeSwap({
+              poolId,
```
nventuro, line 163: Related to the comment above, now that we have chained references we could also structure this in a different way by sending the wstETH back to the sender and swapping from their wallet, avoiding the approval. This would make e.g. joining a pool with wstETH + other token easier.

pkg/standalone-utils/contracts/interfaces/IBaseRelayerLibrary.sol
```diff
@@ -23,6 +23,20 @@ import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
 abstract contract IBaseRelayerLibrary {
     function getVault() public view virtual returns (IVault);
 
+    function approveVault(IERC20 token, uint256 amount) public virtual;
```
nventuro, line 26: I didn't see the need for this before, but now I do: if the relayer might be sender, it needs to have allowance.

The problem with this interface is that we'd always provide infinite allowance, as we wouldn't be able to pass a chained reference here. Would it make sense to instead have swap, join, etc. all do manual approvals for the exact amount before that operation if the relayer was the sender? It'd be extra work for us (meaning we'll need to add functionality to those functions), but way more gas efficient and easier to use for the front-end.

pkg/standalone-utils/contracts/LidoBatchRelayerLibrary.sol
```diff
@@ -15,36 +15,19 @@
...
+import "./relayer/VaultPermit.sol";
+
+/**
+ * @title Batch Relayer Library
+ * @notice This contract is not a relayer by itself and calls into it directly will fail.
+ * The associated relayer can be found by calling `getEntrypoint` on this contract.
+ */
+contract LidoBatchRelayerLibrary is BaseRelayerLibrary, LidoWrapping, VaultActions, VaultPermit {
+    constructor(IVault vault, IERC20 wstETH) BaseRelayerLibrary(vault) LidoWrapping(wstETH) {
+        // solhint-disable-previous-line no-empty-blocks
```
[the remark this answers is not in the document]
TomAFrench: Ah sorry, this was a stepping stone to keep the old tests working while I was setting up Lido support. Agree on just having a single canonical relayer (seeing as it's the whole point of this project)

pkg/standalone-utils/contracts/relayer/LidoWrapping.sol
```diff
@@ -0,0 +1,149 @@
...
+    IERC20 private immutable _stETH;
+    IERC20 private immutable _wstETH;
+
+    /**
+     * @dev The zero address may be passed as wstETH to safely disable this module
+     * @param wstETH - the address of the Lido's wrapped stETH contract
+     */
+    constructor(IERC20 wstETH) {
+        // Safely disable stETH wrapping if no address has been passed for wstETH
+        _stETH = wstETH != IERC20(0) ? IERC20(IwstETH(address(wstETH)).stETH()) : IERC20(0);
```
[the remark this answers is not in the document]
TomAFrench: Yes, this is to allow us to test other features of the relayer without having to deploy a working `wstETH` contract.

pkg/standalone-utils/test/LidoRelayer.test.ts
TomAFrench: I'd say this is non ideal as it would require the user to have an approval on the vault for wstETH whereas having the relayer be the sender allows users to trade while only having a stETH approval.

TomAFrench:
force pushed as accidently pushed a temp commit

pkg/standalone-utils/test/LidoRelayer.test.ts
TomAFrench: I've started adding these tests for primitives although I think it's helpful to have some more complex tests in cases where we have a well defined expected usecase for the module being tested (e.g. for `LidoWrapping` it would be wrap steth and trade using wstETH)

pkg/standalone-utils/contracts/interfaces/IBaseRelayerLibrary.sol
TomAFrench: Thoughts on adding a `_peekChainedReferenceValue()` function for cases like this? This would read a referenced value but not zero it out.

TomAFrench: The issue with bundling an approval into the vault actions is that it then requires that we disallow any GIVEN_OUT actions as these would leave dust approvals on the vault. (unless we purposefully zeroed out the allowance afterwards.)

nventuro: Let's keep infinite allowance for now. Created https://github.com/balancer-labs/balancer-v2-monorepo/issues/991

pkg/standalone-utils/test/LidoRelayer.test.ts
nventuro: Yes, testing the expected use cases makes sense as that may reveal things we didn't notice (such as the approval for wstETH that was mentioned in some other comment). But I still believe a better balance would be to have full unit test style coverage of the primitives, and then a few 'integration' tests at the end (like we do when chaining swaps and joins).

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
+          } = expectTransferEvent(
+            receipt,
+            {
+              from: relayerIsRecipient ? ZERO_ADDRESS : relayer.address,
+              to: relayerIsRecipient ? relayer.address : TypesConverter.toAddress(tokenRecipient),
+              value: expectedWstETHAmount,
+            },
+            wstETH
+          );
+          await expectChainedReferenceContents(toChainedReference(0), wstETHAmount);
```
nventuro, line 241:
```suggestion
          await expectChainedReferenceContents(toChainedReference(0), expectedWstETHAmount);
```

We don't need this: we already know `value` (we set it to `expectedWstETHAmount`). And we've already tested for the transfer event in the previous test.

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
-              // amountOut
-            });
-          });
+          const relayerIsRecipient = TypesConverter.toAddress(tokenRecipient) === relayer.address;
+          expectTransferEvent(
+            receipt,
+            {
+              from: relayerIsRecipient ? ZERO_ADDRESS : relayer.address,
+              to: relayerIsRecipient ? relayer.address : TypesConverter.toAddress(tokenRecipient),
+              value: expectedWstETHAmount,
```
nventuro, line 213: We're sort of indirectly testing that the input amount was correct, shouldn't we test for that?

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
+
+          const relayerIsRecipient = TypesConverter.toAddress(tokenRecipient) === relayer.address;
+          expectTransferEvent(
+            receipt,
+            {
+              from: relayerIsRecipient ? ZERO_ADDRESS : relayer.address,
+              to: relayerIsRecipient ? relayer.address : TypesConverter.toAddress(tokenRecipient),
+              value: expectedWstETHAmount,
+            },
+            wstETH
```
nventuro, line 261: In this case, it definitely feels like what we should be testing is that the input is the chained amount, instead of indirectly asserting that via the output.

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
+          await stETH.transfer(relayer, fp(1), { from: sender });
+          tokenSender = relayer;
+          tokenRecipient = relayer;
         });
+        testWrap();
+      });
 
-        describe('swap using stETH as an input', () => {
-          let singleSwap: SingleSwap;
+      context('sender = relayer, recipient = sender', () => {
```
nventuro, line 189: Why no sender = sender, recipient = sender?

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
-              amountIn: singleSwap.amount,
-              // amountOut
-            });
-          });
+          const relayerIsRecipient = TypesConverter.toAddress(tokenRecipient) === relayer.address;
+          expectTransferEvent(
+            receipt,
+            {
+              from: relayerIsRecipient ? ZERO_ADDRESS : relayer.address,
+              to: relayerIsRecipient ? relayer.address : TypesConverter.toAddress(tokenRecipient),
```
nventuro, line 212: Note that we're not testing that the tokens came from the sender: maybe we should have two assertions, one where the sender sends the tokens, and one where the recipient receives them.

nventuro: The same appies to other `testFoo()` tests.

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
-                assetOutIndex: basePoolTokens.findIndexBySymbol('WETH'),
-                amount: fp(1),
-                userData: '0x',
-              },
-            ];
-
-            assets = basePoolTokens.addresses;
-            limits = basePoolTokens.map((token) => (token.symbol === 'wstETH' ? fp(1) : 0));
-          });
+      context('sender = relayer, recipient = sender', () => {
```
nventuro, line 365: This seems incorrect: the sender of stake is always the user (who sends eth to the relayer): we should only have two options, depending on whether the recipient is the relayer or the user.

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
+          const relayerIsRecipient = TypesConverter.toAddress(tokenRecipient) === relayer.address;
+          expectTransferEvent(
+            receipt,
+            {
+              from: relayerIsRecipient ? ZERO_ADDRESS : relayer.address,
+              to: relayerIsRecipient ? relayer.address : TypesConverter.toAddress(tokenRecipient),
+              value: amount,
+            },
+            stETH
+          );
```
nventuro, line 410:
```suggestion
```
This is the same assertion as the one in the previous test, seems we can drop it?

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
-    });
-  });
 
-  describe('joinPool', () => {
-    let joinRequest: JoinPoolRequest;
+      context('sender = relayer, recipient = sender', () => {
+        beforeEach(() => {
+          tokenRecipient = sender;
+        });
+        testStake();
```
nventuro, line 460: Same comment re. there only being a single sender.

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
 
-    sharedBeforeEach('build join request', async () => {
-      joinRequest = {
-        assets: basePoolTokens.addresses,
-        maxAmountsIn: [0, fp(1)],
-        userData: WeightedPoolEncoder.joinExactTokensInForBPTOut([0, fp(1)], 0),
-        fromInternalBalance: false,
-      };
-    });
+      function testStake(): void {
```
nventuro, line 463:
```suggestion
      function testStakeAndWrap(): void {
```

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
+          const relayerIsRecipient = TypesConverter.toAddress(tokenRecipient) === relayer.address;
+          expectTransferEvent(
+            receipt,
+            {
+              from: relayerIsRecipient ? ZERO_ADDRESS : relayer.address,
+              to: relayerIsRecipient ? relayer.address : TypesConverter.toAddress(tokenRecipient),
+              value: expectedWstETHAmount,
+            },
+            wstETH
+          );
```
nventuro, line 503:
```suggestion
```

Same as mentioned before

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
+          const receipt = await (
+            await relayer
+              .connect(sender)
+              .multicall([encodeStakeETHAndWrap(tokenRecipient, toChainedReference(0))], { value: amount })
+          ).wait();
+
+          expectTransferEvent(
+            receipt,
+            { from: ZERO_ADDRESS, to: relayer.address, value: expectedWstETHAmount },
+            wstETH
```
nventuro, line 523: Same re. we should test that the input was the chained reference instead of indirectly asserting that via the output

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
+              }),
+            ])
+          ).wait();
+
+          expectEvent.inIndirectReceipt(receipt, vault.instance.interface, 'Swap', {
+            poolId,
+            tokenIn: tokenIn.address,
+            tokenOut: tokenOut.address,
+            // amountIn: singleSwap.amount,
+            // amountOut
```
nventuro, line 593:
```suggestion
```

pkg/standalone-utils/test/LidoRelayer.test.ts
nventuro: True, but it requires manually granting allowance for the relayer (maybe #991 will improve this). The `wstETH` approval would be a one-off, and the joins would be easier. As long as we support both we'll be fine.

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
+              encodeUnwrap(relayer.address, recipient.address, toChainedReference(0)),
+            ])
+          ).wait();
+
+          expectEvent.inIndirectReceipt(receipt, vault.instance.interface, 'Swap', {
+            poolId,
+            tokenIn: tokenIn.address,
+            tokenOut: tokenOut.address,
+            // amountIn: singleSwap.amount,
+            // amountOut
```
nventuro, line 658:
```suggestion
```

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
+          });
 
-          const wstETHBalanceAfter = await wstETH.balanceOf(sender);
-          expect(wstETHBalanceAfter).to.be.eq(wstETHBalanceBefore);
+          expectEvent.inIndirectReceipt(
+            receipt,
+            new Interface((await getArtifact('v2-solidity-utils/ERC20')).abi),
+            'Transfer',
+            { to: recipient.address }
```
nventuro, line 600: Shouldn't we at least check that the transfer the recipient got was of the token out?

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
-    await WETH.mint(sender, fp(100));
-    await WETH.approve(vault.address, fp(100), { from: sender });
+    await stETH.mint(sender, fp(100));
+    await stETH.approve(vault.address, fp(100), { from: sender });
 
-    await WETH.mint(sender, fp(2500));
-    await WETH.approve(wstETH.address, fp(150), { from: sender });
+    await stETH.mint(sender, fp(2500));
+    await stETH.approve(wstETH.address, fp(150), { from: sender });
     await wstETH.instance.connect(sender).wrap(fp(150));
```
nventuro, line 69: Let's make this pool have a third token: it is very weird that we are doing 1) wrap steth into wsteth 2) swap wsteth for steth, making the tests harder to read. With a third token we'll always be able to reference either wsteth or steth, plus the neutral third one when required.

pkg/standalone-utils/test/LidoRelayer.test.ts
nventuro: Same in the other swap tests.

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
+                kind: SwapKind.GivenIn,
+                tokenIn,
+                tokenOut,
+                amount: toChainedReference(0),
+                sender: relayer,
+                recipient,
+                outputReference: 0,
+              }),
+            ])
+          ).wait();
```
nventuro, line 586: Since all of our tests either assert the end state or the receipt, we could have a `sharedBeforeEach` where we do the action and get the receipt, and then have simpler `it`s that simply make checks on those.

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
+            maxAmountsIn: params.maxAmountsIn,
+            userData: params.userData,
+            fromInternalBalance: false,
+          },
+          0,
+          params.outputReference ?? 0,
+        ]);
+      }
+
+      context('when the relayer is authorized', () => {
```
nventuro, line 874:
```suggestion
```

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
+                tokens.map((token) => (token === wstETH ? toChainedReference(0) : 0)),
+                0
+              ),
+            }),
+          ]);
 
           expectEvent.inIndirectReceipt(await receipt.wait(), vault.instance.interface, 'PoolBalanceChanged', {
             poolId: basePoolId,
-            liquidityProvider: sender.address,
+            liquidityProvider: relayer.address,
```
nventuro, line 896: Let's also check that the join did _something_, e.g. that the recipient got some BPT, or that at least one of the pool deltas was non zero.

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,462 +1,950 @@
...
           });
         });
 
-        it('does not send wstETH to the recipient', async () => {
+        it('does not take wstETH from the sender', async () => {
```
nventuro, line 900:
```suggestion
        it('does not take wstETH from the user', async () => {
```

We have been using 'user' to refer to the `sender` account (since the relayer is sometimes a sender in other contexts). I'd be fine with renaming `sender` and `recipient` to e.g. `senderUser` or `recipientUser`, for clarity.

pkg/standalone-utils/contracts/relayer/VaultActions.sol
```diff
@@ -46,7 +46,7 @@ abstract contract VaultActions is IBaseRelayerLibrary {
         uint256 value,
         uint256 outputReference
     ) external payable returns (uint256) {
-        require(funds.sender == msg.sender, "Incorrect sender");
+        require(funds.sender == msg.sender || funds.sender == address(this), "Incorrect sender");
```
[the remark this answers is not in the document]
nventuro: Created #992

pkg/standalone-utils/test/LidoRelayer.test.ts
TomAFrench: Ah, good catch. I just did a find-and-replace when giving a proper stETH implementation and didn't think about how we were using WETH in the pool as well so it would make the pool weird.

pkg/standalone-utils/contracts/relayer/LidoWrapping.sol
```diff
@@ -0,0 +1,151 @@
...
+
+        // The fallback function on the wstETH contract automatically stakes and wraps any ETH which is sent to it.
+        // We can then safely just send the ETH and just have to ensure that the call doesn't revert.
+        payable(address(_wstETH)).sendValue(amount);
+
+        // Note that the previous step would be dangerous should _wstETH be set to the zero address,
+        // however in this scenario the next line will revert, preventing loss of funds.
+
+        // As the wstETH contract doesn't return how much wstETH was minted we must query this separately.
+        uint256 result = _wstETH.getWstETHByStETH(amount);
```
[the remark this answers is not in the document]
nventuro: (I'm not 100% sure about this, it'd be great if you could double-check)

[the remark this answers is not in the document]
TomAFrench: I don't think that's right.

Staking ETH into stETH shouldn't change the exchange rate wstETH:stETH and neither should wrapping and unwrapping stETH (it would be a bad wrapper if doing this affected the exchange rate)

stETH is 1:1 with ETH so when I stake 10% of the current `totalPooledEther` I'm minted 10% of the current total supply of stETH (and so 10% of `_getTotalShares()`).

```
_ethAmount * (_getTotalShares() / totalPooledEther) -> _ethAmount * (1.1 *_getTotalShares()) / (1.1 * totalPooledEther)
```

The extra factors cancel and we have the same exchange rate as before.

TomAFrench:
Still doing a couple of modifications (stuff like fixing the confusing naming of sender, etc.)

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,461 +1,838 @@
...
+        });
+        testWrap();
+      });
 
-        await vault.authorizer
-          ?.connect(admin)
-          .grantRolesGlobally([manageUserBalanceAction, swapAction], relayer.address);
+      context('sender = relayer, recipient = relayer', () => {
+        beforeEach(async () => {
+          await stETH.transfer(relayer, fp(1), { from: senderUser });
```
nventuro, line 171:
```suggestion
          await stETH.transfer(relayer, amount, { from: senderUser });
```

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,461 +1,838 @@
...
+        });
+        testWrap();
       });
 
-      context('when the user did allow the relayer', () => {
-        sharedBeforeEach('allow relayer', async () => {
-          await vault.instance.connect(sender).setRelayerApproval(sender.address, relayer.address, true);
+      context('sender = relayer, recipient = senderUser', () => {
+        beforeEach(async () => {
+          await stETH.transfer(relayer, fp(1), { from: senderUser });
```
nventuro, line 180:
```suggestion
          await stETH.transfer(relayer, amount, { from: senderUser });
```

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,461 +1,838 @@
...
+          tokenRecipient = senderUser;
         });
+        testUnwrap();
       });
 
-      context('when the user did not allow the relayer', () => {
-        let singleSwap: SingleSwap;
+      context('sender = relayer, recipient = relayer', () => {
+        beforeEach(async () => {
+          await wstETH.transfer(relayer, fp(1), { from: senderUser });
```
nventuro, line 284:
```suggestion
          await wstETH.transfer(relayer, amount, { from: senderUser });
```

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,461 +1,838 @@
...
-            kind: SwapKind.GivenIn,
-            poolId: basePoolId,
-            assetIn: basePoolTokens.WETH.address,
-            assetOut: basePoolTokens.findBySymbol('wstETH').address,
-            amount: fp(1),
-            userData: '0x',
-          };
+      context('sender = relayer, recipient = senderUser', () => {
+        beforeEach(async () => {
+          await wstETH.transfer(relayer, fp(1), { from: senderUser });
```
nventuro, line 293:
```suggestion
          await wstETH.transfer(relayer, amount, { from: senderUser });
```

pkg/standalone-utils/test/LidoRelayer.test.ts
```diff
@@ -1,461 +1,838 @@
...
-              tokenOut: assets[swaps[0].assetOutIndex],
-              amountIn: swaps[0].amount,
-              // amountOut
-            });
-          });
+      function testStake(): void {
+        it('stakes with immediate amounts', async () => {
+          const receipt = await (
+            await relayer.connect(senderUser).multicall([encodeStakeETH(tokenRecipient, amount)], { value: amount })
+          ).wait();
```
nventuro, line 391: Slightly redundant, but it might not hurt to have an extra test here where `value > amount`, and we check that the excess ETH is returned.

nventuro (approved):
Fantastic work! I think I'll have nightmares about reviewing these tests for the next week :sweat_smile:

nventuro:
Addressing review comments myself as they are quite straightforward, in the interest of unblocking other tasks that depend on this PR.

PR #983, opened 2021-11-03
fix: load setupTests in @balancer-labs/v2-common

TomAFrench:
This ensures that VSCode knows about the contents of `setupTests.ts` to fix the warnings on usage of `sharedBeforeEach`


nventuro:
Topological dev ordering doesn't work as our dependency graph is cyclical.

TomAFrench:
Closing because this one-line fix is actually full of pain.

PR #987, opened 2021-11-05
Allow relayers to revoke approval from other relayers on behalf of user

TomAFrench:
I've added this for the situation in which we're performing a migration of the canonical relayer. In this case we'd ideally revoke approval for the previous relayer to minimise the attack surface so I've added an additional `relayer` parameter to `setRelayerApproval` to allow us to change the approval of relayers on the user's behalf.

Giving a relayer the ability to approval other relayers is very powerful so I've restricted this to only allow revocation of the approval of other addresses.


nventuro (approved):
This is a great idea!

>Giving a relayer the ability to approval other relayers is very powerful so I've restricted this to only allow revocation of the approval of other addresses.

:cheff's-kiss:

PR #989, opened 2021-11-08
feat: add relayer module for wrapping Aave tokens

TomAFrench:
This needs tests but I want to make this available for if we need to get something out for the frontend team to test with

### DIFF
--- a/pkg/standalone-utils/contracts/relayer/AaveWrapping.sol
+++ b/pkg/standalone-utils/contracts/relayer/AaveWrapping.sol
@@ -1,0 +1,89 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


pkg/standalone-utils/contracts/relayer/AaveWrapping.sol
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        // The unwrap caller is the implicit sender of tokens, so if the goal is for the tokens
+        // to be sourced from outside the relayer, we must first them pull them here.
+        if (sender != address(this)) {
+            require(sender == msg.sender, "Incorrect sender");
+            _pullToken(sender, staticToken, amount);
+        }
+
+        (, uint256 result) = staticToken.withdraw(recipient, amount, toUnderlying);
```
nventuro, line 83:
```suggestion
        // Static Tokens are outright burned by `withdraw`, so we don't need to grant allowance as there will be no transfer
        (, uint256 result) = staticToken.withdraw(recipient, amount, toUnderlying);
```

nventuro (approved):
Looks fantastic :ok_hand:  :ok_hand:

nventuro:
Force-pushed to rebase this branch on top of #982 being merged.

PR #990, opened 2021-11-09
New BasePool for improved protocol fees

nventuro:
This adds a new version of `BasePool` which has a `_payProtocolFees` function, minting BPT directly to the fee collector, and ignores the `dueProtocolFee` return value of the join/exit hooks. This results in better rates for LPs and liquidity staying in the Pool instead of being drained.

For now, we only intend to use this for Weighted Pools, so we're not entirely replacing the old `BasePool` but rather having two version side by side. Down the line, we're likely also going to add a `BasePremintedBPTPool`, but there's no benefit to doing that now.

In this PR, I renamed the old `BasePool` and related contracts to `LegacyBasePool` etc.: this resulted in new files being created and the changeset being very large, but since these files are copies of the old ones we don't really need to worry about reviewing them. The new functionality has been added to `BasePool`, `BaseWeightedPool`, etc., as that will make reviewing this PR easier. It may be that once this PR is approved we add a final commit undoing this renaming (perhaps calling the new version `BaseBPTFeesPool`?), but for now this makes sense.

---

Finally, the changes are not complete: some work still needs to be done. At the very least, `WeightedPool2Tokens` and its tests need updating to use the new fee computation and paying mechanisms.


pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -413,7 +413,7 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
         // If swaps are disabled, the only join kind that is allowed is the proportional one, as all others involve
         // implicit swaps and alter token prices.
         _require(
-            getSwapEnabled() || userData.joinKind() == JoinKind.ALL_TOKENS_IN_FOR_EXACT_BPT_OUT,
+            getSwapEnabled() || JoinKind(uint256(userData.joinKind())) == JoinKind.ALL_TOKENS_IN_FOR_EXACT_BPT_OUT,
```
EndymionJkb, line 416: We'll have to remember to put these back (temporary artifacts of the renaming)

EndymionJkb (changes requested):

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -27,6 +27,11 @@ contract WeightedPool is BaseWeightedPool {
 
     uint256 private immutable _totalTokens;
 
+    // Due protocol swap fee amounts are computed by measuring the growth of the invariant after a join or exit event
+    // and before the next one - the invariant's growth is due exclusively to swap fees. This avoids spending gas
+    // computing and tracking protocol fees on each individual swap.
```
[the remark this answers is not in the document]
nventuro: Good idea!

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -348,4 +343,55 @@ contract WeightedPool is BaseWeightedPool {
...
+
+    function _afterJoin(
+        uint256[] memory balances,
+        uint256[] memory amountsIn,
+        uint256[] memory normalizedWeights
+    ) internal override {
+        // Update the invariant with the balances the Pool will have after the join, in order to compute the
+        // protocol swap fee amounts due in future joins and exits.
+
+        for (uint256 i = 0; i < balances.length; ++i) {
```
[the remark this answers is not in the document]
nventuro: I think `_mutateAmounts` had at least one other use case: it seemed silly to keep it here just for this small thing.

pkg/pool-weighted/contracts/WeightedPool2Tokens.sol
```diff
@@ -620,17 +615,16 @@ contract WeightedPool2Tokens is
...
-        _downscaleDownArray(dueProtocolFeeAmounts);
 
         // Update cached total supply and invariant using the results after the exit that will be used for future
         // oracle updates, only if the pool was not paused (to minimize code paths taken while paused).
         if (_isNotPaused()) {
             _cacheInvariantAndSupply();
         }
 
-        return (amountsOut, dueProtocolFeeAmounts);
+        return (amountsOut, new uint256[](2));
```
nventuro, line 627:
```suggestion
        // Always return zero dueProtocolFeeAmounts to the Vault - protocol fees will be paid in BPT,
        // by minting directly to the protocolFeeCollector
        return (amountsOut, new uint256[](2));
```

gtaschuk:
Protocol fee is going to a vote friday and it seems like there's a good chance that it will pass

A couple questions about what would happen if it gets turned on:
1) Assuming we would turn on collection basically as soon as the proposal passes, so the `protocolFeeCollector` would start collecting individual tokens from existing pools, and bpt once we start deploying these?
2) Would we attempt to migrate liquidity from existing pools to the new pattern of fee collection?
3) Would we exclude the `protocolFeeCollector` from receiving LM rewards?  (short term, blocklist it, longterm, not stake it's bpt)
4) Any thoughts on how the `protocolFeeCollector` would choose when to exit pools?

pkg/pool-utils/contracts/BasePool.sol
```diff
@@ -36,6 +36,10 @@ import "./BasePoolAuthorization.sol";
  * @dev Reference implementation for the base layer of a Pool contract that manages a single Pool with optional
  * Asset Managers, an admin-controlled swap fee percentage, and an emergency pause mechanism.
  *
+ * This Pool pays protocol fees by minting BPT directly to the ProtocolFeeCollector instead of using the
+ * `dueProtocolFees` return value. This results in better rates for users, as the Pool won't unbalance itself right
+ * before they join/exit. Furthermore, the underlying tokens will continue to provide liquidity for traders.
```
dmf7z, line 41: > This results in better rates for users, as the Pool won't unbalance itself right before they join/exit. 
 
 This may be not clear if you do not know the previous implementation. "Why would it unbalance?" would be my question if I never read the LegacyBasePool.

dmf7z (approved):
Looks good!

pkg/pool-utils/contracts/BasePool.sol
EndymionJkb:
```suggestion
 * This Pool pays protocol fees by minting BPT directly to the ProtocolFeeCollector instead of using the
 * `dueProtocolFees` return value, which would cause the Vault to transfer pool tokens and unbalance the pool.
 * Paying in BPT results in better rates for users, as the Pool won't unbalance itself right before the join/exit.
 * Furthermore, the underlying tokens will continue to provide liquidity for traders.```

EndymionJkb: Suggested change 
 * before they join/exit. Furthermore, the underlying tokens will continue to provide liquidity for traders.
 * This Pool pays protocol fees by minting BPT directly to the ProtocolFeeCollector instead of using the
 * `dueProtocolFees` return value, which would cause the Vault to transfer pool tokens and unbalance the pool.
 * Paying in BPT results in better rates for users, as the Pool won't unbalance itself right before the join/exit.
 * Furthermore, the underlying tokens will continue to provide liquidity for traders.```

EndymionJkb (approved):

pkg/pool-utils/contracts/BasePool.sol
nventuro: Spot on comment, thank you. I've updated the wording.

nventuro:
This seems good to go, but let's hold off merging for a bit until we finalize compliation and deployment of linear and phantom stable pools.

nventuro:
I removed the changes to BaseWeightedPool and derived contracts - those will be coming in separate PRs as they are unrelated to this effort.

This means that for now no pool uses the new mechanism.

PR #993, opened 2021-11-09
PoC of the batch relayer deployment

nventuro:
This is first draft for us to build upon once pending batch relayer issues are addressed.

I created a new deployment task which deploys the relayer library, verifying the code for both library and relayer entrypoint. This uncovered an issue in our verification plugin, as the verified code [includes sources for pretty much all contracts in the repo, including Vault and Pools](https://kovan.etherscan.io/address/0x2fFB7B215Ae7F088eC2530C7aa8E1B24E398f26a#code).

Additionally, I added a fork test that tests an interaction of approvals, swaps and joins. Once #982 and #989 are merged, we should update these to also test the integration with the Lido and Aave wrappers.


nventuro:
Fixed! Our verified code is now much leaner.

nventuro:
I've updated the build info, abis, etc to reflect #982 and #989. I did not add any tests for these due to a small hiccup with hardhat: the WSTETH-WETH pool was created after the London hardfork, and our version of hardhat seems to be having trouble with type 2 transactions.

The verified code is now quite clean: see e.g. on the [relayer](https://kovan.etherscan.io/address/0x495F696430F4A51F7fcB98FbE68a9Cb7A07fB1bA#code) how it only has relevant source file (thanks #996!).

TomAFrench (approved):
Looks good to me :P

PR #996, opened 2021-11-10
Move userdata enums to libraries

nventuro:
This prevents from contracts using the enum importing the entire Pool code, which among other things results in huge code verification (see #993).


TomAFrench (approved):
LGTM

PR #998, opened 2021-11-12
Make LinearPool setTargets stricter

nventuro:
I also ended up reworking the tests to make it clearer what we're testing in each case, since the old tests did not differentiate being below/above only the old or new targets.


dmf7z (approved):
LGTM

PR #1000, opened 2021-11-16
Base pool controller

EndymionJkb:
Introduce a base pool controller with configurable rights (transfer ownership, set swap fees, update metadata).

Ownership transfer follows the safe two-step "Claimable" process; ownership cannot be renounced. Control of swap fees can be delegated.


nventuro:
PR #1000, nice!

pkg/solidity-utils/contracts/helpers/BalancerErrors.sol
```diff
@@ -156,8 +156,10 @@ library Errors {
     uint256 internal constant UNHANDLED_BY_MANAGED_POOL = 339;
     uint256 internal constant UNHANDLED_BY_PHANTOM_POOL = 340;
     uint256 internal constant TOKEN_DOES_NOT_HAVE_RATE_PROVIDER = 341;
-    uint256 internal constant INVALID_INITIALIZATION = 342;
+    uint256 internal constant INVALID_INITIALIZATION = 342;
```
TomAFrench, line 159:
```suggestion
    uint256 internal constant INVALID_INITIALIZATION = 342;
```

pkg/pool-utils/contracts/controllers/BasePoolController.sol
```diff
@@ -61,27 +192,74 @@ contract BasePoolController is IBasePoolController, Ownable {
...
+
+        _swapFeeController = newSwapFeeController;
+    }
+
+    /**
+     * @dev Pass a call to BasePool's setSwapFeePercentage through to the underlying pool, if allowed.
+     */
+    function setSwapFeePercentage(uint256 swapFeePercentage) external virtual override withBoundPool {
+        _require(canChangeSwapFee(), Errors.UNAUTHORIZED_OPERATION);
+        _require(getSwapFeeController() == msg.sender, Errors.SENDER_NOT_ALLOWED);
```
TomAFrench, line 242: Do we want to leave the option of setting the swap fee controller to `0xba1...ba1` to delegate it to governance? 

We'd need to have similar logic to check with the authorizer as in the pools in that case.

pkg/pool-utils/test/BasePoolController.test.ts
```diff
@@ -129,31 +185,122 @@ describe('BasePoolController', function () {
...
+      });
+    });
+
+    context('with update metadata set to false', () => {
+      sharedBeforeEach('deploy controller (update metadata false)', async () => {
+        await deployControllerAndPool(true, true, false);
+        await poolController.initialize(pool.address);
+      });
+
+      it('sets the permission to fals', async () => {
```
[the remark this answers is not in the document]
EndymionJkb: thank :)

pkg/pool-utils/contracts/controllers/BasePoolController.sol
EndymionJkb: I was thinking about that, but balked because it would need that extra logic. The idea was a project might contract with a provider themselves (vs piggy-backing on our current Gauntlet agreement, which has limited slots, etc.), and delegate to whatever actual address they use to control the fees, vs our governance placeholder address.

Open for discussion, though. I thought keeping it simple and generic would scale better. They can write a bot and delegate to that address, or still use a provider like Gauntlet - but the onus is on them to create the relationship.

TomAFrench: Yeah, I don't think that this would be a particularly commonly used feature.

If we exposed a `canPerform` function on the pools we could piggy back of that but seems like we'd have to have a whole separate mechanism (as there's no easy way to get the pool's factory address) which is definitely in "not worth it" territory.

TomAFrench: I was just making sure that we considered this first but I agree with your take.

TomAFrench (approved):

TomAFrench (approved):

PR #1001, opened 2021-11-16
Add LP allowlist to Managed Pools

EndymionJkb:


pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -175,6 +189,20 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
...
+     */
+    function getAllowlistLPs() public view returns (bool) {
+        return _getMiscData().decodeBool(_ALLOWLIST_LPS_OFFSET);
+    }
+
+    /**
+     * @dev Verifies that a given address is allowed to hold tokens.
+     */
+    function isAllowedAddress(address member) public view returns (bool) {
+        return false == getAllowlistLPs() || _allowedAddresses[member];
```
TomAFrench, line 203:
```suggestion
        return !getAllowlistLPs() || _allowedAddresses[member];
```

Checking equality with false feels weird to me.

EndymionJkb: I've gone back and forth on this. The reason I do this in most languages is clarity. The "!" is easy to miss if you're reading it fast.

Since it doesn't appear elsewhere in the codebase, I can follow the apparent convention and use "!", though spotting negated logic is a lot easier with an explicit "false".

I guess the other way would be to invert the function and say "hasPublicLPs" or something like that, but I think mentioning the allowlist is important.

EndymionJkb: I'll also change the names to make it clearer/consistent with future code in the next PR

TomAFrench (commented):
Do we want to allow the pool owner to transition the pool to remove the whitelist down the line?

EndymionJkb:
> Do we want to allow the pool owner to transition the pool to remove the whitelist down the line?

I was trying to keep it simple, but it's up for discussion. Unlike all the rest of the "rights," which can be confined to the controller, this one has to be in the pool itself, for access to the join hooks. The boolean flag is stored in mutable storage (again unlike the immutable controller rights). So it's *possible* to turn off the whitelist. Owners could implement a crude version of a supply cap by turning it off and on. And there is no way to "trap" LPs in the pool, since it only affects joins.

I could add an owner function to clear that flag, and then controllers could implement various strategies, from simply forwarding the call (no restrictions), to only allowing it to be cleared (not set), to accepting a timestamp and only allowing it to be cleared after a certain date, etc.

TomAFrench:
Yeah, that's what I was thinking. I wouldn't try and constrain it to only allow turning off the allowlist, etc. just leave that to controllers.

TomAFrench (approved):
I'm happy with this. We might want to add a test for allowlisting LPs to fail if `mustAllowlistLPs == false` for completeness though.

Issue #958, opened 2021-10-21
Release new version of WeightedPool with improvements


nventuro:
We've made a number of changes and are planning to have even more. Once they are all done, it'd be good to asses whether we want to release a new version of the factory.

The improvements are:
- oracle queries for uninitialized oracles (https://github.com/balancer-labs/balancer-v2-monorepo/pull/770)
- manual oracle init (https://github.com/balancer-labs/balancer-v2-monorepo/pull/772)
- up to 20 tokens (https://github.com/balancer-labs/balancer-v2-monorepo/pull/723)
- reduced bytecode thanks to library (https://github.com/balancer-labs/balancer-v2-monorepo/pull/749)
- autostaking
- protocol fee changes (to be paid in bpt)
- simplified math for 50/50 pools

EndymionJkb:
This tracks a new factory release that was never finalized; no new factory versions will ship on frozen V2. Closing.

PR #1003, opened 2021-11-16
FixedPoint: improve pow to save gas

dmf7z:
This PR reduces gas cost when computing power of 1, 2 or 4 by avoiding fractional exponentiation. 
For **50/50 weighted pool it saves 3.5K of gas on each swap**. 
It also reduces the same amount of gas for half of  80/20 weighted pool swaps. 
And swaps for some other pools depending on their weights.

### DIFF
--- a/pkg/solidity-utils/contracts/math/FixedPoint.sol
+++ b/pkg/solidity-utils/contracts/math/FixedPoint.sol
@@ -21,6 +21,8 @@ import "../helpers/BalancerErrors.sol";
 
 library FixedPoint {
     uint256 internal constant ONE = 1e18; // 18 decimal places
+    uint256 internal constant TWO = 2 * ONE;
+    uint256 internal constant FOUR = 4 * ONE;
     uint256 internal constant MAX_POW_RELATIVE_ERROR = 10000; // 10^(-14)
 
     // Minimum base for the power function when the exponent is 'free' (larger than ONE).
@@ -103,13 +105,21 @@ library FixedPoint {
      * the true value (that is, the error function expected - actual is always positive).
      */
     function powDown(uint256 x, uint256 y) internal pure returns (uint256) {
-        uint256 raw = LogExpMath.pow(x, y);
-        uint256 maxError = add(mulUp(raw, MAX_POW_RELATIVE_ERROR), 1);
-
-        if (raw < maxError) {
-            return 0;
+        if (y == ONE) {
+            return x;
+        } else if (y == TWO) {
+            return mulDown(x, x);
+        } else if (y == FOUR) {
+            return mulDown(x, mulDown(x, mulDown(x, x)));
         } else {
-            return sub(raw, maxError);
+            uint256 raw = LogExpMath.pow(x, y);
+            uint256 maxError = add(mulUp(raw, MAX_POW_RELATIVE_ERROR), 1);
+
+            if (raw < maxError) {
+                return 0;
+            } else {
+                return sub(raw, maxError);
+            }
         }
     }
 
@@ -118,10 +128,18 @@ library FixedPoint {
      * the true value (that is, the error function expected - actual is always negative).
      */
     function powUp(uint256 x, uint256 y) internal pure returns (uint256) {
-        uint256 raw = LogExpMath.pow(x, y);
-        uint256 maxError = add(mulUp(raw, MAX_POW_RELATIVE_ERROR), 1);
+        if (y == ONE) {
+            return x;
+        } else if (y == TWO) {
+            return mulUp(x, x);
+        } else if (y == FOUR) {
+            return mulUp(x, mulUp(x, mulUp(x, x)));
+        } else {
+            uint256 raw = LogExpMath.pow(x, y);
+            uint256 maxError = add(mulUp(raw, MAX_POW_RELATIVE_ERROR), 1);
 
-        return add(raw, maxError);
+            return add(raw, maxError);
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


pkg/solidity-utils/contracts/math/FixedPoint.sol
```diff
@@ -103,13 +105,22 @@ library FixedPoint {
      * the true value (that is, the error function expected - actual is always positive).
      */
     function powDown(uint256 x, uint256 y) internal pure returns (uint256) {
-        uint256 raw = LogExpMath.pow(x, y);
-        uint256 maxError = add(mulUp(raw, MAX_POW_RELATIVE_ERROR), 1);
-
-        if (raw < maxError) {
-            return 0;
+        if (y == ONE) {
```
nventuro, line 108:
```suggestion
        // Optimize for when y equals 1.0, 2.0 or 4.0, as those are very simple to implement and occur often in 50/50 and
        // 80/20 Weighted Pools
        if (y == ONE) {
```

pkg/solidity-utils/contracts/math/FixedPoint.sol
```diff
@@ -118,10 +129,19 @@ library FixedPoint {
      * the true value (that is, the error function expected - actual is always negative).
      */
     function powUp(uint256 x, uint256 y) internal pure returns (uint256) {
-        uint256 raw = LogExpMath.pow(x, y);
-        uint256 maxError = add(mulUp(raw, MAX_POW_RELATIVE_ERROR), 1);
+        if (y == ONE) {
```
nventuro, line 132:
```suggestion
        // Optimize for when y equals 1.0, 2.0 or 4.0, as those are very simple to implement and occur often in 50/50 and
        // 80/20 Weighted Pools
        if (y == ONE) {
```

pkg/solidity-utils/test/FixedPoint.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  const checkPow = async (x: number, pow: number) => {
+    const result = fp(x ** pow);
+    expect(await lib.powDown(fp(x), fp(pow))).to.equal(result);
+    expect(await lib.powUp(fp(x), fp(pow))).to.equal(result);
+  };
+
+  const checkPows = async (pow: number) => {
+    it('handles small numbers', async () => {
+      await checkPow(0.0007, pow);
```
nventuro, line 22:
```suggestion
      await checkPow(0.00000007, pow);
```

pkg/solidity-utils/test/FixedPoint.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    });
+
+    it('handles medium numbers', async () => {
+      await checkPow(15, pow);
+    });
+
+    it('handles big numbers', async () => {
+      await checkPow(158300, pow);
```
nventuro, line 30:
```suggestion
      await checkPow(1583156787139, pow);
```

nventuro (approved):
So silly that we never implemented this before :facepalm: Another think to consider for #958

pkg/solidity-utils/test/FixedPoint.test.ts
dmf7z: This one fails because it loses precision on mulUp when dividing.

pkg/solidity-utils/test/FixedPoint.test.ts
dmf7z: This one also fails because the result is less than 1e-18

pkg/solidity-utils/test/FixedPoint.test.ts
```diff
@@ -13,8 +12,8 @@ describe('FixedPoint', () => {
 
   const checkPow = async (x: number, pow: number) => {
     const result = fp(x ** pow);
-    expect(await lib.powDown(fp(x), fp(pow))).to.equal(result);
-    expect(await lib.powUp(fp(x), fp(pow))).to.equal(result);
+    expectEqualWithError(await lib.powDown(fp(x), fp(pow)), result, 0.00000001);
```
nventuro, line 15: FixedPoint.pow is supposed to guarantee a max relative error of 10^-14, so we should be able to use that here.

nventuro:
Everything seems spot on, and the implementation is so trivial that it must be correct. But I'd sleep a bit better if the new tests were more extensive. Can you look into the `WeighteOracledMath` tests and imitating how those test a large range of numbers accross different orders of magnitude?

nventuro (approved):

DanielSaska:
@nventuro I was just going over the implementation of `powUp()` and it may be worth noting that while this change is mathematically correct (from what I can see), it is not numerically consistent with the prior behaviour: before the change, `maxError` would always be added to the computed `pow()` which in the case of ^1 would still be non-zero (`=MAX_POW_RELATIVE_ERROR+1`) making the returned value different from the optimized version of `powUp()` which returns the exact value. `powDown()` seems to be similarly affected.

That said, the difference in the behaviour is very small and in essence, the updated implementation is more accurate (in the special cases). I am not fully familiar with the project so it is up to you to determine whether this is an issue depending on how the updated source will be deployed along with existing WeightedPools on-chain.

nventuro:
Thanks for the thorough report @DanielSaska!

The reason why both `powUp()` and `powDown()` exist is because the underlying `pow()` function has some error. Some error is fine and tolerable, but we want to make sure that these errors consistently benefit the Pool to the detriment of the actor that is interacting with it. For example, a Pool will quote a slightly worse rate compared to one with zero numerical error: otherwise, an attacker might be able to extract profit from accumulated errors.

For `powDown()` and `powUp()`, we add and subtract amounts large enough that we can guarantee the computed values always lie below or above the true errorless value. As you noted, these optimizations return values with no error (or very little, such as when computing squares). This is fine: the important property that must hold is not some minimum error, but rather the _sign_ of the error.

Hope that helps clarify things!

PR #1004, opened 2021-11-16
MultiDistributor: consolidate transfers of claimed tokens

TomAFrench:
This PR brings back the token transfer consolidation mechanism from #827 to avoid repeated token transfers when claiming multiple distributions. Due to fact that we'll have more onchain users for the `MultiDistributor` contract compared to the `MerkleOrchard` where most users are EOAs I think benefits of the simpler interface outweigh the benefits from having a separate `tokens` array as in #894.


EndymionJkb (approved):
Maybe add a specific test for calling the version of claim() that has multiple distributionIds? Check that you get the total balances and right number of events? (Unless there is one and I'm not seeing it.)

PR #1012, opened 2021-11-17
Linear: Update documentation

EndymionJkb:


pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -29,8 +29,10 @@ import "./LinearMath.sol";
 import "./LinearPoolUserDataHelpers.sol";
 
 /**
- * @dev LinearPool suitable for assets with an equal underlying token with an exact and non-manipulable exchange rate.
- * Requires an external feed of these exchange rates.
+ * @dev Linear Pools are designed to hold two assets: "main" and "wrapped" tokens that have an equal value underlying
+ * token (e.g., DAI and aDAI).
```
nventuro, line 33:
```suggestion
 * @dev Linear Pools are designed to hold two assets: "main" and "wrapped" tokens that have an equal value underlying
 * token (e.g., DAI and aDAI). The Pool will register three tokens in the Vault however: the two assets and the BPT
 * itself, so that BPT can be exchanged (effectively joining and exiting) via swaps.
 *
 * Unlike most other Pools, this one does not attempt to create revenue by charging fees: value is derived by holding
 * the wrapped, yield-bearing asset.
```

pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -154,10 +156,10 @@ contract LinearPool is BasePool, IGeneralPool, LinearMath, IRateProvider {
     /**
      * @dev Finishes initialization of the Linear Pool: it is unusable before calling this function.
      *
-     * Since Linear Pools have preminted BPT which is stored in the Vault, we need for an initial join to occur for the
-     * BPT to be deposited. Unfortunately, this cannot be performed during construction as a join involves calling a
-     * callback function on the Pool, and the Pool will not have any code until construction finishes. Therefore, this
-     * must happen in a separate call to this function.
+     * Since Linear Pools have preminted BPT stored in the Vault, they require an initial join to deposit that BPT.
+     * Unfortunately, this cannot be performed during construction, as a join involves a callback function on the
+     * Pool: and the Pool will not have any code until construction finishes. Therefore, this must happen in a
```
nventuro, line 161:
```suggestion
     * Pool and the Pool will not have any code until construction finishes. Therefore, this must happen in a
```

pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -420,10 +426,11 @@ contract LinearPool is BasePool, IGeneralPool, LinearMath, IRateProvider {
         // triggered. This allows for a simple and safe way to exit the Pool.
         if (kind == ExitKind.EXACT_BPT_IN_FOR_TOKENS_OUT) {
             _ensurePaused();
-            // Note that this will cause for the user's BPT to be burned, which is not something that happens during
+            // Note that this will cause the user's BPT to be burned, which is not something that happens during
             // regular operation of this Pool, and may lead to accounting errors. Because of this, it is highly
             // advisable to not continue using a Pool on which the pause has been turned on and BPT burned once the
             // pause window expires.
+            // advisable to stop using a Pool after it is paused and the pause window expires.
```
nventuro, line 433: Repeated sentence?

nventuro (commented):
Thank you very much for going over this!

pkg/pool-linear/contracts/LinearPool.sol
EndymionJkb: I debated including that... ok; will add (as commit, so lint doesn't get upset)

pkg/pool-linear/contracts/LinearPool.sol
EndymionJkb: Yes - fixed in next commit. Too much copy/pasting to split this into so many PRs

PR #1017, opened 2021-11-17
Linear: Move ExitKind to library, for consistency with other pools

EndymionJkb:


nventuro (approved):
Thanks a lot for being on top of this!

PR #1018, opened 2021-11-17
Linear: Performance/bytecount

EndymionJkb:


nventuro (approved):
Great, thanks! This reduces bytecode as we don't have to embed the constants _mainToken and _wrappedToken constants in the code.

PR #1019, opened 2021-11-17
MultiDistributor: update staking interface

### LINKED ISSUE #1008: Simplify staking interface
We currently have `stake`, `stakeFor` and `stakeWithPermit`. Similarly we have a bunch of `claim` and `exit` functions. It'd be good to reduce the number of these: at the very least we can merge `stake` and `stakeFor`.

TomAFrench:
I've added `sender` and `recipient` arguments for `stake`, `unstake` and `claim` functions on the `Multidistributor` which allow redirecting the output to a different address. Tests for `unstake` and `claim` have been updated to use the same pattern used by `stake` for testing this.

I haven't made the same change for `exit` as `exitWithCallback` currently has the tokens being split (staked to msg.sender, rewards to callback). I'm not sure if we want to change behaviour to sending staked BPT to the callback contract (adds complexity and scope for locked tokens if a callback doesn't expect this) or complicating the interface by allowing two types of "recipient". I don't think it would be the end of the world to keep this function simpler (as a relayer which wants to redirect output could make `unstake` and `claim` calls separately) but I'm open to opinions.

fixes #1008

### DIFF
--- a/pkg/distributors/contracts/MultiDistributor.sol
+++ b/pkg/distributors/contracts/MultiDistributor.sol
@@ -354,23 +354,17 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
      * @dev Stakes tokens
      * @param stakingToken The token to be staked to be eligible for distributions
      * @param amount Amount of tokens to be staked
+     * @param sender The address which provides tokens to stake
+     * @param recipient The address which receives the staked tokens
      */
-    function stake(IERC20 stakingToken, uint256 amount) external override nonReentrant {
-        _stakeFor(stakingToken, amount, msg.sender, msg.sender);
-    }
-
-    /**
-     * @notice Stakes tokens on behalf of other user
-     * @param stakingToken The token to be staked to be eligible for distributions
-     * @param amount Amount of tokens to be staked
-     * @param user The user staking on behalf of
-     */
-    function stakeFor(
+    function stake(
         IERC20 stakingToken,
         uint256 amount,
-        address user
+        address sender,
+        address recipient
     ) external override nonReentrant {
-        _stakeFor(stakingToken, amount, user, msg.sender);
+        require(sender == msg.sender, "INVALID_SENDER");
+        _stake(stakingToken, amount, sender, recipient);
     }
 
     /**
@@ -393,7 +387,7 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
         bytes32 s
     ) external override nonReentrant {
         IERC20Permit(address(stakingToken)).permit(user, address(this), amount, deadline, v, r, s);
-        _stakeFor(stakingToken, amount, user, user);
+        _stake(stakingToken, amount, user, user);
     }
 
     /**
@@ -501,18 +495,18 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
         callbackContract.distributorCallback(callbackData);
     }
 
-    function _stakeFor(
+    function _stake(
         IERC20 stakingToken,
         uint256 amount,
-        address user,
-        address from
+        address sender,
+        address recipient
     ) internal {
         require(amount > 0, "STAKE_AMOUNT_ZERO");
 
         // Before we increase the user's staked balance we need to update all of their subscriptions
-        _updateSubscribedDistributions(stakingToken, user);
+        _updateSubscribedDistributions(stakingToken, recipient);
 
-        UserStaking storage userStaking = _userStakings[stakingToken][user];
+        UserStaking storage userStaking = _userStakings[stakingToken][recipient];
         userStaking.balance = userStaking.balance.add(amount);
 
         EnumerableSet.Bytes32Set storage distributions = userStaking.subscribedDistributions;
@@ -524,12 +518,12 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
             bytes32 distributionId = distributions.unchecked_at(i);
             Distribution storage distribution = _getDistribution(distributionId);
             distribution.totalSupply = distribution.totalSupply.add(amount);
-            emit Staked(distributionId, user, amount);
+            emit Staked(distributionId, recipient, amount);
         }
 
         // We hold stakingTokens in an external balance as BPT needs to be external anyway
         // in the case where a user is exiting the pool after unstaking.
-        stakingToken.safeTransferFrom(from, address(this), amount);
+        stakingToken.safeTransferFrom(sender, address(this), amount);
     }
 
     function _claim(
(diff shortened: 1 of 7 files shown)


pkg/distributors/test/MultiDistributor.test.ts
```diff
@@ -885,359 +903,391 @@ describe('MultiDistributor', () => {
...
+    const itHandlesUnstaking = (unstake: (token: Token, amount: BigNumberish) => Promise<ContractTransaction>) => {
+      context('when the user did specify some amount', () => {
+        const amount = fp(1);
 
-          await distributor.stake(stakingToken, amount, { from: user1 });
-        });
+        context('when the user has previously staked the requested balance', () => {
+          sharedBeforeEach('stake amount', async () => {
+            await stakingTokens.mint({ to: from, amount });
+            await stakingTokens.approve({ to: distributor, amount, from: from });
```
nventuro, line 913:
```suggestion
            await stakingTokens.approve({ to: distributor, amount, from });
```

pkg/distributors/test/MultiDistributor.test.ts
```diff
@@ -885,359 +903,391 @@ describe('MultiDistributor', () => {
...
 
-        const itTransfersTheStakingTokensToTheUser = () => {
-          it('transfers the staking tokens to the user', async () => {
-            await expectBalanceChange(() => distributor.unstake(stakingToken, amount, { from: user1 }), stakingTokens, [
-              { account: user1, changes: { [stakingToken.symbol]: amount } },
-              { account: distributor.address, changes: { [stakingToken.symbol]: amount.mul(-1) } },
-            ]);
+            await distributor.stake(stakingToken, amount, from, from, { from });
```
nventuro, line 915: from from (from). from? from: from

pkg/distributors/test/MultiDistributor.test.ts
```diff
@@ -885,359 +903,391 @@ describe('MultiDistributor', () => {
...
 
         it('reverts', async () => {
-          await expect(distributor.unstake(stakingToken, amount, { from: user1 })).to.be.revertedWith(
-            'UNSTAKE_AMOUNT_UNAVAILABLE'
+          await expect(unstake(stakingToken, amount)).to.be.revertedWith('UNSTAKE_AMOUNT_ZERO');
+        });
+      });
+    };
+
+    describe('unstake', () => {
```
nventuro, line 1260: I'd move this before `itHnadlesUnstaking`: given how complex it is, reading how it is used before going over it seems easier. I was a bit puzzled as to what the `unstake` function meant.

pkg/distributors/test/MultiDistributor.test.ts
```diff
@@ -885,359 +903,391 @@ describe('MultiDistributor', () => {
...
+            from = user1;
+            to = user1;
+          });
 
-      it('reverts', async () => {
-        await expect(distributor.unstake(stakingToken, amount, { from: user1 })).to.be.revertedWith(
-          'UNSTAKE_AMOUNT_ZERO'
-        );
+          itHandlesUnstaking((token: Token, amount: BigNumberish) =>
+            distributor.unstake(token, amount, from, from, { from })
```
nventuro, line 1277:
```suggestion
            distributor.unstake(token, amount, from, to, { from })
```

pkg/distributors/test/MultiDistributor.test.ts
```diff
@@ -2108,114 +2163,146 @@ describe('MultiDistributor', () => {
...
+
+      context("when caller is authorised to act on sender's behalf", () => {
+        context('when sender and recipient are the same', () => {
+          sharedBeforeEach('define sender and recipient', async () => {
+            from = user1;
+            to = user1;
           });
 
-          itIgnoresTheRequest(true);
+          itHandlesClaiming((distribution: string) => distributor.claim(distribution, false, from, from, { from }));
```
nventuro, line 2296:
```suggestion
          itHandlesClaiming((distribution: string) => distributor.claim(distribution, false, from, to, { from }));
```

pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -393,26 +387,29 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
         bytes32 s
     ) external override nonReentrant {
         IERC20Permit(address(stakingToken)).permit(user, address(this), amount, deadline, v, r, s);
-        _stakeFor(stakingToken, amount, user, user);
+        _stake(stakingToken, amount, user, user);
```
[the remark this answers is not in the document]
TomAFrench: Yes, that's right. The only real situation to sign an permit sig for this contract is for staking so we can reasonably safely assume that this is an action which the user wants to take.

We could add an extra signature here to confirm this but we'd then have to have the user sign two messages.

pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -494,25 +500,25 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
...
         IERC20 stakingToken,
         uint256 amount,
-        address user,
-        address from
+        address sender,
+        address recipient
     ) internal {
         require(amount > 0, "STAKE_AMOUNT_ZERO");
 
         // Before we increase the user's staked balance we need to update all of their subscriptions
```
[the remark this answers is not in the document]
TomAFrench: Fixed. I've left the final reference to "user" as it's speaking more generally rather than specific to the users involved in the current staking action.

pkg/distributors/test/MultiDistributor.test.ts
TomAFrench: I'm not quite sure what you'd prefer here as while it's very frommy it's correct. :laughing:

I can stake some tokens for `from` from `other` to potentially make it clearer that we're just getting "some staked tokens" and it doesn't really matter where from. Is that your concern or just checking based on the other `to/from` issues?

pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -426,40 +423,49 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
...
     }
 
     /**
      * @dev Claims earned tokens for a list of distributions to a callback contract
      * @param distributionIds List of distributions to claim
+     * @param sender The address which earned the tokens being claimed
      * @param callbackContract The contract where tokens will be transferred
      * @param callbackData The data that is used to call the callback contract's 'callback' method
      */
     function claimWithCallback(
```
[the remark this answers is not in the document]
TomAFrench: I'm also not really sold on callbacks on this. I think we can review this after this PR however. https://github.com/balancer-labs/balancer-v2-monorepo/issues/1036

pkg/distributors/test/MultiDistributor.test.ts
nventuro: Nah I think this is fine, the line just struck me as funny :joy:

pkg/distributors/contracts/MultiDistributor.sol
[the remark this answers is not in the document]
nventuro: No, two sigs would be overkill. I was thinking that relying on another contract for security seemed a bit odd, but `stakingToken` will be BPT in 99% of the cases, so this should be fine.

nventuro (approved):

PR #1020, opened 2021-11-18
MultiDistributor: gas golfing

### LINKED ISSUE #1013: Improve MultiDistributor storage accesses
Most of our helper functions recompute data that in some context we already have available, wasting gas. See for exaple [this review comment](https://github.com/balancer-labs/balancer-v2-monorepo/pull/968#discussion_r751423194). We should look into these and optimize for gas, specially in `stake`, `unstake` and `claim`.

TomAFrench:
(I've actually got some more changes locally but probably best to wait until we have the benchmarks in master before looking at this.)

fixes #1013 

This PR:
- Moves safety checks on `_unaccountedUnclaimedTokens` to external getters (This is safe as we only call it on subscribed distributions)
- Passes the new `globalTokensPerStake` to the `_unaccountedUnclaimedTokens` call to avoid recalculation
- Switches `_updateDistributionRate` to take a storage pointer rather than looking up the distribution again from its id.

### DIFF
--- a/pkg/distributors/contracts/MultiDistributor.sol
+++ b/pkg/distributors/contracts/MultiDistributor.sol
@@ -159,7 +159,7 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
         UserStaking storage userStaking = _userStakings[distribution.stakingToken][user];
         UserDistribution storage userDistribution = userStaking.distributions[distributionId];
 
-        // If the user is not subscribed to the queried distribution, doesn't have any unaccounted for tokens.
+        // If the user is not subscribed to the queried distribution, they don't have any unaccounted for tokens.
         // Then we can just return the stored number of tokens which the user can claim.
         if (!userStaking.subscribedDistributions.contains(distributionId)) {
             return userDistribution.unclaimedTokens;
--- a/pkg/distributors/contracts/MultiDistributor.sol
+++ b/pkg/distributors/contracts/MultiDistributor.sol
@@ -236,7 +236,7 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
 
         // Before receiving the tokens, we must update the distribution's rate
         // as we are about to change its payment rate, which affects all other rates.
-        _updateDistributionRate(distributionId);
+        _updateDistributionRate(distribution);
 
         // Get the tokens and deposit them in the Vault as this contract's internal balance, making claims to internal
         // balance, joining pools, etc., use less gas.
@@ -304,7 +304,7 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
                 // staked tokens and decrease the global per token rate.
                 // The unclaimed tokens remain unchanged as the user was not subscribed to the distribution
                 // and therefore not eligible to receive any unaccounted-for tokens.
-                userStaking.distributions[distributionId].userTokensPerStake = _updateDistributionRate(distributionId);
+                userStaking.distributions[distributionId].userTokensPerStake = _updateDistributionRate(distribution);
                 distribution.totalSupply = distribution.totalSupply.add(amount);
                 emit Staked(distributionId, msg.sender, amount);
             }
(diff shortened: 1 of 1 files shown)


pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -174,9 +182,16 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
...
      */
     function getClaimableTokens(bytes32 distributionId, address user) external view override returns (uint256) {
-        IERC20 stakingToken = _getDistribution(distributionId).stakingToken;
-        UserStaking storage userStaking = _userStakings[stakingToken][user];
-        return _getUnclaimedTokens(userStaking, distributionId);
+        Distribution storage distribution = _getDistribution(distributionId);
+        UserStaking storage userStaking = _userStakings[distribution.stakingToken][user];
+        UserDistribution storage userDistribution = userStaking.distributions[distributionId];
+
+        // If the user is not subscribed to the queried distribution, doesn't have any unaccounted for tokens.
```
nventuro, line 189:
```suggestion
        // If the user is not subscribed to the queried distribution, they don't have any unaccounted for tokens.
```

pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -409,13 +424,14 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
...
+
         // Before we reduce the user's staked balance we need to update all of their subscriptions
-        _updateSubscribedDistributions(stakingToken, msg.sender);
+        _updateSubscribedDistributions(userStaking);
 
-        UserStaking storage userStaking = _userStakings[stakingToken][msg.sender];
         uint256 currentBalance = userStaking.balance;
         require(currentBalance >= amount, "UNSTAKE_AMOUNT_UNAVAILABLE");
-        userStaking.balance = userStaking.balance.sub(amount);
+        userStaking.balance -= currentBalance;
```
nventuro, line 434:
```suggestion
        userStaking.balance -= amount;
```

nventuro (commented):
Nice, thanks! It'd be good to know what the gas differences are once you have the benchmarks in place.

pkg/distributors/contracts/MultiDistributor.sol
TomAFrench: :grimacing: yeah that's a doozy.

TomAFrench:
Sneak peek of pre vs post golfing (including consolidating transfers)
image

TomAFrench:
@nventuro This should be good to go. There's definitely more opportunities for gas optimisations (main one being storage slot packing in the structs) but that's impacted by other PRs like #1025 so I'm leaving it out for now.

pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -654,41 +673,39 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
     }
 
     /**
-     * @dev Returns the total unclaimed tokens for a user for a particular distribution
+     * @notice Returns the total unclaimed tokens for a user for a particular distribution
+     * @dev Only returns correct results when called on a distribution which the user is subscribed to
```
EndymionJkb, line 677:
```suggestion
     * @dev Only returns correct results when the user is subscribed to the distribution
```

pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -654,41 +673,39 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
...
+                unclaimedTokens
+            );
     }
 
     /**
-     * @dev Returns the tokens earned for a particular distribution between
-     *      the last time the user updated their position and now
+     * @notice Returns the tokens earned for a particular distribution between
+     *         the last time the user updated their position and now
+     * @dev Only returns correct results when called on a distribution which the user is subscribed to
```
EndymionJkb, line 697:
```suggestion
     * @dev Only returns correct results when the user is subscribed to the distribution
```

EndymionJkb (approved):
Just reworded a couple comments

pkg/distributors/contracts/MultiDistributor.sol
TomAFrench: Thanks!

TomAFrench:
Failing test is from unrelated code.

PR #1021, opened 2021-11-18
MultiDistributor: use relayer powers to reuse Vault approval for staking

### LINKED ISSUE #1009: Stake tokens via Vault approval
This will mean that users only ever grant approval to the Vault, will let relayers use this approval, and will mean that new pools (which have infinite Vault allowance) won't even need approval.

TomAFrench:
We now use a `TRANSFER_EXTERNAL` for staking users tokens so we can take advantage of BPTs (in future) having an approval set on the vault by standard.

I don't think it makes sense to do the same with funding distributions. Team multisigs are less likely to have a preexisting token approval on the vault so we go from 2 txs to 3, (including a mysterious `setRelayerApproval` call) which just increases friction.

`stakeWithPermit` is starting to feel a bit out of place here - kinda weird for it to be processing a permit for the vault. I've left it here for now but it'll probably get the chop when we allow relayers/pools to stake on your behalf.

fixes #1009


pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -529,7 +529,15 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
...
-        stakingToken.safeTransferFrom(from, address(this), amount);
+        IVault.UserBalanceOp[] memory ops = new IVault.UserBalanceOp[](1);
+        ops[0] = IVault.UserBalanceOp({
+            asset: IAsset(address(stakingToken)),
+            amount: amount,
+            sender: from,
+            recipient: payable(address(this)),
+            kind: IVault.UserBalanceOpKind.TRANSFER_EXTERNAL
+        });
+        getVault().manageUserBalance(ops);
```
nventuro, line 540: God everytime I see this abomination that we need to put up with in order to create a fixed size array I want to scream until the eardrums of the solc dev team burst

nventuro: But uh, yea, good idea re. transferring assets via the Vault! 

Something to keep in mind however is that this will cause issues with autostaking, as the Pool won't be able to stake during a join because of the Vault reentrancy. Maybe we do need two ways to stake after all if we want to support the autostake scenario.

pkg/distributors/test/MultiDistributor.test.ts
```diff
@@ -32,7 +34,16 @@ describe('MultiDistributor', () => {
...
+    vault = await Vault.create();
+    distributor = await MultiDistributor.create(vault);
+
+    // Authorise distributor to use users' vault token approvals
+    const action = await actionId(vault.instance, 'manageUserBalance');
+    await vault.grantRoleGlobally(action, distributor);
+    await vault.setRelayerApproval(user1, distributor, true);
+    await vault.setRelayerApproval(user2, distributor, true);
+    await vault.setRelayerApproval(user3, distributor, true);
+    await vault.setRelayerApproval(other, distributor, true);
```
nventuro, line 46: Coupled with what I mentioned above, I am now unsure about this feature. Maybe #1009 is not a great idea after all? There, I mention something about relayers, but that comment doesn't quite make sense and just seems incorrect.

TomAFrench: #1009 makes sense to me? BPT will always have an existing approval on the vault but in order to use this we need the staking contract to be a relayer so it can make use of this.

Users are then required to approve the staking contract as a relayer but this is a one-off action which will avoid them having to do a similar approval for each token.

This seems sound to me. But agreed, we'll have to open a backdoor for token contracts to autostake.

pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -392,7 +392,7 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
         bytes32 r,
         bytes32 s
     ) external override nonReentrant {
-        IERC20Permit(address(stakingToken)).permit(user, address(this), amount, deadline, v, r, s);
+        IERC20Permit(address(stakingToken)).permit(user, address(getVault()), amount, deadline, v, r, s);
```
TomAFrench, line 395: Just realised that this opens up a troll vector. If people want to give approvals to the vault for other reasons then someone can decide to stake those tokens for them instead.

We should switch this over to using the method to be created for autostaking as then we can rely on all approvals to represent a desire to stake those tokens.

TomAFrench:
I've returned `stake` to using standard token approvals and created `stakeUsingVault` for use as a relayer. Also created #1038 to consider specialised functions for different variants of the staking interface.

pkg/distributors/test/MultiDistributor.test.ts
```diff
@@ -872,6 +889,40 @@ describe('MultiDistributor', () => {
...
+
+      context("when caller is authorised to act on sender's behalf", () => {
+        context('when sender and recipient are the same', () => {
+          sharedBeforeEach('define sender and recipient', async () => {
+            from = user1;
+            to = user1;
+          });
+
+          itHandlesStaking((token: Token, amount: BigNumberish) =>
+            distributor.stakeUsingVault(token, amount, from, from, { from })
```
nventuro, line 909:
```suggestion
            distributor.stakeUsingVault(token, amount, from, to, { from })
```

nventuro (approved):

pkg/distributors/test/MultiDistributor.test.ts
TomAFrench: :man_facepalming:

PR #1022, opened 2021-11-18
MultiDistributor: remove lowlevel getters

### LINKED ISSUE #1016: Hide low-level MultiDistributor getters
As mentioned [here](https://github.com/balancer-labs/balancer-v2-monorepo/pull/968#discussion_r751262775), we're currently likely exposing an undue amount of low-level detail that is not necessary to operate with the contract, leading to a complex interface, extra code and arguably a system that is harder to understand. I propose we get rid of these non-essential getters.

TomAFrench:
fixes #1016 

I've removed `unaccountedUnclaimedTokens` and `lastTimePaymentApplicable` as mentioned in the issue. 

`globalTokensPerStake` stands out as something that's unnecessary for users however we're using it in testing where it's definitely useful (to check that the saved value is correctly updated). We could have a `TestMultiDistributor which just adds this getter so we don't include it in the deployment but that seems like faff. Any strong feelings?


EndymionJkb (approved):
I feel strongly that faff is a funny word :)

But yes, it seems excessive to subclass to add a getter. It's possible some UI or something gathering metrics could use it, I suppose; no harm leaving it in.

PR #1025, opened 2021-11-18
MultiDistributor: improve usage of fixed point values

### LINKED ISSUE #1010: MultiDistributor: Review usage of fixed-point arithmetic
The default for .mul and .div is fixed-point, but most operations use Math.mul and .div. Given that we're mixing rates and amounts, we should be way clearer in explaining which values are fixed-point, which are not, and why we use each type of operation in each case.

TomAFrench:
I was having a look into how we use fixed point maths in `MultiDistributor` and it seemed to check out but wasn't optimal.

`globalTokensPerStake` was stored as a fixed point number (shown by the `divDown` as noted in https://github.com/balancer-labs/balancer-v2-monorepo/pull/968#discussion_r751421036) however this is properly cancelled off by the `mulDown` in `_unaccountedUnclaimedTokens` when calculating how many tokens a user is entitled to.

We were then calculating all the rewards correctly with only the small issue was that we could have stored `paymentRate` as a fixed point number to get some extra precision when calculating `unpaidAmountPerToken` for `globalTokensPerStake`.

I've switched the default maths library from `FixedPoint` to `Math` to make the usage of fixed point number clearer.

Fixes #1010


pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -263,7 +263,7 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
         // will be merged. In both scenarios we round down to avoid paying more tokens than were received.
         if (block.timestamp >= periodFinish) {
             // Current distribution period has ended so new period consists only of amount provided.
-            distribution.paymentRate = Math.divDown(amount, duration);
+            distribution.paymentRate = FixedPoint.divDown(amount, duration);
```
nventuro, line 266: Let's add comments mentioning a) that `paymentRate` is a fixed point number, and b) that by doing fixed point division on two non-fixed point numbers, we get a fixed point result.

pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -272,8 +272,8 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
 
             // Checked arithmetic is not required due to the if
             uint256 remainingTime = periodFinish - block.timestamp;
-            uint256 leftoverTokens = Math.mul(remainingTime, distribution.paymentRate);
-            distribution.paymentRate = Math.divDown(amount.add(leftoverTokens), duration);
+            uint256 leftoverTokens = FixedPoint.mulDown(remainingTime, distribution.paymentRate);
```
nventuro, line 275: We should have a comment here mentioning that fixed point multiplication of fixed (rate) and non-fixed (time) params results in a non-fixed result.

pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -704,8 +706,9 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
         UserDistribution storage userDistribution,
         uint256 updatedGlobalTokensPerStake
     ) internal view returns (uint256) {
+        // Note `unaccountedTokensPerStake` is a fixed point value
```
nventuro, line 709:
```suggestion
        // Note `updatedGlobalTokensPerStake` is a fixed point value
```

We could also mention below that the fixed point multiplication below results in a non-fixed point value

nventuro: We could also mention somewhere (perhaps with the data structure description) that both 'tokens per stake' (user and global) are fixed point.

PR #1026, opened 2021-11-18
Let PhantomStable have null rate providers

nventuro:
These are supported throughout the codebase, but the constructor didn't allow for them as `getRate` was called inconditionally.


EndymionJkb (commented):
Looks good. provider.getRate() is only called in _updateTokenRateCache, which you're protecting here.

All other references (_cacheTokenRateIfNecessary, setTokenRateCacheDuration, updateTokenRateCache) check that it has a provider before accessing it.

getTokenRate returns 1 if there is no provider.

Maybe add a test for invalid/unknown tokens after "sets no rate cache duration for BPT"?:

```
  it('sets no rate cache duration for an invalid/unknown token', async () => {
    const { duration, expires, rate } = await pool.getTokenRateCache(ZERO_ADDRESS);

    expect(rate).to.be.zero;
    expect(duration).to.be.zero;
    expect(expires).to.be.zero;
  });
```

nventuro:
Eh I'd say a more consistent experience would be for that function to revert for a token with no cache, wouldn't you say? Exposing the zero values is showing too much low level information in my opinion.

EndymionJkb:
> Good point about `getTokenRateCache`! However I'd say a more consistent experience would be for that function to revert for a token with no cache, wouldn't you say? Exposing the zero values is showing too much low level information in my opinion.

Yes - everywhere else we tend to do that. That means the "sets no rate cache duration for BPT" would revert then as well as if you called it with ZERO_ADDRESS? Or should BPT return 0s, and any token other than the 3 reverts? Would be more bytecode to check, that's the only drawback.

dmf7z (approved):

PR #1027, opened 2021-11-18
Make LinearMath a library

nventuro:
Minor thing, for consistency with the others (#863).


EndymionJkb (approved):
Yes, might as well go all the way

PR #1028, opened 2021-11-18
LinearPool: code review

nventuro:
This is the result of my review of `LinearPool` and interactions with `BasePool`. The changes in this PR should not modify the contract's behavior: they only favor readability. Separate PRs will cover bug fixes.

### DIFF
--- a/pkg/pool-linear/contracts/LinearPool.sol
+++ b/pkg/pool-linear/contracts/LinearPool.sol
@@ -188,12 +188,13 @@ contract LinearPool is BasePool, IGeneralPool, LinearMath, IRateProvider {
     }
 
     /**
-     * @dev Finishes initialization of the Linear Pool: it is unusable before calling this function.
+     * @dev Finishes initialization of the Linear Pool: it is unusable before calling this function as no BPT will have
+     * been minted.
      *
-     * Since Linear Pools have preminted BPT stored in the Vault, they require an initial join to deposit that BPT.
-     * Unfortunately, this cannot be performed during construction, as a join involves a callback function on the
-     * Pool, and the Pool will not have any code until construction finishes. Therefore, this must happen in a
-     * separate call.
+     * Since Linear Pools have preminted BPT stored in the Vault, they require an initial join to deposit said BPT as
+     * their balance. Unfortunately, this cannot be performed during construction, as a join involves calling the
+     * `onJoinPool` function on the Pool, and the Pool will not have any code until construction finishes. Therefore,
+     * this must happen in a separate call.
      *
      * It is highly recommended to create Linear pools using the LinearPoolFactory, which calls `initialize`
      * automatically.
@@ -202,10 +203,13 @@ contract LinearPool is BasePool, IGeneralPool, LinearMath, IRateProvider {
         bytes32 poolId = getPoolId();
         (IERC20[] memory tokens, , ) = getVault().getPoolTokens(poolId);
 
-        // During initialization, the Pool will mint the entire BPT supply for itself, and then join with it.
+        // Joins typically involve the Pool receiving tokens in exchange for newly-minted BPT. In this case however, the
+        // Pool will mint the entire BPT supply to itself, and join itself with it.
         uint256[] memory maxAmountsIn = new uint256[](_TOTAL_TOKENS);
-        maxAmountsIn[tokens[0] == IERC20(this) ? 0 : tokens[1] == IERC20(this) ? 1 : 2] = _INITIAL_BPT_SUPPLY;
+        maxAmountsIn[_bptIndex] = _INITIAL_BPT_SUPPLY;
 
+        // The first time this executes, it will call `_onInitializePool` (as the BPT supply will be zero). Future calls
+        // will be routed to `_onJoinPool`, which always reverts, meaning `initialize` will only execute once.
         IVault.JoinPoolRequest memory request = IVault.JoinPoolRequest({
             assets: _asIAsset(tokens),
             maxAmountsIn: maxAmountsIn,
@@ -229,10 +233,12 @@ contract LinearPool is BasePool, IGeneralPool, LinearMath, IRateProvider {
 
         // The full BPT supply will be minted and deposited in the Pool. Note that there is no need to approve the Vault
         // as it already has infinite BPT allowance.
+        uint256 bptAmountOut = _INITIAL_BPT_SUPPLY;
+
         uint256[] memory amountsIn = new uint256[](_TOTAL_TOKENS);
         amountsIn[_bptIndex] = _INITIAL_BPT_SUPPLY;
 
-        return (_INITIAL_BPT_SUPPLY, amountsIn);
+        return (bptAmountOut, amountsIn);
     }
 
     function _onJoinPool(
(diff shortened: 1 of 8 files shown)


nventuro:
I reverted 889d584 as it makes reviewing the PR much harder, but want to re-introduce it later on as the reading order is otherwise a bit jarring.

pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -30,13 +30,22 @@ import "./LinearPoolUserData.sol";
 
 /**
  * @dev Linear Pools are designed to hold two assets: "main" and "wrapped" tokens that have an equal value underlying
- * token (e.g., DAI and aDAI). The Pool will register three tokens in the Vault however: the two assets and the BPT
- * itself, so that BPT can be exchanged (effectively joining and exiting) via swaps.
+ * token (e.g., DAI and waDAI). There must be an external feed available to provide an exact, non-manipulable exchange
+ * rate between the tokens. In particular, any reversible manipulation (e.g. cause for the rate to increase and then
```
EndymionJkb, line 34:
```suggestion
 * rate between the tokens. In particular, any reversible manipulation (e.g. causing the rate to increase and then
```

pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -30,13 +30,22 @@ import "./LinearPoolUserData.sol";
...
+ *
+ * Despite inheriting from BasePool, much of the basic behavior changes. This Pool does not support regular joins and
+ * exits, as the entire BPT supply is 'preminted' during initialization.
  *
- * There must be an external feed available to provide an exact, non-manipulable exchange rate between the tokens.
+ * Unlike most other Pools, this one does not attempt to create revenue by charging fees: value is derived by holding
+ * the wrapped, yield-bearing asset. However, the 'swap fee percentage' value is still used, albeit with a different
+ * meaning. This Pool attempts to hold a certain amount of "main" tokens, between a lower and upper target value.
+ * Moving the balance outside of that range causes proportional fees to be charged to the sender, which are then used to
+ * incentivize other parties to bring back the balance to the desired region by paying them using these collected fees.
```
EndymionJkb, line 47:
```suggestion
 * The pool charges fees on trades that move the balance outside that range, which are then paid back
 * as incentives to traders whose swaps return the balance to the desired region.
```

pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -46,36 +55,53 @@ contract LinearPool is BasePool, IGeneralPool, LinearMath, IRateProvider {
...
-    uint256 private constant _MAX_TOKEN_BALANCE = 2**(112) - 1;
+    // This is the maximum token amount the Vault can hold. In regular operation, the total BPT supply remains constant
+    // and equal to _INITIAL_BPT_SUPPLY, but most of it remains in the Pool, waiting to be exchanged for tokens. The
+    // actual amount of BPT in circulation is the total supply minus the amount held by the Pool, and is known as the
+    // 'virtual supply'.
+    // The only instance in which the total supply might change is if the emergency pause is turned on, enabling an
+    // alternative way to exit the Pool which involves burning BPT. As this is expected to not happen, we optimize for
+    // success by using _INITIAL_BPT_SUPPLY instead of totalSupply(), saving a storage read. This optimization is only
+    // valid if the Pool is never paused: in the case of an emergency that leads to burned tokens, the Pool should not
+    // be unpaused and continue to be used via regular means.
```
EndymionJkb, line 66:
```suggestion
    // valid if the Pool is never paused: in case of an emergency that leads to burned tokens, the Pool should not
    // be used after the buffer period expires and it automatically 'unpauses'.
```

pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -46,36 +55,53 @@ contract LinearPool is BasePool, IGeneralPool, LinearMath, IRateProvider {
...
 
-    // Linear Pools don't lock any BPT, since they fully support having zero main and wrapped token balances
-    uint256 private constant _LINEAR_MINIMUM_BPT = 0;
-    uint256 private constant _MAX_TOKEN_BALANCE = 2**(112) - 1;
+    // This is the maximum token amount the Vault can hold. In regular operation, the total BPT supply remains constant
+    // and equal to _INITIAL_BPT_SUPPLY, but most of it remains in the Pool, waiting to be exchanged for tokens. The
+    // actual amount of BPT in circulation is the total supply minus the amount held by the Pool, and is known as the
+    // 'virtual supply'.
+    // The only instance in which the total supply might change is if the emergency pause is turned on, enabling an
+    // alternative way to exit the Pool which involves burning BPT. As this is expected to not happen, we optimize for
```
EndymionJkb, line 63:
```suggestion
    // The total supply can only change if the emergency pause is activated by governance, enabling an
    // alternative proportional exit that burns BPT. As this is not expected to happen, we optimize for
```

pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -46,36 +55,53 @@ contract LinearPool is BasePool, IGeneralPool, LinearMath, IRateProvider {
...
+    // The indices of each token when registered, which can then be used to access the balances array.
     uint256 private immutable _bptIndex;
     uint256 private immutable _mainIndex;
     uint256 private immutable _wrappedIndex;
 
+    // Both BPT and the main token have a regular, constant scaling factor (equal to FixedPoint.ONE for BPT, and
+    // dependent on the number of decimals for the main token). The wrapped token's scaling factor however is two-fold:
+    // it is composed of both the decimal scaling factor, as well as an externally provided rate that is used to
+    // convert wrapped tokens to an equivalent main token amount. This external rate is expected to be ever increasing,
+    // reflecting the fact that the wrapped token appreciates in value over time (because e.g. it is accruing interest).
```
EndymionJkb, line 81:
```suggestion
    // dependent on the number of decimals for the main token). However, the wrapped token's scaling factor has two components:
    // the usual token decimal scaling factor, and an externally provided rate used to
    // convert wrapped tokens to an equivalent main token amount. This external rate is expected to be ever increasing,
    // reflecting the fact that the wrapped token appreciates in value over time (e.g.; because it is accruing interest).
```

pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -381,10 +425,12 @@ contract LinearPool is BasePool, IGeneralPool, LinearMath, IRateProvider {
...
         // The full BPT supply will be minted and deposited in the Pool. Note that there is no need to approve the Vault
         // as it already has infinite BPT allowance.
+        uint256 bptAmountOut = _INITIAL_BPT_SUPPLY;
+
         uint256[] memory amountsIn = new uint256[](_TOTAL_TOKENS);
-        amountsIn[_bptIndex] = _MAX_TOKEN_BALANCE;
+        amountsIn[_bptIndex] = _INITIAL_BPT_SUPPLY;
 
-        return (_MAX_TOKEN_BALANCE, amountsIn);
+        return (bptAmountOut, amountsIn);
```
EndymionJkb, line 433: Would it save a little to use named return arguments here? Just "bptAmountOut = _INITIAL_BPT_SUPPLY" then. Instead of setting to a constant and immediately returning it.

Maybe the explicit return is clearer/more consistent?

nventuro: Yeah I'm just not a fan of named return arguments in general, but wanted to at least hint at what _MAX_TOKEN_BALANCE stands for here.

dmf7z (approved):

PR #1030, opened 2021-11-19
Linear: use split code factory

nventuro:
This is required due to changes such as https://github.com/balancer-labs/balancer-v2-monorepo/pull/1029 which increase the factory bytecode past the limit.


EndymionJkb (approved):
Looks good, though you might consider commenting the call to initialize. Something like:

LinearPools require initialization with preminted BPT, which involves a callback function that cannot be called during construction. Conventional joins are disabled, since Linear pools implement joins as swaps with the preminted BPT. So this special external function must be called instead.

dmf7z (approved):

PR #1031, opened 2021-11-22
MultiDistributor: Update used terms in RewardsScheduler

TomAFrench:
`RewardsScheduler` (now `DistributionScheduler`) is quite out of date in relation of `MultiDistributor` in that it uses `rewardsToken`, `rewarder`, etc.

I've standardised the terminology so that both contracts match and updated the test suite to allow more comprehensive tests and remove the redundant deployments of pools, etc.

There's still a fair amount of improvements to be done (e.g. we're storing several variables in storage unnecessarily) but I want to restrict this PR to just renaming + organising tests.


pkg/distributors/contracts/DistributionScheduler.sol
```diff
@@ -0,0 +1,142 @@
...
+        address owner;
+        uint256 amount;
+        DistributionStatus status;
+    }
+
+    event DistributionScheduled(
+        bytes32 scheduleId,
+        address indexed owner,
+        IERC20 indexed stakingToken,
+        IERC20 indexed distributionToken,
```
nventuro, line 52: I'd simplify this to just the distribution id.

pkg/distributors/contracts/DistributionScheduler.sol
```diff
@@ -0,0 +1,142 @@
...
+        uint256 startTime,
+        uint256 amount
+    );
+
+    mapping(bytes32 => ScheduledDistribution) private _scheduledDistributions;
+
+    function getScheduledDistributionInfo(bytes32 scheduleId)
+        external
+        view
+        returns (ScheduledDistribution memory reward)
```
nventuro, line 70:
```suggestion
        returns (ScheduledDistribution memory distribution)
```

pkg/distributors/contracts/DistributionScheduler.sol
```diff
@@ -0,0 +1,142 @@
...
+    {
+        return _scheduledDistributions[scheduleId];
+    }
+
+    function startDistributions(bytes32[] calldata scheduleIds) external {
```
nventuro, line 75: Nitpicky, but I'd place this after `schedule`

pkg/distributors/contracts/DistributionScheduler.sol
```diff
@@ -0,0 +1,142 @@
...
+        for (uint256 i; i < scheduleIds.length; i++) {
+            bytes32 scheduleId = scheduleIds[i];
+            ScheduledDistribution memory scheduledDistribution = _scheduledDistributions[scheduleId];
+
+            require(scheduledDistribution.status == DistributionStatus.PENDING, "Distribution cannot be started");
```
nventuro, line 80: Might be good UX to issue more specific messages if the distro is uninit or already started.

pkg/distributors/contracts/DistributionScheduler.sol
```diff
@@ -0,0 +1,142 @@
...
+            require(scheduledDistribution.startTime <= block.timestamp, "Distribution start time is in the future");
```
nventuro, line 81:
```suggestion
            require(block.timestamp >= scheduledDistribution.startTime, "Distribution start time is in the future");
```

:shrug:

pkg/distributors/contracts/DistributionScheduler.sol
```diff
@@ -0,0 +1,142 @@
...
+
+            _scheduledDistributions[scheduleId].status = DistributionStatus.STARTED;
+
+            uint256 allowance = scheduledDistribution.distributionToken.allowance(
+                address(this),
+                address(_multiDistributor)
+            );
+            if (allowance < scheduledDistribution.amount) {
+                scheduledDistribution.distributionToken.approve(address(_multiDistributor), type(uint256).max);
+            }
```
nventuro, line 91: I'd just do `approve(amount)`, as we do in the relayer, which means EIP 2929 savings will kick in (also resulting in simpler code).

pkg/distributors/contracts/DistributionScheduler.sol
```diff
@@ -0,0 +1,142 @@
...
+    function getScheduleId(
+        IERC20 stakingToken,
+        IERC20 distributionToken,
+        address owner,
+        uint256 startTime
+    ) public pure returns (bytes32) {
+        return keccak256(abi.encodePacked(stakingToken, distributionToken, owner, startTime));
+    }
+
+    function scheduleDistribution(
```
nventuro, line 113: This seems incomplete: shouldn't we check that `msg.sender` is the owner of the distro?

pkg/distributors/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,209 @@
...
+  let distributor: MultiDistributor;
+  let scheduler: Contract;
+
+  let stakingToken: Token, stakingTokens: TokenList;
+  let distributionToken: Token, distributionTokens: TokenList;
+
+  let distributionOwner: SignerWithAddress, poker: SignerWithAddress;
+
+  before('setup signers', async () => {
+    [, distributionOwner, poker] = await ethers.getSigners();
```
nventuro, line 33:
```suggestion
    [, distributionOwner, other] = await ethers.getSigners();
```

The 'poker' bit had me do a double take :joy:

pkg/distributors/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,209 @@
...
+    distributionToken = distributionTokens.first;
+
+    await distributionTokens.mint({ to: distributionOwner });
+    await distributionTokens.approve({ to: scheduler, from: distributionOwner });
+  });
+
+  const calcScheduleId = (stakingToken: Token, distributionToken: Token, owner: Account, time: BigNumberish) => {
+    const addresses = TypesConverter.toAddresses([stakingToken, distributionToken, owner]);
+    return solidityKeccak256(['address', 'address', 'address', 'uint256'], [...addresses, time]);
+  };
```
nventuro, line 56: Can't we just use the `view` function?

pkg/distributors/contracts/DistributionScheduler.sol
nventuro: Also let's return the scheduled id here.

pkg/distributors/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    it('creates a scheduled distribution', async () => {
+      const time = (await currentTimestamp()).add(3600 * 24);
+      await scheduler
+        .connect(distributionOwner)
+        .scheduleDistribution(ZERO_BYTES32, stakingToken.address, distributionToken.address, amount, time);
+
+      const scheduleId = calcScheduleId(stakingToken, distributionToken, distributionOwner, time);
+
+      const response = await scheduler.getScheduledDistributionInfo(scheduleId);
```
nventuro, line 69: I'd consider getting the id from the event, but :shrug:

pkg/distributors/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,209 @@
...
+        distributionTokens,
+        [
+          { account: scheduler.address, changes: { [distributionToken.symbol]: amount } },
+          { account: distributionOwner.address, changes: { [distributionToken.symbol]: amount.mul(-1) } },
+        ]
+      );
+    });
+
+    it('emits a DistributionScheduled event', async () => {
+      const time = (await currentTimestamp()).add(3600 * 24);
```
nventuro, line 96: We have HOUR, MINUTE and so on helpers

pkg/distributors/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,209 @@
...
+
+  const calcScheduleId = (stakingToken: Token, distributionToken: Token, owner: Account, time: BigNumberish) => {
+    const addresses = TypesConverter.toAddresses([stakingToken, distributionToken, owner]);
+    return solidityKeccak256(['address', 'address', 'address', 'uint256'], [...addresses, time]);
+  };
+
+  describe('scheduleDistribution', () => {
+    const amount = fp(1);
+
+    it('creates a scheduled distribution', async () => {
```
nventuro, line 61: Consider adding a .skip test where scheduling fails if the sender is not the owner.

pkg/distributors/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,209 @@
...
+  });
+
+  describe('startDistributions', () => {
+    const amount = fp(1);
+
+    context('when distribution is pending', () => {
+      let scheduleId: string;
+      let time: BigNumber;
+      sharedBeforeEach(async () => {
+        // reward duration is important.  These tests assume a very short duration
```
nventuro, line 123: Is it? Can't we advance time to whenever?

pkg/distributors/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,209 @@
...
+          .connect(distributionOwner)
+          .scheduleDistribution(ZERO_BYTES32, stakingToken.address, distributionToken.address, amount, time);
+
+        scheduleId = calcScheduleId(stakingToken, distributionToken, distributionOwner, time);
+      });
+
+      // Skip tests pending support for paying into a distribution owned by another address
+      context.skip('when start time has passed', () => {
+        sharedBeforeEach(async () => {
+          await advanceTime(3600 * 25);
```
nventuro, line 135: Ideally this'd be something like `startTime.add(MINUTE)`.

pkg/distributors/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,209 @@
...
+        time = (await currentTimestamp()).add(3600 * 24);
+        await scheduler
+          .connect(distributionOwner)
+          .scheduleDistribution(ZERO_BYTES32, stakingToken.address, distributionToken.address, amount, time);
+
+        scheduleId = calcScheduleId(stakingToken, distributionToken, distributionOwner, time);
+      });
+
+      // Skip tests pending support for paying into a distribution owned by another address
+      context.skip('when start time has passed', () => {
```
nventuro, line 133: We could also add a .skip test for when the start time has not passed

pkg/distributors/contracts/DistributionScheduler.sol
TomAFrench: Changed this to just pass over uninit or started distributions to avoid race conditions.

pkg/distributors/contracts/DistributionScheduler.sol
TomAFrench: Why would we return the `sheduleIds`?

pkg/distributors/contracts/DistributionScheduler.sol
TomAFrench: This is correct. This preexisting issue is going to be fixed in a later PR which makes the scheduler pull a lot more data from the distributor (e.g. it currently has no guarantees that the provided tokens are relevant for the specified distribution)

pkg/distributors/test/DistributionScheduler.test.ts
TomAFrench: Will handle later as mentioned above

pkg/distributors/test/DistributionScheduler.test.ts
TomAFrench: Will handle later as mentioned above

pkg/distributors/contracts/DistributionScheduler.sol
nventuro: Sorry, I confused this function with `schedule` and thought it was missing a return value. `schedule` does return the scheduled id, so we're fine.

pkg/distributors/contracts/DistributionScheduler.sol
nventuro: Ah I see, thanks.

nventuro (approved):

PR #1034, opened 2021-11-22
MultiDistributor: transfer staked tokens

gtaschuk:
adds `transferStakedTokens` that transfers staked tokens within the distributor as a more gas efficient alternative to withdrawing, transferring, and restaking


TomAFrench:

Issue #985, opened 2021-11-03
Rename AccessControl's roles to actions


nventuro:
OZ's [AccessControl](https://docs.openzeppelin.com/contracts/4.x/access-control#role-based-access-control) is role-based: it can be used to designate roles that can perform multiple actions.

Due to a number of reasons, including bytecode size restrictions, our own authorization mechanism maps external functions to these 'roles' on a 1:1 basis, so we renamed these to 'actions' for more clarity. Instead of an account having a role that can let it perform an action, it simply can perform the action - roles are superfluous.

The initial version of the Authorizer was simply built on top of AccessControl as it didn't add much, but now that we're more thoroughly changing its internals (see #588 and #980) I suggest we drop OZ's version for a more customized one that supports our advanced features natively, and uses our own naming scheme.

Suggested changes:
 - `hasRole` -> `canPerform` (we already use this getter in all contracts)
 - `grantRole` -> `grantPermission`
 - `revokeRole` -> `revokePermission`

---

The only bit that doesn't quite fit this model is that of admins (to authorize grant and revoke), likely because those _are_ much closer to being roles. 

We could support these by having each action not only keep track of who can perform it, but also of a list of its admins, who can freely grant and remove permission both locally and globally. This might mean adding `isAdminFor`, which also helps draw the distinction between admin and non-admin 'roles'.

Ramarti:
I'd propose just moving the functionality from `AccessControl` to the `Authorizer` and remove redundancies.
+1 to having 2 methods to check if Admin or specific permission

nventuro:
For reference, see [this review comment](https://github.com/balancer-labs/balancer-v2-monorepo/pull/980#discussion_r742238887) about how lacking `isAdminFor` might look a bit weird.

PR #1035, opened 2021-11-23
Authoriser: separation of roles into admin rights and permissions

TomAFrench:
This PR handles the renaming of "roles" to "permissions" as mentioned in #985 as well as splitting the concept of "having admin rights over a permission" off from "having a permission" as the original choices in `AccessControl` no longer make sense when the roles map onto calling particular functions 1:1.

To do this the data structure for permissions has been modified to hold an explicit set of admin addresses for each permission which have the power to grant that permission to other addresses.

```
OLD:
struct RoleData {
    EnumerableSet.AddressSet globalMembers;
    mapping(address => EnumerableSet.AddressSet) membersByContract;
    bytes32 adminRole;
}

NEW:
struct PermissionData {
    EnumerableSet.AddressSet globalMembers;
    mapping(address => EnumerableSet.AddressSet) membersByContract;
    EnumerableSet.AddressSet admins;
}
```

Addresses which have admin rights over the permission `bytes32(0)` are global admins and can grant permissions/adminships globally as they desire.

To handle this split, we've had to duplicate several of the functions to add/remove roles. The changes from OZ feel substantial enough to warrant building our own test suite but I'll hold off on this for the time being in case anyone has concerns about this change.


TomAFrench:
On closer look we can transplant most of the Authoriser tests for this so it's not going to be a huge deal to fill out the tests for this.

nventuro:
@facuspagnuolo is basing his work on the authorizer revamp milestone on this and other unmerged PRs, closing as this will be superceded by those.

PR #1041, opened 2021-11-29
MultiDistributor: Add initial subscription-by-relayer support

TomAFrench:
Currently only the user can subscribe/unsubscribe a user from distributions. This prevents us from having a meaningful one-tx flow as relayers (or auto-staking pools) will be unable to start rewards accruing for a user.

This PR extends #1040 to allow relayers to opt users into distributions.


TomAFrench:

PR #1042, opened 2021-11-29
MultiDistributor: add ability for distribution owners to delegate funding to contracts, etc.

TomAFrench:
Builds on #1040 

This PR adds support for relayers to fund a distribution owned by another address. This allows seamless use of the `RewardScheduler` without having to have users subscribe to a whole new distribution.

PR to update `RewardScheduler` to make use of this will come separately.


nventuro (approved):
Lovely! I'd consider adding a comment in the interface mentioning that relayers can also call these functions (see the docs in IVault for inspiration)

PR #1043, opened 2021-11-30
Managed: Use controller in ManagedPoolFactory

EndymionJkb:


pkg/pool-weighted/contracts/smart/ManagedPoolFactory.sol
```diff
@@ -31,38 +35,40 @@ contract ManagedPoolFactory is BasePoolSplitCodeFactory, FactoryWidePauseWindow
...
+                    swapFeePercentage: poolParams.swapFeePercentage,
+                    pauseWindowDuration: pauseWindowDuration,
+                    bufferPeriodDuration: bufferPeriodDuration,
+                    owner: address(poolController),
+                    swapEnabledOnStart: poolParams.swapEnabledOnStart,
+                    mustAllowlistLPs: poolParams.mustAllowlistLPs,
+                    managementSwapFeePercentage: poolParams.managementSwapFeePercentage
+                })
+            )
+        );
```
TomAFrench, line 70: What do you think of this factory just calling into a base `ManagedPoolFactory` instead of deploying the pool directly?

TomAFrench (commented):
I'm not a fan of having a factory which enforces a particular pool controller. We're presumably going to be allowing "naked" managed pools to be deployed as well without a controller so these would need their own factory and then we have two factories which deploy identical pools (up to the choice of owner).

pkg/pool-weighted/contracts/smart/ManagedPoolFactory.sol
EndymionJkb: I thought we specifically *weren't* allowing that. Since we're removing all restrictions from Managed Pools, "naked" ones would be very dangerous. The ManagedPoolController is very configurable, and is intended to be suitable for 90% of the use cases (like the V1 CRP was).

If somebody really needed different functionality, they would make their own controller, but this way the UI would know that anything deployed from this factory used an approved controller, and is safe to show on the UI.

TomAFrench: As we're allowing asset managers on this factory they're equally as dangerous afaik and they'll all need to be individually whitelisted.

My preference for having a single "naked" `ManagedPoolFactory` that all "pool controller factories" call into comes from:
1. In the case of needing another pool controller factory then we can point it at the same underlying `ManagedPoolFactory` and know that the only changes are in the controller rather than the underlying pool
2. When tracking pools we can have an easy way of collecting all `ManagedPools` regardless of the original source by looking at the base `ManagedPoolFactory`.

Each pool controller factory would maintain a record of which pools it has deployed as well so we can keep track of which pool is from which.

TomAFrench:
This isn't quite the way I envisioned it. Here we're inheriting from a `BaseManagedPoolFactory` so each "controller factory" has their own copy of this code. I was imagining a single `BaseManagedPoolFactory` deployed standalone and then have the controller factories literally call into this one.

TomAFrench (approved):
Looks good! We can revisit to see if merging the factories to regain these tokens in future is necessary.

PR #1044, opened 2021-11-30
Relayer: comment review

EndymionJkb:


pkg/standalone-utils/contracts/relayer/AaveWrapping.sol
```diff
@@ -47,13 +47,14 @@ abstract contract AaveWrapping is IBaseRelayerLibrary {
...
         // The wrap caller is the implicit sender of tokens, so if the goal is for the tokens
-        // to be sourced from outside the relayer, we must first them pull them here.
+        // to be sourced from outside the relayer, we must first pull them here.
         if (sender != address(this)) {
             require(sender == msg.sender, "Incorrect sender");
             _pullToken(sender, dynamicToken, amount);
         }
 
         dynamicToken.approve(address(staticToken), amount);
+        // Use 0 for the referral code
```
TomAFrench, line 57: Do we want to set one of these up maybe?

TomAFrench (approved):
Looks good!

pkg/standalone-utils/contracts/relayer/AaveWrapping.sol
EndymionJkb: I was wondering about that. Is it worth passing one in at construction time? I could add that

pkg/standalone-utils/contracts/interfaces/IStaticATokenLM.sol
```diff
@@ -22,8 +22,8 @@ interface IStaticATokenLM is IERC20 {
      *   0 if the action is executed directly by the user, without any middle-man
      * @param fromUnderlying bool
      * - `true` if the msg.sender comes with underlying tokens (e.g. USDC)
-     * - `false` if the msg.sender comes already with aTokens (e.g. aUSDC)
-     * @return uint256 The amount of StaticAToken minted, static balance
+     * - `false` if the msg.sender already has aTokens (e.g. aUSDC)
```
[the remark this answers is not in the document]
TomAFrench: Alternatively we could ask if Aave is interested in these updates seeing as they haven't deployed yet?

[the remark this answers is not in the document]
nventuro: Ah that'd be good, yes.

pkg/standalone-utils/contracts/relayer/AaveWrapping.sol
nventuro: Aave has confirmed that they are not and will not use these referral codes, so we don't need to do anything.

PR #1045, opened 2021-12-01
Linear Pools: remove price rate cache

TomAFrench:
This PR simply removes the price rate cache from the Linear Pool and creates a virtual `_getWrappedTokenRate` function for derived contracts to specify. A `MockLinearPool` has been created with a constant rate to maintain the current tests.

An `AaveLinearPool` will be added in a later PR.


TomAFrench:
As we've removed a couple of arguments from the `LinearPool` constructor we can get away without having a struct to avoid the stack issues.

pkg/pool-linear/test/LinearPool.test.ts
```diff
@@ -662,222 +658,4 @@ describe('LinearPool', function () {
...
-
-      context('with a price rate below 1', () => {
-        sharedBeforeEach('mock rate', async () => {
-          await wrappedTokenRateProvider.mockRate(fp(0.99));
-          await pool.updateWrappedTokenRateCache();
-        });
-
-        itAdaptsTheScalingFactorsCorrectly();
-      });
-    });
```
nventuro, line 735: These tests are important: they test the rate is correctly applied to the scaling factors. Let's add a simple setter to the `MockLinearPool` for the rate, and keep these tests.

pkg/pool-linear/test/LinearPoolFactory.test.ts
```diff
@@ -22,19 +22,17 @@ describe('LinearPoolFactory', function () {
...
-  const WRAPPED_TOKEN_RATE_CACHE_DURATION = MONTH;
 
   before('setup signers', async () => {
     [, owner] = await ethers.getSigners();
   });
 
   sharedBeforeEach('deploy factory & tokens', async () => {
     vault = await Vault.create();
-    factory = await deploy('LinearPoolFactory', { args: [vault.address] });
+    factory = await deploy('MockLinearPoolFactory', { args: [vault.address] });
```
nventuro, line 32: This is fine as a temporary measure to have these tests pass, but in the future we'll want to replace this back with the AaveLinearPoolFactory.

nventuro (commented):
Looks great, love the separation of concerns. Thanks a lot for your help @TomAFrench!

nventuro (approved):

PR #1050, opened 2021-12-03
Batch relayer deployment

nventuro:
The arbitrum relayer is not verified as arbiscan has trouble with the `create` opcode.


TomAFrench (approved):
Confirmed that the build info matches the output from this commit. I've checked that the deployed addresses match the deployment bytecode and constructor args listed here.

It might be worth including the creation code  for the `BalancerRelayer` in `bytecode/` for completeness.

TomAFrench:
Also we probably should have merged #1044 so we get the fixes included on etherscan but not worth dropping another g for typos.

nventuro:
>Also we probably should have merged #1044 so we get the fixes included on etherscan but not worth dropping another g for typos.

Ah totally forgot about that one, my bad. Oh well, next time :)

>I've checked that the deployed addresses match the deployment bytecode and constructor args listed here.

Have you checked that the constructor arguments are correct? Specifically, the wstETH addresses.

nventuro:
>Confirmed that the build info matches the output from this commit.

Note that this process is incomplete: we're not verifying that the deployed bytecode matches the build info. Ideally we'd have some form of automated setup for that, but we haven't built anything like that yet.

TomAFrench:
> Note that this process is incomplete: we're not verifying that the deployed bytecode matches the build info.

True, just did a quick ctrl f on the bytecode between build info and `bytecode/` (what I was using to check etherscan) and it matches.

> Have you checked that the constructor arguments are correct? Specifically, the wstETH addresses.

Yes, these were fine. The only issue is our stETH/wstETH on Kovan don't have support for "staking" ETH so we can't test those features except on mainnet but I don't think that's a huge deal.

PR #1051, opened 2021-12-03
Multidistributor: distinguish between distribution channels and distributions

### LINKED ISSUE #1048: MultiDistributor: Distinguish between distributions and distribution channels
We currently have some terminology overloading between distributions (a tuple of staking token, distribution token, and owner which stakers can subscribe to for rewards) and distributions (a single deposit of distribution token to be distributed to stakers). This will lead to misunderstandings and difficulty in reasoning about the code by outside parties.

We already have the concept of distribution channels in the Merkle orchard which is a natural fit for the first definition. We should rename any of this usage of "distribution".

TomAFrench:
Fixes #1048 

References to "distributions" have been clarified if they refer to a distribution channel.

2 structs have been renamed:
1. `Distribution` has been renamed to `DistributionChannel`
2. `UserDistribution` has been renamed to `UserDistributionInfo`

Variables in tests have not been updated as I feel it leads to excessive noise. I'm planning to tackle in a separate PR but happy to roll it into this one.


pkg/distributors/contracts/interfaces/IMultiDistributor.sol
```diff
@@ -33,48 +33,51 @@ interface IMultiDistributor {
...
+    event DistributionChannelCreated(
+        bytes32 indexed distributionChannel,
         IERC20 stakingToken,
         IERC20 distributionToken,
         address owner
     );
-    event DistributionDurationSet(bytes32 indexed distribution, uint256 duration);
-    event DistributionFunded(bytes32 indexed distribution, uint256 amount);
+    event DistributionDurationSet(bytes32 indexed distributionChannel, uint256 duration);
+    event DistributionFunded(bytes32 indexed distributionChannel, uint256 amount);
```
TomAFrench, line 56: I've left these as just `Distribution` as they are setting values for the next distribution although I'm a little torn on it. Thoughts?

pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -101,70 +101,75 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
...
 
     /**
-     * @dev Returns the information of a distribution
-     * @param distributionId ID of the distribution being queried
+     * @dev Returns the information of a distribution channel
+     * @param distributionId ID of the distribution channel being queried
      */
-    function getDistribution(bytes32 distributionId) external view override returns (Distribution memory) {
-        return _getDistribution(distributionId);
+    function getDistributionChannel(bytes32 distributionId)
```
nventuro, line 107:
```suggestion
    function getDistributionChannel(bytes32 distributionChannelId)
```

There are multiple other instances of this as well

pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -177,83 +182,87 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
...
 
     /**
-     * @dev Creates a new distribution
-     * @param stakingToken The staking token that will be eligible for this distribution
+     * @dev Creates a new distribution channel
+     * @param stakingToken The staking token that will be eligible for this distribution channel
      * @param distributionToken The token to be distributed to users
      * @param duration The duration over which each distribution is spread
      */
     function createDistribution(
```
nventuro, line 190: Shouldn't this be renamed? If distributions last for some time after they are funded and run through a channel, then what this creates is the channel.

pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -288,78 +297,82 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
...
      */
-    function subscribeDistributions(bytes32[] calldata distributionIds) external override {
-        bytes32 distributionId;
-        Distribution storage distribution;
-        for (uint256 i; i < distributionIds.length; i++) {
-            distributionId = distributionIds[i];
-            distribution = _getDistribution(distributionId);
-
-            IERC20 stakingToken = distribution.stakingToken;
+    function subscribeDistributions(bytes32[] calldata distributionChannelIds) external override {
```
nventuro, line 314: Same here

pkg/distributors/contracts/interfaces/IMultiDistributor.sol
```diff
@@ -35,48 +35,51 @@ interface IMultiDistributor {
...
+    function getDistributionChannel(bytes32 distributionChannelId) external view returns (DistributionChannel memory);
 
-    function globalTokensPerStake(bytes32 distributionId) external view returns (uint256);
+    function globalTokensPerStake(bytes32 distributionChannelId) external view returns (uint256);
 
-    function totalSupply(bytes32 distributionId) external view returns (uint256);
+    function totalSupply(bytes32 distributionChannelId) external view returns (uint256);
 
-    function isSubscribed(bytes32 distributionId, address user) external view returns (bool);
+    function isSubscribed(bytes32 distributionChannelId, address user) external view returns (bool);
```
nventuro, line 75: Note that these were changed here but not on the impl

nventuro (commented):
I'm not sure I get the distinction between distribution and distribution channel (which was clearer in the Orchard), as the current changeset seems to not be entirely consistent in this regard

TomAFrench:
> I'm not sure I get the distinction between distribution and distribution channel (which was clearer in the Orchard), as the current changeset seems to not be entirely consistent in this regard

The distinction is that a person wanting to distributed their token creates a distribution channel which is permanent but then makes distributions into this channel which are the bump of tokens to be distributed over a fixed period.

TomAFrench:

PR #1052, opened 2021-12-04
MultiDistributor: add missing event on distribution channel creation

TomAFrench:
We currently don't emit what the duration of a newly created distribution channel is. This PR emits that data


EndymionJkb (approved):
I was trying to find a way to make both the create and setDuration call the same internal function that emits the event, but I don't think that pattern works here. setDistributionDuration uses msg.sender, so has to be external, etc.

PR #1053, opened 2021-12-04
MultiDistributor: Split `Staked` event into staking and subscribing events

TomAFrench:
The current way that we emit `Staked` and `Unstaked` events causes some annoyances for offchain tracking in subgraphs etc. as it requires querying the tracker's internal state to be able to maintain the data in a similar structure to onchain (without duplication).

https://github.com/balancer-labs/balancer-v2-monorepo/blob/7e71e5cdce0478475a96a1143bb3a48a7f456a6a/pkg/distributors/contracts/interfaces/IMultiDistributor.sol#L24-L45

In reality, when tracking a user staking tokens the tracker just needs to know (`user`, `stakingToken`, `amount`) and when subscribing it needs to know (`user`, `distribution`). This PR switches our events to emit this data instead.

This also acts as a gas optimisation when staking/unstaking a token which is subscribed to many distributions as we only need to emit a single `Staked` event rather than one for each distribution channel. This comes at the expense of emitting an extra event when subscribing/unsubscribing from distribution channels for which you have a zero balance of the staking token which I feel is acceptable.

### DIFF
--- a/pkg/distributors/contracts/MultiDistributor.sol
+++ b/pkg/distributors/contracts/MultiDistributor.sol
@@ -323,8 +323,9 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
                     distribution
                 );
                 distribution.totalSupply = distribution.totalSupply.add(amount);
-                emit Staked(distributionId, msg.sender, amount);
             }
+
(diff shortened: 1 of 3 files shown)


TomAFrench:

PR #1055, opened 2021-12-06
MultiDistributor: rename _updateDistributionRate to _updateGlobalTokensPerStake

TomAFrench:
`_updateGlobalTokensPerStake`  better reflects the purpose of this function in relation to `_updateUserTokensPerStake` than `_updateDistributionRate`

### DIFF
--- a/pkg/distributors/contracts/MultiDistributor.sol
+++ b/pkg/distributors/contracts/MultiDistributor.sol
@@ -237,7 +237,7 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
 
         // Before receiving the tokens, we must update the distribution's rate
         // as we are about to change its payment rate, which affects all other rates.
-        _updateDistributionRate(distribution);
+        _updateGlobalTokensPerStake(distribution);
 
         // Get the tokens and deposit them in the Vault as this contract's internal balance, making claims to internal
         // balance, joining pools, etc., use less gas.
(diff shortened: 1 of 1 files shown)


pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -235,9 +235,9 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
         require(distribution.duration > 0, "DISTRIBUTION_DOES_NOT_EXIST");
         require(distribution.owner == msg.sender, "SENDER_NOT_OWNER");
 
-        // Before receiving the tokens, we must update the distribution's rate
-        // as we are about to change its payment rate, which affects all other rates.
-        _updateDistributionRate(distribution);
+        // Before receiving the tokens, we must sync the distribution up to the present as we are about to change
+        // its payment rate, which would otherwise affect the distribution of tokens distributed since the last update
```
EndymionJkb, line 239: > which would otherwise affect the distribution of tokens distributed since the last update

Can this be simplified? It would affect the accounting of tokens distributed since the last update,
or just: "which would otherwise affect the token distribution"

TomAFrench: I've changed to the first option as I think it's important to describe that we're talking about tokens which have already been distributed (and so should be fixed.)

EndymionJkb: Yes - that's what I was confused about. It can't affect the distribution if they've already been distributed, so I wasn't sure if you mean the accounting of the last, or the future distribution. That clears it up.

EndymionJkb (approved):

PR #1057, opened 2021-12-06
MultiDistributor: extra gas golfing

TomAFrench:
I've created some super simple benchmarking scripts for the remaining MultiDistributor actions and made a few optimisations.

Some micro changes have knocked ~300 gas per subscribed distribution off of most operations. Subscribing and unsubscribing from distributions are now ~2000 gas cheaper as I noticed that we're hitting a cold storage slot whereas we can reuse the SLOAD we use for `distribution.stakingToken`

We still have struct packing optimisations left to look at but that needs #1025 to be in so we can be sure on the bounds of certain fields.


EndymionJkb (approved):
Didn't exhaustively review (e.g., what else might be done), but went through commit by commit and everything looks good

PR #1060, opened 2021-12-06
MultiDistributor: optimise storage layout of Distribution struct

TomAFrench:
This PR massively reduces the costs of all actions which update a subscribed distribution channel but I need to do some thinking about what risks it introduces so it's just a draft for now.

Dumping my thoughts below to get them in order (and so others can poke any holes in it.)

--- 
The `Distribution` struct is now packed so `globalTokensPerStake` and `lastUpdateTime` are squeezed into the same slot avoiding an SSTORE when updating a distribution channel. To do this `globalTokensPerStake` and `paymentRate` are made into `uint192` values.

Both of these are 18 decimal fixed points so ~60 bits of 192 are taken up holding the fractional part of these values. This leaves  enough space for it to hold integer values up to 2**132 = ~5*10^39. Therefore in the extreme cases (assuming 18 decimal tokens):

`paymentRate`: We can distribute 10^39 wei or 10^21 tokens every second.
`globalTokensPerStake`: We can distribute 10^39 wei or 10^21 tokens per every wei of `stakingToken`. Equivalent to 5*10^39 tokens of `distributionToken` for every token of `stakingToken`

These limits seems pretty tough to break. An upper limit on `paymentRate` can be easily enforced so that only leaves `globalTokensPerStake`. A max `paymentRate` with 1 wei staked will overflow `globalTokensPerStake` after 2 seconds but if we just have a single full token staked it will require 10^18 + 1 seconds which is a Long Time T.M.

It's then basically impossible to hit these limits without actively trying to. It also doesn't seem possible to exploit a purposefully created overflow to break the isolation between distribution channels. I need to pump HUGE amounts of tokens into the contract in order to cause an overflow in a reasonable timeframe and if I can get my hands on that amount of tokens then a) I've likely found an infinite mint bug and the token is pretty much worthless anyway now. b) I hold much more of this token than is held in the `MultiDistributor` contract already.

We could potentially pack the values in `UserDistribution` as well which would make claiming tokens cheaper but care needs to be taken that `unclaimedTokens` has enough space for inattentive whales so `userTokensPerStake` would need to be shrinked smaller than `uint192`.

### DIFF
--- a/pkg/distributors/contracts/MultiDistributor.sol
+++ b/pkg/distributors/contracts/MultiDistributor.sol
@@ -287,7 +287,7 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
             // Current distribution period has ended so new period consists only of amount provided.
 
             // By performing fixed point (FP) division of two non-FP values we get a FP result.
-            paymentRate = uint192(FixedPoint.divDown(amount, duration));
+            paymentRate = FixedPoint.divDown(amount, duration);
         } else {
             // Current distribution period is still in progress.
             // Calculate number of tokens that haven't been distributed yet and apply to the new distribution period.
@@ -300,7 +300,7 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
             // Fixed point (FP) multiplication between a non-FP (time) and FP (rate) returns a non-FP result.
             uint256 leftoverTokens = FixedPoint.mulDown(remainingTime, distribution.paymentRate);
             // Fixed point (FP) division of two non-FP values we get a FP result.
-            paymentRate = uint192(FixedPoint.divDown(amount.add(leftoverTokens), duration));
+            paymentRate = FixedPoint.divDown(amount.add(leftoverTokens), duration);
         }
 
         require(paymentRate < type(uint192).max, "PAYMENT_RATE_OVERFLOW");
--- a/pkg/distributors/contracts/MultiDistributor.sol
+++ b/pkg/distributors/contracts/MultiDistributor.sol
@@ -185,7 +185,7 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
     function createDistribution(
         IERC20 stakingToken,
         IERC20 distributionToken,
-        uint256 duration
+        uint64 duration
     ) external override returns (bytes32 distributionId) {
         require(address(stakingToken) != address(0), "STAKING_TOKEN_ZERO_ADDRESS");
         require(address(distributionToken) != address(0), "DISTRIBUTION_TOKEN_ZERO_ADDRESS");
@@ -206,7 +206,7 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
      * @param distributionId The ID of the distribution being modified
      * @param duration Duration over which each distribution is spread
      */
-    function setDistributionDuration(bytes32 distributionId, uint256 duration) external override {
+    function setDistributionDuration(bytes32 distributionId, uint64 duration) external override {
         Distribution storage distribution = _getDistribution(distributionId);
         // These values being guaranteed to be non-zero for created distributions means we can rely on zero as a
         // sentinel value that marks non-existent distributions.
@@ -226,7 +226,7 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
     function _setDistributionDuration(
         bytes32 distributionId,
         Distribution storage distribution,
-        uint256 duration
+        uint64 duration
     ) internal {
         require(duration > 0, "DISTRIBUTION_DURATION_ZERO");
         distribution.duration = duration;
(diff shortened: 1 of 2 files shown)


TomAFrench:
Back of the envelope for packing `UserDistribution`:

```
struct UserDistribution {
    uint128 unclaimedTokens;
    uint128 userTokensPerStake;
}
```

This would allow us to store 2^128 `unclaimedTokens` so ~3*10^20 tokens (~3*10^20 wei) which seems plenty. We'd also be able to distribute 2^(128-60) (~10^20) wei of the `distributionToken` per wei of `stakingToken` which seems to be high enough as well (I can't imagine a situation the number of rewards tokens would be billions of billions times greater than the staking tokens.).

pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -290,11 +291,19 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
             // Fixed point (FP) multiplication between a non-FP (time) and FP (rate) returns a non-FP result.
             uint256 leftoverTokens = FixedPoint.mulDown(remainingTime, distribution.paymentRate);
             // Fixed point (FP) division of two non-FP values we get a FP result.
-            distribution.paymentRate = FixedPoint.divDown(amount.add(leftoverTokens), duration);
+            paymentRate = uint192(FixedPoint.divDown(amount.add(leftoverTokens), duration));
         }
 
-        distribution.lastUpdateTime = block.timestamp;
-        distribution.periodFinish = block.timestamp.add(duration);
+        require(paymentRate < type(uint192).max, "PAYMENT_RATE_OVERFLOW");
```
nventuro, line 297: I think for this require to do anything we need to not cast the result of `divDown` to 192, as the casting will force the value to fit in the type: see [here](https://docs.soliditylang.org/en/v0.8.10/types.html?highlight=higher-order%20bits%20are%20cut%20off#explicit-conversions).

pkg/distributors/contracts/interfaces/IMultiDistributor.sol
```diff
@@ -21,17 +21,38 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/EnumerableSet.so
...
+     *  - uint64 duration;
+     * Slot 4
+     *  - uint256 totalSupply;
+     * Slot 5 (fields always written to together)
+     *  - uint192 paymentRate;
+     *  - uint64 periodFinish;
+     * Slot 6 (fields always written to together)
+     *  - uint192 globalTokensPerStake;
+     *  - uint64 lastUpdateTime;
+     */
```
nventuro, line 45: Have you double-checked that this is the actual slot allocation we get?

pkg/distributors/contracts/MultiDistributor.sol
```diff
@@ -701,30 +710,30 @@ contract MultiDistributor is IMultiDistributor, ReentrancyGuard, MultiDistributo
         returns (uint256 updatedGlobalTokensPerStake)
     {
         updatedGlobalTokensPerStake = _globalTokensPerStake(distribution);
-        distribution.globalTokensPerStake = updatedGlobalTokensPerStake;
-        distribution.lastUpdateTime = _lastTimePaymentApplicable(distribution);
+        distribution.globalTokensPerStake = uint192(updatedGlobalTokensPerStake);
+        distribution.lastUpdateTime = uint64(_lastTimePaymentApplicable(distribution));
```
nventuro, line 714: SInce those function are internal, it is fine to have them return the smaller (and correct) types. The places where we want to keep using uint256 are in the external interfaces, where we simply upcast.

In other words, keep the smaller type internally everywhere, only upcast when exposing it in external getters.

nventuro (commented):
I think your reasoning re. limits is correct both for the distribution and user data, but it'd be good to have more people look at it.

pkg/distributors/contracts/MultiDistributor.sol
TomAFrench: Are you sure? I'd say that the opposite is preferable. Use the "true" types for the external interface to take advantage of the automatic max value checking but use `uint256` internally where we know that the values are safe to keep gas costs low.

pkg/distributors/contracts/interfaces/IMultiDistributor.sol
TomAFrench: I'm not sure how I would test this directly however I get the expected drop of gas costs in the benchmarks that I would expect from this allocation.

Based on this along with the specification as described [here](https://docs.soliditylang.org/en/v0.8.10/internals/layout_in_storage.html) I don't see any reason to doubt that we're getting this configuration but if you know an easy way to check directly then happy to check.

pkg/distributors/contracts/MultiDistributor.sol
TomAFrench: Thanks, I missed removing these casts when adding the overflow check here.

TomAFrench:

PR #1061, opened 2021-12-07
Aave-specialised Linear Pool

TomAFrench:
Draft as `LinearPool.create` only creates one LinearPool type so not all the tests can run at once.


pkg/pool-linear/contracts/aave/AaveLinearPoolFactory.sol
```diff
@@ -0,0 +1,70 @@
...
+        IERC20 mainToken,
+        IERC20 wrappedToken,
+        uint256 lowerTarget,
+        uint256 upperTarget,
+        uint256 swapFeePercentage,
+        address owner
+    ) external returns (LinearPool) {
+        (uint256 pauseWindowDuration, uint256 bufferPeriodDuration) = getPauseConfiguration();
+
+        require(address(mainToken) == IStaticAToken(address(wrappedToken)).ASSET(), "TOKENS_MISMATCH");
```
nventuro, line 44: This check is fantastic, but it belongs in `AaveLinearPool`'s constructor

pkg/pool-linear/contracts/test/MockStaticAToken.sol
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    // solhint-disable-next-line func-name-mixedcase
+    function ASSET() external view override returns (address) {
+        return _ASSET;
+    }
+
+    // solhint-disable-next-line func-name-mixedcase
+    function LENDING_POOL() external view override returns (ILendingPool) {
+        return ILendingPool(this);
```
nventuro, line 42: :joy:

nventuro: This is totally fine as a solution, the alternative approach I would've gone with would be to deploy the lending pool in this contract's constructor.

pkg/pool-linear/contracts/aave/AaveLinearPoolFactory.sol
```diff
@@ -0,0 +1,70 @@
...
+pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
+
+import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
+import "@balancer-labs/v2-pool-utils/contracts/factories/BasePoolSplitCodeFactory.sol";
+import "@balancer-labs/v2-pool-utils/contracts/factories/FactoryWidePauseWindow.sol";
+
+import "./AaveLinearPool.sol";
+
+contract AaveLinearPoolFactory is BasePoolSplitCodeFactory, FactoryWidePauseWindow {
```
nventuro, line 24: I'd remove `MockLinearPoolFactory` and change its tests so that it tests this factory instead of the mock one. Those tests are important as they ensure that the `_create` call is correct, which is doubly critical now that we have the manual `abi.encode` of the raw arguments instead of having the constructor struct (which gave us some type safety).

Once we develop a second type of LinearPool and a corresponding second factory we can think about whether we want to share code among factories or how to reuse the tests, but that's not something we need to worry about now.

nventuro (approved):
In it goes!

PR #1063, opened 2021-12-07
Linear: pack targets and swap fee

nventuro:
This reduces gas in all Linear operations by ~2k (one cold sload).


dmf7z (approved):
LGTM! short and clean

PR #1066, opened 2021-12-08
MultiDistributor: Add working implementation of DistributionScheduler

### LINKED ISSUE #959: MultiDistributor: Reimplement rewards scheduler as a relayer
With the introduction of https://github.com/balancer-labs/balancer-v2-monorepo/issues/922, the rewards scheduler needs to a trusted party for each rewarder. We could rely on the Vault's relayer mechanism to do that.

TomAFrench:
No point in breaking this down really as it's so small. Here's a scheduler.

There's probably some renaming to do before this is properly complete (`scheduleId` is a terrible name) but eh.

fixes #959

### DIFF
--- a/pkg/distributors/contracts/DistributionScheduler.sol
+++ b/pkg/distributors/contracts/DistributionScheduler.sol
@@ -130,7 +130,7 @@ contract DistributionScheduler is IDistributionScheduler {
 
         // Refund tokens to distribution owner.
 
-        distributionChannel.distributionToken.transfer(msg.sender, scheduledDistribution.amount);
+        distributionChannel.distributionToken.safeTransfer(msg.sender, scheduledDistribution.amount);
 
         emit DistributionCancelled(scheduledDistribution.distributionId, scheduleId);
     }
--- a/pkg/distributors/contracts/DistributionScheduler.sol
+++ b/pkg/distributors/contracts/DistributionScheduler.sol
@@ -111,4 +111,36 @@ contract DistributionScheduler is IDistributionScheduler {
             );
         }
     }
+
+    function cancelDistribution(bytes32 scheduleId) external override {
+        ScheduledDistribution memory scheduledDistribution = _scheduledDistributions[scheduleId];
+
+        // Check that scheduled distribution is eligible for cancellation.
+
+        require(scheduledDistribution.status != DistributionStatus.UNINITIALIZED, "Distribution does not exist");
+        require(scheduledDistribution.status != DistributionStatus.STARTED, "Distribution has already started");
+        require(
+            scheduledDistribution.status != DistributionStatus.CANCELLED,
+            "Distribution has already been cancelled"
+        );
+
+        _scheduledDistributions[scheduleId].status = DistributionStatus.CANCELLED;
+
+        // Check that caller is distribution owner and refund tokens.
+
+        IMultiDistributor.Distribution memory distributionChannel = _multiDistributor.getDistribution(
+            scheduledDistribution.distributionId
+        );
+
+        require(distributionChannel.owner == msg.sender, "Only distribution owner can cancel");
+
+        distributionChannel.distributionToken.transfer(msg.sender, scheduledDistribution.amount);
+
+        emit DistributionCancelled(
+            scheduledDistribution.distributionId,
+            scheduleId,
+            scheduledDistribution.startTime,
+            scheduledDistribution.amount
+        );
+    }
 }
--- a/pkg/distributors/contracts/DistributionScheduler.sol
+++ b/pkg/distributors/contracts/DistributionScheduler.sol
@@ -17,41 +17,33 @@ pragma experimental ABIEncoderV2;
 
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/SafeERC20.sol";
 
-import "./interfaces/IMultiDistributor.sol";
+import "./interfaces/IDistributionScheduler.sol";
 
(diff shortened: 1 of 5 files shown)


pkg/distributors/contracts/interfaces/IDistributionScheduler.sol
```diff
@@ -0,0 +1,47 @@
...
+    struct ScheduledDistribution {
+        bytes32 distributionId;
+        uint256 startTime;
+        uint256 amount;
+        DistributionStatus status;
+    }
+
+    event DistributionScheduled(bytes32 indexed distributionId, bytes32 scheduleId, uint256 startTime, uint256 amount);
+    event DistributionStarted(bytes32 indexed distributionId, bytes32 scheduleId);
+    event DistributionCancelled(bytes32 indexed distributionId, bytes32 scheduleId);
```
nventuro, line 32: Might want to index by scheduleId as well.

pkg/distributors/contracts/DistributionScheduler.sol
```diff
@@ -17,104 +17,119 @@ pragma experimental ABIEncoderV2;
...
-    ) public returns (bytes32 scheduleId) {
-        scheduleId = getScheduleId(stakingToken, distributionToken, msg.sender, startTime);
+    ) external override returns (bytes32 scheduleId) {
+        scheduleId = getScheduleId(distributionId, startTime);
         require(startTime > block.timestamp, "Distribution can only be scheduled for the future");
 
         require(
             _scheduledDistributions[scheduleId].status == DistributionStatus.UNINITIALIZED,
             "Distribution has already been scheduled"
         );
```
nventuro, line 62:
```suggestion
        scheduleId = getScheduleId(distributionId, startTime);
        require(
            _scheduledDistributions[scheduleId].status == DistributionStatus.UNINITIALIZED,
            "Distribution has already been scheduled"
        );

        require(startTime > block.timestamp, "Distribution can only be scheduled for the future");
```

very nitpicky

pkg/distributors/contracts/DistributionScheduler.sol
```diff
@@ -17,104 +17,119 @@ pragma experimental ABIEncoderV2;
...
 
+        // As funding pushes out the end timestamp of the distribution channel
+        // we only allow the distribution owner to schedule distributions
```
nventuro, line 64: I think the problem is not so much that the duration is reset (which should be fine), but rather that tokens pending distribution are rolled-over into the new, extended duration (e.g. if there's one day pending worth of rewards, re-starting the distribution with 1 wei would dillute that day worth of tokens over e.g. the full week).

pkg/distributors/contracts/DistributionScheduler.sol
```diff
@@ -17,104 +17,119 @@ pragma experimental ABIEncoderV2;
...
+
+        _scheduledDistributions[scheduleId].status = DistributionStatus.CANCELLED;
+
+        // Check that caller is distribution owner and refund tokens.
+
+        IMultiDistributor.Distribution memory distributionChannel = _multiDistributor.getDistribution(
+            scheduledDistribution.distributionId
+        );
+
+        require(distributionChannel.owner == msg.sender, "Only distribution owner can cancel");
```
nventuro, line 129: The end result is the same, but for the benefit of the reader I'd move this require at the top of the function.

pkg/distributors/contracts/DistributionScheduler.sol
```diff
@@ -17,104 +17,119 @@ pragma experimental ABIEncoderV2;
...
+
+        distributionChannel.distributionToken.transfer(msg.sender, scheduledDistribution.amount);
```
nventuro, line 131: For consistency with the rest of the contract we should use `safeTransfer` here.

pkg/distributors/test/DistributionScheduler.test.ts
```diff
@@ -37,164 +50,164 @@ describe('Distribution Scheduler', () => {
...
-          distributionToken.address,
-          amount,
-          distributionStartTime
-        );
+    const distributionStartTime = MAX_UINT256;
 
-      const event = expectEvent.inReceipt(await tx.wait(), 'DistributionScheduled');
+    context('when called by the distribution owner', () => {
+      context('when scheduling a distribution for the future', () => {
+        context('when there is a previously scheduled conflicting distribution', () => {
```
nventuro, line 77:
```suggestion
        context('when there is no conflicting distribution', () => {
```

pkg/distributors/test/DistributionScheduler.test.ts
```diff
@@ -207,34 +220,109 @@ describe('Distribution Scheduler', () => {
...
 
+      it('skips the distribution', async () => {
+        const receipt = await (await scheduler.connect(other).startDistributions([scheduleId])).wait();
+
+        await expectEvent.notEmitted(receipt, 'DistributionStarted');
+      });
+    });
+  });
+
+  describe('cancelDistribution', () => {
```
nventuro, line 244: Let's add a test where we check starting a cancelled distribution does nothing (it is skipped).

nventuro (approved):
Lovely!

pkg/distributors/contracts/DistributionScheduler.sol
TomAFrench: This results in us getting an access control error on non-existent distributions rather than informing the user that the distribution doesn't exist.

PR #1067, opened 2021-12-08
Get rid of unnecessary FP multiplication

nventuro:
Follow up of https://github.com/balancer-labs/balancer-v2-monorepo/pull/1065


TomAFrench (approved):
Looks good to me.

PR #1070, opened 2021-12-09
MultiDistributor: make any internal variables/functions private

TomAFrench:
Certora reported that this reduces gas costs for free seeing as the `MultiDistributor` is a singleton. I can't replicate this using benchmarks but there's no harm in including it.


EndymionJkb (approved):
They're not virtual, so might as well be private vs internal. Even if we wanted to extend this someday, probably don't want subclasses able to modify the mappings

PR #1071, opened 2021-12-14
New oracle weighted pool

EndymionJkb:
Version of WeightedPool2Tokens that is derived from BaseWeightedPool, eliminating a lot of duplication and cleaning up the codebase. This could be renamed to OracleWeightedPool - but for ease of review I'm keeping the name for now.

Some minor tweaks to BaseWeightedPool were necessary: mostly making some private view methods internal. It uses BasePool's miscData, so doesn't need its own storage and library. Changes to the tests were likewise minimal - just removing a special case that's no longer needed, and adjusting for the fact that _miscData is now 192 bits (not including the swap fee percentage).


nventuro:
I removed the duplication on swaps - we now reuse the base weighted pool code.

I also altered how joins and exits are handled, as the original version had a subtle bug where the balances used to update the oracle were not upscaled. We famously lack tests for tokens with non-18 decimals, so we never caught that.

pkg/pool-weighted/contracts/WeightedPool2Tokens.sol
```diff
@@ -800,7 +341,8 @@ contract WeightedPool2Tokens is
         uint256 balanceToken0,
         uint256 balanceToken1
     ) internal {
-        bytes32 miscData = _miscData;
+        bytes32 miscData = _getMiscData();
+
         if (miscData.oracleEnabled() && block.number > lastChangeBlock) {
```
TomAFrench, line 346:
```suggestion
        if (block.number <= lastChangeBlock) {
            return;
        }
        
        bytes32 miscData = _getMiscData();

        if (miscData.oracleEnabled()) {
```

Seems like we can save an SLOAD for pools with the oracle disabled here.

TomAFrench (approved):
Looks fine to me.

pkg/pool-weighted/contracts/WeightedPool2Tokens.sol
nventuro: Note that the SLOAD would be warm because we also keep the swap fee percentage there, but this is still an improvement. Thanks!

nventuro:
In it goes!

PR #1074, opened 2022-01-05
Stable Pools refactor

EndymionJkb:
Just as WeightedPool2Tokens could be pulled under BaseWeightedPool (and eventually renamed OracleWeightedPool), we can create a BaseStablePool and factor out RateProviders, so that regular Stable Pools can also have rate providers (enabling cDAI/aDAI pools and the like). We would eventually also rename MetaStablePool to OracleStablePool. (Since all Stable Pools would then be "meta," and the distinguishing feature of the current MetaStablePool would be the Oracle.)


EndymionJkb:
Overtaken by events.

PR #1075, opened 2022-01-07
Relayer support for unbutton wrapper

aalavandhan:
Added relayer support for the unbutton wrapper is a ERC-20 wrapper which wraps any generic rebasing ERC-20 token into a non-rebasing, fixed balance version.

Ref: https://github.com/balancer-labs/balancer-sor/pull/224

cc @nventuro


TomAFrench (changes requested):
Hey @aalavandhan, thanks for this PR.

Looking at the tests it seems like we don't actually perform any rebases anywhere. If I add a rebase we start to get a number of test failures so the test suite isn't properly handling the class of tokens it's supposed to handle.

That said, it strikes me that we don't really need to have a rebasing token implementation for the tests as the relayer isn't really concerned with how the wrapper handles the rebasing token. We're not going to hold any AMPL on the relayer past the end of the transaction so as far as it is concerned, AMPL is the same as any other token i.e. the relayer will never experience a rebase. As long as it successfully interacts with the ButtonWrapper then it's working correctly. 

I think the best course of action would be to remove `MockRebasingERC20` and use a standard `TestToken` in its place. Alternatively we should ensure that the test suite is robust enough to handle AMPL experiencing a rebase before calling the relayer.

aalavandhan:
@TomAFrench Thanks for your review. 

I agree with what you point out here. The `MockRebasingERC20` is an unnecessary dependency and I have removed it. The underlying token "rebasing" is outside the scope of the relayer. The relayer just deals with calling the right functions on the wrapper to wrap and unwrap. Thinking back this is one of the reasons why I chose not to add any unit tests using the rebase function. 

FYI, The reason some tests start failing when you add the `rebase` line is because there was a bug in the `MockRebasingERC20` contract. [Here](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1075/commits/359531b672cfe7a71096d30d2f0df167ee9846c0) is a historic commit with the fix and all the tests passing (you can checkout locally to have a look if you're curious).

aalavandhan:
Closing for a new PR combining relayer + linear pool changes

PR #1076, opened 2022-01-11
ctrl-c + ctrl-v gauges

TomAFrench:
Pulling the base contracts into the monorepo so we can make further PRs with necessary features.

Changes from curve-dao-contracts:
- Vyper version has been updated to newer version to take advantage of any optimisations ([breaking changes in 0.3.0](https://vyper.readthedocs.io/en/stable/release-notes.html#v0-3-0) do not affect us)
- CRV inflation information has been split off into `BalancerInflation.vy`, note that this contract will not take into account any minting which occurs outside of the gauge system.
- ~~Delegation has been baked into the veCRV token as taken from [veKP3r](https://github.com/keep3r-network/ff/blob/master/contracts/vKP3R.vy)~~


nventuro (approved):
I've verified that the sources match (minor trivial changes incl. version upgrade) the code verified via Etherscan of the contracts mentioned in https://curve.readthedocs.io/ref-addresses.html

Gauge Controller: 0x2F50D538606Fa9EDD2B11E2446BEb18C9D5846bB
Minter: 0xd061D61a4d941c39E5453435B6345Dc261C2fcE0
Voting Escrow: 0x5f3b5DfEb7B28CDbD7FAba78963EE202a494e2A2

Voting Escrow Delegation is at 0xd30DD0B919cB4012b3AdD78f6Dcb6eb7ef225Ac8, from https://github.com/curvefi/curve-veBoost#deployments

I've also removed LiqGaugeV4 from this PR in favor of https://github.com/balancer-labs/balancer-v2-monorepo/issues/1081

PR #1080, opened 2022-01-15
Add public pool comments

EndymionJkb:
Implementation of Markus's idea, to create an open architecture by which anyone can publicly post information about a pool, and apps can selectively use information based on pool, publisher and content


pkg/pool-utils/test/BasePool.test.ts
```diff
@@ -425,4 +425,19 @@ describe('BasePool', function () {
...
+    let pool: Contract;
+
+    sharedBeforeEach('deploy pool', async () => {
+      pool = await deployBasePool();
+    });
+
+    it('anyone can comment', async () => {
+      const receipt = await (await pool.connect(other).addPoolComment(COMMENT)).wait();
+
+      expectEvent.inReceipt(receipt, 'PoolCommentAdded', { comment: COMMENT });
```
nventuro, line 440:
```suggestion
      expectEvent.inReceipt(receipt, 'PoolCommentAdded', { 
        sender: other.address, 
        poolId: await pool.getPoolId(), 
        comment: COMMENT 
      });
```

nventuro (commented):
Looks good! Not sure about the name, but the idea seems sound.

TomAFrench:
Copying across some thoughts from a conversation on this:

Would this be better to have as a separate "comment registry"? Users would provider a pool address or id along with a comment and the registry would check that this corresponds to a proper pool and then emit the comment in an event if so.

Pros:
- A subgraph or other tracker would only need to watch for events from a single address rather than looking at all pools (and needing to resync when a new factory is added).
- More bytecode budget in the pools themselves.

Cons:
- Anyone writing these comments needs to know the address of this separate contract in order to submit them.

nventuro:
I like that idea a lot, as it will also let us have comments for pools we've already deployed. @EndymionJkb I can take care of this myself if you're busy with Managed Pool work.

EndymionJkb:
> Copying across some thoughts from a conversation on this:
> 
> Would this be better to have as a separate "comment registry"? Users would provider a pool address or id along with a comment and the registry would check that this corresponds to a proper pool and then emit the comment in an event if so.
> 
> Pros:
> 
> * A subgraph or other tracker would only need to watch for events from a single address rather than looking at all pools (and needing to resync when a new factory is added).
> * More bytecode budget in the pools themselves.
> 
> Cons:
> 
> * Anyone writing these comments needs to know the address of this separate contract in order to submit them.

Another Pro is we wouldn't have to redeploy new pool factories to get this functionality. We'd have to deploy this registry, of course, but could then immediately comment on any existing V2 pools.

EndymionJkb:
Superceded by #1093

PR #1085, opened 2022-01-19
Add gauge implementation and factory contract

### LINKED ISSUE #1081: Choose gauge version and make proxyable
A quick scan of [Curve deployments](https://curve.readthedocs.io/ref-addresses.html#liquidity-gauges) seems to indicate there are currently no V4 gauges live. We should look at the differences between V3 and V4 and pick whichever makes the most sense.

Additionally, it seems like we'll need to remove its constructor and transform it into an initializer function, since contract factories are not a thing in Vyper.

TomAFrench:
This PR adds a modified version of the V5 gauge as mentioned in #1081, the main changes are as follows:
- Hardcoded addresses are zeroed out pending deployments
- Variables have been renamed to be Balancer specific
- The gauge's token has been renamed to reference Balancer
- The `Factory.admin()` call has been renamed to `Factory.getAuthoriserAdaptor()`
- Removal of unused events

A factory contract for this gauge has been written in solidity. This factory:
- Allows the deployment of a gauge for an arbitrary ERC20 token 
  - We cannot verify onchain whether it corresponds to a valid pool in a meaningful way due to how permissionless the Vault is so we allow anything. This is fine as long as we do not add these gauges to the gauge controller.
- Allows the authoriser to update the template used for future gauges to allow us to add features in future
  -  Existing gauges are unchanged so there's no risk of rugging. 
- Allows the authoriser to replace the `AuthoriserAdaptor` which has admin powers over the gauges (This contract is added in a separate PR)

fixes #1081


pkg/gauges/contracts/LiquidityGaugeFactory.sol
```diff
@@ -0,0 +1,139 @@
...
+pragma experimental ABIEncoderV2;
+
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/Authentication.sol";
+import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/Clones.sol";
+import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/Ownable.sol";
+import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
+
+import "./interfaces/ILiquidityGauge.sol";
+
+contract LiquidityGaugeFactory is Authentication, Ownable {
```
nventuro, line 25: Do we actually need `Ownable`? We're not using `onlyOwner` anywhere.

pkg/gauges/contracts/LiquidityGaugeFactory.sol
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * @notice Sets the address of the gauge implementation to be used for future deployments
+     */
+    function setGaugeImplementation(address newGaugeImplementation) external authenticate {
+        address currentGaugeImplementation = address(_gaugeImplementation);
+
+        _gaugeImplementation = ILiquidityGauge(newGaugeImplementation);
+        emit GaugeImplementationUpdated(currentGaugeImplementation, newGaugeImplementation);
+    }
```
nventuro, line 96: Couldn't we simply deploy a new factory if we needed to do this?

pkg/gauges/contracts/LiquidityGaugeFactory.sol
```diff
@@ -0,0 +1,139 @@
...
+import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
+
+import "./interfaces/ILiquidityGauge.sol";
+
+contract LiquidityGaugeFactory is Authentication, Ownable {
+    IVault private immutable _vault;
+    ILiquidityGauge private _gaugeImplementation;
+
+    mapping(address => bool) private _isGaugeFromFactory;
+    mapping(address => address) private _poolGauge;
```
nventuro, line 30: Note that:

a) we'll have gauges with no associated pools (e.g. the polygon gauge)
b) multiple pools can technically exist in the same contract (if the same contract calls `registerPool()` multiple times)

TomAFrench: a)
My understanding on this is we'll have a separate factory for gauges for each network. The `_poolGauge` mapping is very useful for L1 pools for onchain lookups in order to join+stake.

Agreed that things are less clear cut for L2s.
 
b) 
If such a contract existed it would have to track ownership in the various pools through either separate ERC20 contracts or using ERC1155 (would have to be wrapped into multiple ERC20s anyway). We're then in a situation which this contract supports.

The only thing that wouldn't work is a pool contract which has multiple subpools and enforces that liquidity provision into all of them is atomic. This design is a bit out there and I don't think we need to worry about it.

pkg/gauges/contracts/LiquidityGaugeFactory.sol
TomAFrench: We could however that opens the possibility of having duplicate gauges for the same pool get deployed. We could add a killswitch to this factory to prevent it deploying gauges for pools from after the migration, however the new factory could deploy new gauges for old pools.

The only way to prevent this would be to always call into the previous factories to check if a gauges already exists.

pkg/gauges/contracts/LiquidityGaugeFactory.sol
TomAFrench: This is to avoid essentially copying all of `Ownable` into this contract in order to track the AuthorizerAdaptor (we can just make it the owner of this contract.)

If it's causing confusion however we can just copy code across.

nventuro: Ahh I understand now - Ownable is essentially storage for an address plus events.

I think I'd still do it manually instead of using Ownable, as Ownable also carries the ownership semantics, and implies the owner can do things other accounts cannot, which is not the case here (at least not as directly as `onlyOwner`).

pkg/gauges/contracts/LiquidityGaugeFactory.sol
nventuro: Right, I see the coordination problem now.

Given that the likelihood of us having to deploy a new gauge contract, or of the gauges being somehow faulty are both very low, I think the current design makes sense. Thanks!

pkg/gauges/contracts/LiquidityGaugeFactory.sol
nventuro: a) I wasn't referring to the actual L2 gauges, but rather the fact that we'll have a gauge that represents all of the Polygon mining (to then bridge tokens over). That gauge will not correspond to any single pool, but rather to the network.

b) Agreed, thanks!

TomAFrench: We can't have a single gauge which represents all the Polygon pools as we need the ability to vote for individual pools on Polygon. Each Polygon pool will have a corresponding gauge on mainnet.

nventuro: You're right, I was confused. Given we have address and ID collisions, I propose then that we use an augmented pool ID that includes the network as the identifier, instead of the plain address.

nventuro: @TomAFrench should we augment pool addresses with chain ID? If so, would we do this in this PR?

TomAFrench: I don't think so, this contract will only be deploying the gauges that will be used for mainnet pools. We'd need a separate factory for Polygon pools, and another for Arbitrum. At that point the factory address is a good disambiguator for which network the gauge is for.

PR #1086, opened 2022-01-19
add AuthoriserAdaptor contract

### LINKED ISSUE #1078: Tie gauge system admin powers into the authoriser
A number of the contracts in the gauge systems refer to a single admin address whereas the Balancer ecosystem permissions actions using the authoriser. Many of these actions are expected to be performed by one-off timelock contracts deployed by the authoriser and so we need a "gateway" for these contracts to go through to control the gauge system.

This gateway contract would have full powers to act on the gauge system but would check that function calls to itself against the authoriser. This maintains the authoriser as being in control of the whole system without any of the gauge contracts having to know about it.

TomAFrench:
As described in the source files.
>This contract is intended to act as an adaptor between systems which expect a single admin address
>and the Balancer Authorizer such that the Authorizer may grant/revoke admin powers to unlimited addresses. 
>
>The permissions the Authorizer can grant are granular such they may be global or specific to a particular contract
>
> When calculating the actionId to call a function on a target contract, it must be calculated as if it were 
> to be called on this adaptor. This can be done by passing the function selector to the `getActionId` function.

This is necessary as we will be having a number of contracts in the gauge system which rely on a single admin address.

We have a little bit of assembly for pulling out the function selector for the call which is to be passed through for which I have added a test to ensure that we're calculating the expected action id.

fixes #1078


pkg/gauges/contracts/AuthorizerAdaptor.sol
```diff
@@ -0,0 +1,105 @@
...
+pragma solidity ^0.7.0;
+
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/IAuthentication.sol";
+import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/Address.sol";
+
+import "@balancer-labs/v2-vault/contracts/interfaces/IAuthorizer.sol";
+import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
+
+/**
+ * @title Authorizer Adaptor
```
nventuro, line 24: :gb:

nventuro: (https://writingexplained.org/adapter-vs-adaptor-difference)

pkg/gauges/contracts/AuthorizerAdaptor.sol
```diff
@@ -0,0 +1,105 @@
...
+ * @notice This contract is intended to act as an adaptor between systems which expect a single admin address
+ * and the Balancer Authorizer such that the Authorizer may grant/revoke admin powers to unlimited addresses. 
+ *
+ * The permissions the Authorizer can grant are granular such they may be global or specific to a particular contract
+ * @dev When calculating the actionId to call a function on a target contract, it must be calculated as if it were
```
nventuro, line 29:
```suggestion
 * The permissions the Authorizer can grant are granular, and they may be global or specific to a particular contract.
 *
 * @dev When calculating the actionId to call a function on a target contract, it must be calculated as if it were 
```

pkg/gauges/contracts/AuthorizerAdaptor.sol
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * @notice Returns the Authorizer
+     */
+    function getAuthorizer() external view returns (IAuthorizer) {
+        return _getAuthorizer();
+    }
+
+    function _getAuthorizer() internal view returns (IAuthorizer) {
+        return getVault().getAuthorizer();
+    }
```
nventuro, line 60:
```suggestion
    function getAuthorizer() public view returns (IAuthorizer) {
        return getVault().getAuthorizer();
    }
```

nventuro: We do have that internal function in some contracts of the Authorization tree, but I'd keep this one simple.

pkg/gauges/contracts/AuthorizerAdaptor.sol
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * @dev As all contracts managed by this adaptor share the same action ID disambiguator
+     *      it's then important to be more granular over `where` an `account` is authorized to act.
+     */
```
nventuro, line 65:
```suggestion

```

pkg/gauges/contracts/AuthorizerAdaptor.sol
```diff
@@ -0,0 +1,105 @@
...
+        return _getAuthorizer().canPerform(actionId, account, where);
+    }
+
+    /**
+     * @notice Returns the action ID associated with calling a given function through this adaptor
+     * @dev The contracts managed by this adaptor do not have action ID disambiguators we use the adaptor's globally
+     * @param selector - The 4 byte selector of the function to be called using `performAction`
+     * @return The associated action ID
+     */
+    function getActionId(bytes4 selector) public view override returns (bytes32) {
```
nventuro, line 76:
```suggestion
    /**
     * @notice Returns the action ID associated with calling a given function through this adaptor.
     * @dev The contracts managed by this adaptor do not have action ID disambiguators: we use the adaptor's globally.
     * This means that contracts with the same function selector will have a matching action ID: if granularity is required
     * then permissions must not be granted globally in the Authorizer.
     *
     * @param selector - The 4 byte selector of the function to be called using `performAction`
     * @return The associated action ID
     */
    function getActionId(bytes4 selector) public view override returns (bytes32) {
```

pkg/gauges/contracts/AuthorizerAdaptor.sol
```diff
@@ -0,0 +1,105 @@
...
+            // The function selector encoded in `data` has an offset relative to the start of msg.data of:
+            // - 4 bytes due to the function selector for `performAction`
+            // - 3 words (3 * 32 = 96 bytes) for `target` and the length and offset of `data`
+            // 96 + 4 = 100 bytes
+            selector := calldataload(100)
+        }
+     
+        _require(_canPerform(getActionId(selector), msg.sender, target), Errors.SENDER_NOT_ALLOWED);
+
+        return target.functionCall(data);
```
nventuro, line 103: Recall that we removed `isContract` from `functionCall`, meaning calls to EOAs will succeed here. I think that's fine, but we may want to at least add a comment about it.

pvt/helpers/src/models/misc/actions.ts
```diff
@@ -1,6 +1,7 @@
 import { Contract } from 'ethers';
+import { Interface } from 'ethers/lib/utils';
 
-export const actionId = (instance: Contract, method: string): Promise<string> => {
-  const selector = instance.interface.getSighash(method);
+export const actionId = (instance: Contract, method: string, contractInterface?: Interface): Promise<string> => {
+  const selector = contractInterface ? contractInterface.getSighash(method) : instance.interface.getSighash(method);
```
nventuro, line 5:
```suggestion
  const selector = (contractInterface ?? instance.interface).getSighash(method);
```

pkg/gauges/contracts/AuthorizerAdaptor.sol
```diff
@@ -0,0 +1,105 @@
...
+        // We want to check that the caller is authorized to call the function on the target rather than this function.
+        // We must then pull the function selector from `data` rather than `msg.sig`.
+        bytes4 selector;
+        assembly {
+            // The function selector encoded in `data` has an offset relative to the start of msg.data of:
+            // - 4 bytes due to the function selector for `performAction`
+            // - 3 words (3 * 32 = 96 bytes) for `target` and the length and offset of `data`
+            // 96 + 4 = 100 bytes
+            selector := calldataload(100)
+        }
```
nventuro, line 99: This is slightly magical - it works because internally `bytes4` is a left-aligned 32 byte value (where the least significant 28 bytes are zeroed-out), and so reading 32 calldata bytes and then keeping the most significant 4 does the right thing.

pkg/gauges/contracts/AuthorizerAdaptor.sol
```diff
@@ -0,0 +1,105 @@
...
+        return keccak256(abi.encodePacked(_actionIdDisambiguator, selector));
+    }
+
+    /**
+     * @notice Performs an arbitrary function call on a target contract, provided the caller is authorized to do so.
+     * @param target - Address of the contract to be called
+     * @param data - Calldata to be sent to the target contract
+     * @return The bytes encoded return value from the performed function call
+     */
+    function performAction(address target, bytes calldata data) external returns (bytes memory) {
```
nventuro, line 89: I'd make this `nonReentrant` and `payable`, forwarding `msg.value` to `target`.

nventuro, line 89: We're implicitly allowing empty `data` (the selector will be 0x00). Does this seem right?

nventuro (approved):
Lovely :ok_hand: :ok_hand:

pkg/gauges/contracts/AuthorizerAdaptor.sol
TomAFrench: It's to balance the Authorizer.

pkg/gauges/contracts/AuthorizerAdaptor.sol
TomAFrench: :magic_wand: :sparkles:

I can't think of a clean way to do this which doesn't end up copying all of `data` into memory though. I guess we're not going to be performing particularly calldata heavy functions through this contract but I like to keep stuff in calldata unless it's really needed.

pkg/gauges/contracts/AuthorizerAdaptor.sol
TomAFrench: Agreed on `nonReentrant`. I don't see much value in making this payable as I can't think of any admin functions which require ETH payments but I can add for generality's sake.

pkg/gauges/contracts/AuthorizerAdaptor.sol
nventuro: Agreed - this is definitely the way to go. I think I intended to suggest maybe adding a comment explaining that, but tbh I don't think it's necessary.

pkg/gauges/contracts/AuthorizerAdaptor.sol
TomAFrench: I think that's fine. It's still requires someone to be authorised to perform a call with an empty function selector and in the case where this happens the worst they can do is send ETH.

TomAFrench: I've added a comment to document this though.

PR #1089, opened 2022-01-21
Mutable management fees

EndymionJkb:


pkg/pool-utils/contracts/controllers/ManagedPoolController.sol
```diff
@@ -77,13 +79,15 @@ contract ManagedPoolController is BasePoolController, IControlledManagedPool {
...
+            .insertBool(managedRights.canSetMustAllowlistLPs, _MUST_ALLOWLIST_LPS_OFFSET);
 
         return
             permissions
-                .insertBool(managedRights.canChangeWeights, _CHANGE_WEIGHTS_OFFSET)
-                .insertBool(managedRights.canDisableSwaps, _DISABLE_SWAPS_OFFSET)
-                .insertBool(managedRights.canSetMustAllowlistLPs, _MUST_ALLOWLIST_LPS_OFFSET)
+                .insertBool(managedRights.canChangeMgmtSwapFee, _CHANGE_MGMT_SWAP_FEE_OFFSET)
                 .insertBool(managedRights.canChangeTokens, _CHANGE_TOKENS_OFFSET)
                 .insertBool(managedRights.canSetCircuitBreakers, _CIRCUIT_BREAKERS_OFFSET);
```
nventuro, line 92: I don't understand this change, why are some done on the first pass and some on the second? Stack too deep?

pkg/pool-weighted/test/ManagedPool.test.ts
```diff
@@ -690,6 +691,26 @@ describe('ManagedPool', function () {
...
+      describe('set management fee', () => {
+        context('when the sender is not the owner', () => {
+          it('non-owners cannot set the management fee', async () => {
+            await expect(
+              pool.setManagementSwapFeePercentage(other, NEW_MANAGEMENT_SWAP_FEE_PERCENTAGE)
+            ).to.be.revertedWith('SENDER_NOT_ALLOWED');
+          });
+        });
+
+        context('when the sender is the owner', () => {
```
nventuro, line 703:
```suggestion
        context('when the sender is the owner', () => {
          it('the management fee can be set', async () => {
            await pool.setManagementSwapFeePercentage(owner, NEW_MANAGEMENT_SWAP_FEE_PERCENTAGE);
            expect(await pool.getManagementSwapFeePercentage()).to.equal(NEW_MANAGEMENT_SWAP_FEE_PERCENTAGE);
          });

```

nventuro (approved):

pkg/pool-utils/contracts/controllers/ManagedPoolController.sol
EndymionJkb: > I don't understand this change, why are some done on the first pass and some on the second? Stack too deep?

Yes; that's always the problem when it looks weird. Adding one more was too many. I could add a comment.

PR #1091, opened 2022-01-22
Test the oracle pool with non-18-decimal tokens

EndymionJkb:
Was working on this when the PR got merged out from under me :)

It might not be the most elegant way to do this, but it tests both an 18 and non-18 decimal token, verifying that the join/exit logic is now correct.


nventuro (approved):
Thanks!

PR #1097, opened 2022-01-26
Fix and add small comments

nventuro:


pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -48,6 +48,8 @@ import "./WeightCompression.sol";
  * IL-protection, and more.
  */
 contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
+    // LiquidityBootstrappingPool change their weights over time: these periods are expected to be long enough (e.g.
+    // days) that any timestamp manipulation would achieve very little.
```
EndymionJkb, line 52:
```suggestion
    // LiquidityBootstrappingPools change their weights over time: these periods are expected to be long enough
    // (e.g., days) that any timestamp manipulation would achieve very little.
```

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -23,10 +23,11 @@ import "../BaseWeightedPool.sol";
 import "./WeightCompression.sol";
 
 /**
- * @dev Weighted Pool with mutable weights, designed to support V2 Liquidity Bootstrapping
+ * @dev Weighted Pool with mutable weights, designed to support V2 Liquidity Bootstrapping.
  */
 contract LiquidityBootstrappingPool is BaseWeightedPool, ReentrancyGuard {
-    // The Pause Window and Buffer Period are timestamp-based: they should not be relied upon for sub-minute accuracy.
+    // LiquidityBootstrappingPool change their weights over time: these periods are expected to be long enough (e.g.
+    // days) that any timestamp manipulation would achieve very little.
```
EndymionJkb, line 30:
```suggestion
    // LiquidityBootstrappingPools change their weights over time: these periods are expected to be long enough
    // (e.g., days) that any timestamp manipulation would achieve very little.
```

pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol
```diff
@@ -23,10 +23,11 @@ import "../BaseWeightedPool.sol";
 import "./WeightCompression.sol";
 
 /**
- * @dev Weighted Pool with mutable weights, designed to support V2 Liquidity Bootstrapping
+ * @dev Weighted Pool with mutable weights, designed to support V2 Liquidity Bootstrapping.
```
EndymionJkb, line 26: Do we still need to say "V2" at this point?

EndymionJkb (commented):
Do we still need to say "V2" at this point?

pkg/pool-weighted/contracts/smart/ManagedPool.sol
nventuro: Huh, interesting. [According to this article](https://jakubmarian.com/comma-after-i-e-and-e-g/), this is a British vs American thing. We're doing brit style throughout the entire process, so I'd keep things consistent here and go with the Queen's grammar.

PR #1102, opened 2022-01-27
ManagedPool weight storage

EndymionJkb:
As a first step toward add/remove functionality, allow the sum of ManagedPool weights to diverge from ONE.


pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -721,4 +771,22 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
         // A valid token can't be zero (must have non-zero weights)
         _require(tokenData != 0, Errors.INVALID_TOKEN);
     }
+
+    function _totalWeightEqualsOne() private view returns (bool) {
+        return FixedPoint.ONE == _totalWeight;
+    }
```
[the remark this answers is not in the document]
nventuro: Or actually even remove the optimization altogether, since often the weights will not add up to one.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -349,6 +356,35 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
...
+     */
+    function _setTotalWeight(uint256[] memory normalizedWeights, uint256 totalWeight) internal {
+        InputHelpers.ensureInputLengthMatch(_getTotalTokens(), normalizedWeights.length);
+        _require(totalWeight > WeightedMath._MIN_WEIGHT, Errors.MIN_WEIGHT);
+
+        if (totalWeight == _totalWeight) {
+            return;
+        }
+
+        _totalWeight = totalWeight;
```
nventuro, line 379: It seems like this achieves nothing. What's the point of the `totalWeight` param?

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -637,25 +673,18 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
...
         IERC20[] memory tokens
     ) internal virtual {
+        uint256 totalWeight = _totalWeight;
         uint256 normalizedSum = 0;
-        bytes32 tokenState;
 
         for (uint256 i = 0; i < endWeights.length; i++) {
             uint256 endWeight = endWeights[i];
             _require(endWeight >= WeightedMath._MIN_WEIGHT, Errors.MIN_WEIGHT);
+            normalizedSum = normalizedSum.add(endWeight);
```
nventuro, line 682: If these are normalized, that should be made clear. I'd call these `normalizedStartWeights`, `normalizedWeight`, etc.

nventuro: Also it seems a bit odd and inefficient to normalize everything and then denormalize. Why not just check that the sum adds up to `_totalWeight` instead? That also means we won't have weird precision issues in the normalization/denormalization process.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -87,20 +87,25 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
 
     // Store scaling factor and start/end weights for each token
     // Mapping should be more efficient than trying to compress it further
-    // [ 155 bits|   5 bits |  32 bits   |   64 bits    |
-    // [ unused  | decimals | end weight | start weight |
-    // |MSB                                          LSB|
+    // [ 123 bits |  5 bits  |  64 bits   |   64 bits    |
+    // [ unused   | decimals | end weight | start weight |
+    // |MSB                                           LSB|
```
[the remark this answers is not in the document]
EndymionJkb: I'm not sure we really need to, but I thought it would make it more accurate/consistent (e.g., when we copy end weights to start weights), now that we could have a wider range of values

pkg/pool-weighted/contracts/smart/ManagedPool.sol
EndymionJkb: It's a placeholder for testing. This won't be in the final code. It's a way to simulating adding and removing, which would affect the sum, by just directly setting it.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -721,4 +771,22 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
...
+
+    // Convert from the internal representation to normalized weights (summing to ONE)
+    function _downscaleWeight(uint256 weight) private view returns (uint256) {
+        return _totalWeightEqualsOne() ? weight : weight.divDown(_totalWeight);
+    }
+
+    // converts from normalized form to the internal representation (summing to _totalWeight)
+    function _upscaleWeight(uint256 weight) private view returns (uint256) {
+        return _totalWeightEqualsOne() ? weight : weight.mulUp(_totalWeight);
+    }
```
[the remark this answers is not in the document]
EndymionJkb: I had it that way originally and went back and forth. Ok; will put back the way I had it

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -355,6 +362,35 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
         emit SwapEnabledSet(swapEnabled);
     }
 
+    /**
+     * @dev Getter for the sum of all weights. In initially FixedPoint.ONE, it can be higher or lower
+     * as a result of adds and removes.
+     */
```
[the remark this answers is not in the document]
EndymionJkb: I added this both for general purpose visibility (e.g., debugging/testing), and in case it might be valuable for the SOR or UI to quickly check if it's non-one in case they need to process those differently for some reason. Will add the comment.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
EndymionJkb: Usually when this gets called, the values are passed in from outside: in the constructor, or a call to updateWeightsGradually, and those we do want to sum to 1, to hide the internal normalization.

The only place it normalizes and denormalizes again is when you call updateWeightsGradually.

I can't avoid normalizing, because normalized weights are needed for the event.

Neither can I avoid denormalizing by simply copying compressed and denormalized end weights to start weights, since it could be in the middle of an ongoing change.

This kind of thing is why I thought it would be best to extend end weights to 64 bits so they both have the same precision.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -231,7 +243,9 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
         endWeights = new uint256[](totalTokens);
 
         for (uint256 i = 0; i < totalTokens; i++) {
-            endWeights[i] = _tokenState[tokens[i]].decodeUint32(_END_WEIGHT_OFFSET).uncompress32();
+            endWeights[i] = _normalizeWeight(
+                _tokenState[tokens[i]].decodeUint64(_END_DENORM_WEIGHT_OFFSET).uncompress64(_MAX_DENORM_WEIGHT_SUM)
+            );
```
dmf7z, line 248: I would be useful to add a comment explaining why `uncompress64(_MAX_DENORM_WEIGHT_SUM)`

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -349,6 +363,35 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
...
+        return _denormWeightSum;
+    }
+
+    /**
+     * @dev Placeholder function for testing. Adding or removing a token may cause the total weight to
+     * diverge from ONE.
+     */
+    function _setDenormWeightSum(uint256[] memory normalizedWeights, uint256 denormWeightSum) internal {
+        InputHelpers.ensureInputLengthMatch(_getTotalTokens(), normalizedWeights.length);
+        _require(denormWeightSum > WeightedMath._MIN_WEIGHT, Errors.MIN_WEIGHT);
```
dmf7z, line 380: Shouldn't it be:

`uint256 minWeightSum = _getTotalTokens() * WeightedMath._MIN_WEIGHT;`
`_require(denormWeightSum >= minWeightSum, Errors.MIN_WEIGHT); `

EndymionJkb: That would be more accurate, but this is just a placeholder function for testing that will go away, so I wasn't too worried about it. The weight sum will actually change as a result of add/remove operations, and all the real checks would be done in those functions.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -359,26 +365,25 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
...
     /**
      * @dev Getter for the sum of all weights. In initially FixedPoint.ONE, it can be higher or lower
-     * as a result of adds and removes. Production code should not need to call this function, as the
-     * value it returns is only used internally for weight.
+     * as a result of adds and removes.
      */
-    function getWeightSum() external view returns (uint256) {
-        return _weightSum;
+    function _getDenormWeightSum() internal view returns (uint256) {
+        return _denormWeightSum;
```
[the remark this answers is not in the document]
EndymionJkb: I think this was already changed?

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -83,27 +83,33 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
...
     // We need to work with normalized weights (i.e. they should add up to 100%), but storing normalized weights
     // would require updating all weights whenever one of them changes, for example in an add or remove token
-    // operation. Instead, we keep track of the sum of all weights, and dynamically normalize them by dividing
-    // each weight by the total sum.
-    uint256 private _weightSum = FixedPoint.ONE;
+    // operation. Instead, we keep track of the sum of all denormalized weights, and dynamically normalize them
+    // for I/O by multiplying or dividing by the `_denormWeightSum`.
+    //
+    // In this contract, "weights" mean normalized weights, and "denormWeights" refer to how they are stored internally.
+    uint256 private _denormWeightSum = FixedPoint.ONE;
```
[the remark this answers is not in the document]
EndymionJkb: I moved this to the constructor and commented. I don't want to iterate twice - I know the initial sum is 1 after the _startGradualWeightChange call

nventuro (approved):

PR #1103, opened 2022-01-27
Use immutable variables in vyper contracts

### LINKED ISSUE #1098: Remove hardcoded values
Given [Vyper 3.0 supports `immutable`](https://vyper.readthedocs.io/en/stable/scoping-and-declarations.html#declaring-immutable-variables), we should use it to remove hardcoded values, such as the BAL address.

TomAFrench:
As we're updating the vyper version used by the contracts we can make use of immutable variables. I've gone through all the addresses are hardcoded or held in storage and set them to be immutable variables. Immutable variables don't get getters by default so I've created explicit getters to replace those automatically generated before.

I've also taken the opportunity to pull addresses from onchain where possible to enforce consistency and remove a redundant variable. 

Closes #1098

### DIFF
--- a/pkg/gauges/contracts/GaugeController.vy
+++ b/pkg/gauges/contracts/GaugeController.vy
@@ -68,7 +68,7 @@ MULTIPLIER: constant(uint256) = 10 ** 18
 admin: public(address)  # Can and will be a smart contract
 future_admin: public(address)  # Can and will be a smart contract
 
-TOKEN: immutable(address) # BPT token
+TOKEN: immutable(address) # 80-20 BAL-WETH BPT token
 VOTING_ESCROW: immutable(address)  # Voting escrow
 
 # Gauge parameters
--- a/pkg/gauges/contracts/GaugeController.vy
+++ b/pkg/gauges/contracts/GaugeController.vy
@@ -67,8 +67,8 @@ MULTIPLIER: constant(uint256) = 10 ** 18
 admin: public(address)  # Can and will be a smart contract
 future_admin: public(address)  # Can and will be a smart contract
 
-token: public(address)  # CRV token
-voting_escrow: public(address)  # Voting escrow
+TOKEN: immutable(address) # BPT token
+VOTING_ESCROW: immutable(address)  # Voting escrow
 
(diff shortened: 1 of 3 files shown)


pkg/gauges/contracts/GaugeController.vy
```diff
@@ -67,8 +68,8 @@ MULTIPLIER: constant(uint256) = 10 ** 18
 admin: public(address)  # Can and will be a smart contract
 future_admin: public(address)  # Can and will be a smart contract
 
-token: public(address)  # CRV token
-voting_escrow: public(address)  # Voting escrow
+TOKEN: immutable(address) # BPT token
```
nventuro, line 71:
```suggestion
TOKEN: immutable(address) # 80-20 BAL-ETH BPT token
```

There are other BPTs referenced in the system (the BPTs that are staked), so it doesn't hurt to clarify.

pkg/gauges/contracts/GaugeController.vy
```diff
@@ -592,4 +599,4 @@ def get_weights_sum_per_type(type_id: int128) -> uint256:
     @param type_id Type id
     @return Sum of gauge weights
     """
-    return self.points_sum[type_id][self.time_sum[type_id]].bias
\ No newline at end of file
+    return self.points_sum[type_id][self.time_sum[type_id]].bias
```
nventuro: praise be

pkg/gauges/contracts/VotingEscrow.vy
```diff
@@ -111,27 +110,44 @@ future_admin: public(address)
...
 
 @external
-def __init__(token_addr: address, _name: String[64], _symbol: String[32], _version: String[32]):
+def __init__(token_addr: address, _name: String[64], _symbol: String[32]):
     """
     @notice Contract constructor
     @param token_addr `ERC20CRV` token address
     @param _name Token name
     @param _symbol Token symbol
-    @param _version Contract version - required for Aragon compatibility
```
nventuro, line 120: Thanks people at Curve for this comment

nventuro (approved):

PR #1105, opened 2022-01-29
[WIP] Usdplus pool

mosinnik:
For review

### DIFF
--- a/pkg/pool-linear/contracts/usdplus/UsdPlusLinearPool.sol
+++ b/pkg/pool-linear/contracts/usdplus/UsdPlusLinearPool.sol
@@ -1,0 +1,61 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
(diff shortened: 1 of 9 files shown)


pkg/pool-linear/contracts/usdplus/UsdPlusLinearPool.sol
```diff
@@ -0,0 +1,61 @@
...
+            bufferPeriodDuration,
+            owner
+        )
+    {
+        _require(address(mainToken) == IStaticUsdPlusToken(address(wrappedToken)).mainToken(), Errors.TOKENS_MISMATCH);
+        _wrappedToken = IStaticUsdPlusToken(address(wrappedToken));
+    }
+
+    function _getWrappedTokenRate() internal view override returns (uint256) {
+        uint256 rate = _wrappedToken.rate();
```
gtaschuk, line 55: If we wanted to make this pool generically EIP4626 compatible, we could use function signature assetsPerShare() here and in the wrapper instead of rate() and scale the number appropriately

I think that function name is unlikely to change at this point 

https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4626.md

https://ethereum-magicians.org/t/eip-4626-yield-bearing-vault-standard/7900/46

gtaschuk: Additionally would require changing the wrapper mainToken() function to underlying()

gtaschuk: Here is a WIP branch that generalizes for 4626: https://github.com/balancer-labs/balancer-v2-monorepo/compare/usdplus-erc4626-pool

mosinnik: Looks good, 2 methods on StaticUsdPlus already may be renamed to be compatible.
But here we need fix rate issue corresponding to current assetsPerShare docs: https://github.com/ovnstable/balancer-v2-monorepo/commit/1e32d133075cf7eac88bb448d03c6393ce41e834

gtaschuk:
Closing in favor of #1120 .  Great initial work here

PR #1108, opened 2022-02-01
Managed Pool: Remove token

EndymionJkb:
Implementation of removeToken: sets allowUnlimitedContractSize for tests.


pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -349,6 +352,100 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
...
+    /**
+     * @dev Remove a token from the pool. This is a permissioned function that performs the following operations:
+     * - Verify there are enough tokens, and there is no ongoing weight change
+     * - Calculate the BPT value of the full token balance, and returns it for possible use by the caller
+     * - Exit the pool, returning the full balance of the token to the recipient
+     * - Deregister the token in the Vault (since it now has zero balance)
+     * - Reduce the total weight by the weight of the token removed, and update the number of tokens and `_tokenState`
+     *   (all other weights then scale accordingly)
+     */
+    function removeToken(uint256 tokenIndex, address recipient) external authenticate whenNotPaused returns (uint256) {
```
nventuro, line 364: Not sure about this, multiple calls to remove or add tokens will shift indices around. I'd have `token` in the interface instead of the index.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -116,6 +118,7 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
     event ManagementFeesCollected(IERC20[] tokens, uint256[] amounts);
     event AllowlistAddressAdded(address indexed member);
     event AllowlistAddressRemoved(address indexed member);
+    event TokenRemoved(IERC20 indexed token);
```
nventuro, line 121: Let's also add the amount.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -349,6 +352,100 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
...
+        //(uint256 bptAmountIn, uint256 minAmountOut, uint256 normalizedWeight) = getTotalBptForToken(tokenIndex);
+        // Exit the pool, returning the full balance of the token to the recipient
+        (IERC20[] memory tokens, uint256[] memory rawBalances, ) = getVault().getPoolTokens(getPoolId());
+        _require(tokenIndex < tokens.length, Errors.OUT_OF_BOUNDS);
+        _require(tokens.length > 2, Errors.MIN_TOKENS);
+        // Verify there is no ongoing weight change
+        _require(
+            0 == _calculateWeightChangeProgress() || FixedPoint.ONE == _calculateWeightChangeProgress(),
+            Errors.REMOVE_TOKEN_DURING_WEIGHT_CHANGE
+        );
```
nventuro, line 379:
```suggestion
        {
            uint256 weightChangeProgress = _calculateWeightChangeProgress();
            _require(weightChangeProgress == 0 || weightChangeProgress == FixedPoint.ONE);
        }
```

nventuro: Can we remove a token if there is a scheduled weight change in the future?

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -349,6 +352,100 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
...
+     * - Reduce the total weight by the weight of the token removed, and update the number of tokens and `_tokenState`
+     *   (all other weights then scale accordingly)
+     */
+    function removeToken(uint256 tokenIndex, address recipient) external authenticate whenNotPaused returns (uint256) {
+        return _removeToken(tokenIndex, recipient);
+    }
+
+    function _removeToken(uint256 tokenIndex, address recipient) internal returns (uint256) {
+        // Calculate the BPT value of the full token balance (other values returned to avoid duplication)
+        //(uint256 bptAmountIn, uint256 minAmountOut, uint256 normalizedWeight) = getTotalBptForToken(tokenIndex);
```
nventuro, line 370:
```suggestion
```

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -349,6 +352,100 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
...
+
+        // Clean up data structures and update the token count
+        delete _tokenState[removedToken];
+        _setMiscData(_getMiscData().insertUint7(tokens.length - 1, _TOTAL_TOKENS_OFFSET));
+
+        // Decrease the total weight by the weight of the token being removed
+        _totalWeight -= weightOfRemovedToken;
+
+        // The bpt amount corresponding to the token is simply its proportional share of the totalSupply
+        return weightOfRemovedToken.mulDown(totalSupply());
```
nventuro, line 406: We need to multiply by the normalized weight, not the raw one. And we need to normalize using the old weight sum (prior to removal).

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -349,6 +352,100 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
...
+        IERC20[] memory tokens,
+        uint256[] memory normalizedWeights,
+        uint256 tokenIndex,
+        uint256 scalingFactor,
+        uint256 unscaledTokenAmountOut,
+        address recipient
+    ) private returns (uint256) {
+        uint256[] memory minAmountsOut = new uint256[](tokens.length);
+        // In the limit, compensate for BasePool downscaling of amountOut
+        minAmountsOut[tokenIndex] = _downscaleUp(unscaledTokenAmountOut, scalingFactor);
```
nventuro, line 420: Limits are checked in the Vault, where there is no scaling. So we shouldn't have to change anything here, and we specially shouldn't downscale an unscaled amount (we only downscale upscaled amounts).

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -349,6 +352,100 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
...
+            getPoolId(),
+            address(this),
+            payable(recipient),
+            IVault.ExitPoolRequest({
+                assets: _asIAsset(tokens),
+                minAmountsOut: minAmountsOut,
+                userData: abi.encode(
+                    WeightedPoolUserData.ExitKind.REMOVE_TOKEN,
+                    tokenIndex,
+                    _upscale(unscaledTokenAmountOut, scalingFactor) // Upscale raw amount for Vault
```
nventuro, line 432: We always pass unscaled amounts in user data, so I wouldn't upscale anything here.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -584,6 +684,28 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
...
+        // call can be made resides in withdrawCollectedManagementFees.
+        _require(sender == address(this), Errors.UNAUTHORIZED_EXIT);
+
+        // No BPT is required for the exit operation itself. The `removeToken` function calculates and returns
+        // the bptAmountIn, but leaves any further action, such as burning BPT, up to the caller.
+        bptAmountIn = 0;
+
+        (uint256 tokenIndex, uint256 amountOut) = userData.removeToken();
+
+        amountsOut = new uint256[](_getTotalTokens());
```
nventuro, line 705: Here is were we need to upscale, as we do in all other join/exit handlers. That upscale will cancel out the downscale that BasePool does when we return (but more generally, inside these handlers we always work with upscaled amounts).

pkg/pool-weighted/contracts/smart/ManagedPool.sol
EndymionJkb: I considered this; it just made the interfaces a bit harder, but I agree it would be safer. Ok; can just convert it to an index on the first step.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
EndymionJkb: Technically yes, at the pool level, since it's happening immediately and won't interfere. There could be a rebalancing or something involving other tokens that you might still want to happen.

They could always just issue another updateWeightsGradually to stop it if they forgot or it was screwing up; it would be safer to prevent it, though.

At the controller level, if the remove is happening after a delay (TBD where the delays are; maybe they're only on adds), maybe the controller would want to block scheduling a removal if there was a future weight change: and similarly block calls to updateWeightsGradually if there was a pending removal.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
EndymionJkb: It is doing that - it's calling _getNormalizedWeights before doing the remove or adjusting the total

pkg/pool-weighted/contracts/smart/ManagedPool.sol
EndymionJkb: That was leftover from something; removed

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -327,10 +328,92 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+        uint256 burnAmount
+    ) external authenticate nonReentrant whenNotPaused {
+        // Exit the pool, returning the full balance of the token to the recipient
+        (IERC20[] memory tokens, uint256[] memory unscaledBalances, ) = getVault().getPoolTokens(getPoolId());
+        _require(tokens.length > 2, Errors.MIN_TOKENS);
+
+        // Tokens cannot be removed during or before a weight change to reduce complexity of weight interactions
+        _ensureNoWeightChange();
+
+        uint256 tokenIndex = _tokenAddressToIndex(token);
```
TomAFrench, line 352: Is there a reason why we reread the pool's tokens within this function rather than passing in `tokens`?

TomAFrench, line 352:
```suggestion
        // Reverts if token does not exist in pool.
        uint256 tokenIndex = _tokenAddressToIndex(token);
```

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -568,27 +651,64 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+        // This exit function is disabled if the contract is paused.
+
+        // This exit function can only be called by the Pool itself - the authorization logic that governs when that
+        // call can be made resides in removeToken.
+        _require(sender == address(this), Errors.UNAUTHORIZED_EXIT);
+
+        uint256 tokenIndex = userData.removeToken();
+
+        // No BPT is required to remove the token - it is up to the caller do determine under which conditions removing
+        // a token makes sense, and if e.g. burning BPT is required.
```
TomAFrench, line 693:
```suggestion
        // No BPT is required to remove the token - it is up to the caller to determine under which conditions removing
        // a token makes sense, and if e.g. burning BPT is required.
```

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -568,27 +651,64 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+        uint256 bptAmountIn = 0;
+
+        uint256[] memory amountsOut = new uint256[](_getTotalTokens());
```
TomAFrench, line 696:
```suggestion
        uint256[] memory amountsOut = new uint256[](balances.length);
```

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -326,6 +327,89 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+            getPoolId(),
+            address(this),
+            payable(recipient),
+            IVault.ExitPoolRequest({
+                assets: _asIAsset(tokens),
+                minAmountsOut: minAmountsOut,
+                userData: abi.encode(WeightedPoolUserData.ExitKind.REMOVE_TOKEN, tokenIndex),
+                toInternalBalance: false
+            })
+        );
```
TomAFrench, line 372:
```suggestion
        getVault().exitPool(
            getPoolId(),
            address(this),
            payable(recipient),
            IVault.ExitPoolRequest({
                assets: _asIAsset(tokens),
                minAmountsOut: new uint256[](tokens.length),
                userData: abi.encode(WeightedPoolUserData.ExitKind.REMOVE_TOKEN, tokenIndex),
                toInternalBalance: false
            })
        );
```

L360 doesn't actually do anything as the pools balances would have been manipulated before we enter this function (and so before we read `tokenBalance`).

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -326,6 +327,89 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+     * is also forbidden during a weight change, or if one is scheduled to happen in the future.
+     *
+     * The caller may aditionally pass a non-zero `burnAmount` to have some of their BPT be burned, which might be
+     * useful in some scenarios to account for the fact that the Pool now has fewer tokens.
+     */
+    function removeToken(
+        IERC20 token,
+        address recipient,
+        uint256 burnAmount
+    ) external authenticate nonReentrant whenNotPaused {
```
TomAFrench, line 344: As mentioned in the previous comment, we don't have any guarantees on how much of `token` we'll receive. We're then committing to burn `burnAmount` BPT of value in return for a substantially reduced amount of `token`.

It seems like we should allow the pool owner to provide minimum amount of `token` they would accept receiving to avoid losing value to MEV.

TomAFrench: Granted, this could be done in a pool manager contract but performing this in the pool is a one-liner.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -326,6 +327,89 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+        // Exit the pool, returning the full balance of the token to the recipient
+        (IERC20[] memory tokens, uint256[] memory unscaledBalances, ) = getVault().getPoolTokens(getPoolId());
+        _require(tokens.length > 2, Errors.MIN_TOKENS);
+
+        // Tokens cannot be removed during or before a weight change to reduce complexity of weight interactions
+        _ensureNoWeightChange();
+
+        // Reverts if token does not exist in pool.
+        uint256 tokenIndex = _tokenAddressToIndex(tokens, token);
+        uint256 tokenBalance = unscaledBalances[tokenIndex];
```
TomAFrench, line 358:
```suggestion
    ) external authenticate nonReentrant whenNotPaused returns (uint256 tokenAmountOut) {
        // Exit the pool, returning the full balance of the token to the recipient
        (IERC20[] memory tokens, uint256[] memory unscaledBalances, ) = getVault().getPoolTokens(getPoolId());
        _require(tokens.length > 2, Errors.MIN_TOKENS);

        // Tokens cannot be removed during or before a weight change to reduce complexity of weight interactions
        _ensureNoWeightChange();

        // Reverts if token does not exist in pool.
        uint256 tokenIndex = _tokenAddressToIndex(tokens, token);
        tokenAmountOut = unscaledBalances[tokenIndex];
```

This will make it easier for Pool managers to know how many tokens they receive (as opposed to before+after balance checks)

pkg/pool-weighted/contracts/smart/ManagedPool.sol
EndymionJkb: Ok; will put it in

pkg/pool-weighted/contracts/smart/ManagedPool.sol
EndymionJkb: Addressed below

TomAFrench: Marking this as resolved as this will be addressed by https://github.com/balancer-labs/balancer-v2-monorepo/pull/1108#discussion_r857824669

pkg/pool-weighted/contracts/smart/ManagedPool.sol
TomAFrench: I spoke with nico about this and his feelings are that with https://github.com/balancer-labs/balancer-v2-monorepo/pull/1108#discussion_r857835184 we can just leave it to the pool manager contract to handle this.

nventuro: My intention was to add a redundant check that would guarantee the entire balance is removed from the Pool, and not a mitigation against MEV (since we're unable to do such a thing at this layer). However, my extra check is actually unnecessary since `deregisterTokens` would simply fail if the balance was not zeroed out.

Given that we have the `minimumAmountsOut` array already, I think it makes sense to add an argument and handle it here, as Jeff did. It is a bit more consistent with how `exitPool` works as well. However, let's please avoid having multiple people performing edits on the same branch: it makes it extremely annoying to update them.

nventuro: (also new features need tests as well, not just contract code)

pkg/pool-weighted/contracts/smart/ManagedPool.sol
nventuro: Great suggestion, yes. I'll add it in a slightly different way as I don't like output variables :upside_down_face:

PR #1110, opened 2022-02-02
Fix incorrect error

nventuro:
This has no effect on the on-chain code, as the error code is the same. It is just the description that was inaccurate.

edit: we actually did use this error code somewhere legitimately, so I added a new one for the balance scenario.


TomAFrench (dismissed):

pkg/solidity-utils/contracts/openzeppelin/ERC20Burnable.sol
```diff
@@ -33,7 +33,7 @@ abstract contract ERC20Burnable is ERC20 {
      * `amount`.
      */
     function burnFrom(address account, uint256 amount) public virtual {
-        uint256 decreasedAllowance = allowance(account, msg.sender).sub(amount, Errors.ERC20_BURN_EXCEEDS_ALLOWANCE);
+        uint256 decreasedAllowance = allowance(account, msg.sender).sub(amount, Errors.ERC20_BURN_EXCEEDS_BALANCE);
```
TomAFrench, line 36: Seems like we've just moved the error onto this line instead.

nventuro: That's what I get for blindly copy-pasting :facepalm:

TomAFrench (approved):

PR #1116, opened 2022-02-03
Authorizer: Refactor implementation

### LINKED ISSUE #985: Rename AccessControl's roles to actions

facuspagnuolo:
Fixes #985 

This PR introduces a simplification for the authorizer implementation. Mainly what was proposed in #985 to rename "roles" to "permissions". Additionally, I propose a different way of implementing the idea of an admin using the authorizer itself.


pkg/vault/contracts/Authorizer.sol
```diff
@@ -32,258 +32,139 @@ import "./interfaces/IAuthorizer.sol";
 contract Authorizer is IAuthorizer {
     using EnumerableSet for EnumerableSet.AddressSet;
```
nventuro, line 33: You're removing usage of EnumerableSet in favor of plain mappings. If we were to do that, we should also get rid of the import and `using` statement.

However, I think we should keep EnumerableSet, especially in this context, as it can be very useful and powerful for automated permission migrations. See for example how we rely on it to [migrate permissions from the BAL token to the BalancerTokenAdmin](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/gauges/contracts/BalancerTokenAdmin.sol#L86).

pkg/vault/contracts/Authorizer.sol
```diff
@@ -32,258 +32,139 @@ import "./interfaces/IAuthorizer.sol";
...
-        address[] calldata where
+        address[] memory where
     ) external {
-        _require(where.length > 0, Errors.INPUT_LENGTH_MISMATCH);
-        for (uint256 i = 0; i < roles.length; i++) {
-            _require(isAdmin(roles[i], msg.sender), Errors.GRANT_SENDER_NOT_ADMIN);
+        for (uint256 i = 0; i < actions.length; i++) {
             for (uint256 j = 0; j < where.length; j++) {
-                _grantRole(roles[i], account, where[j]);
+                grantPermission(actions[i], account, where[j]);
```
nventuro, line 94: I'm generally not a big fan of calling public non-view functions from other functions, as you need to be very careful about `msg.sender`. This case is simple enough that it can be fine.

However I'd keep things where we left them in #1109, and just keep the array functions.

pkg/vault/contracts/Authorizer.sol
```diff
@@ -32,258 +32,139 @@ import "./interfaces/IAuthorizer.sol";
...
+        bytes32[] memory actions,
         address account,
-        address[] calldata where
+        address[] memory where
     ) external {
-        _require(where.length > 0, Errors.INPUT_LENGTH_MISMATCH);
-        for (uint256 i = 0; i < roles.length; i++) {
-            _require(isAdmin(roles[i], msg.sender), Errors.GRANT_SENDER_NOT_ADMIN);
+        for (uint256 i = 0; i < actions.length; i++) {
             for (uint256 j = 0; j < where.length; j++) {
```
nventuro, line 93: I found this very unintuitive, and different from what I expected (even though it does the same thing as the original code). I thought such a call to grantRoles would check that the length of roles and wheres matched, and then iterated over those as if they were a single array of (roles, account, where) tuples.

pkg/vault/contracts/Authorizer.sol
facuspagnuolo: Didn't know about the migration. However, the issue is we will need to denormalize the stored data into two or three data structures as before which IMO is unintuitive and error prone. Otherwise, we could make the migration depend on a list of permissions to be granted or revoked. 

But again, didn't know about the migration, if that's a must we can have it back.

nventuro: Hm, now that I give this closer attention I realize that the previous solution wouldn't quite work, as we will now have people with permissions over multiple addresses. So we'd need to keep sets not only of accounts with permission, but also of all of the addresses where at least one account has said permission.

Let's skip this for now and think more about it later. Created https://github.com/balancer-labs/balancer-v2-monorepo/issues/1129

nventuro (approved):

PR #1117, opened 2022-02-03
Remove temporary no-protocol-fee LBPs

nventuro:
Follow up of https://github.com/balancer-labs/balancer-v2-monorepo/pull/1049: the factories have been deployed, there's no point in keeping the code around since we're not going to use it any more.


TomAFrench (approved):

folex:
I see a point in keeping that code in repo. Right now, if I am to use NoProtocolFeeLiquidityBootstrappingPoolFactory as specified in https://dev.balancer.fi/references/contracts/contract-addresses, I'd like to be able to read its source code and learn the difference between `NoProtocolFee` and https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPoolFactory.sol that wasn't updated since before `Feb 3`.

Right now, all I see is the ABI and source code can be found only in this PR. Which is rather hard to find if you clone the repo and read source code.

Hope it makes sense! Thank you

nventuro:
Hello @folex! Thanks for sharing your thoughts.

Unfortunately, even if we did not remove the files, you'd have no guarantee that the code in the repo has not been changed since deployment|: you might be looking at an outdated version. And since each contract has multiple dependencies, we'd be forced to effectively keep multiple copies of the repo, which is burdensome.

There's two ways you can look at the code that was actually deployed:
 - go to the deployment commit and read the source code there. Fee-less LBPs were deployed in https://github.com/balancer-labs/balancer-v2-monorepo/pull/1049
 - read the build-info for the deployment, which contains the entire source code used in the compilation off the files (albeit in json format): https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/deployments/tasks/20211202-no-protocol-fee-lbp/build-info/NoProtocolFeeLiquidityBootstrappingPoolFactory.json

folex:
Hi, @nventuro. Thanks for the answer!

Yeah, that's true, code can be outdated. But no reason to keep several versions, git commit history will serve you for that. 

Thanks for the note about build-info, didn't know you had that. Nice tool for preserving deployment setup. Not very readable though :D

RE going to the PR and looking at the code there - that's what I ended up doing, but only after my colleague got a bright idea to search through issues and PRs. I was reading the code cloned locally, and I couldn't find `NoProtocolFeeLiquidityBootstrappingPoolFactory.sol` anywhere. 

But anyway, you sure know better how to organise your repo. Just wanted to share some feedback :) Cheers!

PR #1123, opened 2022-02-04
Managed Pool: weight change library

dmf7z:
This PR unifies weight change logic into a library.


EndymionJkb (commented):
Simplifying the tests is a nice benefit here. Do we need the currentTime argument in WeightChange.getNormalizedWeight though? It's always going to be block.timestamp, so we could just get it in that function and the signature would be more straightforward: start/end weight, start/end time

TomAFrench:
True, I quite liked the idea of making the library completely stateless to avoid "magic" but seems like that might be not worth the hassle in this case.

nventuro (approved):

PR #1125, opened 2022-02-04
Fix gauges package tests

TomAFrench:
We haven't picked up a couple of breakages in the test suite because we're not actually running CI for the gauges package.

I've added it to CI and fixed up any issues.


nventuro (approved):
Ah I was just looking into this, thanks!

PR #1128, opened 2022-02-06
wrapped aave ampl - wrapped ampl linear pool

aalavandhan:
Linear pool between wrapped aave AMPL and wrapped AMPL. This pool allows LPs earn AMPL yield on aave while providing AMPL liquidity. Since both AMPL and aAMPL are rebasing tokens, they are wrapped using the "unbutton" non-rebasing wrapper.


aalavandhan:
@rabmarut Thanks will have a look

aalavandhan:
Closing in favour of #1152 1152

PR #1131, opened 2022-02-08
Managed Pool: Add token

EndymionJkb:
Add a token to a Managed Pool at a given weight; other token weights will decrease proportionately.


pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -444,6 +445,268 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+            normalizedWeight.mulUp(weightSumAfterAdd),
+            tokenAmountIn,
+            assetManager
+        );
+
+        // Transfer tokens from the sender to this contract, since the sender for the join must be the pool
+        token.transferFrom(sender, address(this), tokenAmountIn);
+        token.approve(address(getVault()), tokenAmountIn);
+
+        _joinAddToken(tokens, tokenIndex, tokenAmountIn, maxAmountsIn, recipient);
```
TomAFrench, line 685: Let's inline this to be consistent with `removeToken`

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -774,17 +1037,49 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+
+        // No BPT will be issued for the join operation itself. The `addToken` function calculates and returns
+        // the bptAmountOut, but leaves any further action, such as minting BPT, up to the caller.
+        uint256 bptAmountOut = 0;
+
+        // Note that there is no maximum amountsIn parameter: this is handled by `IVault.joinPool`.
+
+        (uint256 tokenIndex, uint256 amountIn) = userData.addToken();
+
+        uint256[] memory amountsIn = new uint256[](_getTotalTokens());
```
TomAFrench, line 1079:
```suggestion
        uint256[] memory amountsIn = new uint256[](scalingFactors.length);
```

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -444,6 +445,268 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+     *   (all other weights then scale accordingly)
+     * - Return the BPT value of the new token, for possible use by the caller
+     */
+    function addToken(
+        IERC20 token,
+        uint256 normalizedWeight,
+        uint256 tokenAmountIn,
+        address assetManager,
+        uint256 minBptPrice,
+        address sender,
```
TomAFrench, line 470: I'm not entirely sold on the idea that allowing a user provided `sender` here is a good idea. It's not as bad as a user provided `sender` for `removeToken` (as that doesn't need an approval onto the pool contract) but it feels icky.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -444,6 +445,268 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+     * - Increase the total weight, and update the number of tokens and `_tokenState`
+     *   (all other weights then scale accordingly)
+     * - Return the BPT value of the new token, for possible use by the caller
+     */
+    function addToken(
+        IERC20 token,
+        uint256 normalizedWeight,
+        uint256 tokenAmountIn,
+        address assetManager,
+        uint256 minBptPrice,
```
TomAFrench, line 469: Let's consider having a `minBptAmountOut` parameter. It feels weird to have a slippage limit set as a rate rather than a concrete amount like we do elsewhere.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -444,6 +445,268 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+     * - Join the pool, pulling in the tokenAmountIn from the sender
+     * - Increase the total weight, and update the number of tokens and `_tokenState`
+     *   (all other weights then scale accordingly)
+     * - Return the BPT value of the new token, for possible use by the caller
+     */
+    function addToken(
+        IERC20 token,
+        uint256 normalizedWeight,
+        uint256 tokenAmountIn,
+        address assetManager,
```
TomAFrench, line 468: This is scary.

pkg/balancer-js/src/utils/errors.ts
```diff
@@ -72,11 +72,14 @@ const balancerErrorCodes: Record<string, string> = {
...
-  '349': 'CHANGE_TOKENS_PENDING_WEIGHT_CHANGE',
-  '350': 'MAX_MANAGEMENT_AUM_FEE_PERCENTAGE',
+  '346': 'CHANGE_TOKENS_DURING_WEIGHT_CHANGE',
+  '347': 'CHANGE_TOKENS_PENDING_WEIGHT_CHANGE',
+  '348': 'MAX_WEIGHT',
+  '349': 'UNAUTHORIZED_JOIN',
+  '350': 'MIN_BPT_PRICE_ADD_TOKEN',
+  '351': 'MAX_MANAGEMENT_AUM_FEE_PERCENTAGE',
+  '352': 'SET_SWAP_FEE_DURING_FEE_CHANGE',
+  '353': 'SET_SWAP_FEE_PENDING_FEE_CHANGE',
```
TomAFrench, line 82: I managed to rearrange these during the squash so lets fix the ordering.

pkg/balancer-js/src/pool-weighted/encoder.ts
```diff
@@ -100,6 +101,14 @@ export class ManagedPoolEncoder {
     // eslint-disable-next-line @typescript-eslint/no-empty-function
   }
 
+  /**
+   * Encodes the userData parameter for adding a new token to a WeightedPool
+   * @param token - the token being added
+   * @param amountIn - the amount of the tokens to send to the pool as its initial balance
+   */
+  static joinForAddToken = (token: string, amountIn: BigNumberish): string =>
+    defaultAbiCoder.encode(['uint256', 'uint256', 'uint256'], [WeightedPoolJoinKind.ADD_TOKEN, token, amountIn]);
```
TomAFrench, line 110:
```suggestion
    defaultAbiCoder.encode(['uint256', 'address', 'uint256'], [WeightedPoolJoinKind.ADD_TOKEN, token, amountIn]);
```

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -444,6 +445,268 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+        // Now we need to make sure the other token weights don't get scaled down below the minimum
+        // normalized weight[i] = denormalized weight[i] / weightSumAfterAdd
+        (IERC20[] memory tokens, , ) = getVault().getPoolTokens(getPoolId());
+        for (uint256 i = 0; i < tokens.length; i++) {
+            _require(
+                _getTokenData(tokens[i]).decodeUint64(_END_DENORM_WEIGHT_OFFSET).uncompress64(_MAX_DENORM_WEIGHT).divUp(
+                    weightSumAfterAdd
+                ) >= WeightedMath._MIN_WEIGHT,
+                Errors.MIN_WEIGHT
+            );
```
TomAFrench, line 536: I think we should break this up as I find it hard to read as is.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -444,6 +445,268 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+        }
```
nventuro, line 537: Processing _all_ tokens seems unnecessarily wasteful. All we need to do is check that no token goes below the minimum weight, but we can do that much more efficiently by keeping track of the smallest weight, and comparing against that.

Weights only change in add token, remove token, and gradual weight change. In a gradual weight change we process all tokens, so we can easily find the smallest weight and store it. We only care about the smallest end weight, since add/remove token is disabled until the weight change finishes. Remove token also changes weights by increasing them by a known factor, and add token similarly decreases them by a factor: we can apply this same factor to the smallest weight.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -444,6 +445,268 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+        //
+        uint256 weightSumRatio = weightSumAfterAdd.divDown(weightSumBeforeAdd);
+        bptAmountOut = totalSupply().mulDown(weightSumRatio.sub(FixedPoint.ONE));
+
+        // Validate that the actual BPT price
+        uint256 actualBptPrice = totalSupply().mulDown(weightSumRatio).mulDown(normalizedWeight).divUp(
+            _upscale(tokenAmountIn, _computeScalingFactor(token))
+        );
+
+        _require(actualBptPrice >= minBptPrice, Errors.MIN_BPT_PRICE_ADD_TOKEN);
```
nventuro, line 575: I'm unsure about this whole thing. Should we receive a `minBPTPrice`? Is that the equivalent to `minTokensOut` in `removeToken`?

If so, it seems this belongs in WeightedMath, not here.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -444,6 +445,268 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+
+        // It will put it at the end, so iterate backwards to find it faster
+        for (uint256 i = tokens.length - 1; i >= 0; i--) {
+            if (tokens[i] == token) {
+                // This is the new one we're joining with
+                maxAmountsIn[i] = tokenAmountIn;
+                tokenIndex = i;
+                break;
+            }
+        }
```
nventuro, line 611: This is unnecessary, since as noted the token will be at the end of the array.

However, this made me notice a bug. We derive BaseWeightedPool, which selects the pool type based on the number of tokens: minimal swap info if >2, and two token if == 0. This is wrong: a two token pool cannot add or remove tokens, we should use minimal swap info always.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -444,6 +445,268 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+
+        (IERC20[] memory tokens, uint256 tokenIndex, uint256[] memory maxAmountsIn) = _registerNewToken(
+            token,
+            normalizedWeight.mulUp(weightSumAfterAdd),
+            tokenAmountIn,
+            assetManager
+        );
+
+        // Transfer tokens from the sender to this contract, since the sender for the join must be the pool
+        token.transferFrom(sender, address(this), tokenAmountIn);
```
nventuro, line 682: I'd transfer from `msg.sender`, it's a bit weird that we simply pull from an arbitrary `sender.`

pkg/pool-weighted/contracts/smart/ManagedPool.sol
TomAFrench: This causes stack too deep issues but this might be fixed by some later changes.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -444,6 +445,268 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+        //
+        // For instance, if adding 60% DAI to our example pool with $10k of value (at $1/DAI), you would add
+        // 10k * 2.5/1.0 * 0.6 = 15,000 DAI
+        // The BPT price would be 525.3056 * 0.6 / 15000 = 0.021, and the controller could set a minimum of 0.02
+        // (lower BPT price = higher maxAmountIn).
+        //
+        // In the commit stage, the actual desired balance would be passed in, and addToken would verify
+        // the final BPT price.
+        //
+        // The controller might also impose other limitations, such as not allowing (or allowlisting) asset managers.
```
nventuro, line 701: This is about controllers, it should probably just be removed.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -444,6 +445,268 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+
+        _denormWeightSum = weightSumAfterAdd;
+
+        emit TokenAdded(token, normalizedWeight, tokenAmountIn);
+
+        return bptAmountOut;
```
nventuro, line 707: Returning this value seems odd - if anything we'd want to mint BPT?

pkg/pool-weighted/contracts/WeightedPoolUserData.sol
```diff
@@ -56,6 +62,11 @@ library WeightedPoolUserData {
         (, bptAmountOut) = abi.decode(self, (JoinKind, uint256));
     }
 
+    // The token must be registered before the add join, so we use the index
```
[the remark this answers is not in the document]
nventuro: I think it's pointing out that the argument is an index and not an address, because by the time of the join the token has already been registered and is therefore in the array. I'd remove it.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -444,6 +445,268 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+
+        // Do not allow adding tokens if there is an ongoing or pending gradual weight change
+        uint256 currentTime = block.timestamp;
+        bytes32 poolState = _getMiscData();
+
+        if (currentTime < poolState.decodeUint32(_WEIGHT_START_TIME_OFFSET)) {
+            _revert(Errors.CHANGE_TOKENS_PENDING_WEIGHT_CHANGE);
+        } else if (currentTime < poolState.decodeUint32(_WEIGHT_END_TIME_OFFSET)) {
+            _revert(Errors.CHANGE_TOKENS_DURING_WEIGHT_CHANGE);
+        }
```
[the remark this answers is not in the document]
nventuro: The remove token PR added one, yes.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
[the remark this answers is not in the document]
EndymionJkb: If it's guaranteed that it always appends, we don't need it.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
EndymionJkb: lint mangled it

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -444,6 +445,268 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+    function addToken(
+        IERC20 token,
+        uint256 normalizedWeight,
+        uint256 tokenAmountIn,
+        address assetManager,
+        uint256 minBptPrice,
+        address sender,
+        address recipient
+    ) external authenticate whenNotPaused returns (uint256) {
+        return _addToken(token, normalizedWeight, tokenAmountIn, assetManager, minBptPrice, sender, recipient);
```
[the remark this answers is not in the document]
TomAFrench: This causes issues for MockManagedPool (but we can dump this if we do https://github.com/balancer-labs/balancer-v2-monorepo/pull/1131#discussion_r859196290)

pkg/pool-weighted/contracts/smart/ManagedPool.sol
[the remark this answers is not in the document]
TomAFrench: We have one set up for weight changes but we also have similar code for swap fee changes (which is the same except for offsets and revert messages).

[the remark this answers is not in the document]
TomAFrench: Yep, I used this in [8dc47ef](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1131/commits/8dc47ef8aaa17259f242833692f7db3dbba156cc)

pkg/pool-weighted/contracts/smart/ManagedPool.sol
TomAFrench: Good catch.

pkg/balancer-js/src/pool-weighted/encoder.ts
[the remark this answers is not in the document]
TomAFrench: Thoughts on just removing this entirely? We know the token will be at the position `tokens.length - 1` so this is unnecessary but that may be overly implicit for this case?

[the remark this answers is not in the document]
TomAFrench: Fixed for the time being in [9991678](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1131/commits/9991678a64be4986c01b79681c39c20d39ce82f3).

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -150,6 +150,7 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
         uint256 startSwapFeePercentage,
         uint256 endSwapFeePercentage
     );
+    event TokenAdded(IERC20 indexed token, uint256 weight, uint256 initialBalance);
```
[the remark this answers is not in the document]
TomAFrench: [b846cca](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1131/commits/b846cca2da901ddb8425fa89b54a74ba8231d238)

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -444,6 +445,268 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+        // The denormalized weight of the new token, 0.6y, is then 0.6(2.5) = 1.5
+        // Since the denorm weights of the original tokens stay the same, the final state is:
+        // |----0.8----|-0.2-|---1.5---|  denormalized weights (as stored)
+        //    0.8/2.5  0.2/2.5 1.5/2.5
+        //      32%      8%      60%      normalized weights (as calculated: W[i]/weightSum)
+        //
+        // The added token is 60%, as desired, and the original 80/20 weights are scaled down
+        // proportionately to 32/8, to fit within the remaining 40%
+        //
+        //uint256 weightSumMultiplier = FixedPoint.ONE.divDown(FixedPoint.ONE - normalizedWeight);
```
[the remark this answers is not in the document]
TomAFrench: Jeff's calculation is correct here. It's easier if you think just in terms of normalized weights after the add:

```
weightSumAfterAdd = weightSumBeforeAdd * (sum of all token weights) / (sum of all token weights minus new token weight)
weightSumAfterAdd = weightSumBeforeAdd * 1 / (1 - normalizedWeight)
```

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -443,6 +444,120 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+        tokensToAdd[0] = token;
+
+        getVault().registerTokens(getPoolId(), tokensToAdd, new address[](1));
+
+        // Tokens array is now different
+        (IERC20[] memory tokens, , ) = getVault().getPoolTokens(getPoolId());
+
+        // `_encodeTokenState` performs an external call to `token` (to get it's decimals value), however this is
+        // reentrancy safe as view functions are called in a STATICCALL context and so will revert if it modifies state.
+        _tokenState[token] = _encodeTokenState(token, denormalizedWeight, denormalizedWeight);
```
TomAFrench, line 555: I've broken this in [51ba378](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1131/commits/51ba3783f22f8f34f58d206ccbed573c3b18fd52) as using `_encodeTokenState` takes in a normalized weight and uses the **old** denormWeightSum when writing it to storage.

TomAFrench: Fixed in [41234db](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1131/commits/41234dbe6a5da09d49185de551bcb6efd54be854) as we now pass in the normalized weight along with the new `denormWeightSum` value.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
TomAFrench: I've added a `mutableTokens` arg to `BaseWeightedPool` to allow us to override this behaviour in [3d63cb3](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1131/commits/3d63cb32876c6f55192124c837223ee7482adb33)

pkg/pool-utils/test/ManagedPoolController.test.ts
```diff
@@ -209,18 +209,6 @@ describe('ManagedPoolController', function () {
...
-        await poolController.connect(manager).withdrawCollectedManagementFees(manager.address);
-      });
-
-      it('reverts if non-manager collects management fees', async () => {
-        await expect(poolController.connect(other).withdrawCollectedManagementFees(other.address)).to.be.revertedWith(
-          'CALLER_IS_NOT_OWNER'
-        );
-      });
-    });
-
```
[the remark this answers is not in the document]
TomAFrench: Must have gotten lost in the squash.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
TomAFrench: I don't think this makes sense from a gas perspective.

With this method we're performing `numTokens` warm SLOADs. Best case scenario we're saving `(2 * numTokens - 1) * 100` gas (assuming we're packing this cached weight so the SLOAD is warm).

The flipside is that we need to update the cache on gradual weight updates which change the weight of the lowest weighted token, which costs 5000 gas (I don't see a way to pack things which allows us to do this SSTORE for free).

Weight updates are going to be way more common than adding new tokens to the pool and unless you have 20+ tokens in your pool then you'll wipe out the savings by changing the weights just once.

TomAFrench: Plus less code is good.

nventuro (approved):

PR #1132, opened 2022-02-09
Safemath token admin

TomAFrench:


nventuro (approved):
I've checked that with this, the only instances of the add and remove operators are the `++i` in the for loops.

Issue #1113, opened 2022-02-03
Consider allowing the minter to claim rewards other than BAL from gauges


TomAFrench:
Currently to claim all of their rewards a user has to:

1. call to the BalancerMinter to claim BAL rewards for all gauges
2. make a separate call to each gauge to claim any other rewards for that gauge.

Step 2 can be improved by having a contract which allows a user to call into many gauges in a single transaction (anyone can claim on a user's behalf so there's no permissioning necessary here). **We'll need this on L2**

What's undecided is whether we want to allow the minter to trigger these reward claims to handle both steps 1 and 2 in a single transaction.

nventuro:
Could we have the user call `setMinterApproval` on the BatchRelayer, and use it to combine both steps?

TomAFrench:
Yes, that would work.

nventuro:
...could we add `setMinterApprovalWithSignature`? :sweat_smile:

PR #1134, opened 2022-02-10
Add setMinterApprovalWithSignature

nventuro:
The implementation is rather straightforward, as we rely on OZ's EIP712 for most of the heavy lifting. The rest of the code is adapted from SignaturesValidator.

The `setMinter` name can be confusing, since the contract we're calling _is_ the minter, but oh well. Given this will be user-facing (as they'll need to sign an EIP712 blob called 'SetMinterApproval'), we may want to consider an alternative name for this.

Fixes https://github.com/balancer-labs/balancer-v2-monorepo/issues/1113


TomAFrench (commented):
Looks good to me. One of the testcases seems like it could be made stronger but the implementation looks correct.

PR #1138, opened 2022-02-11
Add abstract "preminted BAL" gauge

TomAFrench:


pkg/gauges/contracts/gauges/PremintedGauge.sol
```diff
@@ -0,0 +1,137 @@
...
+
+        _INITIAL_RATE = tokenAdmin.INITIAL_RATE();
+        _RATE_REDUCTION_TIME = tokenAdmin.RATE_REDUCTION_TIME();
+        _RATE_REDUCTION_COEFFICIENT = tokenAdmin.RATE_REDUCTION_COEFFICIENT();
+        _RATE_DENOMINATOR = tokenAdmin.RATE_DENOMINATOR();
+
+        // Because we calculate the rate locally, this gauge cannot
+        // be used prior to the start of the first emission period
+        uint256 rate = tokenAdmin.rate();
+        require(rate != 0);
```
nventuro, line 60: This would be `0xffffff...fff` in our case, no?

pkg/gauges/contracts/gauges/PremintedGauge.sol
```diff
@@ -0,0 +1,137 @@
...
+            }
+        }
+
+        return true;
+    }
+
+    function _currentPeriod() internal view returns (uint256) {
+        return (block.timestamp / 1 weeks) - 1;
+    }
+
```
nventuro, line 126: We seem to be missing the `future_epoch_time()` function

pkg/gauges/contracts/gauges/PremintedGauge.sol
```diff
@@ -0,0 +1,137 @@
...
+            lastPeriod += 1;
+            uint256 nextEpochTime = _startEpochTime + _RATE_REDUCTION_TIME;
+            for (uint256 i = lastPeriod; i < lastPeriod + 255; ++i) {
+                if (i > currentPeriod) break;
+
+                uint256 periodTime = i * 1 weeks;
+                uint256 periodEmission = 0;
+                uint256 gaugeWeight = _gaugeController.gauge_relative_weight(address(this), periodTime);
+
+                if (nextEpochTime >= periodTime && nextEpochTime < periodTime + 1 weeks) {
```
nventuro, line 86:
```suggestion
                if (nextEpochTime >= periodTime && nextEpochTime < (periodTime + 1 weeks)) {
```

Dependency on operator precedence makes me very anxious.

pkg/gauges/contracts/gauges/PremintedGauge.sol
```diff
@@ -0,0 +1,137 @@
...
+                    // If the period crosses an epoch, we calculate a reduction in the rate
+                    // using the same formula as used in `ERC20CRV`. We perform the calculation
+                    // locally instead of calling to `ERC20CRV.rate()` because we are generating
```
nventuro, line 89:
```suggestion
                    // using the same formula as used in `BalancerTokenAdmin`. We perform the calculation
                    // locally instead of calling to `BalancerTokenAdmin.rate()` because we are generating
```

pkg/gauges/contracts/gauges/PremintedGauge.sol
```diff
@@ -0,0 +1,137 @@
...
+                    periodEmission += (gaugeWeight * rate * (periodTime + 1 weeks - nextEpochTime)) / 10**18;
+
+                    _rate = rate;
+                    _startEpochTime = nextEpochTime;
+                    nextEpochTime += _RATE_REDUCTION_TIME;
+                } else {
+                    periodEmission = (gaugeWeight * rate * 1 weeks) / 10**18;
+                }
+
+                //log PeriodEmission(period_time, period_emission)
```
nventuro, line 107: We should have this event, no?

pkg/gauges/contracts/gauges/PremintedGauge.sol
```diff
@@ -0,0 +1,137 @@
...
+                newEmissions += periodEmission;
+            }
+
+            _period = currentPeriod;
+            _emissions += newEmissions;
+
+            if (newEmissions > 0 && !isKilled) {
```
nventuro, line 114: Let's add a function to kill the gauge.

nventuro:
I propose we call this `SingleRecipientGauge`. Even if the gauge is not technically single recipient, that is what we'll use it for in all derived contracts (sending funds to either the LM comittee, the L2 bridges, etc.).

pkg/gauges/contracts/gauges/PremintedGauge.sol
TomAFrench: We have `rate = 0` when the `tokenAdmin` is unactivated so this is correct. I think you're mixing this up with `startEpochTime` being `0xffffff...fff`

pkg/gauges/contracts/gauges/PremintedGauge.sol
TomAFrench: This is handled by the type system however, `true + 1 weeks` would throw an error.

pkg/gauges/contracts/gauges/PremintedGauge.sol
nventuro: You're right, thanks!

pkg/gauges/contracts/gauges/PremintedGauge.sol
nventuro: C scarred me for life

pkg/gauges/contracts/gauges/StakelessGauge.sol
```diff
@@ -0,0 +1,143 @@
...
+    function user_checkpoint(address) external pure override returns (bool) {
+        return true;
+    }
+
+    function integrate_fraction(address user) external view override returns (uint256) {
+        require(user == address(this), "Gauge can only mint for itself");
+        return _emissions;
+    }
+
+    function set_killed(bool _isKilled) external {
```
TomAFrench, line 140: We're missing any authentication here. I'm slightly torn between using the AuthorizerAdaptor (for consistency with other gauges) vs just using the Authorizer natively although I'm leaning towards the former, thoughts?

pkg/gauges/contracts/gauges/PremintedGauge.sol
TomAFrench: This doesn't seem to follow. Why would we want this function?

pkg/liquidity-mining/contracts/gauges/StakelessGauge.sol
```diff
@@ -0,0 +1,157 @@
...
+import "../interfaces/ILiquidityGauge.sol";
+
+abstract contract StakelessGauge is ILiquidityGauge, ReentrancyGuard {
+    IERC20 internal immutable _balToken;
+    IBalancerTokenAdmin private immutable _tokenAdmin;
+    IBalancerMinter private immutable _minter;
+    IGaugeController private immutable _gaugeController;
+    address private immutable _authorizerAdaptor;
+
+    event Checkpoint(uint256 periodTime, uint256 periodEmissions);
```
nventuro, line 32: Would it make sense to make `periodTime` indexed?

TomAFrench: Yeah, I think so. We can be a bit liberal with indexing anyway as we're calling this function rarely

PR #1146, opened 2022-02-14
Add relayer module for interacting with gauge system

TomAFrench:


nventuro (commented):
Looks great!

The implementation is simple enough that we _could_ merge with no tests, though I'd like to have at least some before we deploy and grant relayer powers. We could create a new issue to track this.

Issue #1144, opened 2022-02-11
Add single recipient gauge


nventuro:
Derived from #1115, this will be used to send funds to the LiquidityMiningComittee, and potentially to a holder of funds allocated to veBAL holders until more elaborate distribution systems are set up.

PR #1149, opened 2022-02-16
Add BAL Token Holder contract

nventuro:
This contract is meant to act as the recipient of a single recipient gauge (#1144). I've thought of alternative ways to tackle the problem, such as embeddeding this in the gauge itself, but I wanted to keep the `checkpoint()` interface in the gauge, and so this pattern seemed to make sense. 

I don't think the need for multiple calls (`checkpoint` + `withdrawFunds`) is a problem, since these can always be scripted in a contract if desired, or even batched by e.g. a Gnosis Safe, which we expect the authorized parties to use.


TomAFrench (commented):
I imagine the main usecase for this would be for the funds which the LM committee is supposed to distribute, correct? If so it might be worth adding another function to allow other tokens which are sent to this address to be recovered.

Justification being that there's a decent chance that if the pools have co-incentives with another token then there's a decent chance that someone sometime is going to send that token to this contract.

I'd keep this as a separate function to BAL so that we can have separate authorisation for withdrawing BAL vs other tokens.

pkg/standalone-utils/test/BALTokenHolderFactory.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    return await deployedAt('BALTokenHolder', holder);
+  }
+
+  describe('creation', () => {
+    it('emits an event', async () => {
+      const receipt = await (await factory.create('holder')).wait();
+      expectEvent.inReceipt(receipt, 'BALTokenHolderCreated', { name: 'holder' });
+    });
+
+    it('creates a holder with the same BAL and vualt addresses', async () => {
```
TomAFrench, line 51:
```suggestion
    it('creates a holder with the same BAL and vault addresses', async () => {
```

TomAFrench (approved):

PR #1152, opened 2022-03-02
Wrapped AMPL support

aalavandhan:
Supporting wrapped AMPL through the generic unbutton non-rebasing wrapper, which wraps any generic rebasing ERC-20 token into a non-rebasing, fixed balance version.

* Relayer support for all "unbuttoned" tokens
* Linear pool between wrapped aave AMPL and wrapped AMPL. This pool allows LPs earn AMPL yield on aave while providing AMPL liquidity. Since both AMPL and aAMPL are rebasing tokens, they are wrapped using the "unbutton" non-rebasing wrapper. 

Merged previous PRs (#1128, #1075)


rabmarut (commented):
Hey, @aalavandhan! Thanks for putting this all together.

This PR needs a pool factory to go with the pool, so I made my own PR to add one. Please review when you get a chance: https://github.com/ampleforth/balancer-v2-monorepo/pull/5

In the process, I also discovered a fair amount of code duplication which I think resulted from the merge of two PRs, so I cut some things away and cleaned up more generally. I think we're approaching a final form! Thanks for all your hard work so far.

rabmarut (approved):

PR #1153, opened 2022-03-03
Authorizer: Ensure delays correctness

### LINKED ISSUE #1083: Ensure Authorizer Timelock correctness
(to be implemented after https://github.com/balancer-labs/balancer-v2-monorepo/issues/588)

Contracts don't store the Authorizer's address: they ask the Vault for the current Authorizer, and then forward the call there. This is because the Authorizer can be changed by calling `setAuthorizer` on the Vault. Therefore, timelocks are meaningless as long as the Authorizer can be arbitrarily changed to a new one, which may not have the timelock constraints.

For the system to be correctly set up, we'd need for there to be a timelock on the `setAuthorizer` permission, which should be greater or equal to the longest timelock in the entire system. I think we could special-case the `setAuthorizer` permission (obtained via `vault.getActionId(IVault.setAuthorizer.selector)`), and add a check so that whenever a timelock is set, we first check that the `setAuthorizer` timelock duration is greater or equal. This way, in order to add a longer timelock we'd first need to manually extend `setAuthorizer`'s timelock, which makes this property hard to ignore or forget about.

facuspagnuolo:
Fixes #1083 

Note that we can make the authorizer depend on the Vault since this will be a new version of it, otherwise we would have had circular dependencies. 

I think we miss denoting that `IVault` extends `IAuthentication` in `IVault.sol`. Should we add it? (cc @nventuro)

### DIFF
--- a/pkg/vault/contracts/Authorizer.sol
+++ b/pkg/vault/contracts/Authorizer.sol
@@ -20,6 +20,7 @@ import "@balancer-labs/v2-solidity-utils/contracts/helpers/BalancerErrors.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/helpers/IAuthentication.sol";
 
 import "./interfaces/IAuthorizer.sol";
+import "./interfaces/IVault.sol";
 
 /**
  * @dev Basic Authorizer implementation, based on OpenZeppelin's Access Control.
@@ -50,6 +51,7 @@ contract Authorizer is IAuthorizer {
         uint256 executableAt;
     }
 
+    IAuthentication public immutable vault;
     ScheduledAction[] public scheduledActions;
     mapping(bytes32 => bool) public permissionGranted;
     mapping(bytes32 => uint256) public delays;
@@ -84,9 +86,10 @@ contract Authorizer is IAuthorizer {
      */
     event PermissionRevoked(bytes32 indexed action, address indexed account, address indexed where);
 
-    constructor(address admin) {
-        _grantPermission(GRANT_PERMISSION, admin, EVERYWHERE);
-        _grantPermission(REVOKE_PERMISSION, admin, EVERYWHERE);
+    constructor(address _admin, IAuthentication _vault) {
+        vault = _vault;
+        _grantPermission(GRANT_PERMISSION, _admin, EVERYWHERE);
+        _grantPermission(REVOKE_PERMISSION, _admin, EVERYWHERE);
     }
 
     /**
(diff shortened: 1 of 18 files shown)


pkg/vault/contracts/Authorizer.sol
```diff
@@ -129,6 +132,10 @@ contract Authorizer is IAuthorizer {
      */
     function setDelay(bytes32 action, uint256 delay) external {
         _require(msg.sender == address(this), Errors.SENDER_NOT_ALLOWED);
+
+        bytes32 setAuthorizerId = vault.getActionId(IVault.setAuthorizer.selector);
```
TomAFrench, line 136: This is a fixed value so why not store the action id instead of the vault's address in order to query it each time?

facuspagnuolo: I thought about that too... the problem is that the interface does not ensure it is actually constant, it would be wrong conceptually speaking.

pkg/vault/test/Authorizer.test.ts
```diff
@@ -27,7 +27,14 @@ describe('Authorizer', () => {
   const NOT_WHERE = ethers.Wallet.createRandom().address;
 
   sharedBeforeEach('deploy authorizer', async () => {
-    authorizer = await Authorizer.create({ admin });
+    const oldAuthorizer = await Authorizer.create({ admin });
+
+    vault = await deploy('Vault', { args: [oldAuthorizer.address, ZERO_ADDRESS, 0, 0] });
+    authorizer = await Authorizer.create({ admin, vault });
+
+    const setAuthorizerAction = await vault.getActionId(vault.interface.getSighash('setAuthorizer'));
```
nventuro, line 35: We have a helper called `actionId` that does the same as this.

pkg/vault/test/Authorizer.test.ts
```diff
@@ -27,7 +27,14 @@ describe('Authorizer', () => {
...
+    await oldAuthorizer.grantPermissions(setAuthorizerAction, admin, vault, { from: admin });
+    await vault.connect(admin).setAuthorizer(authorizer.address);
```
nventuro, line 37: We will need to do all of this whenever we want to run a test that involves timelocks, correct? Otherwise we'd either have a circular dependency, or a zero vault in the authorizer which will cause all calls to `setDelay` to fail.

If that is true, then I wonder if we should try to either improve this or script this into a helper. Though perhaps that's unnecessary, since most (all?) timelock tests should live in this file.

nventuro (commented):
Looks great! :raised_hands: 

>Note that we can make the authorizer depend on the Vault since this will be a new version of it, otherwise we would have had circular dependencies.

Yes, this is an interesting consequence. I'd consider renaming the new Authorizer to simething like TimelockAuthorizer to reduce confusion.

>I think we miss denoting that IVault extends IAuthentication in IVault.sol. Should we add it?

It is possible IAuthentication didn't exist when we first created IVault. I'd add that dependency, yes.

pkg/vault/test/Authorizer.test.ts
facuspagnuolo: Yea, I don't think it would be useful for now... this is the only place where we do it

pkg/vault/contracts/Authorizer.sol
```diff
@@ -50,6 +51,7 @@ contract Authorizer is IAuthorizer {
         uint256 executableAt;
     }
 
+    IAuthentication public immutable vault;
```
TomAFrench, line 54: Can we change this to private and have a `getVault` getter? This'll match the style used in other contracts.

nventuro (approved):

PR #1155, opened 2022-03-04
ERC 4626 Linear pool factory deployment

gtaschuk:
This sets up the task to deploy, it can be executed as:

```
$> npx hardhat deploy --id 20220304-erc4626-linear-pool --key <your key here> --network polygon
```


rabmarut (commented):
Verified build-info against `e9d7f553ee007409c699cd1d507cc6eb` generated in `pool-linear` at commit 538e7ff6dd7c0492822a55fe430a4ad9038f9039

Verified ABIs against same build

I would argue against deploying this V1 to mainnet unless we know we need it - the final version should be ready for mainnet within the next month or so. I will make a small commit to remove the mainnet reference.

@gtaschuk, if you could please deploy this to Polygon and then commit the output file, I will approve and merge.

rabmarut (approved):
Verified contract at output address, good to go!

PR #1156, opened 2022-03-07
Authorizer: Use `hasPermission` instead of `canPerform` internally

facuspagnuolo:
The authorizer was using `canPerform` unnecessarily for internal methods, there were no timelocks supported for the authorizer itself, thus it doesn't make sense to use it. We should use `hasPermission` instead.


nventuro (approved):
This seems trivially correct, but it goes to show that the current naming scheme can be quite confusing :sweat_smile: Once #1157 is in, I'd seriosuly consider going over all concepts, choosing well-defined names for each of them, and revamp the language of this contract.

PR #1157, opened 2022-03-07
Authorizer: Add grant/revoke timelocks

facuspagnuolo:
### MERGE AFTER #1156

This introduces new methods to allow defining delays to grant and revoke methods on the authorizer.

Potential issue, with this change, only a single delay value is allowed to be set per target contract (where). This might not make a lot of sense now. Previously, we simply had the idea of a "grant permission" per target, but now since we are defining delays we may want to have action granularity.


TomAFrench:
We have a `require(delay > 0, "CANNOT_SCHEDULE_ACTION")` line on the `schedule` function which doesn't exist in the `scheduleGrantPermission`/`scheduleRevokePermission` functions. We can then schedule grants/revocations of permissions which will then always fail (as in `canPerform` we'll go down the immediate action branch).

I'm inclined to say that we should just move the `delay > 0` requirement inside `_schedule` as scheduling an actions where the delay is zero is always pointless and it would be better to fail at the start rather than only finding out at the end of the timelock.

facuspagnuolo:
> We have a `require(delay > 0, "CANNOT_SCHEDULE_ACTION")` line on the `schedule` function which doesn't exist in the `scheduleGrantPermission`/`scheduleRevokePermission` functions. We can then schedule grants/revocations of permissions which would be immediately executable.
> 
> I'm inclined to say that we should just move the `delay > 0` requirement inside `_schedule` as scheduling an action when the delay is zero is always pointless.

Thanks @TomAFrench for the suggestion, nice catch! :)

nventuro (approved):
The implementation looks good for the current implementation, though we've discussed offline that this is not the final implementation we're going for. Marking as approved so we can also get https://github.com/balancer-labs/balancer-v2-monorepo/pull/1167 included and then review this.

For the record, we agreed on having 'grant permission' delays be associated with the action id (e.g. the 'mint BAL' action id will have a 1 year grant and revoke delay regardless of where the permission is granted). If we were to also include the `where` field, we'd need to include the 'everywhere' case, which makes things more complicated.

PR #1158, opened 2022-03-08
Add GaugeAdder contract

### LINKED ISSUE #1139: Create GaugeAdder contract
Instead of having an EOA or multisig directly call `add_gauge` on the `GaugeController` (via `AuthorizerAdaptor`), we should build a contract that does this. This will allow us to enforce certain properties about the gauge: critically, that it comes from one of the known factories, and that the gauge type is the expected one for a gauge of said factory. This will reduce operational risk when adding gauges.

Sample pseudo-code:

```
contract GaugeAdder {
    mapping (GaugeType => ILiquidityGaugeFactory[]) public gaugeFactories;
    
    // These could be dynamic instead of hard-coded, though we'd need to be careful with the mapping.
    enum GaugeType {
        LiquidityMiningCommittee,
        veBAL,
        Ethereum,
        Polygon,
        Arbitrum
    }

    // If the types are static and not dynamic, I'd suggest having an explicit function for each one of them, to reduce the potential
    // for user error. Not all types would allow for gauge creation, e.g. veBAL and LMComittee.
    // With these checks in place, this function is relatively risk-free, but we may want to still keep it permissioned to avoid 
    // gauge-spamming.
    function addEthereumGauge(gauge) authenticate {
        require(isGaugeFromFactory(gauge, GaugeType.Ethereum);

        authorizerAdaptor.performAction(
            gaugeController, abi.encodeWithSelector(gaugeController.add_gauge.selector, gauge, GaugeType.Ethereum, 0)
        );
    }

    function isGaugeFromFactory(gauge, gaugeType) public view returns (bool) {
        for (uint256 i = 0; i < gaugeFactories.length; ++i) {
            if (gaugeFactories[i].isGaugefromFactory(gauge)) {
                return true;
            }
        }

        return false;
    }

    // This function is very sensitive, especially if `addGauge` is not permissioned
    function addGaugeFactory(factory, gaugeType) authenticate {
        // check factory is not yet added
        // add factory
    }
}
```

TomAFrench:
This PR implements a `GaugeAdder` contract as described in #1139.

This contract allows various factory contracts to be allowlisted for specific gauge types which exist on the gauge controller. Authorised users may then add any address which a gauge factory lists as one of its deployed gauges to the gauge controller.

Separate functions are exposed for adding gauges of each type to allow granular permissions over which gauge types a user may add gauges to the controller for.

This is made with the assumption that gauge types will very rarely change and as such we can just redeploy the gauge adder when necessary. For this reason there's no function to remove a gauge factory once added, however on adding a new gauge factory address it is checked that it responds properly to `isGaugeFromFactory` function calls to prevent the gauge adder being inadvertently bricked.

fixes #1139

### DIFF
--- a/pkg/gauges/contracts/AuthorizerAdaptor.sol
+++ b/pkg/gauges/contracts/AuthorizerAdaptor.sol
@@ -21,6 +21,8 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/ReentrancyGuard.
 import "@balancer-labs/v2-vault/contracts/interfaces/IAuthorizer.sol";
 import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
 
+import "./interfaces/IAuthorizerAdaptor.sol";
+
 /**
  * @title Authorizer Adaptor
  * @notice This contract is intended to act as an adaptor between systems which expect a single admin address
@@ -31,7 +33,7 @@ import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
  * @dev When calculating the actionId to call a function on a target contract, it must be calculated as if it were
  * to be called on this adaptor. This can be done by passing the function selector to the `getActionId` function.
  */
-contract AuthorizerAdaptor is IAuthentication, ReentrancyGuard {
+contract AuthorizerAdaptor is IAuthorizerAdaptor, IAuthentication, ReentrancyGuard {
     using Address for address;
 
     bytes32 private immutable _actionIdDisambiguator;
@@ -87,7 +89,13 @@ contract AuthorizerAdaptor is IAuthentication, ReentrancyGuard {
      * @param data - Calldata to be sent to the target contract
      * @return The bytes encoded return value from the performed function call
      */
-    function performAction(address target, bytes calldata data) external payable nonReentrant returns (bytes memory) {
+    function performAction(address target, bytes calldata data)
+        external
+        payable
+        override
+        nonReentrant
+        returns (bytes memory)
+    {
         bytes4 selector;
 
         // We want to check that the caller is authorized to call the function on the target rather than this function.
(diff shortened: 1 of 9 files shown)


pkg/liquidity-mining/contracts/interfaces/IGaugeAdder.sol
```diff
@@ -0,0 +1,69 @@
...
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/IAuthentication.sol";
+
+import "./IAuthorizerAdaptor.sol";
+import "./IGaugeController.sol";
+import "./ILiquidityGauge.sol";
+import "./ILiquidityGaugeFactory.sol";
+
+interface IGaugeAdder is IAuthentication {
+    // TODO: Ensure that these align with the types as set on the GaugeController
+    // Failure to do so will result in an irrecoverably broken GaugeController state.
```
nventuro, line 26: We can achieve this by using this enum in the veBALDeploymentCoordinator

pkg/liquidity-mining/contracts/GaugeAdder.sol
```diff
@@ -0,0 +1,168 @@
...
+    /**
+     * @dev Adds `gauge` to the GaugeController with type `gaugeType` and an initial weight of zero
+     */
+    function _addGauge(address gauge, GaugeType gaugeType) private {
+        require(isGaugeFromValidFactory(gauge, gaugeType), "Invalid gauge");
+
+        // `_gaugeController` enforces that duplicate gauges may not be added so we do not need to check here.
+        _authorizerAdaptor.performAction(
+            address(_gaugeController),
+            abi.encodeWithSelector(IGaugeController.add_gauge.selector, gauge, gaugeType, 0)
```
nventuro, line 165:
```suggestion
            abi.encodeWithSelector(IGaugeController.add_gauge.selector, gauge, gaugeType)
```

This is technically incorrect since we're using the version of `add_gauge` that lacks a weight argument. abi.encode does not check times, so the compiler doesn't complain. If the Controller was written in Solidity, this would work as extra calldata is ignored - I'm not sure about Vyper.

nventuro (approved):

PR #1161, opened 2022-03-10
Expose getter for BAL token on BalancerMinter

TomAFrench:
Simple change to just store the BAL token address on the BalancerMinter as it doesn't really make sense for the minter to not be able to provide the address of the token which is mints.


EndymionJkb (approved):
Some of the external functions use @dev (like most of the rest of the codebase), while most use @notice. The veBAL-related packages seem to use full NatSpec.

PR #1164, opened 2022-03-13
Managed Pool: AUM-based management fees

EndymionJkb:
This introduces a second kind of management fee: one that mints BPT to the manager proportional to the AUM (totalSupply), targeting an annualized percentage. It reuses the join and exit callbacks: separating them into all before/after join/exit combinations, for greater flexibility and future-proofing. (For this particular use case, it *could* just use _beforeJoinExit.)


pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -689,4 +722,62 @@ contract ManagedPool is BaseWeightedPool, ReentrancyGuard {
...
+            uint256 elapsedTime = currentTime - lastCollection;
+            // Similar to BPT swap fee calculation, collect fees equal to the AUM % after minting
+            // F is the AUM fee percentage, S is the totalSupply, and x is the amount to mint after 1 year:
+            // S + x = S + F(S + x)
+            // x = F(S + x)
+            // x(1 - F) = FS
+            // x = S * F/(1 - F); per annual time period
+            // Final value needs to be annualized: multiply by elapsedTime/(365 days)
+            uint256 feePct = _managementAumFeePercentage.divDown(_managementAumFeePercentage.complement());
+            uint256 timePeriodPct = elapsedTime.divDown(365 days);
```
dmf7z, line 773: This is fixed point division on timestamps which are not in fixed point.

TomAFrench:
I've made a slight change to behaviour in [425ffe5](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1164/commits/425ffe54b8f2b9a22dd6afbd13ffc93a3b443841) to save some gas on joins and exits.

We were performing a (warm) SLOAD to check if we were in the initialisation call in order to set the `_lastAumFeeCollectionTimestamp`. I've removed this so that we rely on it being set in the first call to `_collectAumManagementFees`. This forfeits any management fees between initialisation and then but it's likely that the pool manager will be the only LP at this point so this isn't an issue in my mind.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -354,7 +376,42 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+        _managementAumFeePercentage = managementAumFeePercentage;
+        emit ManagementAumFeePercentageChanged(managementAumFeePercentage);
+    }
+
+    /**
+     * @notice Collect any accrued AUM fees and send them to the pool manager.
+     * @dev This can be called by anyone to collect accrued AUM fees - and will be called automatically on
+     * joins and exits.
+     */
+    function collectAumManagementFees() public whenNotPaused nonReentrant {
```
TomAFrench, line 409:
```suggestion
    function collectAumManagementFees() public whenNotPaused {
```

We could technically remove this modifier as we don't make any external calls in this function.

TomAFrench:
I think this should be ready to be merged now. I also think we should merge this before #1108 as we'll need to properly collect the AUM fees before adding or removing tokens if it changes the pool's total supply.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -674,6 +733,56 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+
+        // Collect fees based on the time elapsed
+        if (currentTime > lastCollection) {
+            // Reset the collection timer to the current block
+            _lastAumFeeCollectionTimestamp = currentTime;
+
+            uint256 managementAumFeePercentage = getManagementAumFeePercentage();
+
+            // `lastCollection == 0` means that we're in the first attempt to collect AUM fees
+            // For gas reasons we only collect AUM fees from this point onwards so perform an early return if so.
```
nventuro, line 762: These are not 'gas reasons': we can't collect fees the first time because we don't know what the elapsed time is (as we don't know when the pool was created).

We should also note in comments that since this function is called after all joins/exits, lastCollection will be automatically initialized to the time of the first non-initialziation join.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -350,7 +372,43 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+    /**
+     * @notice Collect any accrued AUM fees and send them to the pool manager.
+     * @dev This can be called by anyone to collect accrued AUM fees - and will be called automatically on
+     * joins and exits.
+     */
+    function collectAumManagementFees() public whenNotPaused {
+        // It only makes sense to collect AUM fees after the pool is initialized (as before then the AUM is zero).
+        // We can query if the pool is initialized by checking for a nonzero total supply.
+        // Performing an early return here prevents zero value AUM fee collections causing bogus events.
+        if (totalSupply() == 0) return;
```
nventuro, line 409: This is such an edge-case that I'd consider simply reverting with e.g. 'pool uninitialized'

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -674,6 +733,56 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+                return;
+            }
+
+            uint256 elapsedTime = currentTime - lastCollection;
+            // Similar to BPT swap fee calculation, collect fees equal to the AUM % after minting
+            // F is the AUM fee percentage, S is the totalSupply, and x is the amount to mint after 1 year:
+            // S + x = S + F(S + x)
+            // x = F(S + x)
+            // x(1 - F) = FS
+            // x = S * F/(1 - F); per annual time period
```
nventuro, line 774: I found this hard to understand and had to go back to the explanation of BPT fees in WeightedMath: instead of mentioning how this is similar to that scenario, I'd instead have here a longer-form explanation like the one that exists on that file.

pkg/pool-weighted/test/ManagedPool.test.ts
```diff
@@ -869,8 +861,94 @@ describe('ManagedPool', function () {
...
+              const balanceBefore = await pool.balanceOf(owner);
+
+              expectEvent.inIndirectReceipt(
+                await collectAUMFees(),
+                pool.instance.interface,
+                'ManagementAumFeeCollected'
+              );
+
+              const balanceAfter = await pool.balanceOf(owner);
+              expect(balanceAfter.sub(balanceBefore)).to.equalWithError(expectedManagementFeeBpt, 0.0001);
```
nventuro, line 923:
```suggestion
              const balanceAfter = await pool.balanceOf(owner);
              const actualFees = balanceAfter.sub(balanceBefore);
              expect(actualFees).to.equalWithError(expectedManagementFeeBpt, 0.0001);
              
              expectEvent.inIndirectReceipt(
                await collectAUMFees(),
                pool.instance.interface,
                'ManagementAumFeeCollected',
                { bptAmount: actualFees }
              );

```

pkg/pool-weighted/test/ManagedPool.test.ts
```diff
@@ -869,8 +861,94 @@ describe('ManagedPool', function () {
...
+            it('cannot be set above the maximum AUM fee', async () => {
+              await expect(pool.setManagementAumFeePercentage(owner, fp(0.2))).to.be.revertedWith(
+                'MAX_MANAGEMENT_AUM_FEE_PERCENTAGE'
+              );
+            });
+          });
+        });
+      });
+
+      describe('management aum fee collection', () => {
```
nventuro, line 877: We should add tests for 
 - manual collection before initialization
 - manual collection when the pool is paused
 - automatic collection when the pool is paused (proportional exits)

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -674,6 +733,56 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
     }
 
+    // Join/exit callbacks
+
+    function _beforeJoinExit(
+        uint256[] memory,
+        uint256[] memory,
+        uint256
+    ) internal virtual override {
+        _collectAumManagementFees();
```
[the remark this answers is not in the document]
TomAFrench: Resolved in [592d454](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1164/commits/592d4547cf62e97bbde9e81486472a57a7664257)

pkg/pool-weighted/contracts/smart/ManagedPool.sol
TomAFrench: [8d2d780](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1164/commits/8d2d78006eb412c1f389fe799389205092caf960)

pkg/pool-weighted/contracts/smart/ManagedPool.sol
TomAFrench: > These are not 'gas reasons': we can't collect fees the first time because we don't know what the elapsed time is (as we don't know when the pool was created).

Well tbf we don't know that because it costs us gas on each join in order to check if we are initialising the pool.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -350,7 +372,43 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+        _managementAumFeePercentage = managementAumFeePercentage;
+        emit ManagementAumFeePercentageChanged(managementAumFeePercentage);
+    }
+
+    /**
+     * @notice Collect any accrued AUM fees and send them to the pool manager.
+     * @dev This can be called by anyone to collect accrued AUM fees - and will be called automatically on
+     * joins and exits.
+     */
+    function collectAumManagementFees() public whenNotPaused {
```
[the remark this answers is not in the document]
TomAFrench: [ca4229b](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1164/commits/ca4229bee41717f2c39adac7402b4be942e5d9f8)

pkg/pool-weighted/test/ManagedPool.test.ts
TomAFrench: [9806e29](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1164/commits/9806e29da8a2c7e99e759e3b160c5db8d1c08e1e)

pkg/pool-weighted/contracts/smart/ManagedPool.sol
TomAFrench: I've added a longer explanation in [9be13c5](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1164/commits/9be13c5b36520c8537ce76fc82bf78995e97b749)

pkg/pool-utils/test/ManagedPoolController.test.ts
```diff
@@ -337,9 +352,9 @@ describe('ManagedPoolController', function () {
...
 
       it('reverts if the manager tries to change the management fee', async () => {
-        await expect(poolController.connect(manager).setManagementSwapFeePercentage(NEW_MGMT_FEE)).to.be.revertedWith(
-          'UNAUTHORIZED_OPERATION'
-        );
+        await expect(
+          poolController.connect(manager).setManagementSwapFeePercentage(NEW_MGMT_SWAP_FEE)
+        ).to.be.revertedWith('UNAUTHORIZED_OPERATION');
       });
     });
```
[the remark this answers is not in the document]
TomAFrench: https://github.com/balancer-labs/balancer-v2-monorepo/issues/1267

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -674,6 +741,71 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+        // Collect fees based on the time elapsed
+        if (currentTime > lastCollection) {
+            // Reset the collection timer to the current block
+            _lastAumFeeCollectionTimestamp = currentTime;
+
+            uint256 managementAumFeePercentage = getManagementAumFeePercentage();
+
+            // If `lastCollection` has not been set then we don't know what period over which to collect fees.
+            // We then perform an early return after initializing it so that we can collect fees next time.
+            // We also perform an early return if the Pool has been paused or the or the AUM fee is zero.
```
nventuro, line 775:
```suggestion
            // If `lastCollection` has not been set then we don't know what period over which to collect fees.
            // We then perform an early return after initializing it so that we can collect fees next time. This
            // means that AUM fees are not collected for any tokens the Pool is initialized with until the first
            // non-initialization join or exit.
            // We also perform an early return if the AUM fee is zero, to save gas.
            //
            // If the Pool has been paused, all fee calculation and minting is skipped to reduce execution
            // complexity to a minimum (and therefore likelihood of errors). We do still update the last
            // collection timestamp however, to avoid potentially collecting extra fees if the Pool were to
            // be later unpaused. Any fees that would be collected while the Pool is paused are lost.
```

pkg/pool-weighted/test/ManagedPool.test.ts
TomAFrench: Addressed over commits [9b9aebb](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1164/commits/9b9aebba1ef5c17cc11d4d3d2bfa04f04f77db5e) through [0a76238](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1164/commits/0a7623805dccf285cb01ef8016d2c117ad9b9d58)

nventuro (approved):

PR #1166, opened 2022-03-14
Add veBALDeploymentCoordinator

TomAFrench:
Draft PR as this is still in development but having a PR open helps discussion.

This PR adds the veBALDeploymentCoordinator contract which implements much of the setup of the veBAL system to avoid mistakes during setup and enforcing certain timings.

### DIFF
--- a/pkg/liquidity-mining/contracts/veBALDeploymentCoordinator.sol
+++ b/pkg/liquidity-mining/contracts/veBALDeploymentCoordinator.sol
@@ -30,6 +30,7 @@ contract veBALDeploymentCoordinator is Authentication, ReentrancyGuard {
     IVault private immutable _vault;
     IBalancerToken private immutable _balancerToken;
     IBalancerMinter private immutable _balancerMinter;
+    IGaugeController private immutable _gaugeController;
 
     enum DeploymentStage { PENDING, FIRST_STAGE_DONE, SECOND_STAGE_DONE }
 
@@ -66,6 +67,7 @@ contract veBALDeploymentCoordinator is Authentication, ReentrancyGuard {
         _vault = balancerTokenAdmin.getVault();
         _balancerToken = balancerTokenAdmin.getBalancerToken();
         _balancerMinter = balancerMinter;
+        _gaugeController = IGaugeController(balancerMinter.getGaugeController());
 
         _activationScheduledTime = activationScheduledTime;
     }
@@ -92,6 +94,13 @@ contract veBALDeploymentCoordinator is Authentication, ReentrancyGuard {
         return _balancerMinter;
     }
 
+    /**
+     * @notice Returns the address of the Gauge Controller
+     */
+    function getGaugeController() external view returns (IGaugeController) {
+        return _gaugeController;
+    }
+
     function getCurrentDeploymentStage() external view returns (DeploymentStage) {
         return _currentDeploymentStage;
     }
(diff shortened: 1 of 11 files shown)


pkg/liquidity-mining/contracts/veBALDeploymentCoordinator.sol
```diff
@@ -0,0 +1,242 @@
...
+        // Step 3: setup the Liquidity Mining program by creating the different gauge types on the Gauge Controller.
+        //
+        // All gauge types are initially created with a weight of 0, which will allow for gauges to be deployed and LPs
+        // to vote and stake for them, without yet minting any BAL. This is intended to provide a grace period for LPs
+        // to migrate to the new system before it is fully activated. The gauge type weights will be set to their actual
+        // values on the last stage of deployment.
+        {
+            // Admin functions on the Gauge Controller have to be called via the the AuthorizerAdaptor, which acts as
+            // its admin.
+            IAuthorizerAdaptor authorizerAdaptor = gaugeController.admin();
```
TomAFrench, line 149: This is currently assuming that ownership over the `GaugeController` has been properly transferred to the `AuthorizerAdaptor`. The Curve contracts will initially have the deployer key set as the admin and there's a two step process to transfer ownership.

We should either have the coordinator make the AuthorizerAdaptor claim ownership over the various contracts in `performFirstStage` or add a `prepareForFirstStage` function which will just perform all of these ownership transfers to get the system in a "proper uninitialised state". I'd lean towards the second as we can do additional verification before hitting the button but it's not a huge deal.

TomAFrench: From the looks of things the only contracts for which we need to do this is `VotingEscrow` and `GaugeController`

pkg/liquidity-mining/contracts/veBALDeploymentCoordinator.sol
```diff
@@ -0,0 +1,274 @@
...
+
+        // grant self power to add gauges
+        // deploy lm, vebal, merkle gauges. add gauges to gauge controller
+
+        // grant gauge adder authority over controller
+        // grant self power over guage adder
+        // add l1 and l2 factories to gauge adder
+        // deploy l1 and l2 gauges, add to controller via gauge adder
+
+        // add tokens to l1 and l2 gauges
```
TomAFrench, line 207: I have a local branch which implements the setup of a [GaugeAdder](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1158) contract, allowlist the factory contract for Ethereum gauges and deploy gauges for a preset array of pools.

The comments here mention adding factories for L2 gauges and deploying single recipient gauges. L2 gauges are definitely out of scope for being deployed in stage 1 but we can potentially add in single recipient gauges as well.

TomAFrench: I'm not sure what gas costs might end up being if we're deploying a large number of gauges here (granted these are proxies). It might be worth splitting off step 4 to be another function call which is immediately callable by anyone after `performFirstStage` has been called.

pkg/liquidity-mining/contracts/veBALDeploymentCoordinator.sol
nventuro: Addressed in #1171

nventuro (approved):

PR #1167, opened 2022-03-15
Authorizer: Rename and improve wording

### LINKED ISSUE #1159: Revamp Authorizer terminology
With the introduction of timelocks and global permissions, we are now beginning to have an over abundance of new terms that are not necessarily obvious, such as permission, permission ids,  action, action ids, scheduled action ids, perform vs permission, etc.

I think it'd be a good idea to go over this concepts and to try to arrive at a minimal set of clearly defined terms, with the goal of simplifying jargon in the new contract.

facuspagnuolo:
### MERGE AFTER #1157 

Fixes #1159


pkg/vault/contracts/TimelockAuthorizer.sol
```diff
@@ -0,0 +1,368 @@
...
+        bytes32 actionId,
+        uint256 newDelay,
+        address[] memory executors
+    ) external returns (uint256 scheduledActionId) {
+        require(newDelay <= MAX_DELAY, "DELAY_TOO_LARGE");
+        bytes32 scheduleDelayActionId = keccak256(abi.encodePacked(SCHEDULE_DELAY_ACTION_ID, actionId));
+        _require(hasPermission(scheduleDelayActionId, msg.sender, address(this)), Errors.SENDER_NOT_ALLOWED);
+
+        uint256 actionDelay = delaysPerActionId[actionId];
+        bytes memory data = abi.encodeWithSelector(this.setDelay.selector, actionId, newDelay);
```
nventuro, line 182:
```suggestion
        bytes memory data = abi.encodeWithSelector(this.setDelay.selector, actionId, newDelay);
        // The delay change is scheduled to execute after the current delay for the action has elapsed. This is critical, as
        // otherwise it'd be possible to execute an action with a delay shorter than its current one by first changing it to a
        // smaller (or zero) value.
        uint256 actionDelay = delaysPerActionId[actionId];
```

pkg/vault/contracts/TimelockAuthorizer.sol
```diff
@@ -0,0 +1,368 @@
...
+        require(scheduledActionId < scheduledActions.length, "ACTION_DOES_NOT_EXIST");
+        ScheduledAction storage scheduledAction = scheduledActions[scheduledActionId];
+        require(!scheduledAction.executed, "ACTION_ALREADY_EXECUTED");
+        require(!scheduledAction.cancelled, "ACTION_ALREADY_CANCELLED");
+
+        // solhint-disable-next-line not-rely-on-time
+        require(block.timestamp >= scheduledAction.executableAt, "ACTION_NOT_EXECUTABLE");
+        bytes32 executeActionId = _getExecuteActionId(scheduledActionId);
+        bool isAllowed = !scheduledAction.protected || hasPermission(executeActionId, msg.sender, address(this));
+        _require(isAllowed, Errors.SENDER_NOT_ALLOWED);
```
nventuro, line 213:
```suggestion
        if (scheduledAction.protected) {
            bytes32 executeActionId = _getExecuteActionId(scheduledActionId);
            _require(hasPermission(executeActionId, msg.sender, address(this)), Errors.SENDER_NOT_ALLOWED);
        }
```

pkg/vault/contracts/TimelockAuthorizer.sol
```diff
@@ -0,0 +1,368 @@
...
+     * @dev Cancels a scheduled action `scheduledActionId`
+     */
+    function cancel(uint256 scheduledActionId) external {
+        require(scheduledActionId < scheduledActions.length, "ACTION_DOES_NOT_EXIST");
+        ScheduledAction storage scheduledAction = scheduledActions[scheduledActionId];
+
+        require(!scheduledAction.executed, "ACTION_ALREADY_EXECUTED");
+        require(!scheduledAction.cancelled, "ACTION_ALREADY_CANCELLED");
+
+        bytes32 actionId = IAuthentication(scheduledAction.where).getActionId(_decodeSelector(scheduledAction.data));
```
nventuro, line 230:
```suggestion
        // The permission to cancel a scheduled action is the same one used to schedule it
        bytes32 actionId = IAuthentication(scheduledAction.where).getActionId(_decodeSelector(scheduledAction.data));
```

pkg/vault/contracts/TimelockAuthorizer.sol
```diff
@@ -0,0 +1,368 @@
...
+        _grantPermission(getActionId(TimelockAuthorizer.grantPermissions.selector), admin, EVERYWHERE);
+        _grantPermission(getActionId(TimelockAuthorizer.revokePermissions.selector), admin, EVERYWHERE);
+    }
+
+    /**
+     * @dev Tells the action ID for a certain function selector
+     */
+    function getActionId(bytes4 selector) public view override returns (bytes32) {
+        return keccak256(abi.encodePacked(bytes32(uint256(address(this))), selector));
+    }
```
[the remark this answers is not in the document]
facuspagnuolo: This will change once we change grant/revoke actions IDs to consider the action being granted/revoked

pkg/vault/contracts/TimelockAuthorizer.sol
facuspagnuolo: Not sure why this change

nventuro: We skip having to read the execute action id if the action is protected. But more importantly, I think the version with the `if` statement better conveys that the requirement on the caller only exists if the action is protected.

nventuro (approved):

PR #1168, opened 2022-03-15
Refactor relayer tests

TomAFrench:
We've got a lot of duplicated code in the relayer tests at the minute which is annoying me a bit and will only get worse once we add relayer libraries for ERC4626, veBAL, etc.

This PR just pulls some of the duplicated code out into helper files and cleans up some code smells so should be pretty straightforward to review.

`expectEvent` has also had a fix applied where it was treating addresses as case sensitive when filtering.


EndymionJkb (approved):
I especially liked the event cleanup.

PR #1169, opened 2022-03-16
Unseeded LBP PoC

### LINKED ISSUE #1520: Add capital-free LBPs

EndymionJkb:
PoC LiquidityBootstrappingPool that requires no capital. (I left the original LiquidityBootstrappingPool and factory in place, to not break the tests.)

The underlying pool is limited to two tokens, and pays protocol fees in BPT (so far, we need this anyway). It also allows the reserve token to have an asset manager.

There is a pool controller (derived from BaseController, so the manager is transferrable, etc.), which doubles as the asset manager.

There are two factories: SeededLiquidityBootstrappingPoolFactory just deploys the new pool type - but hard-codes the asset manager to zero, and sets the owner to be the deployer: no controller, just a normal LBP, funded in the usual way that requires capital.

UnseededLiquidityBootstrappingPoolFactory creates an AssetManagedLiquidityBootstrappingPoolController, and wires it up to a new AssetManagedLiquidityBootstrappingPool, so that the controller is both the pool owner and reserve token asset manager. (The manager of the pool controller is passed in.)

The result will be an apparently fully funded LBP, but with zero cash balance, so that tokens cannot be "sold back" at first.

Resolves #1520.


pkg/pool-utils/contracts/controllers/AssetManagedLBPController.sol
```diff
@@ -0,0 +1,158 @@
...
+        _projectTokenIndex = projectTokenFirst ? 0 : 1;
+        _reserveTokenIndex = projectTokenFirst ? 1 : 0;
+        _vault = vault;
+    }
+
+    /**
+     * @dev Call this function instead of the usual `joinPool` (init). It will borrow the necessary amount
+     * of reserve tokens using a DODO flashloan.
+     */
+    function fundPool(uint256[] memory initialBalances) external onlyManager {
```
Great-Odin: In the case of using fundPool instead of `joinPool`, how are the parameters we used to be passed to join pool process since fundPool only takes the initialBalances?
ex:
```
        bytes memory userData = abi.encode(0, poolConfig.amounts); // JOIN_KIND_INIT = 0
        Vault(VAULT).joinPool(
            poolId,
            address(this), // sender
            address(this), // recipient
            Vault.JoinPoolRequest(poolConfig.tokens, poolConfig.amounts, userData, false)
        );
      ```

File: pkg/pool-utils/contracts/controllers/AssetManagedLBPController.sol [hunk 2]
```diff
@@ -0,0 +1,158 @@
...
+ * will be supplied using an AAVE flashloan, then immediately withdrawn using the asset manager
+ * functionality to repay the loan.
+ *
+ * The result is an apparently fully funded pool, but with a zero cash balance in the reserve token.
+ * This means project tokens cannot be "sold back" into the pool - at least until enough are sold to
+ * build up the real balance.
+ *
+ * Note that there is a flash loan fee, so the manager will need enough reserve tokens to pay it (and
+ * must approve the controller to pull them during the `fundPool` call). The amount of the fee can be
+ * calculated using the `getFlashLoanFeePercentage` function on the controller.
```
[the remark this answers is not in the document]
EndymionJkb: The current version doesn't need a flash loan (though it would be possible to do it that way): I updated the comment. There shouldn't be any references to flash loans in the code. Is this from an older commit?

pkg/pool-utils/contracts/controllers/AssetManagedLBPController.sol
EndymionJkb: No other arguments are required. It's doing an initialization join, which only requires the initial balances. In the internal join call on the Vault, the sender is the controller contract, and the recipient is the manager. It pulls project tokens from - and sends BPT to - the manager.

pkg/pool-utils/contracts/controllers/AssetManagedLBPController.sol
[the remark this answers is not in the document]
Great-Odin: Charging a fee for this service could also be a good idea to filter out non reputable projects

pkg/pool-utils/contracts/controllers/AssetManagedLBPController.sol
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * @dev The initial funding results in a non-zero managed balance. When the cash balance is greater,
+     * this function can be called to return the managed funds to cash.
+     * If there is a managed balance, and the cash balance is greater, convert the managed balance to cash.
+     */
+    function restorePool() external {
+        (uint256 cash, uint256 managed, , ) = _vault.getPoolTokenInfo(getPoolId(), _reserveToken);
+
+        if (managed > 0 && cash >= _getVirtualReserveAmount()) {
```
alchtruss: Is there any risk in calling this function, when the cash just exceeds or equals the _getVirtualReserveAmount. And then say like have a user sell a bunch of project tokens back to the pool shortly after. I suppose i.e. does calling restorePool() in such case result in the inability to use virtual funds and a low balance of real funds?

EndymionJkb: That function isn't there any more. You're looking at *very* old versions; not sure how you're able to comment on non-current code

EndymionJkb:
Deprioritized for now. Archive for future reference.

EndymionJkb:
De-prioritized for now. It's at a decent stopping point; Notion doc under contracts details future plans.

PR #1176, opened 2022-03-18
Proxify StakelessGauges and add SingleRecipientGauge

TomAFrench:
This PR changes the `StakelessGauge` contract to be compatible with deployment as a proxy contract. 

To demonstrate how initialisation of this contract would look like, I've added a concrete implementation of a `StakelessGauge` which simply forwards the BAL to a set address (`SingleRecipientGauge`).

I've also changed the `ILiquidityGauge` interface to only refer to the shared functionality for all mainnet gauges (The LiquidityGaugeV5 is now a `IStakingLiquidityGauge` which details the deposit/withdraw interface).

Fixes https://github.com/balancer-labs/balancer-v2-monorepo/issues/1144


pkg/liquidity-mining/contracts/gauges/SingleRecipientGaugeFactory.sol
```diff
@@ -0,0 +1,64 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
+
+import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/Clones.sol";
+
+import "./SingleRecipientGauge.sol";
+
+contract LiquidityGaugeFactory {
```
nventuro, line 22:
```suggestion
contract SingleRecipientLiquidityGaugeFactory {
```

pkg/liquidity-mining/contracts/gauges/SingleRecipientGauge.sol
```diff
@@ -0,0 +1,40 @@
...
+    using SafeERC20 for IERC20;
+
+    address private _recipient;
+
+    constructor(IBalancerMinter minter) StakelessGauge(minter) {
+        // solhint-disable-previous-line no-empty-blocks
+    }
+
+    function initialize(address recipient) external override {
+        __StakelessGauge_init();
```
nventuro, line 32:
```suggestion
        __StakelessGauge_init(); // This will revert in all calls except the first one
```

pkg/liquidity-mining/contracts/gauges/SingleRecipientGaugeFactory.sol
```diff
@@ -0,0 +1,64 @@
...
+    ISingleRecipientLiquidityGauge private _gaugeImplementation;
+
+    mapping(address => bool) private _isGaugeFromFactory;
+
+    event GaugeCreated(address indexed gauge, address indexed recipient);
```
nventuro, line 27: What do you think about renaming this to `SingleRecipientGaugeCreated`?

pkg/liquidity-mining/contracts/interfaces/ILiquidityGauge.sol
```diff
@@ -14,24 +14,16 @@
...
 // solhint-disable func-name-mixedcase
 
-interface ILiquidityGauge is IERC20 {
-    function initialize(address lpToken) external;
-
-    function lp_token() external view returns (IERC20);
-
-    function deposit(uint256 value, address recipient) external;
-
-    function withdraw(uint256 value) external;
```
nventuro, line 30: Why are we moving these out of the base interface?

TomAFrench: This is because the `StakelessGauge` implements the `ILiquidityGauge` interface and doesn't contain these concepts. I'm trying to keep this interface as just having functions which are shared across _all_ mainnet gauges.

pkg/liquidity-mining/contracts/gauges/SingleRecipientGaugeFactory.sol
TomAFrench: Hmm, not sure. I'm guessing the argument in favour is that it avoids us matching the event signature for the `LiquidityGaugeFactory` but with a semantically different address emitted (recipient rather than pool)?

nventuro: Yes, that's what I was thinking. Being explicit in the name also reduces the likelihood of getting them mixed up.

TomAFrench: Ok, in that case I'm inclined to go whole hog with renaming `LiquidityGaugeV5` to `StakingLiquidityGauge` in that case. It feels weird to have one implementation having the same name with the entire class of contracts.

TomAFrench: (I'd do this in a separate PR though)

nventuro (approved):

PR #1120, opened 2022-02-03
ERC4626 Linear Pool

gtaschuk:
This implements a linear pool for a ERC4626 token

Written in collaboration with the Overnight.fi team - This was created to work with https://github.com/ovnstable/ovnstable-app/tree/static-usdplus

This supercedes #1105 

References:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4626.md

https://ethereum-magicians.org/t/eip-4626-yield-bearing-vault-standard/7900/46

**NOTE: This works for fee-free ERC4626 tokens, and we are working on using `totalAssets/totalSupply` rather than `assetsPerShare` which would incorporate fees, which would better be accounted for in the swap fees**

PR #1177, opened 2022-03-18
Linear Pool: ERC4626 Finalized

rabmarut:
[EIP-4626](https://eips.ethereum.org/EIPS/eip-4626) has been finalized as of yesterday. This PR updates the `ERC4626LinearPool` in light of new additions to the standard. It builds upon the previous work from #1120.

Notably, after the initial development effort of the linear pool, the EIP-4626 standard added `convertToAssets` and `convertToShares` functions, which should be more reliable than the pool's former rate calculation method of `totalAssets/totalSupply`. These functions provide assets<>shares conversions free of any deposit/withdrawal fees, slippage, or caller-specific logic, which was exactly the spirit of the pool's original rate calculation.

Note that, although some additional pieces of the interface have been renamed, and the `Withdraw` event now has an extra parameter, the `deposit` and `redeem` ABIs have not changed, and so no changes are required in the `ERC4626Wrapping` relayer contract.

### DIFF
--- a/pkg/pool-linear/contracts/erc4626/ERC4626LinearPool.sol
+++ b/pkg/pool-linear/contracts/erc4626/ERC4626LinearPool.sol
@@ -31,7 +31,7 @@ contract ERC4626LinearPool is LinearPool {
         string memory name,
         string memory symbol,
         IERC20 mainToken,
-        IERC20 wrappedToken,
+        IERC4626 wrappedToken,
         uint256 upperTarget,
         uint256 swapFeePercentage,
         uint256 pauseWindowDuration,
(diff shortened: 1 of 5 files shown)


pkg/pool-linear/contracts/erc4626/ERC4626LinearPool.sol
```diff
@@ -67,29 +67,23 @@ contract ERC4626LinearPool is LinearPool {
         // whereas mainToken is DAI. But the 1:1 relationship holds, and
         // the pool is still valid.
 
-        // _getWrappedTokenRate is scaled e18, so we may need to scale totalAssets/totalSupply
+        // _getWrappedTokenRate is scaled e18, so we may need to scale IERC4626.convertToAssets()
         uint256 wrappedTokenDecimals = ERC20(address(wrappedToken)).decimals();
         uint256 mainTokenDecimals = ERC20(address(mainToken)).decimals();
-        uint256 digitsDifference = Math.add(18, wrappedTokenDecimals).sub(mainTokenDecimals);
+        uint256 digitsDifference = wrappedTokenDecimals.sub(mainTokenDecimals);
```
gtaschuk, line 73: assume it's true that `wrappedTokenDecimals` is always > `mainTokenDecimals`?

gtaschuk (approved):

pkg/pool-linear/contracts/erc4626/ERC4626LinearPool.sol
rabmarut: Argh no, it's not always true. It's arbitrarily true in the test, so it wasn't caught. Thanks for noticing - I'll fix it

rabmarut:
In order to fix the decimal comparison, I decided to load two uint256s on each rate calculation. Is this egregious? I don't want to gas golf it to the point of sacrificing readability for what should be some dead-simple math. It seems the only way to get around this is to pack either these two numbers, or else a number and a scaling direction, into one storage slot. Worth the effort or no?

gtaschuk (approved):

PR #1178, opened 2022-03-18
Create new package for governance script contracts

TomAFrench:
This PR moves the `veBALDeploymentCoordinator` contract into the `governance-scripts` package.

This new package will contain any script-like contracts we write in order for Balancer governance to action complex proposals. Each script would live in a separate folder similar to how we currently handle the `deployments` package.

The main difference is the date included in the folder name is to be the date at which the script goes up for a vote on Snapshot. I've then left the date for the `veBALDeploymentCoordinator` as `202303xx` for now and we can update once we know on what day we'll vote on this contract.


pkg/governance-scripts/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "!contracts/test/*"
+  ],
+  "scripts": {
+    "build": "yarn compile",
+    "compile": "hardhat compile && rm -rf artifacts/build-info",
+    "compile:watch": "nodemon --ext sol --exec yarn compile",
+    "lint": "yarn lint:typescript && yarn lint:solidity",
+    "lint:typescript": "eslint . --ext .ts --ignore-path ../../.eslintignore  --max-warnings 0",
+    "lint:solidity": "solhint 'contracts/**/*.sol'",
+    "test": "yarn compile && mocha --extension ts --require hardhat/register --require @balancer-labs/v2-common/setupTests --recursive",
```
nventuro, line 26: We should add these tests to the CI file at the top level

nventuro (approved):

pkg/governance-scripts/package.json
TomAFrench: True

TomAFrench: I always forget that

nventuro: That's how I remember :grin: we always have these followed up by the 'oops CI was not running' PR

nventuro (approved):

TomAFrench:
I'm going to hold off on merging this until we get confirmation of the date to go in the folder name

PR #1180, opened 2022-03-18
Prevent multiple SingleRecipientGauges for the same recipient

TomAFrench:
This PR aims to improve onchain discover-ability of SingleRecipientGauges. The `SingleRecipientGaugeFactory` now maintains a record of the gauge deployed for a given recipient and prevents duplicated gauges for a single recipient.

This does raise a little bit of complexity for the inbetween period where we're merkle dropping rewards for Polygon and Arbitrum as we'll need separate addresses to claim each network's rewards but I think a small amount of pain there is outweighed by the benefits in future.

This change is also motivated by the fact that we'll need this behaviour on the Polygon/Arbitrum gauge factories to help prevent duplicated gauges being added to the gauge factory.

### DIFF
--- a/pkg/liquidity-mining/contracts/gauges/SingleRecipientGaugeFactory.sol
+++ b/pkg/liquidity-mining/contracts/gauges/SingleRecipientGaugeFactory.sol
@@ -23,6 +23,7 @@ contract SingleRecipientLiquidityGaugeFactory {
     ISingleRecipientLiquidityGauge private _gaugeImplementation;
 
     mapping(address => bool) private _isGaugeFromFactory;
+    mapping(address => address) private _recipientGauge;
 
     event SingleRecipientGaugeCreated(address indexed gauge, address indexed recipient);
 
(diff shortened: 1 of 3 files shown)


pkg/liquidity-mining/contracts/gauges/SingleRecipientGaugeFactory.sol
```diff
@@ -44,6 +45,20 @@ contract SingleRecipientLiquidityGaugeFactory {
...
+    function getRecipientGauge(address recipient) external view returns (ILiquidityGauge) {
+        return ILiquidityGauge(_recipientGauge[recipient]);
+    }
+
+    /**
+     * @notice Returns the recipient of `gauge`.
+     */
+    function getGaugeRecipient(address gauge) external view returns (address) {
+        return ISingleRecipientLiquidityGauge(gauge).getRecipient();
+    }
```
nventuro, line 60: Is this for lazy callers? :stuck_out_tongue:

nventuro (approved):

pkg/liquidity-mining/contracts/gauges/SingleRecipientGaugeFactory.sol
TomAFrench: Yeah, I'm starting to think that assuming anyone looking up info on the gauge system is lazy is a good habit in general.

PR #1181, opened 2022-03-18
Add PolygonRootGauge

TomAFrench:
Merge after #1180 

This PR adds a new `StakelessGauge` implementation which will bridge the minted BAL to a recipient on Polygon.

For context how to use the Polygon POS bridge to verify `_postMintAction`, see [here](https://docs.polygon.technology/docs/develop/ethereum-polygon/pos/calling-contracts/erc20)


pkg/liquidity-mining/contracts/gauges/SingleRecipientGaugeFactory.sol
```diff
@@ -52,11 +67,14 @@ contract SingleRecipientLiquidityGaugeFactory {
      * @return The address of the deployed gauge
      */
     function create(address recipient) external returns (address) {
+        require(_recipientGauge[recipient] == address(0), "Gauge already exists");
```
nventuro, line 70: This seems like an odd restriction, though I don't see why we'd have gauges with the same type and same recipient (why not use the same one?), or different type and same recipient (since each type is quite different).

pkg/liquidity-mining/contracts/gauges/polygon/PolygonRootGauge.sol
```diff
@@ -0,0 +1,70 @@
...
+        IERC20 token,
+        bytes calldata depositData
+    ) external;
+}
+
+contract PolygonRootGauge is ISingleRecipientLiquidityGauge, StakelessGauge {
+    IPolygonRootChainManager private immutable _polygonRootChainManager;
+    address private immutable _polygonERC20Predicate;
+
+    address private _recipient;
```
nventuro, line 33:
```suggestion
    // This value is kept in storage and not made immutable to allow for this contract to be proxyable
    address private _recipient;
```

nventuro (commented):
Looks great, thanks for including the link to the Polygon docs!

I'm wondering how extensive our testing efforts for these things should be. I think it might be good to have a simple bridge mock, but testing gauges requires interacting with the rest of the veBAL system, which may be problematic.

pkg/liquidity-mining/contracts/gauges/SingleRecipientGaugeFactory.sol
TomAFrench: Yeah, I was slightly torn on this.

Gauges are defined by a "key"-gauge mapping and for the standard L1 and L2 gauges this mapping is naturally 1-to-1 as you can't stake into multiple gauges and two root gauges bridging funds to a single streamer is equivalent to one. 

`SingleRecipientGauge` is the only situation where it kinda makes sense to relax this to be 1-to-many. Example of this being the temporary gauges for veBAL/Polygon/Arbitrum gauge types can't just send funds to the LM committee multisig but need to be proxied through another contract/multisig.

I'm erring on the side of maintaining 1-to-1 for reasons of consistency, discouraging commingling of funds where possible and an ability to look up the gauge for a given recipient easily.

TomAFrench: That said, this change is already in master.

nventuro: I think those trade-offs make sense. Thanks!

nventuro:
I think this can be merged as-is, but we should have some level of testing in the form of fork-tests (as unit tests would be a bit cumbersome to do) before deploying.

TomAFrench:
Merge conflicts are fixed now

PR #1185, opened 2022-03-21
Prevent duplicate gauges for a pool being added to GaugeController

TomAFrench:
This PR implements a global mapping for BPT addresses to the canonical gauge address. This is required as because we support multiple factories it's possible to have multiple gauges for the same pool eligible to be added to the GaugeController. Without this PR, we'd then have to rely on the entities authorised to call `addEthereumGauge` not to do so for these gauges.

~~Having this mapping is also very useful for building an onchain registry for the veBAL system as we can allow a trustless migration of this data to a standalone registry contract.~~ Actually we could just do it from gauge to address using the GaugeContoller.

Note that this will prevent us from deploying a new gauge implementation for a particular pool unless we migrate the GaugeAdder contract but I think that this is a relatively low priority issue and we can work around it.

### DIFF
--- a/pkg/liquidity-mining/contracts/GaugeAdder.sol
+++ b/pkg/liquidity-mining/contracts/GaugeAdder.sol
@@ -129,10 +129,9 @@ contract GaugeAdder is IGaugeAdder, Authentication, ReentrancyGuard {
         // We then check here to see if the new gauge's pool already has a gauge on the Gauge Controller
         address pool = address(gauge.lp_token());
         require(_poolGauge[pool] == ILiquidityGauge(0), "Duplicate gauge");
+        _poolGauge[pool] = gauge;
 
         _addGauge(address(gauge), GaugeType.Ethereum);
-
-        _poolGauge[pool] = gauge;
     }
 
     /**
--- a/pkg/liquidity-mining/contracts/GaugeAdder.sol
+++ b/pkg/liquidity-mining/contracts/GaugeAdder.sol
@@ -21,6 +21,7 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/ReentrancyGuard.
 import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
 
 import "./interfaces/IGaugeAdder.sol";
+import "./interfaces/IStakingLiquidityGauge.sol";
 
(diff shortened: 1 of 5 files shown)


pkg/liquidity-mining/contracts/GaugeAdder.sol
```diff
@@ -109,8 +123,16 @@ contract GaugeAdder is IGaugeAdder, Authentication, ReentrancyGuard {
...
+    function addEthereumGauge(IStakingLiquidityGauge gauge) external override authenticate {
+        // Each gauge factory prevents deploying multiple gauges for the same Balancer pool
+        // however two separate factories can each deploy their own gauge for the same pool.
+        // We then check here to see if the new gauge's pool already has a gauge on the Gauge Controller
+        address pool = address(gauge.lp_token());
+        require(_poolGauge[pool] == ILiquidityGauge(0), "Duplicate gauge");
+
+        _addGauge(address(gauge), GaugeType.Ethereum);
+
+        _poolGauge[pool] = gauge;
```
nventuro, line 135:
```suggestion
        _poolGauge[pool] = gauge;
        _addGauge(address(gauge), GaugeType.Ethereum);
```

This prevents reentrancy issues

pkg/liquidity-mining/test/GaugeAdder.test.ts
```diff
@@ -172,9 +172,15 @@ describe('GaugeAdder', () => {
   });
 
   describe('addEthereumGauge', () => {
+    let gauge: string;
```
nventuro, line 175: We should probably have a test for duplicate gauge prevention

nventuro (commented):
> Actually we could just do it from gauge to address using the GaugeContoller.

I don't follow that statement, could you expand?

TomAFrench:
> > Actually we could just do it from gauge to address using the GaugeContoller.
> 
> I don't follow that statement, could you expand?

This was my confusing way of saying that the `_poolGauge` mapping doesn't actually help us migrate information on the links between pools and gauges into a standalone registry contract as we can just iterate through gauges on the GaugeController and then query on the gauge which pool it is the gauge for.

Going from gauge -> pool rather than pool -> gauge.

TomAFrench:
@nventuro are we gtg on this?

nventuro (approved):

PR #1186, opened 2022-03-21
Add contracts for distributing BAL rewards on L2s

### LINKED ISSUE #1142: Add L2 Gauges
These are the same as the L1 gauges, except without the calculations related to BAL minting.

### LINKED ISSUE #1143: Add L2 Gauge Factory
This will deploy gauges from #1142, following a similar pattern to the current gauge factory (ability to change the implementation, etc.). Concerns from #1140 apply.

On gauge creation, we should also create a simple GaugeBALSeeder contract associated with the gauge, which when called adds BAL minted on L1 to the gauge. This seeder will be the recipient of the gauges described in #1115 (tentatively called Single Recipient Gauges).

The timing of when the seeder adds tokens to the gauge is important, as undistributed tokens from previous weeks are rolled-over into the next week. However, we also need to make sure that the L1 funds have been minted and sent before scheduling any action. We could initially make this action permissioned and have a team on top of operations, but ideally we'd devise a way for this to be pubicly callable.

TomAFrench:
#1181 adds a `PolygonRootGauge` which mints BAL and bridges it to a recipient on the Polygon network, but we then need some way to distribute this BAL to LPs of the relevant pool on Polygon. This PR adds the required contracts to do so.

The "child" chain system is composed of two main contracts: a `ChildChainStreamer` and a `RewardsOnlyGauge`. These separate concerns of "vesting tokens over a set period" and "distributing vested tokens proportionally to stake" respectively.

The `RewardsOnlyGauge` doesn't have any concept of the rate to pay out tokens as `LiquidityGaugeV5` does, it instead relies on `ChildChainStreamer` from which it claims all currently vested tokens from the `ChildChainStreamer` on every interaction and distributes it proportionally to stakers.

The `ChildChainStreamer` receives BAL from the `PolygonRootGauge` each week where it sits until someone calls `notify_reward_amount(token)`, this balance of BAL is then slowly released to the `RewardsOnlyGauge` over the course of the week. To prevent abuse, if there is an reward stream in progress for a token then only the distributor for that token can call this function.

fixes #1142 
fixes #1143

### DIFF
--- a/pkg/liquidity-mining/contracts/gauges/ChildChainStreamer.vy
+++ b/pkg/liquidity-mining/contracts/gauges/ChildChainStreamer.vy
@@ -109,6 +109,8 @@ def remove_reward(_token: address, _recipient: address):
             self.reward_tokens[i] = self.reward_tokens[idx]
             self.reward_tokens[idx] = ZERO_ADDRESS
             self.reward_count = idx
+            log RewardDistributorUpdated(_token, ZERO_ADDRESS)
+            log RewardDurationUpdated(_token, 0)
             return
     raise  # this should never be reached
 


TomAFrench:
We likely need to modify the `RewardsOnlyGauge` so that the frontend can pull the reward data in the same way as mainnet.

pkg/liquidity-mining/contracts/gauges/RewardsOnlyGauge.vy
```diff
@@ -0,0 +1,613 @@
...
+    rate: uint256
+    last_update: uint256
+    integral: uint256
+
+MAX_REWARDS: constant(uint256) = 8
+CLAIM_FREQUENCY: constant(uint256) = 3600
+
+# keccak256("isValidSignature(bytes32,bytes)")[:4] << 224
+ERC1271_MAGIC_VAL: constant(bytes32) = 0x1626ba7e00000000000000000000000000000000000000000000000000000000
+VERSION: constant(String[8]) = "v5.0.0"
```
nventuro, line 62: This is not a v5 :stuck_out_tongue:

pkg/liquidity-mining/contracts/gauges/RewardsOnlyGauge.vy
```diff
@@ -0,0 +1,613 @@
...
+    _from: indexed(address)
+    _to: indexed(address)
+    _value: uint256
+
+event Approval:
+    _owner: indexed(address)
+    _spender: indexed(address)
+    _value: uint256
+
+struct Reward:
```
nventuro, line 49: Is this just to return a bunch of data in one go?

pkg/liquidity-mining/contracts/gauges/RewardsOnlyGauge.vy
```diff
@@ -0,0 +1,613 @@
...
+        new_claimable: uint256 = 0
+        if integral_for < integral:
+            self.reward_integral_for[token][_user] = integral
+            new_claimable = user_balance * (integral - integral_for) / 10**18
+
+        claim_data: uint256 = self.claim_data[_user][token]
+        total_claimable: uint256 = shift(claim_data, -128) + new_claimable
+        if total_claimable > 0:
+            total_claimed: uint256 = claim_data % 2 ** 128
+            if _claim:
```
nventuro, line 190: It seems there's a difference here between the deployed version I looked at and [the code in the Curve repo](https://github.com/curvefi/curve-dao-contracts/blob/master/contracts/gauges/RewardsOnlyGauge.vy):

image

The change was introduced here: https://github.com/curvefi/curve-dao-contracts/pull/105. It seems like it's the only difference between the code deployed and the code on the repo.

pkg/liquidity-mining/contracts/gauges/ChildChainLiquidityGaugeFactory.sol
```diff
@@ -0,0 +1,155 @@
...
+    function getVault() public view returns (IVault) {
+        return _vault;
+    }
+
+    /**
+     * @dev Returns the address of the Vault's Authorizer.
+     */
+    function getAuthorizer() public view returns (IAuthorizer) {
+        return getVault().getAuthorizer();
+    }
```
nventuro, line 76: This contract does not use neither the Vault nor the Authorizer

pkg/liquidity-mining/contracts/gauges/ChildChainLiquidityGaugeFactory.sol
```diff
@@ -0,0 +1,155 @@
...
+
+        IChildChainStreamer(streamer).initialize(gauge);
+        IRewardsOnlyGauge(gauge).initialize(pool, streamer, _CLAIM_SIG);
+
+        _isGaugeFromFactory[gauge] = true;
+        _poolGauge[pool] = gauge;
+        _gaugeStreamer[gauge] = streamer;
+        emit RewardsOnlyGaugeCreated(gauge, pool, streamer);
+
+        return gauge;
```
nventuro, line 153: I thought about returning the streamer here, which we'd need if we wanted to manage permissions for it, but we can always query it from the gauge instead.

nventuro (commented):
I compared `Child-Chain Streamer` with the implementation deployed at https://polygonscan.com/address/0xC48f4653dd6a9509De44c92beb0604BEA3AEe714#code (from [the Curve docs](https://curve.readthedocs.io/ref-addresses.html#polygon)). and `RewardsOnlyGauge` with the implementation deployed at https://polygonscan.com/address/0x19793B454D3AfC7b454F206Ffe95aDE26cA6912c#code (from [the Curve docs](https://curve.readthedocs.io/ref-addresses.html#polygon)).

pkg/liquidity-mining/contracts/gauges/ChildChainStreamer.vy
```diff
@@ -0,0 +1,217 @@
...
+last_update_time: public(uint256)
+
+
+@external
+def __init__(_bal_token: address, _authorizerAdaptor: address):
+    BAL_TOKEN = _bal_token
+    AUTHORIZER_ADAPTOR = _authorizerAdaptor
+
+    # prevent initialization of implementation
+    self.reward_receiver = 0x000000000000000000000000000000000000dEaD
```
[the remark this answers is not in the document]
TomAFrench: It's vyper so not using proxies would require reimplementing the contract in solidity. As you say these are low cost environments so it's not really worth it for the savings.

pkg/liquidity-mining/contracts/gauges/RewardsOnlyGauge.vy
TomAFrench: Yes, this is so that the `reward_data(address)` function behaves the same on `RewardsOnlyGauge` and `LiquidityGaugeV5` to make things easier on the frontend otherwise they'd have to handle external rewards differently on mainnet vs polygon.

pkg/liquidity-mining/contracts/gauges/ChildChainStreamer.vy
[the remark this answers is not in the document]
nventuro: Ah forgot about mixed language factories. Thanks.

pkg/liquidity-mining/contracts/gauges/RewardsOnlyGauge.vy
TomAFrench: The current implementation looks correct and more gas efficient so I'm inclined to leave it as is.

pkg/liquidity-mining/contracts/gauges/ChildChainLiquidityGaugeFactory.sol
TomAFrench: I've added a lazy getter to make it easier to go from pool to streamer but I think it's good to be consistent that gauge factories will always return the address of the gauge.

pkg/liquidity-mining/contracts/gauges/ChildChainStreamer.vy
```diff
@@ -0,0 +1,217 @@
...
+    amount: uint256 = ERC20(_token).balanceOf(self)
+    response: Bytes[32] = raw_call(
+        _token,
+        concat(
+            method_id("transfer(address,uint256)"),
+            convert(msg.sender, bytes32),
+            convert(amount, bytes32),
+        ),
+        max_outsize=32,
+    )
```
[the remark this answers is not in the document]
TomAFrench: I've added this along with the ability to send tokens to an address other than the authorizer adaptor as we'll always just immediately sweep tokens off of the adaptor.

pkg/liquidity-mining/contracts/gauges/ChildChainStreamer.vy
```diff
@@ -0,0 +1,217 @@
...
+            actual_balance: uint256 = ERC20(token).balanceOf(self)
+
+            if actual_balance > expected_balance:
+                new_amount: uint256 = actual_balance - expected_balance
+                duration: uint256 = self.reward_data[token].duration
+
+                if block.timestamp >= self.reward_data[token].period_finish:
+                    self.reward_data[token].rate = new_amount / duration
+                else:
+                    assert msg.sender in [AUTHORIZER_ADAPTOR, self.reward_data[_token].distributor], "Reward period still active"
```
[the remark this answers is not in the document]
TomAFrench: Eh, I'm a little cautious about making too many modifications to the authentication mechanisms as we'd then need to add proper testing for them.

I felt that allowing the adaptor to poke this contract would be helpful as we can always atomically change the distributor, poke and change back so this just makes things easier.

pkg/liquidity-mining/contracts/gauges/ChildChainStreamer.vy
```diff
@@ -0,0 +1,217 @@
...
+    @notice Add a reward token
+    @param _token Address of the reward token
+    @param _distributor Address permitted to call `notify_reward_amount` for this token
+    @param _duration Number of seconds that rewards of this token are streamed over
+    """
+    assert self.reward_data[_token].distributor == ZERO_ADDRESS, "Reward token already added"
+
+    idx: uint256 = self.reward_count
+    self.reward_tokens[idx] = _token
+    self.reward_count = idx + 1
```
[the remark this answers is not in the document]
TomAFrench: Ugh, yeah. Sorry brainfart.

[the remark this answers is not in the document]
TomAFrench: Yeah, this will revert properly.

nventuro (approved):

PR #1188, opened 2022-03-21
fix: Decimal math errors in stable-pool math.ts

RedFox20:
calculateApproxInvariant used incorrect javascript `>` operator, which resulted in incorrect converge. Additionally converge must happen when BigNumber integer representation of `inv - prevInv` is <= integer 1

### DIFF
--- a/pvt/helpers/src/models/pools/stable/math.ts
+++ b/pvt/helpers/src/models/pools/stable/math.ts
@@ -37,12 +37,12 @@ export function calculateApproxInvariant(
       .add(ampTimesTotal.mul(sum).mul(P_D))
       .div(decimal(totalCoins).add(1).mul(inv).add(ampTimesTotal.sub(1).mul(P_D)));
 
-    // Equality with the precision of 1
-    if (inv > prevInv) {
-      if (inv.sub(prevInv).lte(1)) {
+    // converge with precision of integer 1
+    if (inv.gt(prevInv)) {
+      if (fp(inv).sub(fp(prevInv)).lte(1)) {
         break;
       }
-    } else if (prevInv.sub(inv).lte(1)) {
+    } else if (fp(prevInv).sub(fp(inv)).lte(1)) {
       break;
     }
   }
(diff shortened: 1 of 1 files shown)


RedFox20:
I was writing some Unit Tests based on stable/math.ts, and found that there were major divergences compared to StableMath.sol. The main issue ended up being the convergence condition, which in solidity is done on uint256 integers, while in stable/math.ts it's done on Decimal classes. So stable/math.ts converged prematurely in many cases.

Opened this PR with a quick workaround, though I'm sure something more elegant can be done. There are some math precision differences between StableMath.sol and stable/math.ts, so this workaround does not converge as well. Another workaround could be simply using `.lte(10)` instead of `.lte(1)`. A better one is rewriting some of these to simply use `BigNumber` instead of Decimal.js

axic:
I think [here](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pvt/helpers/src/models/pools/weighted/math.ts#L64) and [here](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pvt/helpers/src/models/pools/weighted/math.ts#L128) may be affected too, because `balanceRatiosWithFee[i]` is a `BigNumber`. And likely [this](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pvt/helpers/src/models/pools/weighted/math.ts#L197) may be safer using the BigNumber operators.

nventuro:
Thanks a lot for the report! We'll look into it :raised_hands:

dmf7z (approved):
Thanks for identifying this @RedFox20. In order to keep it similar to the smart contract math, we will not change the code structure. However we will evaluate using BigNumber instead of Decimal.js for more precision.

dmf7z:
> I think [here](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pvt/helpers/src/models/pools/weighted/math.ts#L64) and [here](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pvt/helpers/src/models/pools/weighted/math.ts#L128) may be affected too, because `balanceRatiosWithFee[i]` is a `BigNumber`. And likely [this](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pvt/helpers/src/models/pools/weighted/math.ts#L197) may be safer using the BigNumber operators.

Thx @axic , we will update it on a different PR.

PR #1191, opened 2022-03-23
Add custom implementation of FeeDistributor contract

TomAFrench:
~~**Note: This is still in development. We're going to need a pretty sizable test suite for this and I have some concerns about whether edge-cases on timestamps are always handled properly.**~~

This PR implements an extension to Curve's [FeeDistributor](https://github.com/curvefi/curve-dao-contracts/blob/master/contracts/FeeDistributor.vy) contract for the distribution of protocol fees to veCRV holders.

While CRV distributes a single token to it veCRV holderes, veBAL holders will receive at least 2 tokens. This gives two possible options:

1. Deploy two copies of `FeeDistributor.vy`, one for BAL and one for bb-a-USD.
2. Add multi-token support to `FeeDistributor`.

Option 2 is fairly simple by just storing the user's veBAL balance in storage rather immediately transferring tokens to the user. This decouples user and token checkpointing such that claiming is just a matter of iterating through each week and multiplying the amount of fees collected by the fraction of the total veBAL supply made up by the user.

This then allows arbitrary numbers of different tokens to be distributed through the contract without any risk from malicious tokens (distributing a new token to veBAL holders is as simple as transferring it to the `FeeDistributor` contract).


pkg/liquidity-mining/contracts/fee-distribution/FeeDistributor.sol
```diff
@@ -0,0 +1,369 @@
...
+    mapping(IERC20 => uint256) private _tokenStartTime;
+    mapping(IERC20 => uint256) private _tokenTimeCursor;
+    mapping(IERC20 => uint256) private _tokenLastBalance;
+    mapping(IERC20 => mapping(uint256 => uint256)) private _tokensPerWeek;
+
+    // User State
+    mapping(address => uint256) private _userTimeCursor;
+    mapping(address => uint256) private _userLastEpochCheckpointed;
+    mapping(address => mapping(uint256 => uint256)) private _userBalanceAtTimestamp;
+    mapping(address => mapping(IERC20 => uint256)) private _userTokenTimeCursor;
```
TomAFrench, line 47: Several of these mappings should be replaced with mappings to packed structs, however this should be done after initial reviews for readability sake.

pkg/liquidity-mining/contracts/fee-distribution/FeeDistributor.sol
```diff
@@ -0,0 +1,410 @@
...
+        _userTokenTimeCursor[user][token] = userTimeCursor;
+
+        if (amount > 0) {
+            _tokenLastBalance[token] -= amount;
+            token.safeTransfer(user, amount);
+            emit TokensClaimed(user, token, amount, userTimeCursor);
+        }
+
+        return amount;
+    }
```
TomAFrench, line 177: As a knock on consequence we need to add this function to handle claiming of tokens with each token getting its own time cursor to ensure that a user can only claim fees from a given week once.

pkg/liquidity-mining/contracts/fee-distribution/FeeDistributor.sol
TomAFrench: I've got these changes in a separate branch so will make a PR after this one to apply these changes.

nventuro:
I've checked that the GitHub code matches the code deployed at https://etherscan.io/address/0xA464e6DCda8AC41e03616F95f4BC98a13b8922Dc#code, which is [their official deployment](https://curve.readthedocs.io/ref-addresses.html#curve-dao).

pkg/liquidity-mining/contracts/fee-distribution/FeeDistributor.sol
```diff
@@ -0,0 +1,501 @@
...
+                // The week being considered lies inside the user epoch described by `oldUserPoint`
+                // we can then use it to calculate the user's balance at the beginning of the week.
+
+                int128 dt = int128(weekCursor - oldUserPoint.ts);
+                uint256 userBalance = oldUserPoint.bias > oldUserPoint.slope * dt
+                    ? uint256(oldUserPoint.bias - oldUserPoint.slope * dt)
+                    : 0;
+
+                // User's lock has expired and they haven't relocked yet.
+                if (userBalance == 0 && userEpoch > maxUserEpoch) break;
```
TomAFrench, line 405: if (userBalance == 0 && userEpoch > maxUserEpoch) {
                    weekCursor = _roundUpTimestamp(block.timestamp);
                    break;
                }

TomAFrench: Suggestion from Nico, we can fast forward the week cursor to now as the clearly hasn't received any rewards in the intervening weeks.

pkg/liquidity-mining/contracts/fee-distribution/FeeDistributor.sol
```diff
@@ -0,0 +1,501 @@
...
+
+    // Helper functions
+
+    /**
+     * @dev Wrapper around `_userTokenTimeCursor` which returns the start timestamp for `token`
+     * if `user` has not attempted to interact with it previously.
+     */
+    function _getUserTokenTimeCursor(address user, IERC20 token) internal view returns (uint256) {
+        uint256 userTimeCursor = _userTokenTimeCursor[user][token];
+        return userTimeCursor > 0 ? userTimeCursor : _tokenStartTime[token];
```
TomAFrench, line 458: Improvement for a later PR, we should also have a user-level start time so that someone who locks veBAL in a year doesn't have to claim 52 empty weeks first.

pkg/liquidity-mining/contracts/fee-distribution/FeeDistributor.sol
```diff
@@ -0,0 +1,501 @@
...
+    function getTotalSupplyAtTimestamp(uint256 timestamp) external view override returns (uint256) {
+        return _veSupplyCache[timestamp];
+    }
+
+    /**
+     * @notice Returns the FeeDistributor's cached balance of `token`.
+     */
+    function getTokenLastBalance(IERC20 token) external view override returns (uint256) {
+        return _tokenLastBalance[token];
+    }
```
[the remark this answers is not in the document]
TomAFrench: Yeah I think so, we're breaking compatibility in a number of places due to needing an extra argument for which token we're interested in so I think we can just go all the way.

As anyone can claim on behalf of others the only thing that a contract _needs_ to be compatible is the ability to sweep multiple tokens.

pkg/liquidity-mining/contracts/fee-distribution/FeeDistributor.sol
```diff
@@ -0,0 +1,496 @@
...
+     * @notice Claims all pending distributions of the provided token for a user.
+     * @dev It's not necessary to explicitly checkpoint before calling this function, it will ensure the FeeDistributor
+     * is up to date before calculating the amount of tokens to be claimed.
+     * @param user - The user on behalf of which to claim.
+     * @param token - The ERC20 token address to be claimed.
+     * @return The amount of `token` sent to `user` as a result of claiming.
+     */
+    function claimToken(address user, IERC20 token) external override nonReentrant returns (uint256) {
+        // Prevent someone from assigning tokens to an inaccessible week.
+        require(block.timestamp > _startTime, "Fee distribution has not started yet");
```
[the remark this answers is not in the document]
TomAFrench: How so? If a token is checkpointed before `_startTime` then any funds will be assigned to a week before the first one which is able to be claimed.

As a sidenote I'm going to move this into the "first interaction with token" block within `_checkpointToken`

pkg/liquidity-mining/contracts/fee-distribution/FeeDistributor.sol
```diff
@@ -0,0 +1,496 @@
...
+        return (timestamp / 1 weeks) * 1 weeks;
+    }
+
+    /**
+     * @dev Rounds the provided timestamp up to the beginning of the next week (Thurs 00:00 UTC)
+     */
+    function _roundUpTimestamp(uint256 timestamp) private pure returns (uint256) {
+        return _roundDownTimestamp(timestamp + 1 weeks - 1);
+    }
+}
```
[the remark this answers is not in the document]
TomAFrench: We _could_ but I don't think it's particularly useful. It makes sense for Curve's case as they're only distributing 3CRV but it's much messier in our case as everything is distributed.

We can't withdraw any tokens which have been checkpointed as we've committed to paying those out to veBAL holders and the last veBAL holder won't get any tokens as their share got paid to earlier claimers.

We can then only withdraw uncheckpointed tokens which causes annoying dynamics as anyone can checkpoint the token to grief (Not to mention that any amount of tokens with any sizeable economic value would be proactively checkpointed by veBAL holders).

pkg/liquidity-mining/contracts/fee-distribution/FeeDistributor.sol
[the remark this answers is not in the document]
nventuro: Ahh that make sense, yes. But your improvement makes even more sense than the comment :stuck_out_tongue:

PR #1193, opened 2022-03-23
Authorizer: Fix missing renames

facuspagnuolo:


pkg/vault/contracts/TimelockAuthorizer.sol
```diff
@@ -74,12 +74,12 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
     /**
      * @dev Emitted when an action `actionId` is executed
```
nventuro, line 75: This is also outdated :stuck_out_tongue:

nventuro (approved):

PR #1197, opened 2022-03-23
Authorizer: Improve grant/revoke delays granularity

### LINKED ISSUE #1198: Improve delays granularity for grant and revoke
Currently the authorizer implementation allows only to define a single delay all grant permissions, and similar for revoke permissions. Ideally we would love to support having different delays depending on what permissions the user is trying to grant or revoke. 

Note that this another level of indirection that will probably end up in defining some sort of root permission.

facuspagnuolo:
Fixes #1198

### DIFF
--- a/pkg/vault/contracts/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/TimelockAuthorizer.sol
@@ -44,8 +44,11 @@ import "./interfaces/IVault.sol";
 contract TimelockAuthorizer is IAuthorizer, IAuthentication {
     using Address for address;
 
-    uint256 public constant MAX_DELAY = 2 * (365 days);
+    bytes32 public constant WHATEVER = 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff;
     address public constant EVERYWHERE = address(-1);
+
+    uint256 public constant MAX_DELAY = 2 * (365 days);
+    uint256 public constant ROOT_TRANSFER_DELAY = 7 days;
 
     struct ScheduledExecution {
         address where;
@@ -59,9 +62,12 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
     // solhint-disable var-name-mixedcase
     bytes32 public immutable GRANT_ACTION_ID;
     bytes32 public immutable REVOKE_ACTION_ID;
+    bytes32 public immutable EXECUTE_ACTION_ID;
     bytes32 public immutable SCHEDULE_DELAY_ACTION_ID;
 
     IAuthentication private immutable _vault;
+
+    address public root;
     ScheduledExecution[] public scheduledExecutions;
     mapping(bytes32 => bool) public isPermissionGranted;
     mapping(bytes32 => uint256) public delaysPerActionId;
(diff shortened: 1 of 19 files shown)


pkg/vault/contracts/TimelockAuthorizer.sol
```diff
@@ -152,6 +200,54 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
...
+    /**
+     * @dev Schedules a root change call
+     */
+    function scheduleRootChange(address newRoot, address[] memory executors)
+        external
+        returns (uint256 scheduledExecutionId)
+    {
+        _require(isRoot(msg.sender), Errors.SENDER_NOT_ALLOWED);
+        bytes32 actionId = getActionId(this.setRoot.selector);
+        bytes32 scheduleDelayActionId = getActionId(SCHEDULE_DELAY_ACTION_ID, actionId);
```
nventuro, line 246: Shouldn't this be the action ID of scheduling a root change?

pkg/vault/contracts/TimelockAuthorizer.sol
```diff
@@ -240,6 +338,22 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
...
+    function manageGrantPermission(
+        bytes32 actionId,
+        address account,
+        address where,
+        bool allowed
+    ) external {
+        bytes32 grantPermissionsActionId = getActionId(GRANT_ACTION_ID, actionId);
+        bool isAllowed = isRoot(msg.sender) || hasPermission(grantPermissionsActionId, msg.sender, where);
+        _require(isAllowed, Errors.SENDER_NOT_ALLOWED);
+        (allowed ? _grantPermission : _revokePermission)(grantPermissionsActionId, account, where);
```
nventuro, line 354: The permission to grant should indeed be `grant + action id`, but shouldn't we call `hasPermission(grant, msg.sender, where, actionId)` instead? Otherwise we would not be able to grant `grant + whatever`.

nventuro: (same with revoke)

pkg/vault/test/TimelockAuthorizer.test.ts
```diff
@@ -1239,4 +1343,49 @@ describe('TimelockAuthorizer', () => {
...
+      it('reverts', async () => {
+        await expect(authorizer.instance.setRoot(grantee.address)).to.be.revertedWith('SENDER_NOT_ALLOWED');
+      });
+    });
+
+    context('when trying to schedule a call', async () => {
+      it('schedules a root change', async () => {
+        const expectedData = authorizer.instance.interface.encodeFunctionData('setRoot', [grantee.address]);
+
+        const id = await authorizer.scheduleRootChange(grantee, [], { from: admin });
```
nventuro, line 1360: Let's add a test where the caller is not the current root

pkg/vault/contracts/TimelockAuthorizer.sol
```diff
@@ -96,14 +102,39 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
...
+        root = admin;
         _vault = vault;
-        GRANT_ACTION_ID = getActionId(TimelockAuthorizer.grantPermissions.selector);
-        REVOKE_ACTION_ID = getActionId(TimelockAuthorizer.revokePermissions.selector);
+
+        bytes32 grantActionId = getActionId(TimelockAuthorizer.grantPermissions.selector);
+        _grantPermission(getActionId(grantActionId, WHATEVER), admin, EVERYWHERE);
+
+        bytes32 revokeActionId = getActionId(TimelockAuthorizer.revokePermissions.selector);
+        _grantPermission(getActionId(revokeActionId, WHATEVER), admin, EVERYWHERE);
```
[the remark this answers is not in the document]
facuspagnuolo: > The goal of this to save `root` from having to manage the grant permission in order to then grant permissions to others, correct?

> I'd consider removing this so that the entire new permissions structure is contained within the migration contract. We might otherwise have situations where e.g. the migration is made root and at the end transfers root to a multisig, but we forget to transfer these permissions the migration would end up with.

Not sure how you can do this. I mean, someone needs to have permissions to grant others, if we create the authorizer without setting the root in the constructor then who is able to grant permissions?

pkg/vault/contracts/TimelockAuthorizer.sol
```diff
@@ -141,6 +175,20 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
...
 
+    /**
+     * @dev Tells whether `account` has explicit permission to perform action `actionId` in `where` with `how`
+     */
+    function hasPermission(
+        bytes32 actionId,
+        address account,
+        address where,
+        bytes32 how
+    ) public view returns (bool) {
```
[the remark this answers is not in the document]
facuspagnuolo: Sure, any good idea so we don't fall in `canPerformWithParams` or sth similar?

pkg/vault/test/TimelockAuthorizer.test.ts
```diff
@@ -179,6 +180,1088 @@ describe('TimelockAuthorizer', () => {
...
+      });
+
+      context('when revoking permission', () => {
+        context('for a specific action', () => {
+          const actionId = ACTION_1;
+
+          context('for a specific contract', () => {
+            const where = WHERE_1;
+
+            it('cannot grant permission for that action in that contract only', async () => {
```
nventuro, line 306: These should read 'revoke', no?

nventuro (approved):

pkg/vault/test/TimelockAuthorizer.test.ts
facuspagnuolo: The way I see it is that if you revoke the permission to grant a permission, then you "cannot grant permission for..." anymore

PR #1202, opened 2022-03-24
Remove broken historical lookup functions from VotingEscrow

TomAFrench:
To be merged if #1201 is rejected :sadge:

There's no point in having these functions as they will revert in any meaningful scenario other than getting the current balance/supply. Including them results just bloats the ABI and causes annoyances when using it in ethers.

We should then remove the overload which allows providing a timestamp to query at.


nventuro:
Closing in favor of #1201 :happyge:

PR #1207, opened 2022-03-28
Add SmartWalletChecker

TomAFrench:
The `VotingEscrow` contract requires any attempts to lock BPT into veBAL for which `tx.origin != msg.sender` (i.e. which come from a contract) to be allowlisted on an external contract known as the `SmartWalletChecker`.

https://github.com/balancer-labs/balancer-v2-monorepo/blob/91049cbed3482d0bdf6982ba8e030c47ec317913/pkg/liquidity-mining/contracts/VotingEscrow.vy#L170-L181

Due to interest from third parties wanting to build on top of veBAL we then need a simple implementation of this contract to allow testing, etc.

This PR adds a very simple `SmartWalletChecker` where the Authorizer can allow certain addresses to add entries to a mapping of allowlisted contracts. These powers should always be executed through timelocks or governance scripts.

### DIFF
--- a/pkg/liquidity-mining/contracts/SmartWalletChecker.sol
+++ b/pkg/liquidity-mining/contracts/SmartWalletChecker.sol
@@ -1,0 +1,71 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+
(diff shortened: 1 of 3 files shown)


pkg/liquidity-mining/contracts/SmartWalletChecker.sol
```diff
@@ -0,0 +1,72 @@
...
+        require(!_allowlistedAddresses[contractAddress], "Address already allowlisted");
+        _allowlistedAddresses[contractAddress] = true;
+        emit ContractAddressAdded(contractAddress);
+    }
+
+    function denylistAddress(address contractAddress) external authenticate {
+        require(_allowlistedAddresses[contractAddress], "Address is not allowlisted");
+        _allowlistedAddresses[contractAddress] = false;
+        emit ContractAddressRemoved(contractAddress);
+    }
```
nventuro, line 65: I'd use EnumerableSet for this: it's likely we'd want to migrate permissions from one checker to another.

pkg/liquidity-mining/test/SmartWalletChecker.test.ts
```diff
@@ -0,0 +1,158 @@
...
+      context('when address is already denylisted', () => {
+        it('reverts', async () => {
+          await expect(smartWalletChecker.connect(caller).denylistAddress(ANY_ADDRESS)).to.be.revertedWith(
+            'Address is not allowlisted'
+          );
+        });
+      });
+
+      context('when address is not currently denylisted', () => {
+        sharedBeforeEach('authorize caller', async () => {
```
nventuro, line 138:
```suggestion
        sharedBeforeEach('allowlist address', async () => {
```

PR #1211, opened 2022-03-30
Cache protocol fee module

### LINKED ISSUE #1210: Create ProtocolFeeCache module
We have this behavior in a couple contracts already, it'd be good to create a small module out of it and reuse the code that way.

EndymionJkb:
Fixes #1210 - factors caching the protocol fee out of ManagedPool and StablePhantomPool


pkg/pool-utils/contracts/ProtocolFeeCache.sol
```diff
@@ -0,0 +1,91 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/BalancerErrors.sol";
+import "@balancer-labs/v2-vault/contracts/interfaces/IVault.sol";
+
+abstract contract ProtocolFeeCache {
+    uint256 public constant DELEGATE_PROTOCOL_FEES_SENTINEL = type(uint256).max;
```
nventuro, line 21: There is no need to make this public, as it is an implementation detail. We can keep it internal.

pkg/pool-utils/contracts/ProtocolFeeCache.sol
```diff
@@ -0,0 +1,91 @@
...
+    // value), so we need to fetch it ourselves from the Vault's ProtocolFeeCollector. However, this value changes so
+    // rarely that it doesn't make sense to perform the required calls to get the current value in every single swap.
+    // Instead, we keep a local copy that can be permissionlessly updated by anyone with the real value.
+    uint256 private _cachedProtocolSwapFeePercentage;
+
+    event CachedProtocolSwapFeePercentageUpdated(uint256 protocolSwapFeePercentage);
+
+    constructor(IVault vault, uint256 protocolSwapFeePercentage) {
+        // Set initial value of the protocolSwapFeePercentage; can be updated externally if it is delegated
+        bool delegatedProtocolFees = protocolSwapFeePercentage == DELEGATE_PROTOCOL_FEES_SENTINEL;
```
nventuro, line 40:
```suggestion
        // Protocol fees are delegated to the value reported by the Fee Collector if the sentinel value is passed.
        bool delegatedProtocolFees = protocolSwapFeePercentage == DELEGATE_PROTOCOL_FEES_SENTINEL;
```

pkg/pool-utils/contracts/ProtocolFeeCache.sol
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * @dev Getter for the protocol swap fee percentage
+     */
+    function getCachedProtocolSwapFeePercentage() public view returns (uint256) {
+        return _cachedProtocolSwapFeePercentage;
+    }
+
+    /**
+     * @dev Returns whether the pool pays protocol fees.
```
nventuro, line 78:
```suggestion
     * @dev Returns whether this Pool tracks protocol fee changes in the Fee Collector.
```

pkg/pool-utils/contracts/ProtocolFeeCache.sol
```diff
@@ -0,0 +1,91 @@
...
+     * Updates the cache to the latest value set by governance.
+     */
+    function updateCachedProtocolSwapFeePercentage() external {
+        if (getProtocolFeeDelegation()) {
+            _updateCachedProtocolSwapFee(_vault);
+        }
+    }
+
+    /**
+     * @dev Getter for the protocol swap fee percentage
```
nventuro, line 71:
```suggestion
     * @dev Returns the current protocol swap fee percentage. If `getProtocolFeeDelegation()` is false, this value is
     * immutable. Alternatively, it will track the global fee percentage set in the Fee Collector.
```

nventuro (commented):
Looks great! `StablePhantomPool` also uses cached protocol fees, could you give updating that contract to use this a try?

We should also add tests: I'd imagine the easiest way to do it is to copy `PhantomStablePool`'s tests (which are regardless quite simple as this is not a complex feature).

pkg/pool-utils/contracts/ProtocolFeeCache.sol
```diff
@@ -0,0 +1,91 @@
...
+            emit CachedProtocolSwapFeePercentageUpdated(protocolSwapFeePercentage);
+        }
+    }
+
+    /**
+     * @dev Can be called by anyone to update the cache fee percentage (when delegated).
+     * Updates the cache to the latest value set by governance.
+     */
+    function updateCachedProtocolSwapFeePercentage() external {
+        if (getProtocolFeeDelegation()) {
```
nventuro, line 65: This should revert if `delegatedProtocolFees` is false.

nventuro:
Once this is merged, we could look into having a separate `immutable` slot to store non-delegated fees, as there is no point in reading from storage in that case. Something like:

```
fn getProtocolFees() {
  if (delegated) {
    return _immutable:
  } else {
    return _storage;
  }
}
```

We can do that in a separate PR however.

pkg/pool-utils/contracts/ProtocolFeeCache.sol
EndymionJkb: I made it public so that factories (e.g., PhantomStablePool) could initialize it just using
ProtocolFeeCache.DELEGATE_PROTOCOL_FEES_SENTINEL()

Which is clearer than hard-coding a value.

nventuro: Ah, fair point!

pkg/pool-weighted/test/ManagedPool.test.ts
```diff
@@ -682,12 +682,53 @@ describe('ManagedPool', function () {
...
 
         // Real Vault will return a zero protocol fee
-        expectEvent.inReceipt(await receipt.wait(), 'ProtocolSwapFeeCacheUpdated', {
+        expectEvent.inReceipt(await receipt.wait(), 'CachedProtocolSwapFeePercentageUpdated', {
           protocolSwapFeePercentage: 0,
         });
       });
     });
 
+    describe('Non-delegated protocol fees', () => {
```
nventuro, line 691:
```suggestion
    describe('non-delegated protocol fees', () => {
```

nventuro: Also, these tests should be in a separate file, and we should remove the duplcate tests that StablePhantomPool has.

pkg/pool-utils/contracts/ProtocolFeeCache.sol
```diff
@@ -0,0 +1,98 @@
...
+        if (delegatedProtocolFees) {
+            _updateCachedProtocolSwapFee(vault);
+        } else {
+            _require(
+                protocolSwapFeePercentage <= _MAX_PROTOCOL_SWAP_FEE_PERCENTAGE,
+                Errors.SWAP_FEE_PERCENTAGE_TOO_HIGH
+            );
+
+            // Set the fixed protocol fee percentage, which can be zero
+            _cachedProtocolSwapFeePercentage = protocolSwapFeePercentage;
```
TomAFrench, line 58: Seems like we're doing an SSTORE for nothing on deployment here. We only read `_cachedProtocolSwapFeePercentage` in `getCachedProtocolSwapFeePercentage` and don't go down the relevant code path when we're not delegating the protocol fee percentage.

EndymionJkb: Yes; this was left over from before I added the immutable storage

TomAFrench:
Ah, I renamed some of the functions/events to be `xProtocolFeeCache` rather than `xCachedProtocolFee` as this will match our other caches but just realised this will affect the ABIs for StablePhantomPools which is a bit annoying.

nventuro:
I'm fine with dropping ABI compat on those events, those are internal things for advanced users - no third party, UI, etc. will be concerned with the internal cache of the fee being updated.

EndymionJkb:
> I'm fine with dropping ABI compat on those events, those are internal things for advanced users - no third party, UI, etc. will be concerned with the internal cache of the fee being updated.

For context, that's why the names were what they were. Since we already had this on stable phantom pools, I took the naming conventions from there for the least disruption.

pkg/pool-utils/test/ProtocolFeeCache.test.ts
```diff
@@ -0,0 +1,97 @@
...
+        deploy('MockProtocolFeeCache', { args: [vault.address, MAX_PROTOCOL_FEE.add(1)] })
+      ).to.be.revertedWith('SWAP_FEE_PERCENTAGE_TOO_HIGH');
+    });
+  });
+
+  context('with delegated fee', () => {
+    sharedBeforeEach('deploy delegated fee cache', async () => {
+      await vault.setSwapFeePercentage(VAULT_PROTOCOL_FEE, { from: admin });
+
+      protocolFeeCache = await deploy('MockProtocolFeeCache', { args: [vault.address, MAX_UINT256] });
```
nventuro, line 42:
```suggestion
      // The sentinel value used to designate delegated fees is MAX_UINT256
      protocolFeeCache = await deploy('MockProtocolFeeCache', { args: [vault.address, MAX_UINT256] });
```

nventuro (approved):

pkg/pool-utils/contracts/ProtocolFeeCache.sol
```diff
@@ -0,0 +1,103 @@
...
+        // As `_fixedProtocolSwapFeePercentage` is immutable we must set a value, but just set to zero if it's not used.
+        _fixedProtocolSwapFeePercentage = delegatedProtocolFees ? 0 : protocolSwapFeePercentage;
+    }
+
+    /**
+     * @dev Can be called by anyone to update the cache fee percentage (when delegated).
+     * Updates the cache to the latest value set by governance.
+     */
+    function updateProtocolSwapFeePercentageCache() external {
+        _require(getProtocolFeeDelegation(), Errors.UNAUTHORIZED_OPERATION);
```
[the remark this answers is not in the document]
EndymionJkb: Yes, was being a bit lazy here, trying to reuse an error. Will add.

nventuro:
Looks great! We can merge once CI passes.

PR #1214, opened 2022-03-31
Managed Pool: Aum protocol fees

EndymionJkb:
Introduce a new protocol level fee for managed pool AUM.

Current version deploys an AumProtocolFeesCollector contract that manages the amount of the fee, and allows it to be changed by governance.

The BaseManagedPoolFactory deploys this (as a singleton), and sets the collector for each pool created by the factory.

When AUM fees are collected, the pool gets the current fee percentage from the collector, and sends any amounts due the protocol to the main Vault ProtocolFeesCollector


pkg/standalone-utils/contracts/AumProtocolFeesCollector.sol
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * @dev This an auxiliary contract to the Vault, deployed by it during construction. It offloads some of the tasks the
+ * Vault performs to reduce its overall bytecode size.
+ *
+ * The current values for all protocol fee percentages are stored here, and any tokens charged as protocol fees are
+ * sent to this contract, where they may be withdrawn by authorized entities. All authorization tasks are delegated
+ * to the Vault's own authorizer.
+ */
+contract AumProtocolFeesCollector is IAumProtocolFeesCollector, Authentication, ReentrancyGuard {
```
TomAFrench, line 33: I'd be inclined to have this be a simple contract which only returns the protocol AUM fee percentage and rely on the standard protocol fee collector to custody the funds.

EndymionJkb: I still wanted it to be controlled by governance, but the consensus seems to be send all funds to the Vault's fee collector, so this version does that. (See the updated comment.)

pkg/pool-utils/contracts/AumProtocolFeeCache.sol
```diff
@@ -0,0 +1,66 @@
...
+     * @dev Can be called by anyone to update the cache AUM fee percentage.
+     * Updates the cache to the latest value set by governance.
+     */
+    function updateProtocolAumFeePercentageCache() external {
+        _updateProtocolAumFeeCache(_aumProtocolFeesCollector);
+    }
+
+    function _updateProtocolAumFeeCache(IAumProtocolFeesCollector protocolFeeCollector) private {
+        uint256 currentProtocolAumFeePercentage = protocolFeeCollector.getAumFeePercentage();
+
```
nventuro, line 61:
```suggestion
    function _updateProtocolAumFeeCache(IAumProtocolFeesCollector aumProtocolFeeCollector) private {
        uint256 currentProtocolAumFeePercentage = aumProtocolFeeCollector.getAumFeePercentage();

```

pkg/pool-utils/test/AumProtocolFeeCache.test.ts
```diff
@@ -0,0 +1,155 @@
...
+        deploy('MockAumProtocolFeeCache', {
+          args: [vault.address, MAX_PROTOCOL_FEE.add(1), aumProtocolFeesCollector.address],
+        })
+      ).to.be.revertedWith('SWAP_FEE_PERCENTAGE_TOO_HIGH');
+    });
+
+    it('reverts when trying to update fixed fee', async () => {
+      await expect(protocolFeeCache.updateProtocolSwapFeePercentageCache()).to.be.revertedWith('INVALID_OPERATION');
+    });
+  });
```
nventuro, line 118: I don't think we need to repeat all of the protocol swap fee cache tests here. If you really want to do it, they should be abstracted away into a behavior and reused that way.

pkg/pool-weighted/contracts/smart/BaseManagedPoolFactory.sol
```diff
@@ -30,8 +31,10 @@ import "./ManagedPool.sol";
  * deploy the pool, passing in the controller as the owner.
  */
 contract BaseManagedPoolFactory is BasePoolSplitCodeFactory, FactoryWidePauseWindow {
+    AumProtocolFeesCollector public immutable aumProtocolFeesCollector;
```
nventuro, line 34: I'd expose this in the typical `getValue()` getter style we use.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -164,6 +165,7 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
         uint256 protocolSwapFeePercentage;
         uint256 managementSwapFeePercentage;
         uint256 managementAumFeePercentage;
+        IAumProtocolFeesCollector aumProtocolFeesCollector; // Should be a constructor parameter, but blew up the stack
```
nventuro, line 168:
```suggestion
        IAumProtocolFeesCollector aumProtocolFeesCollector;
```

nventuro: I found this comment misleading. If anything we could have a comment above the struct saying 'the pool receives so many arguments that it runs into stack-too-deep errors, so they are passed in a memory struct instead to avoid those' or something like that.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -1029,6 +1031,13 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+            // Compute the protocol's share of the AUM fee
+            uint256 protocolBptAmount = bptAmount.mulUp(getProtocolAumFeePercentageCache());
+            bptAmount = bptAmount.sub(protocolBptAmount);
+
+            _payProtocolFees(protocolBptAmount);
+
+            // Only show the amount collected by the manager
             emit ManagementAumFeeCollected(bptAmount);
 
             _mintPoolTokens(getOwner(), bptAmount);
```
nventuro, line 1043:
```suggestion
            uint256 managerBPTAmount = bptAmount.sub(protocolBptAmount);

            _payProtocolFees(protocolBptAmount);

            // Only show the amount collected by the manager
            emit ManagementAumFeeCollected(managerBPTAmount);

            _mintPoolTokens(getOwner(), managerBPTAmount);
```

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -1029,6 +1031,13 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
             uint256 fractionalTimePeriod = elapsedTime.divDown(365 days);
             bptAmount = annualizedFee.mulDown(fractionalTimePeriod);
 
+            // Compute the protocol's share of the AUM fee
+            uint256 protocolBptAmount = bptAmount.mulUp(getProtocolAumFeePercentageCache());
+            bptAmount = bptAmount.sub(protocolBptAmount);
+
+            _payProtocolFees(protocolBptAmount);
+
+            // Only show the amount collected by the manager
```
nventuro, line 1040: Why?

pkg/pool-weighted/test/BaseManagedPoolFactory.test.ts
```diff
@@ -34,16 +36,23 @@ describe('BaseManagedPoolFactory', function () {
...
 
     tokens = await TokenList.create(['MKR', 'DAI', 'SNX', 'BAT'], { sorted: true });
+
+    createTime = await currentTimestamp();
+
+    protocolFeesCollector = factory.aumProtocolFeesCollector();
+  });
+
+  it('factory has the AumProtocolFeesController', async () => {
+    expect(protocolFeesCollector).to.not.equal(ZERO_ADDRESS);
```
nventuro, line 55: This test is not doing what you think it is. We're not `await`ing the getter, so the `feeCollector` is an unresolved promise.

Also, this should be called protocolAUMFeeCollector or something - the feeCollector is a different contract.

pkg/pool-weighted/test/BaseManagedPoolFactory.test.ts
```diff
@@ -144,8 +154,11 @@ describe('BaseManagedPoolFactory', function () {
       const pool = await createPool();
       const { pauseWindowEndTime, bufferPeriodEndTime } = await pool.getPausedState();
 
-      expect(pauseWindowEndTime).to.equal(createTime.add(BASE_PAUSE_WINDOW_DURATION));
-      expect(bufferPeriodEndTime).to.equal(createTime.add(BASE_PAUSE_WINDOW_DURATION + BASE_BUFFER_PERIOD_DURATION));
+      expect(pauseWindowEndTime).to.equalWithError(createTime.add(BASE_PAUSE_WINDOW_DURATION), 0.00001);
+      expect(bufferPeriodEndTime).to.equalWithError(
+        createTime.add(BASE_PAUSE_WINDOW_DURATION + BASE_BUFFER_PERIOD_DURATION),
+        0.00001
+      );
```
nventuro, line 161: Why this change? The actual values should exactly equal the expected ones.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
EndymionJkb: I'd like to group the "factory" parameters (vault, owner, and the temporarily pausable parameters) into a separate struct to keep NewPoolParams pure, but just updating the comment now in the interest of time.

pkg/pool-utils/test/AumProtocolFeeCache.test.ts
EndymionJkb: Just removing for now, then. There are no other tests in the base class, so the base class test would just be one line calling the behavior.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
EndymionJkb: Why show only the manager amount? The event is called ManagementAumFeeCollected. Or do you mean something else?

pkg/pool-weighted/test/BaseManagedPoolFactory.test.ts
EndymionJkb: It seems like the extra deployment (of the AumProtocolFeesCollector) makes it take extra time

nventuro: But we don't measure time that way, we use the block's timestamp. The issues was that the time of measurement was moved here:

PR #1216, opened 2022-04-04
Deploy ERC4626 Linear Pool Factory V2

rabmarut:
This version of the ERC4626LinearPoolFactory reflects the changes introduced in #1177 after finalization of [EIP-4626](https://eips.ethereum.org/EIPS/eip-4626).

This version will be deployed to all networks, whereas an initial deployment was introduced only to Polygon in #1155. That initial factory on Polygon should be considered deprecated for all but one specific use case.


pkg/deployments/tasks/20220304-erc4626-linear-pool/readme.md
```diff
@@ -1,6 +1,6 @@
-# 2022-03-04 - ERC4626 Linear Pool
+# 2022-03-04 - ERC4626 Linear Pool V1
 
-Deployment of the `ERC4626LinearPoolFactory`, for Linear Pools with a ERC4626 wrapped yield bearing token
+Preliminary deployment of the `ERC4626LinearPoolFactory`, for Linear Pools with a ERC4626 yield-bearing token. This version was released prior to finalization of EIP-4626 and should no longer be used.
```
nventuro, line 3: Let's add a link to the replacement deployment.

nventuro (approved):
I've checked that the build info matches that of commit 18d1d22a3ccf521b0a61da6f746dcc9646484aa5. 

I also verified the Arbitrum code, which was unverified, and checked that the code deployed to all networks is the same (and since the Arbitrum verification passes, that also indirectly checks that it matches the code in the repo).

nventuro:
I updated this by deprecating the prior deployment, adding both to the readme, doing the optimism deployment, and adding new infra pieces such as action ids and deployment txs.

PR #1218, opened 2022-04-05
Prevent late lockers from having to claim many weeks of zero-value distributions

TomAFrench:
We currently have a concept of a `tokenStartTime` so that the contract knows from when users should attempt to claim a particular token if they haven't interacted with it before. This means that if we add a new token to be distributed then people don't have to make zero-value claims all the way from `_startTime` in order to sync the contract up to the current week.

However the issue is we need the same concept for users (`userStartTime`), otherwise if someone were to lock veBAL in 2 years time and want to claim BAL tokens then they would have to go through all previous BAL distributions whilst we know that they won't receive any fees from before their first lock.

We then save the timestamp of the first week which the user would be eligible for a distribution and when first claiming a token we can fast forward to this week.

### DIFF
--- a/pkg/liquidity-mining/contracts/fee-distribution/FeeDistributor.sol
+++ b/pkg/liquidity-mining/contracts/fee-distribution/FeeDistributor.sol
@@ -16,6 +16,7 @@ pragma solidity ^0.7.0;
 pragma experimental ABIEncoderV2;
 
 import "@balancer-labs/v2-solidity-utils/contracts/helpers/IAuthentication.sol";
+import "@balancer-labs/v2-solidity-utils/contracts/math/Math.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/ReentrancyGuard.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/SafeERC20.sol";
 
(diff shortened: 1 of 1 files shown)


pkg/liquidity-mining/contracts/fee-distribution/FeeDistributor.sol
```diff
@@ -450,7 +452,12 @@ contract FeeDistributor is IFeeDistributor, ReentrancyGuard {
      */
     function _getUserTokenTimeCursor(address user, IERC20 token) internal view returns (uint256) {
         uint256 userTimeCursor = _userTokenTimeCursor[user][token];
-        return userTimeCursor > 0 ? userTimeCursor : _tokenStartTime[token];
+        if (userTimeCursor > 0) return userTimeCursor;
+        // This is the first time that the user has interacted with this token.
+        // We then start from the latest out of either when `user` first locked veBAL or `token` was first checkpointed.
+        uint256 userStartTime = _userStartTime[user];
+        uint256 tokenStartTime = _tokenStartTime[token];
+        return userStartTime > tokenStartTime ? userStartTime : tokenStartTime;
```
nventuro, line 460:
```suggestion
        return Math.max(_userStartTime[user], _tokenStartTime[token]);
```

(you'd need to also import Math)

pkg/liquidity-mining/contracts/fee-distribution/FeeDistributor.sol
```diff
@@ -450,7 +452,12 @@ contract FeeDistributor is IFeeDistributor, ReentrancyGuard {
      */
     function _getUserTokenTimeCursor(address user, IERC20 token) internal view returns (uint256) {
         uint256 userTimeCursor = _userTokenTimeCursor[user][token];
-        return userTimeCursor > 0 ? userTimeCursor : _tokenStartTime[token];
+        if (userTimeCursor > 0) return userTimeCursor;
```
nventuro, line 455: Not quite related to this PR, but I find the existence of this function a bit strange as it is only used in `_claimToken`. All other functions direcfly access the mapping, including `getUserTokenTimeCursor` (!).

nventuro (approved):

PR #1220, opened 2022-04-06
add DistributionScheduler contract

TomAFrench:
This contract aims to allow partners to schedule distributions of their tokens on gauges without needing to fire off a transaction at the same time every week, instead they can queue up multiple weeks worth of rewards and leave the triggering of the distribution to bots like Defender.

Anyone may provide tokens to the `DistributionScheduler` along with a timestamp at which it is to be sent to the gauge, this is then stored in a linked list of all the other distributions of this token to a particular gauge. We can then just iterate through this list up to the current timestamp in order to start any distributions which are scheduled to do so.

We require that distributions start at Thurs 00:00 UTC for consistency with BAL rewards and also to prevent DOS from having many 1 wei distributions which each need to be processed before the next meaningful distribution.

### DIFF
--- a/pkg/liquidity-mining/contracts/admin/DistributionScheduler.sol
+++ b/pkg/liquidity-mining/contracts/admin/DistributionScheduler.sol
@@ -19,16 +19,17 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/SafeERC20.sol";
 
 import "../interfaces/IStakingLiquidityGauge.sol";
 
+// solhint-disable not-rely-on-time
+
 /**
  * @title DistributionScheduler
  * @notice Scheduler for setting up permissionless distributions of liquidity gauge reward tokens.
- * @dev Any address may send tokens to the DistributionSchedule to be distributed among gauge depositors 
-
+ * @dev Any address may send tokens to the DistributionSchedule to be distributed among gauge depositors.
  */
 contract DistributionScheduler {
     using SafeERC20 for IERC20;
 
-    uint256 private constant MAX_REWARDS = 8;
+    uint256 private constant _MAX_REWARDS = 8;
 
     uint32 private constant _HEAD = 0;
     uint32 private constant _NULL = 0;
@@ -109,7 +110,7 @@ contract DistributionScheduler {
      * @param gauge - The gauge which is to distribute the reward token.
      */
     function startDistributions(IStakingLiquidityGauge gauge) external {
-        for (uint256 i = 0; i < MAX_REWARDS; ++i) {
+        for (uint256 i = 0; i < _MAX_REWARDS; ++i) {
             IERC20 token = gauge.reward_tokens(i);
             if (token == IERC20(0)) break;
 
@@ -198,8 +199,8 @@ contract DistributionScheduler {
             require(rewardAmount <= type(uint224).max, "Reward amount overflow");
             rewardsList[nextNodeKey].amount = uint224(rewardAmount);
         } else {
-            // We're inserting a node in between `currentNodeKey` and `nextNodeKey`.
-            // We then update `currentNodeKey` to point to the newly inserted node and the new node point to `nextNodeKey`.
+            // We're inserting a node in between `currentNodeKey` and `nextNodeKey` so then update
+            // `currentNodeKey` to point to the newly inserted node and the new node to point to `nextNodeKey`.
             rewardsList[insertedNodeKey] = RewardNode(amount, nextNodeKey);
             rewardsList[currentNodeKey].nextNodeKey = insertedNodeKey;
         }
--- a/pkg/liquidity-mining/contracts/admin/DistributionScheduler.sol
+++ b/pkg/liquidity-mining/contracts/admin/DistributionScheduler.sol
@@ -17,7 +17,7 @@ pragma experimental ABIEncoderV2;
 
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/SafeERC20.sol";
 
-import "../interfaces/IStakingLiquidityGauge.sol";
+import "../interfaces/IRewardTokenDistributor.sol";
 
 // solhint-disable not-rely-on-time
 
@@ -52,7 +52,7 @@ contract DistributionScheduler {
      *         - the timestamp of the next scheduled distribution of `token` to `gauge`. Zero if no distribution exists.
      */
     function getRewardNode(
-        IStakingLiquidityGauge gauge,
+        IRewardTokenDistributor gauge,
         IERC20 token,
         uint256 timestamp
     ) external view returns (RewardNode memory) {
@@ -64,7 +64,7 @@ contract DistributionScheduler {
      * @param gauge - The gauge which is to distribute the reward token.
      * @param token - The token which is to be distributed among gauge depositors.
      */
-    function getPendingRewards(IStakingLiquidityGauge gauge, IERC20 token) external view returns (uint256) {
+    function getPendingRewards(IRewardTokenDistributor gauge, IERC20 token) external view returns (uint256) {
         mapping(uint32 => RewardNode) storage rewardsList = _rewardsLists[_getRewardsListId(gauge, token)];
 
         (, uint256 amount) = _getPendingRewards(rewardsList, block.timestamp);
@@ -82,7 +82,7 @@ contract DistributionScheduler {
      * @param startTime - The timestamp at the beginning of the week over which to distribute tokens.
      */
     function scheduleDistribution(
-        IStakingLiquidityGauge gauge,
+        IRewardTokenDistributor gauge,
         IERC20 token,
         uint256 amount,
         uint256 startTime
@@ -111,7 +111,7 @@ contract DistributionScheduler {
      * @notice Process all pending distributions for a gauge to start distributing the tokens.
      * @param gauge - The gauge which is to distribute the reward token.
      */
-    function startDistributions(IStakingLiquidityGauge gauge) external {
+    function startDistributions(IRewardTokenDistributor gauge) external {
         for (uint256 i = 0; i < _MAX_REWARDS; ++i) {
             IERC20 token = gauge.reward_tokens(i);
             if (token == IERC20(0)) break;
@@ -129,7 +129,7 @@ contract DistributionScheduler {
      * @param gauge - The gauge which is to distribute the reward token.
      * @param token - The token which is to be distributed among gauge depositors.
      */
-    function startDistributionForToken(IStakingLiquidityGauge gauge, IERC20 token) public {
+    function startDistributionForToken(IRewardTokenDistributor gauge, IERC20 token) public {
         mapping(uint32 => RewardNode) storage rewardsList = _rewardsLists[_getRewardsListId(gauge, token)];
 
         (uint32 firstUnprocessedNodeKey, uint256 rewardAmount) = _getPendingRewards(rewardsList, block.timestamp);
@@ -144,7 +144,7 @@ contract DistributionScheduler {
 
     // Internal functions
 
-    function _getRewardsListId(IStakingLiquidityGauge gauge, IERC20 rewardToken) internal pure returns (bytes32) {
+    function _getRewardsListId(IRewardTokenDistributor gauge, IERC20 rewardToken) internal pure returns (bytes32) {
         return keccak256(abi.encodePacked(gauge, rewardToken));
     }
 
--- a/pkg/liquidity-mining/contracts/admin/DistributionScheduler.sol
+++ b/pkg/liquidity-mining/contracts/admin/DistributionScheduler.sol
@@ -1,0 +1,214 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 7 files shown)


pkg/liquidity-mining/contracts/test/MockRewardTokenDistributor.sol
```diff
@@ -0,0 +1,72 @@
...
+pragma experimental ABIEncoderV2;
+
+import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/IERC20.sol";
+
+// solhint-disable func-name-mixedcase, var-name-mixedcase, not-rely-on-time
+
+/**
+ * @dev This contract is designed to mock LiquidityGaugeV5's interface for distributing external tokens.
+ */
+contract MockRewardTokenDistributor {
```
nventuro, line 25: Let's also inherit from the interface, given that's what we're mocking and interacting with in the tests.

pkg/liquidity-mining/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,394 @@
...
+    });
+
+    context('when DistributionScheduler is the distributor for reward token', () => {
+      sharedBeforeEach('set reward token distributor to DistributionScheduler', async () => {
+        await rewardTokenDistributor.add_reward(rewardToken.address, distributionScheduler.address);
+      });
+
+      context('when distribution is scheduled in the past', () => {
+        it('reverts', async () => {
+          await expect(scheduleDistribution(1, 0)).to.be.revertedWith(
```
nventuro, line 111:
```suggestion
          await expect(scheduleDistribution(1, (await currentTimestamp()).sub(1))).to.be.revertedWith(
```

pkg/liquidity-mining/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,394 @@
...
+
+    context("when distribution starts at a timestamp which doesn't fit in a uint32", () => {
+      it('reverts', async () => {
+        await expect(scheduleDistribution(1, MAX_UINT32.add(1))).to.be.revertedWith('Reward timestamp overflow');
+      });
+    });
+
+    context('when reward token does not exist on gauge', () => {
+      it('reverts', async () => {
+        await expect(scheduleDistribution(1, 0)).to.be.revertedWith('Reward token does not exist on gauge');
```
nventuro, line 88: I'd consider changing these hardcoded 1's to e.g. `const amount = 42`. This is specially useful in later tests.

pkg/liquidity-mining/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,394 @@
...
+
+            context('when a reward is being added after the last existing reward', () => {
+              let insertedTime: BigNumber;
+
+              sharedBeforeEach('set insertedTime', async () => {
+                insertedTime = startTime.add(20 * WEEK);
+              });
+
+              it('updates the previous node to point at the new node', async () => {
+                const prevNodeKey = insertedTime.sub(2 * WEEK);
```
nventuro, line 175: This is a bit magical as it relies on how nodes were constructed in the `for` loop. I'd consider replacing this with a search to find the last node key.

pkg/liquidity-mining/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,394 @@
...
+              for (let i = 0; i < 10; i++) {
+                await scheduleDistribution(100, startTime.add(i * 2 * WEEK));
+              }
+            });
+
+            context('when a reward is being added after the last existing reward', () => {
+              let insertedTime: BigNumber;
+
+              sharedBeforeEach('set insertedTime', async () => {
+                insertedTime = startTime.add(20 * WEEK);
```
nventuro, line 171:
```suggestion
                insertedTime = startTime.add(20 * WEEK);
```

Same here, we're _right_ at the edge of constructed values. I'd consider doing something like `9999 * WEEK`.

pkg/liquidity-mining/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,394 @@
...
+
+              it('writes the expected node', async () => {
+                const oldRewardNode = await getRewardNode(insertedTime);
+                expect(oldRewardNode.amount).to.be.eq(0);
+
+                await scheduleDistribution(1, insertedTime);
+
+                const newRewardNode = await getRewardNode(insertedTime);
+                expect(newRewardNode.amount).to.be.eq(1);
+                expect(newRewardNode.nextNodeKey).to.be.eq(insertedTime.add(WEEK));
```
nventuro, line 219:
```suggestion
                expect(newRewardNode.nextNodeKey).to.be.eq(oldRewardNode.nextNodeKey);
```

pkg/liquidity-mining/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,394 @@
...
+            });
+
+            it('writes the expected node', async () => {
+              const oldRewardNode = await getRewardNode(startTime);
+              expect(oldRewardNode.amount).to.be.eq(0);
+
+              await scheduleDistribution(1, startTime);
+
+              const newRewardNode = await getRewardNode(startTime);
+              expect(newRewardNode.amount).to.be.eq(1);
```
nventuro, line 156:
```suggestion
              expect(newRewardNode.amount).to.be.eq(1);
              expect(newRewardNode.nextNodeKey)to.be.eq(NULL);
```

pkg/liquidity-mining/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,394 @@
...
+              });
+
+              it('writes the expected node', async () => {
+                const oldRewardNode = await getRewardNode(insertedTime);
+                expect(oldRewardNode.amount).to.be.eq(100);
+
+                await scheduleDistribution(1, insertedTime);
+
+                const newRewardNode = await getRewardNode(insertedTime);
+                expect(newRewardNode.amount).to.be.eq(101);
```
nventuro, line 246: If we go with `amount` I'd rewrite this to e.g. `oldRewardNode.amount + 1` (or even `+ newAmount`), which is a bit more readable.

pkg/liquidity-mining/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,394 @@
...
+                expect(newRewardNode.nextNodeKey).to.be.eq(oldRewardNode.nextNodeKey);
+              });
+
+              context('when providing more tokens would cause an overflow on the node', () => {
```
nventuro, line 250: I wrote a note on my notebook to check if this test case existed while reviewing the code. Well done :grin:

pkg/liquidity-mining/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,394 @@
...
+                it('reverts', async () => {
+                  const { amount: existingRewardAmount } = await getRewardNode(insertedTime);
+                  const rewardAmount = MAX_UINT224.sub(1);
```
nventuro, line 253:
```suggestion
                  const rewardAmount = MAX_UINT224.sub(existingRewardAmount).add(1);
```

pkg/liquidity-mining/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,394 @@
...
+      return tx.wait();
+    }
+
+    sharedBeforeEach('set reward token distributor to DistributionScheduler', async () => {
+      await rewardTokenDistributor.add_reward(rewardToken.address, distributionScheduler.address);
+      startTime = roundUpTimestamp(await currentTimestamp());
+      for (let i = 0; i < 10; i++) {
+        await scheduleDistribution(100, startTime.add(i * WEEK));
+      }
+      await advanceToTimestamp(startTime.add(1));
```
nventuro, line 287: I think we don't need this `add(1)`?

pkg/liquidity-mining/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,394 @@
...
+
+        await startDistributionForToken();
+
+        expect(await getNextNodeKey(HEAD)).to.be.eq(startTime.add(WEEK));
+      });
+    });
+
+    context('when the last scheduled distribution is processed', () => {
+      sharedBeforeEach('advance time past the last scheduled distribution', async () => {
+        await advanceToTimestamp(startTime.add(10 * WEEK));
```
nventuro, line 328: Same comment re. scheduled construction in the for loop and this magical `10`

pkg/liquidity-mining/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,394 @@
...
+
+    sharedBeforeEach('deploy Token', async () => {
+      rewardToken2 = await Token.create('REWARD2');
+      await rewardToken2.mint(caller);
+      await rewardToken2.approve(distributionScheduler, MAX_UINT256, { from: caller });
+    });
+
+    sharedBeforeEach('initialize scheduled distributions', async () => {
+      await rewardTokenDistributor.add_reward(rewardToken.address, distributionScheduler.address);
+      await rewardTokenDistributor.add_reward(rewardToken2.address, distributionScheduler.address);
```
nventuro, line 353: Let's add a third reward with another distributor and check that it is correctly skipped

pkg/liquidity-mining/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,394 @@
...
+      const tx = await distributionScheduler.startDistributions(rewardTokenDistributor.address);
+      const receipt = await tx.wait();
+
+      expectEvent.inIndirectReceipt(
+        receipt,
+        rewardToken.instance.interface,
+        'Transfer',
+        {
+          from: distributionScheduler.address,
+          to: rewardTokenDistributor.address,
```
nventuro, line 377: We could also add `amount` here, populated with `getPendingRewards`

nventuro (commented):
Lovely :grin: You've earned comp-sci fun permission!

pkg/liquidity-mining/contracts/admin/DistributionScheduler.sol
```diff
@@ -0,0 +1,215 @@
...
+        uint256 startTime
+    ) external {
+        require(amount > 0, "Must provide non-zero number of tokens");
+
+        // Ensure that values won't overflow when put into storage.
+        require(amount <= type(uint224).max, "Reward amount overflow");
+        require(startTime <= type(uint32).max, "Reward timestamp overflow");
+
+        // Ensure that a user doesn't add a reward token which becomes locked on scheduler
+        address rewardDistributor = gauge.reward_data(token).distributor;
```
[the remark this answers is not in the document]
TomAFrench: I'm not a huge fan of this as I think it's overly restrictive. We'll need to special case any gauges which receive rewards but don't exist on the gauge controller.

We've had situations recently where partners (Gnosis in this case) want to get rewards started the same day and we obviously can't add gauges to the controller on that sort of turnaround. They then need to do hacky workarounds passing the distributor role to various addresses in order to get the first rewards out and then get it set up properly once the gauge is on the controller.

tl;dr I see more potential for friction on the ops side from adding this than risk from someone yeeting tokens into the void not checking the address they're sending it to.

pkg/liquidity-mining/contracts/admin/DistributionScheduler.sol
```diff
@@ -0,0 +1,219 @@
...
+     */
+    function _insertReward(
+        mapping(uint32 => RewardNode) storage rewardsList,
+        uint32 insertedNodeKey,
+        uint224 amount
+    ) private {
+        // We want to find two nodes which sit either side of the new node to be created so we can insert between them.
+
+        uint32 currentNodeKey = _HEAD;
+        uint32 nextNodeKey = rewardsList[currentNodeKey].nextTimestamp;
```
nventuro, line 185: Are these intentionally not renamed?

pkg/liquidity-mining/test/DistributionScheduler.test.ts
```diff
@@ -0,0 +1,424 @@
...
+          await expect(scheduleDistribution(amount, (await currentTimestamp()).sub(1))).to.be.revertedWith(
+            'Distribution can only be scheduled for the future'
+          );
+        });
+      });
+
+      context('when distribution is scheduled in the future', () => {
+        let startTime: BigNumber;
+
+        sharedBeforeEach('schedule some existing distributions', async () => {
```
nventuro, line 122:
```suggestion
        sharedBeforeEach('set the first valid timestamp to schedule rewards', async () => {
```

pkg/liquidity-mining/contracts/admin/DistributionScheduler.sol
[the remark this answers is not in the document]
nventuro: That's a very good point, thank.

In such a scenario, the LM committee would still need to manually add the reward tokens though, right?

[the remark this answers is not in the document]
TomAFrench: Yeah, they'd need to add the token and set the scheduler as the distributor but that would be one tx rather than several.

nventuro (approved):

PR #1224, opened 2022-04-08
Authorizer: Model executor

### LINKED ISSUE #1199: Try moving the scheduled executions outside the authorizer itself
Simply to avoid having the authorizer executing calls, it would be nice to explore how convenient it could be moving the execution logic to external contracts. These should simple executors with basic logic, we could use factories to create them for every call if necessary.

facuspagnuolo:
Fixes #1199

### DIFF
--- a/pkg/vault/contracts/authorizer/Executor.sol
+++ b/pkg/vault/contracts/authorizer/Executor.sol
@@ -1,0 +1,32 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+
+import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/Address.sol";
+
+import "./TimelockAuthorizer.sol";
+
+contract Executor {
+    TimelockAuthorizer public immutable timelock;
+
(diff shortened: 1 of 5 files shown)


pkg/vault/contracts/authorizer/Executor.sol
```diff
@@ -0,0 +1,32 @@
...
+import "./TimelockAuthorizer.sol";
+
+contract Executor {
+    TimelockAuthorizer public immutable timelock;
+
+    constructor() {
+        timelock = TimelockAuthorizer(msg.sender);
+    }
+
+    function execute(address target, bytes memory data) external returns (bytes memory result) {
```
nventuro, line 28: Do you think there's any reason to make this `nonReentrant`? If we do add it to be doubly sure of correctness, would we be impeding any reasonable use cases?

pkg/vault/contracts/authorizer/Executor.sol
```diff
@@ -0,0 +1,32 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+
+import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/Address.sol";
+
+import "./TimelockAuthorizer.sol";
+
+contract Executor {
```
nventuro, line 21: I'd rename this to something like TimelockExecutor, or ScheduledActionExecutor (I quite like this second one)

nventuro (approved):

pkg/vault/contracts/authorizer/Executor.sol
facuspagnuolo: I named it exactly `TimelockExecutor` initially but I ended up changing it to simply `Executor` since there is nothing related to timelocks in this contract. Let me know your thoughts.

nventuro: There is a lot about timelocks: only the authorizer can call this, which it does because of timelocks :stuck_out_tongue:

pkg/vault/contracts/authorizer/Executor.sol
```diff
@@ -0,0 +1,32 @@
...
+    TimelockAuthorizer public immutable timelock;
```
nventuro, line 22: I'd rename this to `authorizer`.

PR #1225, opened 2022-04-08
Authorizer: Improve granter/revoker interface

facuspagnuolo:
### MERGE AFTER #1224

### DIFF
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -228,28 +228,25 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
     }
 
     /**
-     * @dev Tells whether `account` can perform action `actionId` in `where` with `how`
+     * @dev Tells whether `account` can grant permissions for action `actionId` in `where`
      */
-    function canPerformOrWhatever(
+    function canGrant(
         bytes32 actionId,
         address account,
-        address where,
-        bytes32 how
+        address where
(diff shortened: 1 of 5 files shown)


nventuro (approved):
_Nice_, it's amazing how much this simple improvement makes things more readable and easier to understand!

facuspagnuolo:
Glad you liked it @nventuro! I have just pushed another minor enhacement

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -343,12 +351,8 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
...
         require(block.timestamp >= scheduledExecution.executableAt, "ACTION_NOT_EXECUTABLE");
         if (scheduledExecution.protected) {
-            bool isAllowed = hasPermissionOrWhatever(
-                EXECUTE_ACTION_ID,
-                msg.sender,
-                address(this),
-                bytes32(scheduledExecutionId)
-            );
+            bytes32 executeScheduledActionId = getActionId(EXECUTE_ACTION_ID, bytes32(scheduledExecutionId));
+            bool isAllowed = hasPermission(executeScheduledActionId, msg.sender, address(this));
```
nventuro, line 355: This is not the same as the old version, is it? With this change it is not possible to have global execution rights

nventuro: (I don't think we needed those, just pointing that out)

PR #1228, opened 2022-04-11
Miscellaneous FeeDistributor improvements

TomAFrench:
It'll be best to go through this PR commit by commit as they're all fairly isolated.

The commit messages should be fairly explanatory.


nventuro:
The first part of 118da2089e63726dec44945e73684bad90f5f557 did nothing (the initial assingment, both the original code and the edit), so I just removed that. I was not convinced of correctness of the second part, and it is just a simple comparison on a stack value, so I reverted that change.

nventuro (approved):

PR #1231, opened 2022-04-13
Add deployment task for L2 gauges

TomAFrench:


nventuro (approved):
I've checked that the addresses of BAL in both Polygon and Arbitrum are correct. The ChildChainGaugeLiquidityFactory build info corresponds to building commit `740bc174b5876eac5c7c922c985bee6dec5adf4e`. The build info for RewardOnlyGauge and ChildChainStreamer had to be verified manually as those are hand-built Vyper build infos, but they also match the artifacts at said commit.

PR #1232, opened 2022-04-13
Add PolygonRootGaugeFactory deployment

nventuro:
As mentioned in the readme, depending on how the vote goes this deployment might require updating the `BalancerMinter` input. The build info, abi, other input arguments, etc. however can all be verified already.


nventuro:
(this is also pending fork tests)

nventuro:
I added relatively comprehensive fork tests, in which: a) a gauge is created, b) the factory is added to the gauge adder, c) the gauge is then added to the gauge controller via the gauge adder, d) the gauge is voted on by a veBAL holder, e) a week passes and the gauge mints nothing due to the one week delay, f) another week passes and the gauge successfully mints and bridges.

To test the bridge behavior, I performed the same steps manually on the bridge. I first [approved BAL](https://etherscan.io/tx/0x8d09254acae138deb65d6d01e21e284a78801e178ce880f9923071e3bef801ba) on the predicate, and then [deposited](https://etherscan.io/tx/0x345a9de3d920bc35dcd0bf84f56837f826b66cb1b5872b0edf8eba05948a7dd3) on the gateway, setting the recipient to a different address as the sender. The funds arrived on Polygon [in this transaction](https://polygonscan.com/tx/0x6bfe724cdc3abc3aee9ac4d69c083e9490a8ec7c0e352a4f2a21bd33cab0e68c).

The actual deposit triggered a [`LockedERC20` event](https://etherscan.io/tx/0x345a9de3d920bc35dcd0bf84f56837f826b66cb1b5872b0edf8eba05948a7dd3#eventlog), with the arguments being depositor, recipient, token and amount. These are the same values I tested for in the fork test.

pkg/deployments/tasks/20220413-polygon-root-gauge-factory/test/task.fork.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    const expectedEmissions = gaugeRelativeWeight.mul(weeklyRate).div(FP_SCALING_FACTOR);
+    expectEqualWithError(actualEmissions, expectedEmissions, 0.001);
+
+    // Tokens are minted for the gauge
+    expectEvent.inIndirectReceipt(await mintTx.wait(), BAL.interface, 'Transfer', {
+      from: ZERO_ADDRESS,
+      to: gauge.address,
+      value: actualEmissions,
+    });
```
TomAFrench, line 168: We have so many of these littered about we should really wrap it up into an `expectTransferEvent` globally at some point.

pkg/deployments/tasks/20220413-polygon-root-gauge-factory/test/task.fork.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    const calldata = gauge.interface.encodeFunctionData('checkpoint');
+
+    // Even though the gauge has relative weight, it cannot mint yet as it needs for the epoch to finish
+    const zeroMintTx = await authorizerAdaptor.connect(admin).performAction(gauge.address, calldata);
+    expectEvent.inIndirectReceipt(await zeroMintTx.wait(), gauge.interface, 'Checkpoint', {
+      periodTime: firstMintWeekTimestamp.sub(WEEK), // Process past week, which had zero votes
+      periodEmissions: 0,
+    });
+    // No token transfers are performed if the emissions are zero, but we can't test for a lack of those
```
TomAFrench, line 146:
```suggestion
    expectEvent.notEmitted(receipt, "Transfer");
```

pvt/helpers/src/time.ts
```diff
@@ -8,6 +8,11 @@ export const currentTimestamp = async (): Promise<BigNumber> => {
   return bn(timestamp);
 };
 
+export const currentWeekTimestamp = async (): Promise<BigNumber> => {
+  const { timestamp } = await network.provider.send('eth_getBlockByNumber', ['latest', true]);
+  return bn(timestamp).div(WEEK).mul(WEEK);
```
TomAFrench, line 13: Similarly with this, we're going to have to round up and down timestamps a reasonable amount.

TomAFrench (approved):
I've built commit 8f408a2c and can confirm that the buildinfo and bytecode match what is contained in this PR.

pvt/helpers/src/time.ts
TomAFrench: This is also going to conflict with a similar change in #1234

pkg/deployments/tasks/20220413-polygon-root-gauge-factory/test/task.fork.ts
nventuro: This won't work because the contract we call (the authorizer) is not the one who emits the event: we need the equivalent of `inIndirectReceipt` but for `notEmitted`.

nventuro: (I mean, the assertion will pass, but it won't do the right thing)

pkg/deployments/tasks/20220413-polygon-root-gauge-factory/test/task.fork.ts
nventuro: We do use `expectTokenTransfer` extensively, but our forking infrastructure and associated helpers could really use some help, yes (as can be seen in the huge task loading boilerplate).

PR #1234, opened 2022-04-14
 Add ArbitrumRootGaugeFactory deployment

nventuro:
This PR is extremely similar to https://github.com/balancer-labs/balancer-v2-monorepo/pull/1232, except for Arbitrum.

One important difference is that the fork test uncovered a bug in the bridge: we were incorrectly encoding the `bytes` array sent as the last argument to the GatewayRouter. This can be observed (in a somewhat cryptic manner) in the [original Curve code, on line 200](https://etherscan.io/address/0x9044E12fB1732f88ed0c93cfa5E9bB9bD2990cE5#code). This second `bytes` argument is not used, but the `abi.decode` call on the router expects it, as can be observed [in a debugger on this failed transaction](https://dashboard.tenderly.co/tx/mainnet/0x096486d241dd716e908df2e306393456b2cb0139da7b694cdf34a6cb83d6490a?trace=0.2.0.0).

I introduced the fix in commit add32dc44b3edc3441f5dac34546e16198d2fe2c, which is the commit that should be used to verify the build info.

---

To test the bridge behavior, I performed the same steps manually on the bridge. I first [approved BAL](https://etherscan.io/tx/0xe8509d4846babd91160d5c115e937aae4312c47666cc8e1e21d756e0a40b970e) on the gateway, and then [deposited](https://etherscan.io/tx/0x83fc20bdea8114eec9d10ab8d4b05397b331798d77d7f52fe3040187ad5f8cce) on the gateway router, setting the recipient to a different address as the sender. The funds arrived on Arbitrum [in this transaction](https://arbiscan.io/tx/0x6c417742d9162685d8dd60b5f0ce1871a3d8d4d9dd14530111035436f5821c93).

The actual deposit triggered a [DepositInitiated event](https://etherscan.io/tx/0x83fc20bdea8114eec9d10ab8d4b05397b331798d77d7f52fe3040187ad5f8cce#eventlog), with the arguments being depositor, recipient, token and amount. These are the same values I tested for in the fork test.


nventuro:
PR #1234 :raised_hands:

pkg/deployments/tasks/20220413-arbitrum-root-gauge-factory/test/task.fork.ts
```diff
@@ -0,0 +1,185 @@
...
+    // Even though the gauge has relative weight, it cannot mint yet as it needs for the epoch to finish
+    const bridgeETH = await gauge.getTotalBridgeCost();
+    const zeroMintTx = await authorizerAdaptor
+      .connect(admin)
+      .performAction(gauge.address, calldata, { value: bridgeETH });
+    expectEvent.inIndirectReceipt(await zeroMintTx.wait(), gauge.interface, 'Checkpoint', {
+      periodTime: firstMintWeekTimestamp.sub(WEEK), // Process past week, which had zero votes
+      periodEmissions: 0,
+    });
+    // No token transfers are performed if the emissions are zero, but we can't test for a lack of those
```
TomAFrench, line 149: https://github.com/balancer-labs/balancer-v2-monorepo/pull/1232#discussion_r850389518

pkg/liquidity-mining/contracts/gauges/arbitrum/ArbitrumRootGauge.sol
```diff
@@ -66,7 +66,7 @@ contract ArbitrumRootGauge is ISingleRecipientGauge, StakelessGauge {
             mintAmount,
             gasLimit,
             gasPrice,
-            abi.encode(maxSubmissionCost)
+            abi.encode(maxSubmissionCost, "")
```
TomAFrench, line 69: Ugh, this is annoying.

PR #1235, opened 2022-04-14
veBAL L2 gauges coodinator

TomAFrench:


nventuro:
I introduced some fixes, set a temporary checkpointer multisig (the one that has permission to pause boosted pools), and added a bunch of fork tests. The only things I'm not thoroughly testing in the fork tests are the creation of the gauges, which seems like it shouldn't require much testing.

nventuro:
I removed equal gauge type setting and handled that in #1237 to avoid issues with that one running first, and this one then undoing that.

pkg/governance-scripts/contracts/20220415-veBAL-L2/veBALL2GaugeSetupCoordinator.sol
```diff
@@ -0,0 +1,307 @@
...
+        authorizer.grantRole(checkpointGaugeRole, GAUGE_CHECKPOINTER_MULTISIG);
+    }
+
+    function _addNewEthereumGauges() private {
+        // All these addresses are required to be mainnet gauges which have already been deployed
+        address payable[3] memory newGauges = [
+            0xa57453737849A4029325dfAb3F6034656644E104, // 80HAUS-20WETH
+            0xA6468eca7633246Dcb24E5599681767D27d1F978, // 50COW-50GNO
+            0x158772F59Fe0d3b75805fC11139b46CBc89F70e5 // 50COW-50WETH
+            // 80NOTE-20WETH may be eligible to be added here as well soon
```
TomAFrench, line 165: Do we need to insert this address here? Not sure what votes have completed.

TomAFrench:
I've checked and the build-info committed matches what I get by building [15e6884](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1235/commits/15e68843362ca2cc5637deb6445dee4d4d60e8e7)

PR #1238, opened 2022-04-15
Authorizer: Code Cleanup

rabmarut:


pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -272,15 +281,15 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
...
     function setRoot(address newRoot) external onlyExecutor {
         root = newRoot;
         emit RootSet(newRoot);
     }
 
     /**
-     * @dev Schedules a root change call
+     * @dev Schedules an execution to change the root address to `newRoot`.
      */
     function scheduleRootChange(address newRoot, address[] memory executors)
```
rabmarut, line 294: It's not clear to me why `SCHEDULE_DELAY_ACTION_ID` is included here - I thought it was only applicable to `scheduleDelayChange`.

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -383,8 +392,8 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
     }
 
     /**
-     * @dev Manages `account`'s granter condition for action `actionId` in `where`.
-     * Note that pairs can manage themselves, even banning the root, but the root can allow himself back at any time
```
rabmarut, line 387: I think this comment hints at the undesirable functionality we uncovered. The "pair" management piece seems to imply that a granter has authority to manage all identical granter permissions.

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
nventuro: The first argument of `scheduleWithDelay` is only emitted in an event - it is otherwise not used. It simply infroms that a delay change has been scheduled for `this.setRoot`. But it does look a bit wonky.

nventuro:
Fixing linter isses and merging as these cosmetical changes are useful by themselves.

PR #1239, opened 2022-04-15
Authorizer: Fix actionId emitted for scheduleRootChange

rabmarut:
The action ID emitted by `scheduleRootChange` is deceptive and misleading. It hashes the true action ID (`setRoot`) with `SCHEDULE_DELAY_ACTION_ID`, which is the action ID for `scheduleDelayChange`. This second ID should not be used outside of its corresponding function and seems to be arbitrarily included here as a result of nomenclature confusion.


nventuro (approved):
Well spotted! This change makes the behavior consistent with usage of `schedule`.

PR #1240, opened 2022-04-18
Run fork tests on CI

nventuro:
Adding a cache layer, only testing modified directories, etc.


nventuro:
(marked as ready to review to have CI run)

TomAFrench:
Everything here should be good to go except a number of the tests rely on a specific fork block. e.g `veBALDeploymentCoordinator` tests require the network to be forked before its activation block but the rest of the veBAL fork tests require the network to be forked later.

nventuro:
There's a couple more hidden dependencies in that some tests impersonate accounts, and those accounts have moved their tokens elsewhere, making the tests fail if forking from too recent a time.

nventuro:
Superceded by https://github.com/balancer-labs/balancer-v2-monorepo/pull/1501

PR #1246, opened 2022-04-20
Smart wallet coordinator deployment

nventuro:


TomAFrench:
I checked a966703 and the outputs match what I get from compiling the commit

PR #1253, opened 2022-04-22
Managed Pools: gradual swap fee

EndymionJkb:
Implements a gradual swap fee with an interface equivalent to the gradual weight update. The controller should impose limits on this functionality (e.g., a maximum swap fee, minimum/fixed update duration).

For instance, the controller might calculate the initial high swap fee (e.g., 80%), set it to that value, then initiate a gradual swap fee change down to the original fee - or a new value.


pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -221,6 +235,52 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+        // Load current pool state from storage
+        bytes32 poolState = _getMiscData();
+
+        startTime = poolState.decodeUint32(_FEE_START_TIME_OFFSET);
+        endTime = poolState.decodeUint32(_FEE_END_TIME_OFFSET);
+        endSwapFeePercentage = _endSwapFeePercentage;
+    }
+
+    function _setSwapFeePercentage(uint256 swapFeePercentage) internal virtual override {
+        // Do not allow setting if there is an ongoing fee change
```
TomAFrench, line 266: Should we just overwrite the ongoing gradual weight change? A pool manager can currently just do `updateSwapFeeGradually(block.timestamp, block.timestamp, newSwapFeePercentage)` to instantly set the swap fee percentage to a new value.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -223,6 +237,62 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+     * @dev Return start time, end time, and endWeights as an array.
+     * Current weights should be retrieved via `getNormalizedWeights()`.
+     */
+    function getGradualFeeUpdateParams()
+        external
+        view
+        returns (
+            uint256 startTime,
+            uint256 endTime,
+            uint256 endSwapFeePercentage
```
TomAFrench, line 266: I think we should include the `startSwapFeePercentage` here as otherwise you can't calculate the current swap fee from the return value from this function.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
EndymionJkb: Well the intent here was to keep the interface and events consistent. setSwapFeePercentage is in the interface from BasePool, so if they call it, it should emit the normal event instead of the gradual change one. V1 also worked like this - to minimize the interactions (and test cases, and attack surface), it only let you do one thing at a time.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
EndymionJkb: Yes; sorry - was getting late :)

pkg/pool-weighted/contracts/smart/ManagedPool.sol
TomAFrench: Fair enough. Events are kinda tricky on this contract as you need to track both the instantaneous and gradual change events to get the current value of the swap fee.

I was just thinking on whether we can save the two error code slots but this seems reasonable.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -223,6 +237,64 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+                super.getSwapFeePercentage(),
+                _endSwapFeePercentage,
+                startTime,
+                endTime
+            );
+    }
+
+    /**
+     * @dev Return start time, end time, and endWeights as an array.
+     * Current weights should be retrieved via `getNormalizedWeights()`.
```
TomAFrench, line 258:
```suggestion
     * Current swap fee percentage should be retrieved via `getSwapFeePercentage()`.
```

TomAFrench, line 258:
```suggestion
     * @dev Return start time, end time, initial swap fee and final swap fee.
     * Current swap fee percentage should be retrieved via `getSwapFeePercentage()`.
```

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -278,6 +350,23 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+        uint256 startTime,
+        uint256 endTime,
+        uint256 endSwapFeePercentage
+    ) external authenticate whenNotPaused nonReentrant {
+        _require(endSwapFeePercentage >= _getMinSwapFeePercentage(), Errors.MIN_SWAP_FEE_PERCENTAGE);
+        _require(endSwapFeePercentage <= _getMaxSwapFeePercentage(), Errors.MAX_SWAP_FEE_PERCENTAGE);
+
+        startTime = GradualValueChange.resolveStartTime(startTime, endTime);
+
+        _startGradualSwapFeeChange(startTime, endTime, super.getSwapFeePercentage(), endSwapFeePercentage);
```
TomAFrench, line 367: I've got some question marks on the usage of `super.getSwapFeePercentage()` here. It seems like we'll end up reading the start swap fee from `_miscData` rather than using the current swap fee. This'll work fine to begin with but we'll always revert back to the swap fee set in the last call to `_setSwapFeePercentage` or deployment.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -278,6 +350,23 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
 
+    /**
+     * @dev Schedule a gradual swap fee change, from the current value to the given ending fee percentage,
+     * over startTime to endTime.
+     */
+    function updateSwapFeeGradually(
+        uint256 startTime,
+        uint256 endTime,
+        uint256 endSwapFeePercentage
+    ) external authenticate whenNotPaused nonReentrant {
```
TomAFrench, line 361: Do we want to give the option to set a `startSwapFeePercentage` here? It didn't make sense to do this for weights as you "never" want to change it instantaneously but it seems like it would be useful here.

Main example being where we're restarting trading and want to bump the swap fee up to 100% and let it decay back to its initial value, we currently need to make 2 calls to do this.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -617,12 +706,42 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+        uint256 startTime,
+        uint256 endTime,
+        uint256 startSwapFeePercentage,
+        uint256 endSwapFeePercentage
+    ) private {
+        _setMiscData(
+            _getMiscData()
+                .insertUint32(startTime, _FEE_START_TIME_OFFSET)
+                .insertUint32(endTime, _FEE_END_TIME_OFFSET)
+                .insertUint64(startSwapFeePercentage, _SWAP_FEE_PERCENTAGE_OFFSET)
```
TomAFrench, line 739: Related to https://github.com/balancer-labs/balancer-v2-monorepo/pull/1253/files#r857521915

This last line is a noop. `BasePool._setMiscData` explicitly clears out these top bits to stop us from changing the swap fee percentage so even if we fix the related issue then we'll get bad swap fee changes.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
TomAFrench: Fixed in [d103215](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1253/commits/d10321572a79df371f18b57382a88d3d5620894d)

pkg/pool-weighted/contracts/smart/ManagedPool.sol
TomAFrench: I've added a failing testcase in [1f3a723](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1253/commits/1f3a72373b382a68ba2b51761e1a31c176931d43)

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -265,7 +326,9 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
     }
 
     function _getTotalTokens() internal view virtual override returns (uint256) {
-        return _getMiscData().decodeUint7(_TOTAL_TOKENS_OFFSET);
+        (IERC20[] memory tokens, , ) = getVault().getPoolTokens(getPoolId());
```
TomAFrench, line 329: My first instinct would be to cache this in storage on the pool contract. We're going to be performing a lot of SLOADs to get the number of tokens here.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -75,14 +75,17 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
...
+    uint256 private constant _WEIGHT_START_TIME_OFFSET = 0;
+    uint256 private constant _WEIGHT_END_TIME_OFFSET = 32;
+    uint256 private constant _END_SWAP_FEE_PERCENTAGE_OFFSET = 64;
+    uint256 private constant _FEE_START_TIME_OFFSET = 128;
+    uint256 private constant _SWAP_ENABLED_OFFSET = 159;
+    uint256 private constant _FEE_END_TIME_OFFSET = 160;
+    uint256 private constant _MUST_ALLOWLIST_LPS_OFFSET = 191;
+    uint256 private constant _SWAP_FEE_PERCENTAGE_OFFSET = 192;
 
     // 7 bits is enough for the token count, since _MAX_MANAGED_TOKENS is 50
```
TomAFrench, line 90:
```suggestion
```

TomAFrench (approved):

PR #1256, opened 2022-04-22
Normalise weights after interpolation instead of before

TomAFrench:
By normalising after we interpolate the weight of a token we save a warm SLOAD per weight calculated. This saves 2 SLOADs on swaps and `numTokens` on weight updates. As a bonus the code looks prettier.

My only concern with this is if `_denormWeightSum` is small then we're compressing the range of values we're interpolating over and so we're more likely to see rounding errors. My thoughts on this however is that:

- You'd have to work pretty hard to reduce `_denormWeightSum` in such a way that this was significant based on how many decimal places we have to work with and that's assuming that it's even possible.
- Before these rounding errors became significant you'd likely be changing your weights by so much over such a small period of time that you'd be arbed hugely anyway.

**That said this change is hugely insignificant in the grand scheme of things so feel free to kick this can down the road.**

### DIFF
--- a/pkg/pool-weighted/contracts/smart/ManagedPool.sol
+++ b/pkg/pool-weighted/contracts/smart/ManagedPool.sol
@@ -370,18 +370,14 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
 
     function _getNormalizedWeight(IERC20 token) internal view override returns (uint256) {
         bytes32 tokenData = _getTokenData(token);
-        uint256 startWeight = _normalizeWeight(
-            tokenData.decodeUint64(_START_DENORM_WEIGHT_OFFSET).uncompress64(_MAX_DENORM_WEIGHT)
-        );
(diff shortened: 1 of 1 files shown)


EndymionJkb (commented):
It was only like this to preserve the interface to the original normalizeWeights function - now that everything's been generalized, it's clear we can do this optimization.

To me the gas savings at high token counts (one of the major features of Managed Pools) outweigh possible slight rounding errors.

I tested some pretty small weight sums (like 0.00001), without a problem the old way. As long as it works as least as well, this should be fine.

TomAFrench:
Yep, I just ran the `GradualWeightChange` tests after manually denormalising the "weights" to have a really small `_denormWeightSum` and everything worked fine.

EndymionJkb:
> Yep, I just ran the `GradualWeightChange` tests after manually denormalising the "weights" to have a really small `_denormWeightSum` and everything worked fine.

I also tested it with 50 tokens at a weight sum of 0.000000001; the error was < 0.01%, so I think we're fine.

PR #1257, opened 2022-04-23
Add Distribution Scheduler deployment

nventuro:
I fixed a discrepance between the actual gauge and our Solidity interface, plus added an utility query function.


TomAFrench:
Confirming that I get the same build-info, bytecode and ABI when compiling [718baa3](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1257/commits/718baa383d021422e72bea82512c7ead437e183e)

TomAFrench (approved):

PR #1258, opened 2022-04-25
Managed Pool: library cleanup

EndymionJkb:
Somehow had leftover "weight" references in the generalized value change library


TomAFrench (approved):
Thanks! I swear I changed these so maybe got lost in a merge conflict?

PR #1260, opened 2022-04-25
Add tests for GradualValueChange with zero duration

TomAFrench:
This PR adds tests for the GradualValueChange library to ensure that it correctly interpolates values when there is a zero duration change.


TomAFrench:
Merging as the failing tests are unrelated.

PR #1265, opened 2022-04-25
Deploy: UnbuttonAaveLinearPoolFactory

rabmarut:
Deploy the work from #1152, which introduces a twist on Aave linear pools where both the main and wrapped tokens are [UnbuttonTokens](https://github.com/buttonwood-protocol/button-wrappers/blob/main/contracts/UnbuttonToken.sol). An UnbuttonToken wraps a rebasing token and abstracts away the balance changes.

This is immediately useful for creating a wAMPL-waAMPL linear pool, but it is not limited to AMPL as an underlying token.


nventuro (approved):
I verified that the build info corresponds to commit e23d937ddfb3202f0c2b7c4363dd1fa166f12507.

I also did the optimism deployment, updated readmes, and added deployment txs.

PR #1266, opened 2022-04-25
Deploy: BatchRelayer V2 to mainnet and arbitrum

rabmarut:
The BatchRelayer V2 was previously deployed only to Polygon in #1182, pending a potential V3 deployment to other networks. Given changing priorities and increasing delays in the V3 deployment, this PR pushes V2 to the remaining networks for use with both ERC-4626 and Unbutton linear pools.

This is still a draft because I'm having difficulty deploying to arbitrum at the moment.


nventuro:
I did the arbitrum deployment and added the deployment tx hashes for both mainnet and arbitrum, so that deployment verification would succeed.

rabmarut:
Thanks for pushing forward, looks like this got pretty stale. Everything looks good to me but it's my PR so I can't approve it.

nventuro (approved):
With CI passing we can now merge.

PR #1269, opened 2022-04-25
Avoid SLOAD when updating weights

TomAFrench:
The optimisooooor


EndymionJkb (commented):
I added a `_totalTokensCache` in the pool contract (See [comment](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1253#discussion_r858067655))

The only place it's used here is `updateWeightsGradually`, but will be used a lot in add/remove.

EndymionJkb (approved):

nventuro:
>I added a _totalTokensCache in the pool contract (See https://github.com/balancer-labs/balancer-v2-monorepo/pull/1253#discussion_r858067655)

I am fairly certain we don't need to store this at all - we'll likely do a pass later removing it from all places where it shows up

PR #1270, opened 2022-04-26
doc: minor ManagedPool cleanup

EndymionJkb:


pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -515,9 +515,8 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
         }
     }
 
-    // We override _onJoinPool and _onExitPool as we need to not compute the current invariant and calculate protocol
-    // fees, since that mechanism does not work for Pools in which the weights change over time. Instead, this Pool
-    // always pays zero protocol fees.
+    // We override _onJoinPool and _onExitPool as we do not need to compute the current invariant and calculate protocol
+    // fees, since that mechanism does not work for Pools where the weights change over time.
```
TomAFrench, line 519:
```suggestion
    // We override _onJoinPool and _onExitPool as we do not need to compute the current invariant and calculate protocol
    // fees, since that mechanism does not work for Pools where the weights change over time. We instead calculate protocol fees on each swap.
```

EndymionJkb: Pushed up separately for lint

PR #1271, opened 2022-04-26
fix: fix pool-utils CI to allow it to deploy oversized ManagedPool

EndymionJkb:
The risk of getting used to broken CI... we don't always notice when other things are broken


TomAFrench:
The tests still seem to be failing (and don't seem to be related to the pool being too large to deploy)

EndymionJkb:
> The tests still seem to be failing (and don't seem to be related to the pool being too large to deploy)

This is for additional failures on pool-utils; the Oracle ones are still there.

PR #1272, opened 2022-04-26
Optimise checks for if a gradual change is scheduled/ongoing

TomAFrench:
This PR changes the checks for any ongoing gradual changes to first check the end time and only if we fail that determine which condition we're failing on.

This saves us from decoding the start time and performing a comparison against the current time if we know it's going to succeed (as we know we're past the end time.)


nventuro:
Well spotted!

PR #1273, opened 2022-04-27
Refactor _encodeTokenState to allow passing a custom denormalisation factor

TomAFrench:
Currently we perform a lot of warm SLOADs whenever we perform a weight update as we're reading `_denormWeightSum` twice per token. This isnt ideal as we can easily get >50 reads for large pools.

Another factor feeding into this is when we are adding a new token we need to write the denormalised weight of the new token into storage using the value of `_denormWeightSum` **after** the token is added rather than before.

### DIFF
--- a/pkg/pool-weighted/contracts/smart/ManagedPool.sol
+++ b/pkg/pool-weighted/contracts/smart/ManagedPool.sol
@@ -857,13 +857,14 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
     ) internal virtual {
         uint256 normalizedSum;
 
+        uint256 denormWeightSum = _denormWeightSum;
         for (uint256 i = 0; i < endWeights.length; i++) {
             uint256 endWeight = endWeights[i];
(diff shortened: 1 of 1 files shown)


pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -1046,4 +1051,9 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
     function _denormalizeWeight(uint256 weight) private view returns (uint256) {
         return weight.mulUp(_denormWeightSum);
     }
+
+    // converts from normalized form to the internal representation (summing to customDenormWeightSum)
+    function _denormalizeWeight(uint256 weight, uint256 customDenormWeightSum) private pure returns (uint256) {
+        return weight.mulUp(customDenormWeightSum);
+    }
```
TomAFrench, line 1058: This may be overkill and we can inline this if you think readability wouldn't be impacted too much.

nventuro (approved):

PR #1275, opened 2022-04-28
Use cached denormWeightSum for all normalization operations

TomAFrench:
Follows on from #1273 to pass in `_denormWeightSum` on all calls to `_normalizeWeight` and `_denormalizeWeight`.

In most calls to these functions we're iterating over all tokens so we might as well optimise for this case and remove the unary forms of these functions.

### DIFF
--- a/pkg/pool-weighted/contracts/smart/ManagedPool.sol
+++ b/pkg/pool-weighted/contracts/smart/ManagedPool.sol
@@ -330,9 +330,11 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
 
         endWeights = new uint256[](totalTokens);
 
+        uint256 denormWeightSum = _denormWeightSum;
         for (uint256 i = 0; i < totalTokens; i++) {
             endWeights[i] = _normalizeWeight(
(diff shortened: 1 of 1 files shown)


EndymionJkb (approved):
It looks fine, but the runtime performance would be the same, right? Is this just saving bytecode by having fewer functions? (It only saves 7 bytes 25404 -> 25397; so maybe the compiler is already doing some optimization there.)

I like renaming the argument: "custom" sounds like you might sometimes use one different from _denormWeightSum.

Having these functions is good for clarity/self-documentation (vs inlining), and having 2 functions instead of 4 is less cluttered

TomAFrench:
> It looks fine, but the runtime performance would be the same, right? Is this just saving bytecode by having fewer functions? (It only saves 7 bytes 25404 -> 25397; so maybe the compiler is already doing some optimization there.)

It's not the same for `_normalizedWeight(uint256)` performs a warm SLOAD for `_denormWeightSum` on each call but we can change that to just one and pass it in to `_normalizedWeight(uint256, uint256)` each time. This'll save `100 * numTokens` gas for pool joins/exits.

> I like renaming the argument: "custom" sounds like you might sometimes use one different from _denormWeightSum.

Well sometime's we do :) This is pretty helpful when adding tokens.

PR #1280, opened 2022-05-04
Managed Pool: separate max swap fee

EndymionJkb:
Setting the swap fee to 100% causes some join/exit swaps to revert. Separate the maximum swap fee from the maximum *management* swap fee (which can be 100%).

It can also revert with much lower values under some circumstances; further changes may follow.


pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -69,6 +69,9 @@ contract ManagedPool is BaseWeightedPool, AumProtocolFeeCache, ReentrancyGuard {
     // creation gas consumption.
     uint256 private constant _MAX_MANAGED_TOKENS = 38;
 
+    // Must be < ONE for correct behavior of join/exit swaps
```
TomAFrench, line 72:
```suggestion
    // Swap fees close to 100% can cause reverts when performing join/exit swaps. 80% is a safe value while still high.
```

TomAFrench (approved):
I'd clarify what we mean by "correct behaviour" in the comment but otherwise this seems fine.

PR #1284, opened 2022-05-05
Check deployment outputs

nventuro:
With #1283, tasks now have different modes. This PR adds a new mode `CHECK`, which is used to verify that the addresses stored in the `outputs` directories correspond to the build infos plus inputs. This is a work in progress, as it requires interacting with a Tenderly endpoint that is not yet public.

The CHECK mode compares the data of the output contract creation transaction (which Tenderly reports) against the creation code + arguments that the task attempts to deploy. On success, it returns the previously deployed instance to support tasks that deploy multiple contracts, call functions on them, use their addresses, etc.

We will likely have to add a list of known exceptions for e.g. the relayer entrypoints, which are automatically deployed by relayer libraries.

### DIFF
--- a/pkg/deployments/ci/prepare-config.ts
+++ b/pkg/deployments/ci/prepare-config.ts
@@ -12,7 +12,7 @@ if (process.env.CI) {
       },
       "polygon": {
         "url": "${process.env.INFURA_POLYGON_NONARCHIVE_ENDPOINT}"
-      }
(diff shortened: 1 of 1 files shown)


TomAFrench:
@nventuro Can you push those extra commits from yesterday?

TomAFrench (approved):

PR #1285, opened 2022-05-06
Interface Refactor

EndymionJkb:
Create a balancer-interfaces package with all interfaces (and closely related/referenced libraries), and make everything dependent on it, vs. the current tight coupling that leads to circular dependencies and hardhat issues.

See this [dependency diagram](https://drive.google.com/file/d/19hH_wLF6PJXVeykgT9XmvCtSLmenrlYl/view?usp=sharing). (Almost everything also depends on solidity-utils; most of those links removed for readability.)


pkg/solidity-utils/contracts/helpers/SingletonAuthentication.sol
```diff
@@ -0,0 +1,62 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+
+import "@balancer-labs/v2-interfaces/contracts/liquidity-mining/IAuthorizerAdaptor.sol";
+import "@balancer-labs/v2-interfaces/contracts/vault/IVault.sol";
+
+import "./Authentication.sol";
+
+abstract contract SingletonAuthentication is Authentication, IAuthorizerAdaptor {
```
nventuro, line 22: Woops :stuck_out_tongue:

nventuro (commented):
Thanks a lot @EndymionJkb! Seems like some SingletonAuthentication files got added here - once those are removed, we can merge.

pkg/solidity-utils/contracts/helpers/SingletonAuthentication.sol
EndymionJkb: Looks like just the file itself and the Mock (not any usage of it). Yeah, the risk of giant commits with hundreds of files, and switching branches back and forth :)

PR #1286, opened 2022-05-06
change 'factory' to 'NoProtocolFeeLiquidityBootstrappingPoolFactory'

gerrrg:
done for consistency w/ other outputs


nventuro:
Thanks @gerrrg! 

Can you also update the task source (`index.ts`) so that it doesn't force that name? It likely is outdated and implements a lot of logic, while it should be able to be simplified greatly (look at the index for the Vault deployment for an example).

TomAFrench (approved):
Looks good. I'll fix any linting issues after the merge to save gerg having to do it.

PR #1287, opened 2022-05-06
Reduce managed pool tests

nventuro:
This PR alters some basic managed pool tests and makes them fatster by:
 - removing a test I deemeed unnecessary (the weight sum, which is an internal detail)
 - improving transparency on how we test weights
 - removing unnecessary tests for all token combinations (multiple tokens are handled with for loops - we don't need to test all combinations like we do for weighted pool)


EndymionJkb (approved):
The weightSum argument was left over from before a previous merge, when I had to edit and test it manually. Reducing the number of runs also makes it significantly faster.

PR #1292, opened 2022-05-06
Fix flaky ManagedPool tests

nventuro:
This addresses RNG issues, as well as incomplete intializations. With these, all ManagedPool tests pass.


EndymionJkb (approved):
Ok; looks like the occasional zeroes from plain "random()" were messing with the deployer's normalization. Will remember that!

PR #1293, opened 2022-05-06
Fix failing weighted pool tests

nventuro:
:upside_down_face:


nventuro:
Advancing so far into the future pushes us past the [Year 2038 problem](https://www.wikiwand.com/en/Year_2038_problem) start date (we're closer to 2038 than to 1970), which causes our 32 bit time measurements to not suffice.

nventuro:
_grumble_ looks like there's one more managed pool test to conquer

EndymionJkb (approved):

PR #1295, opened 2022-05-10
Add ChildChainGaugeTokenAdder contract

### LINKED ISSUE #1291: Build GaugeRewardTokenAdder for L2s
Adding reward tokens to the L2 gauges is slightly involved, as we need to not only update the streamers but also the gauges themselves, which must be kept in sync. 

We should check the following:
 - that the gauge comes from the factory
 - that the gauge's streamer comes from the factory
 - that the gauge is the streamer's recipient

We'll want to have multiple tests on this mechanism. The Curve contracts seem contradictory in this regard: the gauge attempts to checkpoint to 'verify that claims are working', but this doesn't actually happen: the CLAIM_FREQUENCY prevents two successive claims from transferring tokens :shrug:

TomAFrench:
fixes #1291

### DIFF
--- a/pkg/liquidity-mining/contracts/admin/ChildChainGaugeTokenAdder.sol
+++ b/pkg/liquidity-mining/contracts/admin/ChildChainGaugeTokenAdder.sol
@@ -18,8 +18,7 @@ pragma experimental ABIEncoderV2;
 import "@balancer-labs/v2-solidity-utils/contracts/helpers/Authentication.sol";
 
 import "../interfaces/IAuthorizerAdaptor.sol";
-import "../interfaces/IChildChainStreamer.sol";
-import "../interfaces/IRewardsOnlyGauge.sol";
+import "../interfaces/IChildChainLiquidityGaugeFactory.sol";
 
 /**
  * @title ChildChainGaugeTokenAdder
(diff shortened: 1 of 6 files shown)


pkg/liquidity-mining/contracts/admin/ChildChainGaugeTokenAdder.sol
```diff
@@ -0,0 +1,122 @@
...
+        IRewardsOnlyGauge gauge,
+        IERC20 rewardToken,
+        address distributor
+    ) external authenticate {
+        require(_gaugeFactory.isGaugeFromFactory(address(gauge)), "Invalid gauge");
+        IChildChainStreamer streamer = IChildChainStreamer(_gaugeFactory.getGaugeStreamer(address(gauge)));
+        require(streamer == gauge.reward_contract(), "Not original gauge streamer");
+
+        // We first add the new token to the streamer so that the gauge can claim it when checkpointing.
+        _addTokenToStreamer(streamer, rewardToken, distributor);
```
nventuro, line 83: This actually doesn't happen: the gauge checkpoints before and after adding the token, but the second checkpoint does nothing because of the 1 hour maximum frequency.

nventuro (approved):

pkg/liquidity-mining/contracts/admin/ChildChainGaugeTokenAdder.sol
TomAFrench: I'm meaning this in the ongoing sense here rather than for the next claim specifically.

PR #1296, opened 2022-05-11
Add ChildChainDistributionScheduler

TomAFrench:
Very quick and dirty migration of the DistributionScheduler to handle L2 gauges.

~~We'll need a couple of tests for recovering invalid tokens but will add those to this PR later on.~~

The main need for this new contract is that the L2 gauges handle rewards differently compared to gauges on L1. Instead of calling a function on the gauge which then pulls tokens in from the scheduler, we must actively push them to the streamer contract and then update it to begin pushing those rewards to the gauge.

Another important difference is the fact that it's possible for a reward token to be removed from a gauge (by removing it from the gauge's streamer). This gives the opportunity for funds to be locked if there are distributions pending when the token is removed from the streamer (by calling `remove_reward`) while they are sitting on the scheduler. As we don't track ownership of funds inside the scheduler we must have a authenticated escape hatch which allows Balancer governance to retrieve these funds.


TomAFrench:
We should add the check from #1459 here as well

jubeira:
Closing for now.

We're moving forward with different child chain gauges. 
A distribution scheduler might still make sense with the new contracts, but it's not a priority for now. We can always reopen / recycle this in the future if needed.

PR #1297, opened 2022-05-12
Test case showcasing the error #321 revert

danielmkm:
The test tries to recreate the circumstances of the beethoven x UST 4 pool that became blocked May 11, 2022.

The interesting part of this test is that the failure is non-deterministic. The error revert occurs at different balances on each test run.


EndymionJkb:
Thanks! Will be incorporated in the StablePhantom tests.

PR #1298, opened 2022-05-12
Add explicit getters for TimelockAuthorizer

TomAFrench:
This PR replaces any public state variables with private versions and a getter.

One question this raises is whether we want to be more careful about potential confusion between `isPermissionGranted` and `hasPermission` (I've added some natspec on this). One option is to not expose the `isPermissionGranted` getter but this would cause confusion when trying to figure out if someone's permission is local or global.

### DIFF
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -77,7 +77,7 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
     IAuthentication private immutable _vault;
     uint256 private immutable _rootTransferDelay;
 
-    address public root;
+    address private _root;
     ScheduledExecution[] public scheduledExecutions;
     mapping(bytes32 => bool) public isPermissionGranted;
     mapping(bytes32 => uint256) public delaysPerActionId;
@@ -127,7 +127,7 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
         IAuthentication vault,
         uint256 rootTransferDelay
     ) {
-        root = admin;
+        _root = admin;
         _vault = vault;
         _executor = new TimelockExecutor();
         _rootTransferDelay = rootTransferDelay;
@@ -148,7 +148,7 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
      * @dev Returns true if `account` is the root.
      */
     function isRoot(address account) public view returns (bool) {
-        return account == root;
+        return account == _root;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -197,6 +212,21 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
         return keccak256(abi.encodePacked(actionId, account, where));
     }
 
+    /**
+     * @dev Returns true if `account` is has the permission defined by action `actionId` and target `where`.
```
EndymionJkb, line 216:
```suggestion
     * @dev Returns true if `account` has the permission defined by action `actionId` and target `where`.
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -197,6 +212,21 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
...
+     * This function is specific for the strict permission defined by the tuple `(actionId, where)`, `account` may also
+     * hold the global permission for the action `actionId` which would allow them to perform this action on `where`.
+     * For this reason, it's recommended to use `hasPermission` if checking whether `account` is allowed to perform
+     * a given action.
+     */
+    function isPermissionGranted(
```
EndymionJkb, line 222: It would break the name consistency, but would "isPermissionGrantedOnTarget" be clearer?

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -284,7 +321,7 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
      * @dev Sets the root address to `newRoot`.
      */
     function setRoot(address newRoot) external onlyExecutor {
```
EndymionJkb, line 323: Just throwing out there whether this is important enough to justify a claim step, vs just having the timelock. setRoot would then just set _rootCandidate, and the new root account would have to call claimRoot(), which would set _root and emit the event.

EndymionJkb (changes requested):
Looks like getRootTransferDelay() could be external; it's not called in the contract. And getRoot() is external (getting an immutable, so cheap when it's read directly in this contract).

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
TomAFrench: Yeah, I'd agree that this is something we want to have as many failsafes as possible on. I'm made an issue to track this in #1312 but I think it falls outside the scope of this PR.

PR #1299, opened 2022-05-12
Update Authorizer tests

TomAFrench:
I noticed that we had a couple of contexts descriptions backwards so fixed this.

Something that made me a little uneasy about the current tests is they always call into the Vault which is a "special" contract in the context of the Authorizer (calling into the `setAuthorizer` function in particular). If instead we have a generic authenticated contract we don't have to think about whether we're testing something specific for that function or not.

Swapping out the contract which we test against also makes it easier for us to test handling the return value in later changes.

### DIFF
--- a/pkg/vault/contracts/test/MockAuthenticatedContract.sol
+++ b/pkg/vault/contracts/test/MockAuthenticatedContract.sol
@@ -1,0 +1,36 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


pkg/vault/contracts/test/MockAuthenticatedContract.sol
```diff
@@ -0,0 +1,36 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
+
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/SingletonAuthentication.sol";
+
```
EndymionJkb, line 19: Add something like this to explain the purpose?

```
/*
 * @author Balancer Labs
 * @notice Generic authenticated contract
 * @dev A general purpose contract that can be used for testing permissioned functions in a more abstract way,
 * to test Authorizer functionality independent of specific Vault functions.
 */
```

EndymionJkb (approved):

PR #1301, opened 2022-05-16
Add implementation + deployment of the ProtocolFeesWithdrawer

TomAFrench:
This PR adds the ProtocolFeesWithdrawer which allows us to prevent withdrawals of certain tokens from the ProtocolFeesCollector. This is helpful in the case where the ProtocolFeesCollector holds a token which must not be distributed under any circumstances (e.g. it's actually user funds rather than protocol fees) as an emergency subDAO can denylist the token to temporarily freeze it in the collector until they can be safely dealt with.

The initial list of denylisted tokens is just SNX and sBTC. We'll likely want to flesh that out with a full set of Sythetix tokens before deployment.


pkg/deployments/tasks/20220516-protocol-fee-withdrawer/readme.md
```diff
@@ -0,0 +1,12 @@
+# 2022-05-13 - Protocol Fee Withdrawer
```
nventuro, line 1:
```suggestion
# 2022-05-16 - Protocol Fee Withdrawer
```

pkg/deployments/tasks/20220516-protocol-fee-withdrawer/readme.md
```diff
@@ -0,0 +1,12 @@
...
+
+Deployment of the Protocol Fee Withdrawer, which is used to prevent any withdrawals of certain tokens from the Protocol Fee Collector. This is useful for certain situations where the Protocol Fee Collector holds tokens which should not be treated as regular protocol fees.
+
+See [the Double Entrypoint Fix Relayer](../20220513-double-entrypoint-fix-relayer/) for information on an example of a situation where this is necessary.
+
+## Useful Files
+
+- [Ethereum mainnet addresses](./output/mainnet.json)
+- [Polygon addresses](./output/polygon.json)
+- [Arbitrum addresses](./output/arbitrum.json)
```
nventuro, line 11: Let's also do optimism

TomAFrench:
If I build [289ac5f](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1301/commits/289ac5f4a7b5855e04b844f7a435d1a8dcb64cd9) I get a buildinfo with the hash `b068bb7aff7bd2bc4a519998ca0e67b1` which doesn't match what we have in this commit.

nventuro:
I pushed e2115f158731ba01f11eea6412018e3883e3c88e, which reverts the change from [e9b66e7](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1301/commits/e9b66e72b04a12403e9dd19d35e99d487406e4e5) in an unrelated contract in the same package, changing the build info. The latest commit (e2115f1) now has correct build info.

I'll undo e2115f1 separately in master, so that this PR has the correct code.

TomAFrench:
Thanks, If I build e2115f158731ba01f11eea6412018e3883e3c88e, I now get a buildinfo with the hash 72ed588df7b18e41b79aeecffdd00819 as expected.

PR #1307, opened 2022-05-17
Managed Pool: Circuit breaker PoC

EndymionJkb:
This is a direct port of the old #746 prototype (from last July) into the current codebase, so that it is slightly more usable as a baseline reference to what the original thinking was back then. The code base has changed enormously (e.g., add/remove), and we might well do circuit breakers completely differently, but for what it's worth, here's an update of that old draft PR.


pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -1277,6 +1419,45 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, AumProtocolFeeCache,
...
+    function _setCircuitBreakerRatio(
+        IERC20 token,
+        uint256 initialPrice,
+        uint256 minRatio,
+        uint256 maxRatio
+    ) internal {
+        // Has to be > minRatio (if equal, encoded value would be 0, indistinguishable from no circuit breaker)
+        _require(minRatio == 0 || minRatio > _MIN_CIRCUIT_BREAKER_RATIO, Errors.MIN_CIRCUIT_BREAKER_RATIO);
+        _require(maxRatio == 0 || maxRatio <= _MAX_CIRCUIT_BREAKER_RATIO, Errors.MAX_CIRCUIT_BREAKER_RATIO);
+        _require(maxRatio >= minRatio, Errors.INVALID_CIRCUIT_BREAKER_RATIOS);
```
gerrrg, line 1431: This line needs to be:
`_require(maxRatio >= minRatio || maxRatio == 0, Errors.INVALID_CIRCUIT_BREAKER_RATIOS);`
otherwise you cannot have a minRatio without a maxRatio

gerrrg:
Just going to start leaving my notes here as I come across items that need addressing

- CB logic for givenOut needs to be flipped
- checkCircuitBreakerUpperBound compares final price to lowerBound
- checkCircuitBreakerLowerBound compares final price to upperBound
- when I've been calculating/comparing bounds, I've been multiplying by the {max,min}Ratio rather than dividing so that I get the price of the BPT wrt the token. You divide and end up with the reciprocal. Not so much an error as much as how we want to organize/think about the CBs. 
- In setting CBs:
    - currently: `_require(maxRatio >= minRatio || maxRatio == 0, Errors.INVALID_CIRCUIT_BREAKER_RATIOS);`
    - change to: `_require(maxRatio >= minRatio || maxRatio == 0, Errors.INVALID_CIRCUIT_BREAKER_RATIOS);`
    - ^ allows you to have a `minRatio` without setting a `maxRatio`
- Currently the only way to set a CB is through `setCircuitBreakerRatio()` which requires values for all tokens in the pool. This needs to be broken out to allow setting singular circuit breakers; otherwise, the act of updating one CB value will end up resetting all `initialPrices`s for all tokens. If a user started out with a desired CB range for TokenA, set their CBs, and then that token drifted near the end of that range, setting CB range for TokenB will reinitialize TokenA's initialPrice, which makes the current value (a nearly bad value) the new reference point of the range, allowing worse values than intended. A notable case for which a user might want to set only only CB is when adding a token

EndymionJkb:
Thanks! It's a bit broken at the moment (by some merges: I actually didn't
intend to push it that last time). Will update it soon.

gerrrg:
Yeah no urgent rush on these items, just wanted to have all these things cataloged so we don't lose track

EndymionJkb:
Superseded by #1792 (and others to follow)

PR #1309, opened 2022-05-18
Update SNX impl input

nventuro:
The target was changed three days ago.


TomAFrench (approved):
Not sure there's anything for me to verify other than the updated SNX implementation address (which looks good.)

PR #1310, opened 2022-05-18
Automatically save the deployment tx hash on deployment

TomAFrench:
We now automatically save the mapping which previously sat in `networks.ts` into separate json files by network.

We only save deployment transactions for mainnet, polygon, arbitrum and optimism to avoid getting a bunch of testnet transactions saved.


pkg/deployments/src/task.ts
```diff
@@ -92,6 +92,7 @@ export default class Task {
     const output = this.output({ ensure: false });
     if (force || !output[name]) {
       const instance = await this.deploy(name, args, from, libs);
+      await saveContractDeploymentTransactionHash(instance.address, instance.deployTransaction.hash, this.network);
```
nventuro, line 95: This should be in the `deploy` function (which is called here in the previous line), along with the call to `save`. Otherwise, tasks that use `deploy` directly won't save their hashes.

TomAFrench: Ah true, I was working on the understanding that `deploy` still didn't write the result to `output` so this wasn't necessary.

PR #1313, opened 2022-05-23
Only allow root to create new granters or revokers

TomAFrench:
This PR aims to fix the "whack-a-mole" problem where giving granter powers to an address allows it to create infinitely more granters such that it's impossible for those granter powers to be meaningfully revoked. (Similarly with revoker powers).

To fix this we only allow the root address to create more granters/revokers, ensuring that an entity can only create more entities if they are "one level down the org chart".

I've retained the ability to granters to remove other granters as these act as a good first line of defense if a granter goes rogue (assuming that it's first action isn't to kill off the other granters) as we expect them to be more agile than the root and will likely become aware of the rogue granter before the root does.

### DIFF
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -431,7 +431,7 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
     /**
      * @dev Sets `account`'s granter status to `allowed` for action `actionId` in target `where`.
      * Note that granters can revoke the granter status of other granters, even banning the root.
-     * However the root can allow themself back at any time while revoking the malicious granters.
+     * However, the root can always rejoin, and then revoke any malicious granters.
      */
     function manageGranter(
         bytes32 actionId,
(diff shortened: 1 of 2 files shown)


pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -430,16 +430,20 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
 
     /**
      * @dev Sets `account`'s granter status to `allowed` for action `actionId` in target `where`.
-     * Note that pairs can manage themselves, even banning the root, but the root can allow itself back at any time.
+     * Note that granters can revoke the granter status of other granters, even banning the root.
+     * However the root can allow themself back at any time while revoking the malicious granters.
```
EndymionJkb, line 434:
```suggestion
     * However, the root can always rejoin, and then revoke any malicious granters.
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -476,16 +480,20 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
 
     /**
      * @dev Sets `account`'s revoker status to `allowed` for action `actionId` in target `where`.
-     * Note that pairs can manage themselves, even banning the root, but the root can allow himself back at any time
+     * Note that revokers can revoke the revoker status of other revokers, even banning the root.
+     * However the root can allow themself back at any time while revoking the malicious revokers.
```
EndymionJkb, line 484:
```suggestion
     * However, the root can always rejoin, and then revoke any malicious revokers.
```

EndymionJkb (changes requested):

PR #1314, opened 2022-05-23
Only allow root to schedule a change to a delay

TomAFrench:
We only really expect a delay to be set up at around the same time at which we're performing the initial setup of a new permission. As it's an infrequent action with potential for griefing (a 2 year delay to withdraw protocol fees wouldn't be fun) it makes sense that only the root can perform this action.

### DIFF
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -360,8 +360,7 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
         address[] memory executors
     ) external returns (uint256 scheduledExecutionId) {
         require(newDelay <= MAX_DELAY, "DELAY_TOO_LARGE");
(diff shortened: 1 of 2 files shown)


EndymionJkb (approved):

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -360,8 +360,7 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
         address[] memory executors
     ) external returns (uint256 scheduledExecutionId) {
         require(newDelay <= MAX_DELAY, "DELAY_TOO_LARGE");
-        bool isAllowed = _hasPermissionOrWhatever(SCHEDULE_DELAY_ACTION_ID, msg.sender, address(this), actionId);
```
nventuro, line 363: With this change we no longer meaningfully use `SCHEDULE_DELAY_ACTION_ID` (only for emitting events) - does it make sense to keep it?

Issue #1312, opened 2022-05-20
Add a 2 step process for transferring root permissions


TomAFrench:
Losing control over the `root` of the Authorizer would be catastrophic so as Jeff [mentions](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1298#discussion_r877403894), we should implement a two step process for transferring this permission.

TomAFrench:
Addressed in #1316

PR #1315, opened 2022-05-23
Introduce 2-step process for changing the authorizer root account

EndymionJkb:
Since the consequences of accidentally setting the root to an invalid account would be catastrophic (see [Issue 1312](https://github.com/balancer-labs/balancer-v2-monorepo/issues/1312), introduce a two-step process (patterned off the similar method in BaseController)


EndymionJkb:
Superseded by #1316

EndymionJkb:
Superseded by #1316

PR #1316, opened 2022-05-23
Switch to a two step process to migrate the root account of the Authorizer

TomAFrench:
Fixes #1315 

The new flow for migrating the root account is now as follows:
1. root account calls `scheduleRootChange`passing the new root address
2. A executor executes this action, the new root address is then stored as `_pendingRoot`.
3. `_pendingRoot` must then call `claimRoot` on the Authorizer in order to become the new root address.

I've also migrated the migrator over to use this pattern safely. Previously `finalizeMigration` would atomically set the new root as well as swap out the old authorizer for the new on the Vault (disasterous if we *somehow* didn't realise that root was set incorrectly beforehand). We now execute a root transfer action to set `_pendingRoot` on the new Authorizer but `finalizeMigration` is blocked until it has been claimed, ensuring that it is an account which can interact with the new Authorizer. 

Something to consider is whether we want the `root` account to be able to immediately clear out the `_pendingRoot` variable in the case where we wish to abort a root transfer without having to wait `_rootTransferDelay`  to be able to reset `_pendingRoot` to zero.

### DIFF
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizerMigrator.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizerMigrator.sol
@@ -71,6 +71,9 @@ contract TimelockAuthorizerMigrator {
         bytes32 id = bytes32(uint256(address(_newAuthorizer)));
         GRANT_PERMISSION_ACTION_ID = keccak256(abi.encodePacked(id, TimelockAuthorizer.grantPermissions.selector));
         REVOKE_PERMISSION_ACTION_ID = keccak256(abi.encodePacked(id, TimelockAuthorizer.revokePermissions.selector));
+
+        // Enqueue a root change execution in the new authorizer to set it to the desired root address
+        rootChangeExecutionId = _newAuthorizer.scheduleRootChange(_root, _arr(address(this)));
     }
 
     /**
@@ -86,7 +89,6 @@ contract TimelockAuthorizerMigrator {
      */
     function migrate(uint256 n) external {
         require(!isComplete(), "MIGRATION_COMPLETE");
-        _beforeMigrate();
         _migrate(n == 0 ? rolesData.length : n);
         _afterMigrate();
     }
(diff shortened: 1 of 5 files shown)


EndymionJkb (approved):

pkg/vault/contracts/authorizer/TimelockAuthorizerMigrator.sol
```diff
@@ -158,6 +153,11 @@ contract TimelockAuthorizerMigrator {
             newAuthorizer.grantPermissions(actionIds, defaultAdmin, wheres);
         }
         newAuthorizer.revokePermissions(actionIds, address(this), wheres);
+
+        // Finally trigger the first step of transferring root ownership over the TimelockAuthorizer to `root`.
+        // Before the migration can be finalized, `root` must call `claimRoot` on the `TimelockAuthorizer`.
+        require(newAuthorizer.canExecute(rootChangeExecutionId), "CANNOT_TRIGGER_ROOT_CHANGE_YET");
+        newAuthorizer.execute(rootChangeExecutionId);
```
nventuro, line 160: This is slightly problematic in that we won't be able to do the role migration until the root delay has passed. Ideally the new contract will be complete before the migration is finalized so that addresses and permissions can be checked, etc. before the transfer actually occurs.

pkg/vault/test/authorizer/TimelockAuthorizerMigrator.test.ts
```diff
@@ -42,6 +42,9 @@ describe('TimelockAuthorizerMigrator', () => {
     newAuthorizer = await deployedAt('TimelockAuthorizer', await migrator.newAuthorizer());
     const setAuthorizerActionId = await actionId(vault, 'setAuthorizer');
     await oldAuthorizer.grantRolesToMany([setAuthorizerActionId], [migrator.address]);
+
+    const CHANGE_ROOT_DELAY = await newAuthorizer.getRootTransferDelay();
+    await advanceTime(CHANGE_ROOT_DELAY);
```
nventuro, line 47: This reflects how nothing can be done until the delay passes, which will be annoying in the real world.

nventuro (commented):
Necroing this PR for a small comment

pkg/vault/contracts/authorizer/TimelockAuthorizerMigrator.sol
TomAFrench: We'll be able to migrate all the roles except for the last one. Good point though, we should allow the roles to fully migrate without calling `_afterMigrate`

pkg/vault/test/authorizer/TimelockAuthorizerMigrator.test.ts
TomAFrench: We *could* move this down further but it's just here for simplicity.

PR #1319, opened 2022-05-25
Ensure that global permissions are properly transferred along with root status.

TomAFrench:
When we deploy the `TimelockAuthorizer` we grant permissions for the first `root` to grant or revoke all actions (`WHATEVER` on `EVERYWHERE`). This allows them to grant permissions to other accounts without having to first make themselves a granter for that permission.

I noticed while working on the migrator that we don't clean up these permissions when transferring the `root` account. If we transfer `root` then the old `root` will retain it's global grant/revoke permissions and the new `root` will need to manually give themselves these permissions before they can do anything.

This seems like a recipe for accidentally leaving very powerful permissions on an old `root` account which can then be used to grief later down the line so I've updated `claimRoot` to properly transfer these to the new `root`.


TomAFrench:
I've cut out some of the tests for the migrator as they don't really make sense in the new model of root transfers. The transfer of admin roles is handled by `claimRoot` so we don't need to test it again.

TomAFrench:
I've removed some unnecessary tests in #1321 so if we merge that it'll make the diff in this PR simpler.

EndymionJkb (approved):

PR #1320, opened 2022-05-25
Managed Pool: More informative management fee events

EndymionJkb:


TomAFrench:
I'm not sure if this is really necessary. Thoughts @nventuro ?

EndymionJkb:
Decided not to do it. Other fees don't give both values (e.g., in ProtocolFeesProvider); subgraph does have enough info, just a little more work.

PR #1322, opened 2022-05-25
Managed Pools: Consistent LP allowlist behavior

EndymionJkb:
Require the allowlist to be enabled for both adds and removes. (Otherwise the comment about the list being retained while off would not necessarily be correct.)

### DIFF
--- a/pkg/pool-weighted/contracts/smart/ManagedPool.sol
+++ b/pkg/pool-weighted/contracts/smart/ManagedPool.sol
@@ -436,8 +436,8 @@ contract ManagedPool is BaseWeightedPool, AumProtocolFeeCache, ReentrancyGuard {
 
     /**
      * @dev Can enable/disable the LP allowlist. Note that any addresses added to the allowlist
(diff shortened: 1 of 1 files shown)


pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -434,7 +436,8 @@ contract ManagedPool is BaseWeightedPool, AumProtocolFeeCache, ReentrancyGuard {
 
     /**
      * @dev Can enable/disable the LP allowlist. Note that any addresses added to the allowlist
-     * will be retained if the allowlist is toggled off and back on again.
+     * will be retained if the allowlist is toggled off and back on again, because removing addresses
+     * is not allowed while the allowlist is disabled.
```
TomAFrench, line 440:
```suggestion
     * will be retained if the allowlist is toggled off and back on again, because adding or removing addresses
     * is not allowed while the allowlist is disabled.
```

TomAFrench (approved):

PR #1325, opened 2022-05-26
Rework TimelockAuthorizerMigrate

TomAFrench:
This PR updates the TimelockAuthorizerMigrator to fully support the new TimelockAuthorizer (support for granters/revokers) and reworks how it transfers permissions across.

Main differences:
1. `RoleData` is no longer used to migrate all users with a particular permission on the old Authorizer to the new one but instead has an explicit address to grant the new permission to. We'll need to enumerate all permissions to be granted but we perform a check that this permission exists on the old Authorizer so we don't have to worry about slipping in a malicious permission here.
2. We now support setting up granters and revokers as part of the migration (passed in the constructor as `_grantersData` and `_revokersData`). As these concepts don't exist on the old Authorizer as do normal permissions we can't verify these onchain and so we must manually check that these are set up properly before performing the migration.

Other than this the migrator works in much the same way as before. The other major changes to the migrator came in #1316 and #1319 which increased safety / reduced the complexity of interacting with the TimelockAuthorizer

### DIFF
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizerMigrator.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizerMigrator.sol
@@ -50,6 +50,9 @@ contract TimelockAuthorizerMigrator {
         address target;
     }
 
+    /**
+     * @dev Reverts if _rolesData contains a role for an account which doesn't hold the same role on the old Authorizer.
+     */
     constructor(
         IVault _vault,
         address _root,
@@ -68,14 +71,19 @@ contract TimelockAuthorizerMigrator {
 
         for (uint256 i = 0; i < _rolesData.length; i++) {
             RoleData memory roleData = _rolesData[i];
+            // We require that any permissions being copied from the old Authorizer must exist on the old Authorizer.
+            // This simplifies verification of the permissions being added to the new TimelockAuthorizer.
             require(_oldAuthorizer.canPerform(roleData.role, roleData.grantee, roleData.target), "UNEXPECTED_ROLE");
             rolesData.push(roleData);
         }
         for (uint256 i = 0; i < _grantersData.length; i++) {
-            grantersData.push(_grantersData[i]);
+            // There's no concept of a "granter" on the old Authorizer so we cannot verify these onchain.
+            // We must manually verify that these permissions are set sensibly.
+            grantersData.push(RoleData(_grantersData[i].grantee, _grantersData[i].role, _grantersData[i].target));
         }
         for (uint256 i = 0; i < _revokersData.length; i++) {
-            revokersData.push(_revokersData[i]);
+            // Similarly to granters, we must manually verify that these permissions are set sensibly.
+            revokersData.push(RoleData(_revokersData[i].grantee, _revokersData[i].role, _revokersData[i].target));
         }
 
         // Enqueue a root change execution in the new authorizer to set it to the desired root address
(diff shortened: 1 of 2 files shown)


pkg/vault/contracts/authorizer/TimelockAuthorizerMigrator.sol
```diff
@@ -61,27 +70,41 @@ contract TimelockAuthorizerMigrator {
...
     }
 
     /**
-     * @dev Tells whether the migration has been completed or not
+     * @notice Returns whether the migration has been completed or not
      */
     function isComplete() public view returns (bool) {
-        return migratedRoles >= rolesData.length;
+        // As we set up the revoker roles last we can use them to determine whether the full migration is complete.
+        return revokersMigrated >= revokersData.length;
```
EndymionJkb, line 98: Seems a bit hacky to check this way. It depends on the implementation in _migrate(). Could use a boolean flag for this, and set it at the end of _migrate(). Or at least document in _migrate() that there is an order dependency, for future proofing. If somebody ever added something after, or changed the order, this could break

pkg/vault/contracts/authorizer/TimelockAuthorizerMigrator.sol
```diff
@@ -103,26 +126,68 @@ contract TimelockAuthorizerMigrator {
         vault.setAuthorizer(newAuthorizer);
     }
 
+    // Internal Functions
+
+    /**
+     * @notice Migrates to TimelockAuthorizer by setting up roles from the old Authorizer and new granters/revokers.
+     * @param n The number of permissions to set up on the new TimelockAuthorizer.
```
EndymionJkb, line 133: Maybe document that n == MAX_UINT256 means migrate all, here and wherever "n" appears? Otherwise people might wonder what all the min's are doing.

Assuming we have this "n" business in the first place, allowing partial migrations, to ensure we don't run out of gas trying to do them all? (e.g., if there were 300 permissions.)

pkg/vault/contracts/authorizer/TimelockAuthorizerMigrator.sol
```diff
@@ -61,27 +70,41 @@ contract TimelockAuthorizerMigrator {
...
+        return revokersMigrated >= revokersData.length;
     }
 
     /**
      * @dev Migrate roles from the old authorizer to the new one
-     * @param n Number of roles to migrate, use 0 to migrate all the remaining ones
+     * @param n Number of roles to migrate, use MAX_UINT256 to migrate all the remaining ones
      */
     function migrate(uint256 n) external {
         require(!isComplete(), "MIGRATION_COMPLETE");
```
EndymionJkb, line 106: The requires that check the completion status could be made into modifiers. Even though they're local to this contract, it would be a bit more readable.

`function migrate(uint256 n) external onlyBeforeComplete {...}`

TomAFrench: We only make these checks on two functions I don't think modifiers give us much benefits here as opposed to just inlining the check.

pkg/vault/contracts/authorizer/TimelockAuthorizerMigrator.sol
TomAFrench: I've addressed this in [810869a](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1325/commits/810869ac454bfb70a156752b75452b739fa6f9e8)

nventuro (approved):

PR #1326, opened 2022-05-26
Make MockMortalFactory inherit from a proper factory.

TomAFrench:
I wanted to do a dryrun of integrating `MortalFactory` into a real factory contract so changed `MockMortalFactory` to inherit from `MockPoolFactory`.

We've got some annoying inheritance issues but I don't think there's a way round without while keeping `MortalFactory` as a mixin. Let me know what you think.

### DIFF
--- a/pkg/pool-utils/contracts/factories/BasePoolFactory.sol
+++ b/pkg/pool-utils/contracts/factories/BasePoolFactory.sol
@@ -37,7 +37,7 @@ abstract contract BasePoolFactory {
     /**
      * @dev Returns the Vault's address.
      */
-    function getVault() public view returns (IVault) {
+    function getVault() public view virtual returns (IVault) {
         return _vault;
     }
 
--- a/pkg/pool-utils/contracts/test/MockMortalFactory.sol
+++ b/pkg/pool-utils/contracts/test/MockMortalFactory.sol
@@ -16,12 +16,22 @@ pragma solidity ^0.7.0;
 
 import "../factories/MortalFactory.sol";
 
-contract MockMortalFactory is MortalFactory {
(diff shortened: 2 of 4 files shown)


EndymionJkb:
I did find the code I had for using this with a real factory. I guess this is fine for testing; do we need it? Maybe just add some documentation? (I'll do that anyway.)

I just changed three lines (inherit from MortalFactory, call the base constructor, and add the check in create), added all the MortalFactory tests, and everything worked:

contract ManagedPoolFactory is MortalFactory {
    constructor(address baseFactory) MortalFactory ((BaseManagedPoolFactory(baseFactory).getVault())) {
        baseManagedPoolFactory = baseFactory;
    }

    function create(...) {
        _ensureEnabled();
        ....
   }
}

TomAFrench:
I'm not sure I follow that code snippet you shared there. Are you expecting the MortalFactory and the base factory to be separate contracts, i.e. MortalFactory is just a proxy calling into the real factory?

EndymionJkb:
> I'm not sure I follow that code snippet you shared there. Are you expecting the MortalFactory and the base factory to be separate contracts, i.e. MortalFactory is just a proxy calling into the real factory?

We're talking past each other somehow. MortalFactory is a Mixin you inherit from to make your factory killable. Maybe ManagedPoolFactory isn't the best example:

```
contract SomeFactory is BasePoolFactory, FactoryWidePauseWindow, MortalFactory {
    constructor(IVault vault) BasePoolFactory(vault) MortalFactory(vault) {
       ...
    }

    function create(...) external returns (address) {
        _ensureEnabled();
        ...
    } 
}

```

EndymionJkb:
Superseded by #1350

PR #1327, opened 2022-05-26
Do child chain streamer and gauge optimism deployment

nventuro:
See https://github.com/ethereum-optimism/ethereum-optimism.github.io/pull/89/files for the BAL address.


TomAFrench (approved):
As we're repeating an already verified deployment, CI should be sufficient for verifying this deployment.

PR #1330, opened 2022-05-30
Curvier math

### LINKED ISSUE #1341: Extend StableMath convergence range
We already use the same algorithm as Curve. Use a higher precision calculation that more closely parallels the Curve implementation, and duplicates its performance.

EndymionJkb:
Calculate the invariant with higher precision, to extend to range of convergence. Add more comprehensive math tests, and extend stable pool tests to all 5 tokens.

Closes #1341


pkg/pool-stable/contracts/StableMath.sol
```diff
@@ -68,33 +68,35 @@ library StableMath {
 
         // We support rounding up or down.
 
-        uint256 sum = 0;
         uint256 numTokens = balances.length;
+        uint256 prevInvariant;
+        uint256 sum;
```
nventuro, line 73: Can you link the Curve file to compare this against?

pkg/pool-stable/contracts/StableMath.sol
```diff
@@ -77,24 +79,29 @@ library StableMath {
...
 
         for (uint256 i = 0; i < 255; i++) {
-            uint256 P_D = balances[0] * numTokens;
-            for (uint256 j = 1; j < numTokens; j++) {
-                P_D = Math.div(Math.mul(Math.mul(P_D, balances[j]), numTokens), invariant, roundUp);
+            uint256 D_P = invariant;
+
+            for (uint256 j = 0; j < numTokens; j++) {
+                // (D_P * invariant) / (balances[j] * numTokens)
+                D_P = Math.div(Math.mul(D_P, invariant), Math.mul(balances[j], numTokens), roundUp);
```
nventuro, line 91: If I understand this correctly, D_P seems to be the inverse of P_D (or at least close - there seems to be a missing factor of `invariant^2`. If this is true, then likely we'd want to have rounding in the opposite direction as we originally did.

@sergioyuhjtman could you take a look at this?

pkg/pool-stable/test/StableMath.test.ts
```diff
@@ -27,18 +26,33 @@ describe('StableMath', function () {
...
-
-        expectEqualWithError(result, expectedInvariant, MAX_RELATIVE_ERROR);
-      });
+    async function checkInvariant(balances: BigNumber[], amp: number): Promise<void> {
+      const ampParameter = bn(amp).mul(AMP_PRECISION);
+
+      const result = await mock.invariant(ampParameter, balances, true);
+      const expectedInvariant = calculateInvariant(balances, ampParameter);
+
+      expectEqualWithError(result, expectedInvariant, MAX_RELATIVE_ERROR);
```
nventuro, line 35:
```suggestion
      const actualInvariant = await mock.invariant(ampParameter, balances, true);
      const expectedInvariant = calculateInvariant(balances, ampParameter);

      expectEqualWithError(actualInvariant, expectedInvariant, MAX_RELATIVE_ERROR);
```

pkg/pool-stable/test/StableMath.test.ts
```diff
@@ -27,18 +26,33 @@ describe('StableMath', function () {
...
+      for (let numTokens = 2; numTokens <= 5; numTokens++) {
+        let balances = Array(numTokens).fill(300);
+        balances = balances.map((b) => (random(100) > 50 ? b + random(50) : b - random(50)));
+
+        it(`computes the invariant for ${numTokens} tokens`, async () => {
+          for (let amp = 100; amp <= 5000; amp += 100) {
+            await checkInvariant(
+              balances.map((b) => fp(b)),
+              amp
+            );
```
nventuro, line 48:
```suggestion
        const balances = Array(numTokens).fill(300)
          .map((b) => (random(100) > 50 ? b + random(50) : b - random(50)))
          .map(fp);

        it(`computes the invariant for ${numTokens} tokens`, async () => {
          for (let amp = 100; amp <= 5000; amp += 100) {
            await checkInvariant(balances, amp);
```

nventuro: I find this slightly more straightforward, with all of the balances mutation in a single place

nventuro:
CI is failing for unrelated reasons. Merging!

PR #1333, opened 2022-05-30
Standardise bytecode output

TomAFrench:
We've managed to be inconsistent with how we save the contract creation code, resulting in some files having the key "bytecode" and some having "creationCode". We also have some of the bytecode strings prepended with `0x` and some not.

I've standardised these to match the original format as used in the first tasks.


nventuro:
Tom, slayer of prefixes

PR #1334, opened 2022-05-30
Check that the ABIs and bytecode from tasks matches what's in the build-info

TomAFrench:


.github/workflows/ci.yml
```diff
@@ -224,6 +224,28 @@ jobs:
       - name: Test
         run: yarn workspace @balancer-labs/v2-governance-scripts test
 
+  check-artifacts:
```
nventuro, line 227: Maybe rename these to 'check-deployment-artifacts' and 'check-deployment-addresses'

pkg/deployments/tasks/20210418-vault/abi/BalancerHelpers.json
```diff
@@ -68,7 +68,7 @@
         "type": "uint256[]"
       }
     ],
-    "stateMutability": "view",
+    "stateMutability": "nonpayable",
```
nventuro, line 71: Ah unfortunate. We have the `overrideQueryFunctions` helper change this, but not on the build info itself.

.github/workflows/ci.yml
TomAFrench: Done, I don't really want to change the hardhat tasks though as it'll be unwieldy on the CLI.

nventuro: That's fine - the deployments context is already there since those belong to that package

PR #1335, opened 2022-05-30
Cheaper range checking

EndymionJkb:
Propagate Tom's math insight to the rest of the codebase.


TomAFrench (approved):
The optimizoooor

EndymionJkb:
I tested it just to be sure - it saves about 60 gas

PR #1336, opened 2022-05-30
Deploy Tribe delegation fix

nventuro:
This implements [the Tribe proposal](https://forum.balancer.fi/t/tribe-dao-boost-delegation/3218), by creating a new delegation contract with some preseeded values.

The contract is the same as the current one, except some functions are made internal so that they can be called from another function, and the list of preseeded calls was added. The ABI had to be manually edited due to https://github.com/ethers-io/ethers.js/issues/3032, as the constructor was otherwise not callable from ethers.


nventuro:
We may want to also include https://github.com/curvefi/curve-veBoost/pull/13 it if proves correct, in which case we'd need to do a redeployment.

TomAFrench (approved):

nventuro:
Let's keep this in hold until we decide on https://github.com/curvefi/curve-veBoost/pull/13

PR #1337, opened 2022-05-31
Also refactor WeightCompression

EndymionJkb:
Why stop at WordCodec?


pkg/pool-weighted/contracts/lib/WeightCompression.sol
```diff
@@ -22,77 +22,37 @@ import "@balancer-labs/v2-solidity-utils/contracts/math/FixedPoint.sol";
...
-    }
-
-    /**
-     * @dev Convert a 64-bit value to full FixedPoint
-     */
-    function uncompress64(uint256 value, uint256 maxValue) internal pure returns (uint256) {
-        return value.mulUp(maxValue).divUp(type(uint64).max);
-    }
+    // Reverse a compression operation (normalized to a given value).
+    function uncompress(
```
TomAFrench, line 49: Seems like we should rename this be `decompress`?

TomAFrench (approved):
Looks good. Made a couple of notes on things that we might want to take the opportunity to change.

pkg/pool-weighted/contracts/lib/WeightCompression.sol
```diff
@@ -22,77 +22,37 @@ import "@balancer-labs/v2-solidity-utils/contracts/math/FixedPoint.sol";
  * so heavier compression (fewer bits) results in fewer decimals.
```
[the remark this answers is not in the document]
EndymionJkb: Will fix before merging.

pkg/pool-weighted/contracts/lib/WeightCompression.sol
EndymionJkb: This is like unregister vs deregister. It was kind of a toss-up there, but we decided on deregister.

decompress *is* usually the word used for the opposite of compression (e.g., in other software), so I'm happy to change it, unless there are strong objections.

Just in straight English terms, "uncompressed" is normally an adjective meaning it was never compressed. You can make a verb out of anything, but decompress is the more commonly used verb for reversing a compression operation.

EndymionJkb: For a people divided by a common language, we're agreeing a lot on English lately.

I need to decompress after a long day of PR wrangling!

pkg/pool-weighted/contracts/lib/WeightCompression.sol
```diff
@@ -17,82 +17,42 @@ pragma solidity ^0.7.0;
...
-    }
-
-    /**
-     * @dev Compress a FixedPoint value to 31 bits
-     */
-    function compress31(uint256 value) internal pure returns (uint256) {
-        return value.mulUp(_UINT31_MAX).divUp(FixedPoint.ONE);
-    }
+    // If a normalization factor is given (e.g., 10_000), normalize against that.
+    function compress(
```
nventuro, line 33: The docs for the full `compress` version should explain what it does, and what each argument means. The docs for the overloaded version with reduced arity should then explain that it's just a simpler version where `normalizeTo` is set to FP.ONE.

nventuro (commented):
Necroing this to add a small comment and suggest https://github.com/balancer-labs/balancer-v2-monorepo/issues/685 is also tackled if we're going to modify this code. 

@EndymionJkb since you've already been looking at these tests this should be fairly low hanging fruit for you.

PR #1339, opened 2022-05-31
Remove non split code pool factory

nventuro:
There's no point in keeping the simple factory, so better to just get rid of it.


pkg/pool-utils/contracts/factories/BasePoolSplitCodeFactory.sol
```diff
@@ -20,7 +20,14 @@ import "@balancer-labs/v2-interfaces/contracts/vault/IVault.sol";
...
 /**
- * @dev Same as `BasePoolFactory`, for Pools whose creation code is so large that the factory cannot hold it.
+ * @dev Base contract for Pool factories.
+ *
+ * Pools are deployed from factories to allow third parties to reason about them. Unknown Pools may have arbitrary
+ * logic: being able to assert that a Pool's behavior follows certain rules (those imposed by the contracts created by
+ * the factory) is very powerful.
+ *
+ * By using the split code mechanism, we can deploy Pools with creation code so large that a regular factory
+ * contract would not be able to store it.
```
TomAFrench, line 30:
```suggestion
 * @notice Base contract for Pool factories.
 *
 * Pools are deployed from factories to allow third parties to reason about them. Unknown Pools may have arbitrary
 * logic: being able to assert that a Pool's behavior follows certain rules (those imposed by the contracts created by
 * the factory) is very powerful.
 *
 * @dev By using the split code mechanism, we can deploy Pools with creation code so large that a regular factory
 * contract would not be able to store it.
```

PR #1350, opened 2022-06-01
Add disable function to BasePoolSplitCodeFactory

### LINKED ISSUE #1127: Add pool factory killswitch
It'd be interesting to be able to deprecate old pool factories entirely, by preventing them from creating more Pools. This would've proven useful with LBPs, for which we had to [release a hotfix](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1049), but still see usage of the old, deprecated factories.

Note that #1049 includes such a killswitch.

EndymionJkb:
Having removed the non-split factory in #1339, we now have only one. Add the "mortality" functionality directly to this base class (requiring no modifications to derived contracts), instead of doing a separate Mixin. 

Closes #1127

### DIFF
--- a/pkg/pool-utils/test/factories/BasePoolSplitCodeFactory.test.ts
+++ b/pkg/pool-utils/test/factories/BasePoolSplitCodeFactory.test.ts
@@ -60,7 +60,7 @@ describe('BasePoolSplitCodeFactory', function () {
     });
   });
 
-  describe('mortality', () => {
+  describe('disable', () => {
     context('when enabled', () => {
       it('disabled should be false', async () => {
         expect(await factory.isDisabled()).to.be.false;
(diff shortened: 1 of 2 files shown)


pkg/pool-utils/test/factories/BasePoolSplitCodeFactory.test.ts
```diff
@@ -48,4 +59,40 @@ describe('BasePoolSplitCodeFactory', function () {
       expect(await factory.isPoolFromFactory(other.address)).to.be.false;
     });
   });
+
+  describe('mortality', () => {
```
nventuro, line 63:
```suggestion
  describe('disabling', () => {
```

nventuro (approved):

PR #1354, opened 2022-06-02
Add more tests to WordCodec

nventuro:
This also greatly improves the validation logic, checking that the combination of bitlength and offset makes sense, and fixes a range check for negative integers.

### DIFF
--- a/pkg/solidity-utils/contracts/helpers/WordCodec.sol
+++ b/pkg/solidity-utils/contracts/helpers/WordCodec.sol
@@ -66,8 +66,7 @@ library WordCodec {
         uint256 offset,
         uint256 bitLength
     ) internal pure returns (bytes32) {
-        _require(bitLength > 0 && bitLength < 256, Errors.OUT_OF_BOUNDS);
-        _require(value >> bitLength == 0, Errors.CODEC_OVERFLOW);
+        _validateEncodingParams(value, offset, bitLength);
         uint256 mask = (1 << bitLength) - 1;
 
         bytes32 clearedWord = bytes32(uint256(word) & ~(mask << offset));
@@ -88,8 +87,7 @@ library WordCodec {
         uint256 offset,
         uint256 bitLength
     ) internal pure returns (bytes32) {
-        _require(bitLength > 0 && bitLength < 256, Errors.OUT_OF_BOUNDS);
-        _validateSignedInts(value, bitLength);
+        _validateEncodingParams(value, offset, bitLength);
 
         uint256 mask = (1 << bitLength) - 1;
         bytes32 clearedWord = bytes32(uint256(word) & ~(mask << offset));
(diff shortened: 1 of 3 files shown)


pkg/solidity-utils/test/WordCodec.test.ts
```diff
@@ -12,121 +12,291 @@ describe('WordCodec', () => {
...
+      async function assertSignedEncoding(value: BigNumberish, offset: number, bits: number) {
+        const result = await lib.encodeInt(value, offset, bits);
+
+        // We must be able to restore the original value
+        expect(await lib.decodeInt(result, offset, bits)).to.equal(value);
+        // All other bits should be clear
+        expect(negate(bn(1).shl(bits).sub(1).shl(offset)).and(bn(result))).to.equal(0);
+      }
+
+      // We want to be able to use 2 bit values, so we can only go up to offset 254. We only covert part of the offset
```
EndymionJkb, line 109:
```suggestion
      // We want to be able to use 2 bit values, so we can only go up to offset 254. We only cover part of the offset
```

pkg/solidity-utils/test/WordCodec.test.ts
```diff
@@ -12,121 +12,291 @@ describe('WordCodec', () => {
...
+        // We must be able to restore the original value
+        expect(await lib.decodeUint(result, offset, bits)).to.equal(value);
+        // All other bits should match the original word
+        const mask = negate(bn(1).shl(bits).sub(1).shl(offset));
+        const clearedResult = mask.and(result);
+        const clearedWord = mask.and(word);
+        expect(clearedResult).to.equal(clearedWord);
+      }
+
+      // We want to be able to use 2 bit values, so we can only go up to offset 254. We only covert part of the offset
```
EndymionJkb, line 191:
```suggestion
      // We want to be able to use 2 bit values, so we can only go up to offset 254. We only cover part of the offset
```

pkg/solidity-utils/test/WordCodec.test.ts
```diff
@@ -12,121 +12,291 @@ describe('WordCodec', () => {
...
+        // We must be able to restore the original value
+        expect(await lib.decodeInt(result, offset, bits)).to.equal(value);
+        // All other bits should match the original word
+        const mask = negate(bn(1).shl(bits).sub(1).shl(offset));
+        const clearedResult = mask.and(result);
+        const clearedWord = mask.and(word);
+        expect(clearedResult).to.equal(clearedWord);
+      }
+
+      // We want to be able to use 2 bit values, so we can only go up to offset 254. We only covert part of the offset
```
EndymionJkb, line 249:
```suggestion
      // We want to be able to use 2 bit values, so we can only go up to offset 254. We only cover part of the offset
```

pkg/solidity-utils/contracts/helpers/WordCodec.sol
```diff
@@ -197,10 +144,77 @@ library WordCodec {
...
+
+    /**
+     * @dev Decodes and returns a boolean shifted by an offset from a 256 bit word.
+     */
+    function decodeBool(bytes32 word, uint256 offset) internal pure returns (bool) {
+        return (uint256(word >> offset) & _MASK_1) == 1;
+    }
+
+    /**
+     * @dev Inserts 192 bit shifted by an offset into a 256 bit word, replacing the old value. Returns the new word.
```
EndymionJkb, line 157:
```suggestion
     * @dev Inserts a 192 bit value shifted by an offset into a 256 bit word, replacing the old value. Returns the new word.
```

PR #1356, opened 2022-06-03
Streamline StablePool

### LINKED ISSUE #1343: Streamline Pause functionality
After Recovery Mode infrastructure #1342, and applying it to StablePool #1353 pause mode will be modified to remove all the special-casing associated with paused pools and the standard proportional exit, which was pressed into service to fill the "emergency exit" role. Pause will then put the pool in "suspended animation" by halting everything, including the standard proportional exit. Users can then use the new exit method, crafted especially for this purpose.

EndymionJkb:
Remove special casing of proportional exit when the pool is paused. All normal exits work the same (i.e., are blocked) when paused.

Closes #1343


pkg/pool-stable/contracts/StablePool.sol
```diff
@@ -395,28 +395,20 @@ contract StablePool is BaseGeneralPool, LegacyBaseMinimalSwapInfoPool, IRateProv
         internal
         virtual
         override
+        whenNotPaused
```
nventuro, line 398: Can we add this modifier on BaseLegacyPool?

pkg/pool-stable/test/StablePool.test.ts
```diff
@@ -529,11 +529,11 @@ describe('StablePool', function () {
           await multiExitGivenIn();
         });
 
-        it('does not revert if paused', async () => {
+        it('reverts if paused', async () => {
           await pool.pause();
 
           const bptIn = previousBptBalance.div(2);
-          await expect(pool.multiExitGivenIn({ from: lp, bptIn })).not.to.be.reverted;
+          await expect(pool.multiExitGivenIn({ from: lp, bptIn })).to.be.revertedWith('PAUSED');
```
nventuro, line 536: I'd remove all of the pause tests here and instead have tests on LegacyBasePool that check that exits and joins are disabled when paused.

pkg/pool-stable/contracts/StablePool.sol
EndymionJkb: No - it's pure virtual there, so you can't add modifiers.

pkg/pool-stable/test/StablePool.test.ts
nventuro: Sure, but we can call `_ensureNotPaused()` before calling `_onExitPoll(...)` in the BasePool.

pkg/pool-stable/contracts/StablePool.sol
```diff
@@ -256,10 +256,7 @@ contract StablePool is BaseGeneralPool, LegacyBaseMinimalSwapInfoPool, IRateProv
         address,
         uint256[] memory scalingFactors,
         bytes memory userData
-    ) internal virtual override whenNotPaused returns (uint256, uint256[] memory) {
-        // It would be strange for the Pool to be paused before it is initialized, but for consistency we prevent
-        // initialization in this case.
```
nventuro, line 261: This comment was correct, it should not be removed.

pkg/pool-utils/contracts/LegacyBasePool.sol
```diff
@@ -244,6 +244,11 @@ abstract contract LegacyBasePool is IBasePool, BalancerPoolToken, TemporarilyPau
     ) public virtual override onlyVault(poolId) returns (uint256[] memory, uint256[] memory) {
         uint256[] memory scalingFactors = _scalingFactors();
 
+        // Joins are unsupported when paused
+        // It would be strange for the Pool to be paused before it is initialized, but for consistency we prevent
+        // initialization in this case.
```
nventuro, line 249:
```suggestion
```

pkg/pool-stable/test/meta/MetaStablePool.test.ts
```diff
@@ -258,30 +258,8 @@ describe('MetaStablePool', function () {
...
-          context('when the oracle is enabled', () => {
-            itDoesNotUpdateTheOracleData(action, lastChangeBlockOffset);
-            itDoesNotCacheTheLogInvariantAndSupply(action, lastChangeBlockOffset);
-          });
-
-          context('when the oracle is disabled', () => {
-            itDoesNotDoAnythingWhenDisabled(action, lastChangeBlockOffset);
-          });
+        it('all exits are disabled', async () => {
+          await expect(action(0)).to.be.revertedWith('PAUSED');
```
nventuro, line 262: I don't understand this test - can't we remove this?

pkg/pool-stable/contracts/StablePool.sol
EndymionJkb: It got moved up to LegacyBasePool, where it's disabled, not removed.

pkg/pool-stable/test/meta/MetaStablePool.test.ts
EndymionJkb: Yes; don't need it now.

pkg/pool-stable/contracts/StablePool.sol
EndymionJkb: Moving the check from StablePool up to LegacyBasePool also affects Linear & PhantomStable (had to skip tests that would fail). Should we put it back? Maybe it should only be moved after all the other pools support recovery mode.

PR #1357, opened 2022-06-04
goerli deployment

mikemcdonald:
Full goerli testnet deployment. Doesn't use all of the veBAL fix tasks but should be in an almost identical end state


nventuro:
I created a new CI job to check the testnet deployments, and added secrets to the repo for these endpoints. The Kovan one is not an archive node as Alchemy no longer supports it, but we don't need archive nodes for this anyway.

We still need to populate the transaction files for both Kovan and Goerli for the jobs to run though.

PR #1358, opened 2022-06-05
Add  WeightCompression tests

### LINKED ISSUE #685: Add tests for WeightCompression

EndymionJkb:
This adds the long-awaited tests for WeightCompression - and basic input validation - in the spirit of WordCodec. Of course it is much simpler (e.g., only two functions, everything is unsigned). Testing over the full input range led to adjusting the rounding directions to increase accuracy. There is no difference in the "normal" operating range (e.g, 16, 32, 64 bits).

Closes #685

### DIFF
--- a/pkg/pool-weighted/contracts/test/MockWeightCompression.sol
+++ b/pkg/pool-weighted/contracts/test/MockWeightCompression.sol
@@ -1,0 +1,46 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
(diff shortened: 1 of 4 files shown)


pkg/pool-weighted/contracts/test/MockWeightCompression.sol
```diff
@@ -0,0 +1,46 @@
...
+
+    // If no normalization value is given, assume it is 1.
+    function compress(uint256 value, uint256 bitLength) external pure returns (uint256) {
+        return WeightCompression.compress(value, bitLength, FixedPoint.ONE);
+    }
+
+    // Reverse a compression operation (normalized to 1).
+    function decompress(uint256 value, uint256 bitLength) external pure returns (uint256) {
+        return WeightCompression.decompress(value, bitLength, FixedPoint.ONE);
+    }
```
TomAFrench, line 45:
```suggestion
    function compress(uint256 value, uint256 bitLength) external pure returns (uint256) {
        return WeightCompression.compress(value, bitLength);
    }

    // Reverse a compression operation (normalized to 1).
    function decompress(uint256 value, uint256 bitLength) external pure returns (uint256) {
        return WeightCompression.decompress(value, bitLength);
    }
```

We're not testing the overloads here.

TomAFrench (commented):
It feels like we could remove some duplication in the tests with a `itCompressesCorrectly` helper.

TomAFrench:
Agreed, I took a quick shot at it and with the special cases you end up with a similar amount of complexity.

TomAFrench (approved):

PR #1361, opened 2022-06-06
Decouple pause and recovery mode

EndymionJkb:
Upon further consideration, pausing should not automatically enter recovery mode. Recovery mode has other implications, such as canceling protocol fees, and should not be entered automatically (even though this means there would be a time during which LPs would not be able to exit).


pkg/pool-utils/test/LegacyBasePool.test.ts
```diff
@@ -658,23 +658,19 @@ describe('LegacyBasePool', function () {
     let sender: SignerWithAddress;
 
     function pausingEntersRecoveryMode() {
```
nventuro, line 660: I'd rename this helper as well

PR #1366, opened 2022-06-08
Deploy fixed PreseededVotingEscrowDelegation contract

TomAFrench:
This PR pulls in the fix from https://github.com/curvefi/curve-veBoost/pull/15 so that we don't have the same issue with boosts which expire on consecutive weeks.


nventuro:
I verified the build info is correct

PR #1367, opened 2022-06-08
Improve recovery mode tests

nventuro:
This removes unnecessary pause/recovery mode interaction tests (there's none), and adds a critical test that checks that recovery mode exits work even if the pool is paused.


EndymionJkb (approved):
Yes, this is better; I was struggling with the pause window (it kept failing because it was past the window, I guess due to order dependencies). This is cleaner.

I have a fork test in the deployment making sure it can exit while paused

PR #1369, opened 2022-06-08
Pay no protocol fees on recovery mode

nventuro:
This makes all pools process a zero-valued protocol swap fee percentage. StablePool was already resetting last invariant on recovery mode deactivation, so we don't need to worry about that. 

The diff is best reviewed with no whitespace changes.


EndymionJkb (approved):
Mine was very similar; just went through Vault.ts to set the swap fee percentage, and had a separate zero fee event vs a parameter value for InnerOnJoin/ExitPoolCalled.

PR #1376, opened 2022-06-10
Add tests for BaseCordinator

TomAFrench:
This PR adds tests for the BaseCoordinator contract with some minor changes to functionality to accommodate this.

### DIFF
--- a/pkg/governance-scripts/contracts/test/TestCoordinator.sol
+++ b/pkg/governance-scripts/contracts/test/TestCoordinator.sol
@@ -17,7 +17,7 @@ pragma solidity ^0.7.0;
 import "../BaseCoordinator.sol";
 
 contract TestCoordinator is BaseCoordinator {
-    event StagesRegistered();
+    event RegisterStagesHookCalled();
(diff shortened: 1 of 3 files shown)


pkg/governance-scripts/contracts/test/TestCoordinator.sol
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    // Coordinator Setup
+
+    function _registerStages() internal override {
+        _registerStage(_performStageOne);
+        _registerStage(_performStageTwo);
+        _registerStage(_performStageThree);
+        _registerStage(_performStageFour);
+        emit StagesRegistered();
```
nventuro, line 35: I'd rename this to `RegisterStagesCalled` or 'hook called' etc

nventuro: Otherwise the tests make it seem like this is an official event

nventuro (approved):
Lovely! :clap:

PR #1380, opened 2022-06-10
Check that factory impls come from factories

nventuro:
This checks that action ids saved via https://github.com/balancer-labs/balancer-v2-monorepo/pull/1352 for pools that were created from a factory actually do correspond to pools from that factory.

### DIFF
--- a/pkg/deployments/src/actionId.ts
+++ b/pkg/deployments/src/actionId.ts
@@ -8,7 +8,7 @@ import Task, { TaskMode } from './task';
 
 const ACTION_ID_DIRECTORY = path.join(__dirname, '../action-ids');
 
-type ContractActionIdData = { useAdaptor: boolean; exampleContractAddress?: string; actionIds: Record<string, string> };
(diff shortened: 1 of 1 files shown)


pkg/deployments/src/actionId.ts
```diff
@@ -8,7 +8,7 @@ import Task, { TaskMode } from './task';
 
 const ACTION_ID_DIRECTORY = path.join(__dirname, '../action-ids');
 
-type ContractActionIdData = { useAdaptor: boolean; exampleContractAddress?: string; actionIds: Record<string, string> };
+type ContractActionIdData = { useAdaptor: boolean; factoryImplementation?: string; actionIds: Record<string, string> };
```
nventuro, line 11: tbh i'm not a fan of this name

TomAFrench: `factoryProduct` seems ok for now but maybe `factoryOutput`?

TomAFrench (approved):

pkg/deployments/src/actionId.ts
nventuro: I like it.

Issue #1364, opened 2022-06-07
StablePool tests assume 18-decimals


EndymionJkb:
This is a perennial problem, and is probably an issue elsewhere as well. There is a varyDecimals option when creating a TokenList that will guarantee non-18-decimal tokens, which in my opinion should be the default. Otherwise, it's possible to miss issues caused by forgetting to upscale/downscale properly.

The specific way this fails in Stable Pool is in calculating the invariant. Lots of functions depend on that, and if you pass unscaled balances, you will get an incorrect result. For the moment, the varyDecimals is off, so everything is 18-decimals, but we need to turn it on, and then fix all the broken tests.

There is (or soon will be) some upscale/downscale infrastructure in BasePool.ts now to help with this.

PR #1384, opened 2022-06-13
Extend StablePool tests to handle non-18-decimal tokens

EndymionJkb:
Addresses #1364. Adds some infrastructure and improvements to tests so that they also test non-18-decimal tokens in more cases.

Some are more difficult (e.g., using the MockVault, making assumptions about the history/lastInvariant), and would have to be rewritten completely.

StablePhantom also has similar issues, so I wouldn't consider 1364 closed, but this should be an incremental improvement.


EndymionJkb:
Revisit later? It would be good to resolve the flakiness.

PR #1388, opened 2022-06-14
Only run deployment tasks in master or if deployments package has changed

TomAFrench:
We're getting some rate limiting errors on the CI jobs which check the deployments package. As this isn't something that needs to be checked on each commit we can safely restrict these checks to commits in master or in PRs which modify the deployments package.


EndymionJkb (approved):
Yes; I'm seeing this too. Thanks!

PR #1389, opened 2022-06-14
Generalize stable invariant dependency logic

EndymionJkb:
Breaking up the StablePhantom recovery mode PR into smaller units for ease of review


pkg/pool-stable-phantom/contracts/test/MockStablePhantomPool.sol
```diff
@@ -15,14 +15,46 @@
...
         // solhint-disable-previous-line no-empty-blocks
     }
 
     function mockCacheTokenRateIfNecessary(IERC20 token) external {
         _cacheTokenRateIfNecessary(token);
     }
+
+    function getRate() public view virtual override whenInvariantConverges returns (uint256) {
+        return super.getRate();
+    }
```
nventuro, line 33: Let's take this a step further and instead of overriding `getRate` let's make `_scalingFactors()` revert.

nventuro: Maybe that should be an alternative thing though, as the scaling factors would revert if a dependency is broken, not the pool. Perhaps `InvariantDependency` can be renamed to e.g. `FailureModes`?

pkg/pool-stable-phantom/contracts/test/MockStablePhantomPool.sol
```diff
@@ -15,14 +15,46 @@
...
+
+    function _onSwapGivenIn(
+        SwapRequest memory request,
+        uint256[] memory balancesIncludingBpt,
+        uint256 indexIn,
+        uint256 indexOut
+    ) internal virtual override returns (uint256 amountOut) {
+        if (request.tokenIn != IERC20(this) && request.tokenOut != IERC20(this)) {
+            _ensureInvariantConverges();
+        }
```
nventuro, line 43: The invariant is still computed on single sided joins and exits (i.e. when token in or token out is BPT), so I'm not sure we need this `if`.

nventuro (commented):
Great initiative! :raised_hands:

pkg/pool-stable-phantom/contracts/test/MockStablePhantomPool.sol
EndymionJkb: We don't - I didn't look closely enough at the full code path (lots going on in those functions). Will just use the modifier.

pkg/pool-stable-phantom/contracts/test/MockStablePhantomPool.sol
EndymionJkb: Good idea - will generalize to rate providers failing

pkg/pool-stable-phantom/contracts/test/MockStablePhantomPool.sol
```diff
@@ -15,14 +15,57 @@
...
         _cacheTokenRateIfNecessary(token);
     }
+
+    function getRate()
+        public
+        view
+        virtual
+        override
+        whenNotInFailureMode(FailureMode.INVARIANT)
+        whenNotInFailureMode(FailureMode.PRICE_RATE)
```
nventuro, line 37: I'd remove the price rate modifier from here and add it to `_scalingFactors`. Same with swaps and joins. That'll let us more accurately simulate what actually happens (i.e. that getting the scaling factors fails).

pkg/pool-stable-phantom/contracts/test/MockStablePhantomPool.sol
```diff
@@ -15,14 +15,57 @@
...
 
-contract MockStablePhantomPool is StablePhantomPool {
+contract MockStablePhantomPool is StablePhantomPool, MockFailureModes {
     constructor(NewPoolParams memory params) StablePhantomPool(params) {
         // solhint-disable-previous-line no-empty-blocks
     }
 
-    function mockCacheTokenRateIfNecessary(IERC20 token) external {
+    function mockCacheTokenRateIfNecessary(IERC20 token) external whenNotInFailureMode(FailureMode.PRICE_RATE) {
         _cacheTokenRateIfNecessary(token);
```
nventuro, line 28: We should also add the modifier not to this external function but to the internal function that gets the rates and caches them. Otherwise, even if we set the price rate failure mode, the cache would not fail (which is not what would actually happen in production).

pkg/pool-utils/contracts/test/MockFailureModes.sol
```diff
@@ -0,0 +1,57 @@
...
+
+    modifier whenNotInFailureMode(FailureMode mode) {
+        _ensureNotFailed(mode);
+        _;
+    }
+
+    constructor() {
+        // Initialize error messages
+        _failureCode[FailureMode.INVARIANT] = Errors.STABLE_INVARIANT_DIDNT_CONVERGE;
+        _failureCode[FailureMode.PRICE_RATE] = Errors.LOW_LEVEL_CALL_FAILED;
```
nventuro, line 37: I'm torn on this. Intuitively I'd think it'd be better to error out with a custom message so that we can be sure that the cause for the failure was this module and not something inside the contracts. @TomAFrench wdyt?

EndymionJkb: The idea was to mimic the behavior of production so that we could potentially reuse tests that already check for failures with the "real" messages, but I can see both sides.

It's actually flexible enough now for both. I could initialize all modes with a custom 'INDUCED_FAILURE' message - but if you wanted to override it (e.g., with STABLE_INVARIANT_DIDNT_CONVERGE, to reuse a test), you could. How about that?

pkg/pool-stable-phantom/contracts/test/MockStablePhantomPool.sol
EndymionJkb: Ok; also getTokenRate

pkg/pool-utils/contracts/test/MockFailureModes.sol
nventuro: No, let's make it explicit that it is the mock contract that is failing. I don't really see many use cases where it'd make sense to reuse real tests.

nventuro (approved):

PR #1390, opened 2022-06-14
Add ChildChainGaugeRewardHelper

### LINKED ISSUE #1351: Add L2 gauge claimer utility
Tokens can be claimed for a RewardsOnlyGauge directly, but these are subject to a restriction in that they only get tokens from their streamer at most CLAIM_FREQUENCY seconds apart - meaning gauges only get tokens at most every hour.

However, the streamer transfer can be initiated by any account, not just the gauge. We could then build a simple permissionless utility that would let anyone claim on multiple gauges all of their reward tokens, without being subject to this delay. On top of that, we could provide a helper `view` method that'd return the number of tokens due, which front-end applications could call to display data.

TomAFrench:
This PR implements a helper contract as described in #1351 to allow claiming from multiple RewardsOnlyGauges in a single transaction while adding a workaround for the `CLAIM_FREQUENCY` issue.

Closes #1351.


pkg/liquidity-mining/contracts/gauges/ChildChainGaugeRewardHelper.sol
```diff
@@ -0,0 +1,52 @@
...
+
+    function claimRewardsFromGauge(IRewardsOnlyGauge gauge, address user) external {
+        _claimRewardsFromGauge(gauge, user);
+    }
+
+    function claimRewardsFromGauges(IRewardsOnlyGauge[] calldata gauges, address user) external {
+        for (uint256 i = 0; i < gauges.length; i++) {
+            _claimRewardsFromGauge(gauges[i], user);
+        }
+    }
```
nventuro, line 32: Let's add natspec docs, as it is not unlikely these will be used from etherscan.

pkg/liquidity-mining/contracts/gauges/ChildChainGaugeRewardHelper.sol
```diff
@@ -0,0 +1,52 @@
...
+
+    function _claimRewardsFromGauge(IRewardsOnlyGauge gauge, address user) internal {
+        if (gauge.last_claim() + CLAIM_FREQUENCY < block.timestamp) {
+            gauge.claim_rewards(user);
+        } else {
```
nventuro, line 37: I doubt this saves any gas - we could just remove the `if`.

pkg/liquidity-mining/contracts/gauges/ChildChainGaugeRewardHelper.sol
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    function pendingRewards(
+        IRewardsOnlyGauge gauge,
+        address user,
+        address token
+    ) external returns (uint256) {
+        gauge.reward_contract().get_reward();
+        return gauge.claimable_reward_write(user, token);
+    }
```
nventuro, line 51: Let's add a comment explaining that this function is expected to be called by off-chain clients via `eth_call` and not `eth_sendTransaction`, as if it were `view`.

pkg/liquidity-mining/test/ChildChainGaugeRewardHelper.test.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    await balToken.mint(streamerTwo.address, rewardAmount);
+    await streamerTwo.connect(distributor).notify_reward_amount(balToken.address);
+  });
+
+  describe('claimRewardsFromGauge', () => {
+    context('when the gauge has not claimed tokens from streamer recently', () => {
+      it('claims tokens from the streamer', async () => {
+        const tx = await gaugeHelper.claimRewardsFromGauge(gaugeOne.address, user.address);
+        expectTransferEvent(await tx.wait(), { from: streamerOne.address, to: gaugeOne.address }, balToken);
```
nventuro, line 105: We can augment these tests by calling `pendingRewards` and checking that the amount transfered matches exactly the pending rewards.

We can also add tests where we check that `pendingRewards` increases with the expected rate as time goes by, even if there's been a recent claim.

TomAFrench: > We can augment these tests by calling pendingRewards and checking that the amount transfered matches exactly the pending rewards.

I think this is something which is covered by the tests for the `RewardsOnlyGauge`. It's the responsibility of that contract to properly pay out rewards once the tokens are sent to it, this contract just ensures those tokens have been sent.

nventuro:
I added some rough tests that try to implement what I mentioned [here](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1390#discussion_r897308695), but don't understand why they fail.

TomAFrench:
> I added some rough tests that try to implement what I mentioned [here](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1390#discussion_r897308695), but don't understand why they fail.

You're assuming that all the user's tokens are sitting on the streamer.

```
existingRewards = Amount of tokens sitting on the gauge waiting to be claimed.
totalCumulativeRewards = Total amount of tokens to be paid out by timestamp T
expectedNewRewardsAmount = totalCumulativeRewards - existingRewards = Amount of rewards which should be claimable but are currently sitting on the streamer.

reportedPendingRewards = Amount of tokens sitting on the gauge after sending all tokens on the streamer which are ready to be paid out at timestamp T = totalCumulativeRewards
```

This means that `reportedPendingRewards != expectedNewRewardsAmount` unless the user has claimed all their tokens from the gauge in the same block at which you query. You should instead test `        expect(reportedPendingRewards.sub(existingRewards)).to.be.almostEqual(expectedNewRewardsAmount);
`

PR #1391, opened 2022-06-14
Rename error message for clarity

### LINKED ISSUE #1267: Consider renaming `UNAUTHORIZED_OPERATION` revert message
See: https://github.com/balancer-labs/balancer-v2-monorepo/pull/1164#discussion_r857807683

EndymionJkb:
As Tom pointed out in the issue referenced below, UNAUTHORIZED_OPERATION is misleading in the context of fixed pool controller rights. "Unauthorized" implies the caller doesn't have permission, where in fact the revert is due to an immutable property of the pool, set on deployment.

Closes #1267


nventuro (approved):
Looks like this was introduced in https://github.com/balancer-labs/balancer-v2-monorepo/pull/1001 and only used for ManagedPools. Looks great!

PR #1392, opened 2022-06-14
Deprecate deployments

### LINKED ISSUE #1323: Add "deprecated" folder in `deployments/tasks`
We've got a good number of deployment tasks which don't correspond to a production deployment anymore. As time progresses we're accumulating these which adds overhead to finding the relevant task you're looking for.

For example we have various merkle redeems, old relayers, etc. which are relatively harmless. However we also have a number of pool factories which contain vulnerabilities: LBPs, Stable pools, etc. Having these mixed in with the production code causes some risks for people to use the wrong factory for deployments, etc.

I think that we should moved these to a `deprecated` subdirectory to make it clear that these tasks should not be used. We'll have to make sure that no "active" tasks reference these deprecated tasks as they'll no longer be resolvable by the task apparatus. We should however continue to perform the `check-deployments` CI on them.

nventuro:
I also took the opportunity to create a `scripts` subdirectory under `tasks` for coordinators. The main readme has been updated to reflect these changes.

The task code changes were fairly minimal since we write good code that is easy to maintain :upside_down_face: 

Closes #1323


pkg/deployments/README.md
```diff
@@ -57,28 +57,16 @@ Returns the address of a contract's canonical deployment.
...
 
-## Past Deployments
+## Active Deployments
 
 | Description                                          | Task ID                                                                                              |
 | ---------------------------------------------------- | ---------------------------------------------------------------------------------------------------- |
 | Authorizer, governance contract                      | [`20210418-authorizer`](./tasks/20210418-authorizer)                                                 |
 | Vault, main protocol contract                        | [`20210418-vault`](./tasks/20210418-vault)                                                           |
 | Weighted Pools of up to 8 tokens                     | [`20210418-weighted-pool`](./tasks/20210418-weighted-pool)                                           |
-| Weighted Pools with two tokens and price oracle      | [`20210418-weighted-pool`](./tasks/20210418-weighted-pool)                                           |
```
TomAFrench, line 67: Would you say this is deprecated right now? afaik we use this on the frontend when deploying a 2 token pool atm.

pkg/deployments/README.md
```diff
@@ -87,20 +75,47 @@ Returns an object with all contracts from a deployment and their addresses.
...
+| Coordination of setup of L2 gauges for veBAL system | [`20220415-veBAL-L2-gauge-setup-coordinator`](./tasks/scripts/20220415-veBAL-L2-gauge-setup-coordinator) |
+| Coordination of veBAL gauges fix (Option 1)         | [`20220418-veBAL-gauge-fix-coordinator`](./tasks/scripts/20220418-veBAL-gauge-fix-coordinator)           |
+| veBAL Smart Wallet Checker Coordinator              | [`20220421-smart-wallet-checker-coordinator`](./tasks/scripts/20220421-smart-wallet-checker-coordinator) |
+| Tribe BAL Minter Coordinator                        | [`20220606-tribe-bal-minter-coordinator`](./tasks/scripts/20220606-tribe-bal-minter-coordinator)         |
+| Coordination of the double entrypoint issue fix     | [`20220610-snx-recovery-coordinator`](./tasks/scripts/20220610-snx-recovery-coordinator)                 |
+
+## Deprecated Deployments
+
+These deployments have been depreacted because they're either oudated and have been replaced by newer versions, or because they no longer form part of the current infrastructure. **In almost all cases they should no longer be used,** and are only kept here for historical reasons.
```
TomAFrench, line 105:
```suggestion
These deployments have been deprecated because they're either outdated and have been replaced by newer versions, or because they no longer form part of the current infrastructure. **In almost all cases they should no longer be used,** and are only kept here for historical reasons.
```

pkg/deployments/README.md
TomAFrench: I know that oracles are not being maintained going forwards but there's no reason to deprecate this deployment right at the minute.

nventuro: I mean the pools are fine in the sense that they are not incorrect, but discouraging people from creating more of them seems to be in the spirit of what we're doing.

TomAFrench: Fair, I'd say that if we're actively allowing new pools to be created on the frontend then we haven't deprecated it though. If we wanted to do that then we should stop the frontend from interacting with this factory surely?

nventuro: Yes - this PR is only part of the effort :stuck_out_tongue:

PR #1385, opened 2022-06-13
Remove oracle from Stable hierarchy

EndymionJkb:
Removes MetaStablePool and related infrastructure.

PR #1393, opened 2022-06-14
Remove remaining oracle infrastructure

EndymionJkb:
After #1385  and #1386, there are no remaining dependencies on the oracle infrastructure (PoolPriceOracle, QueryProcessor, interfaces), so they can be removed.


pkg/pool-utils/contracts/RecoveryMode.sol
```diff
@@ -41,11 +41,11 @@ import "./BasePoolAuthorization.sol";
...
- * to safely compute the invariant for the special case of proportional withdrawals, without running the usual more
- * complex code necessary in general (e.g., for non-proportional joins/exits and swaps). This might be enough to
- * support functions like `getRate`, allowing dependent pools to function normally. (Otherwise, the dependent pools
- * can themselves be placed into Recovery Mode.)
+ * In some pools, such as those with Price Rates, Recovery mode takes advantage of mathematical properties to safely
+ * compute the invariant for the special case of proportional withdrawals, without running the usual more complex code
+ * necessary in general (e.g., for non-proportional joins/exits and swaps). This might be enough to support functions
+ * like `getRate`, allowing dependent pools to function normally. (Otherwise, the dependent pools can themselves be
+ * placed into Recovery Mode.)
  *
```
nventuro, line 49: The blurb update is alright, but now that I look at this paraghraph it seems like we should remove it? After all, we don't try to compute `getRate` differently or anything like that.

nventuro (approved):

pkg/pool-utils/contracts/RecoveryMode.sol
EndymionJkb: Yes; this is mostly left over from when we had a more expansive scope for Recovery Mode, and is probably overkill now.

PR #1394, opened 2022-06-14
Use infura for CI tasks which don't need archive access

TomAFrench:
Looks like #1388 isn't enough so lets split across different RPCs.

### DIFF
--- a/.github/workflows/deployment-checks.yml
+++ b/.github/workflows/deployment-checks.yml
@@ -54,10 +54,10 @@ jobs:
       - name: Prepare Config
         run: yarn workspace @balancer-labs/v2-deployments ci:prepare-config
         env:
-          ALCHEMY_MAINNET_ARCHIVE_ENDPOINT: ${{ secrets.ALCHEMY_MAINNET_ARCHIVE_ENDPOINT }}
-          ALCHEMY_POLYGON_ARCHIVE_ENDPOINT: ${{ secrets.ALCHEMY_POLYGON_ARCHIVE_ENDPOINT }}
-          ALCHEMY_ARBITRUM_ARCHIVE_ENDPOINT: ${{ secrets.ALCHEMY_ARBITRUM_ARCHIVE_ENDPOINT }}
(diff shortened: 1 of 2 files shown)


.github/workflows/deployment-checks.yml
```diff
@@ -54,10 +54,10 @@ jobs:
...
-          ALCHEMY_MAINNET_ARCHIVE_ENDPOINT: ${{ secrets.ALCHEMY_MAINNET_ARCHIVE_ENDPOINT }}
-          ALCHEMY_POLYGON_ARCHIVE_ENDPOINT: ${{ secrets.ALCHEMY_POLYGON_ARCHIVE_ENDPOINT }}
-          ALCHEMY_ARBITRUM_ARCHIVE_ENDPOINT: ${{ secrets.ALCHEMY_ARBITRUM_ARCHIVE_ENDPOINT }}
-          ALCHEMY_OPTIMISM_ARCHIVE_ENDPOINT: ${{ secrets.ALCHEMY_OPTIMISM_ARCHIVE_ENDPOINT }}
+          MAINNET_RPC_ENDPOINT: ${{ secrets.ALCHEMY_MAINNET_ARCHIVE_ENDPOINT }}
+          POLYGON_RPC_ENDPOINT: ${{ secrets.ALCHEMY_POLYGON_ARCHIVE_ENDPOINT }}
+          ARBITRUM_RPC_ENDPOINT: ${{ secrets.ALCHEMY_ARBITRUM_ARCHIVE_ENDPOINT }}
+          OPTIMISM_RPC_ENDPOINT: ${{ secrets.ALCHEMY_OPTIMISM_ARCHIVE_ENDPOINT }}
       - name: Check Deployment Addresses
         run: yarn workspace @balancer-labs/v2-deployments check-deployments
```
nventuro, line 62: Checking deployments doesnt require access to an archive node - all we're doing is fetching tx receipts. The only jobs that need an archive node are fork tests, as those query past state.

PR #1397, opened 2022-06-16
PhantomStable recovery mode

### LINKED ISSUE #1403: Implement Recovery Mode in StablePhantomPool
Propagate StablePool recovery mode to PhantomStable, which is different in several ways. We do not need to compare invariants on joins/exits, so some special processing is unnecessary. However, we need to override the default proportional exit calculations to account for the BPT token.

EndymionJkb:
Replaces the former proportional exit (on pause only), with support for recovery mode. Recovery mode is reversible; at the cost of some gas per swap (~2.3k), BPT accounting is maintained. The pool does not charge protocol fees when recovery mode is enabled.

This first version has no "regular" joins/exits. Closes #1403


pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -1141,23 +1326,33 @@ describe('StablePhantomPool', () => {
...
       });
 
-      // Skip until recovery mode implemented
-      context.skip('when paused', () => {
-        context('one lp', () => {
-          sharedBeforeEach('pause pool', async () => {
-            await pool.pause();
-          });
+      context('when in recovery mode', () => {
+        sharedBeforeEach('enable recovery mode (and simulate utter failure)', async () => {
```
TomAFrench, line 1350: I'm not seeing how we're simulating failure here. Surely that's happening on L1446?

TomAFrench (approved):
I made a couple of commits to move some related functions closer together and clear up the diff a little. With that done, I'm happy with just a couple of potential comment changes.

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
EndymionJkb: That got left over after shuffling things around. Will remove; 1446 covers that case.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -511,96 +521,80 @@ contract StablePhantomPool is StablePool, ProtocolFeeCache {
...
-        bptOut = 0;
-        amountsIn = new uint256[](totalTokens);
-
-        // Due protocol fees are all zero except for the BPT amount, which is then zeroed out.
-        dueProtocolFeeAmounts = new uint256[](totalTokens);
-        dueProtocolFeeAmounts[_bptIndex] = _dueProtocolFeeBptAmount;
-        _dueProtocolFeeBptAmount = 0;
-    }
-
-    /**
```
nventuro, line 537: Try to avoid moving blocks around in complicated PRs, as it makes the review unnecessarily harder (e.g. I had to check whether this block had changed at all).

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -511,96 +521,80 @@ contract StablePhantomPool is StablePool, ProtocolFeeCache {
...
-        view
-        returns (uint256, uint256[] memory)
-    {
-        // This proportional exit function is only enabled if the contract is paused, to provide users a way to
-        // retrieve their tokens in case of an emergency.
-        //
-        // This particular exit function is the only one available because it is the simplest, and therefore least
-        // likely to be incorrect, or revert and lock funds.
+    // We cannot use the default implementation here, since we need to account for the BPT token
+    function _doRecoveryModeExit(
```
nventuro, line 549: I think there's a simpler way to achieve this and rely more on the base impl, highlighting the differences. Something like:

```
function _doRecoveryModeExit(
    uint256[] memory balances,
    uint256,
    bytes memory userData
) internal virtual override returns (uint256, uint256[] memory) {
    // Since this Pool uses preminted BPT, we need to replace the total supply with the virtual total supply, and
    // adjust the balances array by removing BPT from it.
    (uint256 bptAmountIn, uint256[] memory amountsOut) = super._doRecoveryModeExit(
        _dropBptItem(balances), 
        _getVirtualSupply(balances[_bptIndex])
    );

    return (bptAmountIn, _addBptItem(amountsOut, 0));
}
```

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -511,96 +521,80 @@ contract StablePhantomPool is StablePool, ProtocolFeeCache {
...
             _getVirtualSupply(balances[_bptIndex])
         );
 
         return (bptAmountIn, _addBptItem(amountsOut, 0));
     }
 
+    // Override the implementation in StablePool, which has special processing to "reset" protocol fee
+    // calculation after disabling recovery mode. This is unnecessary here, as protocol fees are
+    // acccumulated on each swap (or not, if recovery mode is enabled).
+    function _setRecoveryMode(bool enabled) internal virtual override {
```
nventuro, line 571: Good catch. We'll be able to get rid of these things thankfully once we remove StablePool. If we chose to kept it, we could also introduce a BaseStablePool, so this doesn't seem too concerning.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -794,14 +788,15 @@ contract StablePhantomPool is StablePool, ProtocolFeeCache {
...
+        // virtualSupply = _MAX_TOKEN_BALANCE - amounts[_bptIndex] + _dueProtocolFeeBptAmount;
         //
-        // However, since this Pool never mints or burns BPT outside of the initial supply (except in the event of an
-        // emergency pause), we can simply use `_MAX_TOKEN_BALANCE` instead of `totalSupply()` and save
-        // gas.
-        virtualSupply = _MAX_TOKEN_BALANCE - amounts[_bptIndex] + _dueProtocolFeeBptAmount;
+        // However, while the Pool is in recovery mode, LPs can exit proportionally and burn BPT. Since recovery
+        // mode is reversible - and unlike pausing, allows normal operation while enabled, including swaps -
+        // we cannot use this optimization. Otherwise, burning BPT in recovery mode would make the accounting
+        // permanently inaccurate (and the pool unsafe).
```
nventuro, line 798: I think this will create more confusion than anything. I don't think we need to explain why we are not making a hypothetical optimization that'd be available of the system worked differently. The fact that this used to be possible is irrelevant.

I'd delete these two paragraphs.

EndymionJkb:
> Left some comments on the tests.
> 
> I don't see any tests that check the protocol fee override works correctly though - am I missing anything? I think we should test a) swaps during recovery mode don't accrue fees, b) disabling recovery mode correctly continues with fee accrual and collection as previously.

There are some in "protocol swap fees / accounting," but I can improve them.

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -85,6 +85,11 @@ describe('StablePhantomPool', () => {
       expect(await pool.inRecoveryMode()).to.be.true;
     }
 
+    async function disableRecoveryMode(pool: StablePhantomPool): Promise<void> {
+      await pool.disableRecoveryMode(admin);
+      expect(await pool.inRecoveryMode()).to.be.false;
+    }
```
nventuro, line 91: These helpers are ok, but feel free to just call `disableRecoveryMode` and not check the result afterwards - we don't really need that test.

nventuro (approved):

PR #1399, opened 2022-06-17
WeightedPool simulator

EndymionJkb:
Written to investigate the behavior of extreme swap fees, the infrastructure created for the fee simulation could be reused for other purposes. (In any case I didn't want to lose it.)

This was written before the Python simulation tools being developed by Integrations, but having a "native" TS way to do this right in the repo might be useful as well. You can create a WeightedPool with arbitrary tokens and weights, and "seed" it with a given set of prices and desired total dollar amount of liquidity, so that the starting balances will be realistic.

This could be used to write new benchmark tests.


pvt/simulators/weighted-pools/weightedPool.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  // These "simulate" functions were designed for a very specific purpose - testing the behavior over
+  // the full range of swap fee values (1%-99%). They write CSV files in a format that can be imported
+  // to Excel and graphed.
+  //
+  // Of course it is more general: you can create a simulated pool with any number of tokens and starting
+  // liquidity (it will calculate the balances, given the token prices in USD), and perform operations to
+  // inspect the results.
+  //
+  // Running `yarn hardhat run weighted-pools/weightedPool.ts` from the /simulators directory should produce
```
TomAFrench, line 32:
```suggestion
  // Running `yarn ts-node weighted-pools/weightedPool.ts` from the /simulators directory should produce
```

We don't need hardhat in this package as we're not running a chain. We can also remove the config file + dependencies.

pvt/simulators/weighted-pools/model.ts
```diff
@@ -0,0 +1,388 @@
...
+  tokens: string[];
+  weights: number[];
+  balances: number[];
+  swapFee: number;
+
+  totalSupply: number;
+  numTokens: number;
+  prices: number[];
+  /* eslint-disable @typescript-eslint/no-explicit-any */
+  indices: any;
```
TomAFrench, line 21:
```suggestion
  indices: Record<string, number>;
```

pvt/simulators/weighted-pools/swaps.ts
```diff
@@ -0,0 +1,46 @@
+import WeightedPool from './model';
+import fs from 'fs';
+
+export enum SwapKind {
+  GivenIn = 0,
+  GivenOut,
+}
```
TomAFrench, line 7:
```suggestion
import fs from 'fs';
import { SwapKind } from '@balancer-labs/balancer-js';
import WeightedPool from './model';
```

TomAFrench (commented):
I'd move `weighted-pools/weightedPool.ts` to a `scripts` directory or similar. Unless you dig through the files then it's not obvious on how to run the simulator.

pvt/simulators/weighted-pools/model.ts
EndymionJkb: Thanks! Was wondering how to fix that

pvt/simulators/scripts/weightedPool.ts
```diff
@@ -0,0 +1,99 @@
...
+    );
+  }
+}
+
+main()
+  .then(() => process.exit(0))
+  .catch((error) => {
+    console.error(error);
+    process.exit(1);
+  });
```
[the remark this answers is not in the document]
EndymionJkb: Because I copied it from an existing script and didn't think of that :)

PR #1407, opened 2022-06-22
Replicate StablePool tests in Phantom

nventuro:
#1404 made Phantom no longer depend on Stable, but didn't change anything on the test suite. This PR adds some tests from Stable's suite that now test Phantom behavior (the biggest thing being the amplification factor).


EndymionJkb (approved):
Yes; looks like there wasn't much.

PR #1409, opened 2022-06-23
Add missing BasePool paused test

nventuro:
Somehow we were not testing for this :shrug:


EndymionJkb (approved):
Hard to believe - maybe it got lost in a merge at some point. In any case good catch.

PR #1410, opened 2022-06-23
Add AssetManagerConfig test

EndymionJkb:
Move the Mock from asset-manager-utils up to pool-utils, and add tests to BasePool (including pausing)


nventuro (approved):

pkg/pool-utils/test/BasePool.test.ts
```diff
@@ -794,4 +794,54 @@ describe('BasePool', function () {
...
+      pool = await deployBasePool({ owner: poolOwner, assetManagers });
+    });
+
+    sharedBeforeEach('set permissions', async () => {
+      const pauseAction = await actionId(pool, 'pause');
+      await authorizer.connect(admin).grantPermissions([pauseAction], admin.address, [ANY_ADDRESS]);
+    });
+
+    it('lets the owner set the asset manager config', async () => {
+      await pool.connect(poolOwner).setAssetManagerPoolConfig(tokens.first.address, encodedConfig);
```
nventuro, line 826: Let's emit an event in the MockAssetManager with the pool id and data received, and check that they match the expected values.

PR #1413, opened 2022-06-24
Support 5 tokens in PhantomStable

nventuro:
They artificially only supported up to 4 tokens, but this need not be the case.

### DIFF
--- a/pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
+++ b/pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
@@ -58,8 +58,12 @@ describe('StablePhantomPool', () => {
   });
 
   context('for a 5 token pool', () => {
+    itBehavesAsStablePhantomPool(5);
+  });
+
+  context('for a 6 token pool', () => {
     it('reverts', async () => {
-      const tokens = await TokenList.create(5, { sorted: true });
+      const tokens = await TokenList.create(6, { sorted: true });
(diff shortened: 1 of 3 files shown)


pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -770,7 +774,7 @@ describe('StablePhantomPool', () => {
 
         describe('update', () => {
           const itUpdatesTheRateCache = (action: (token: Token) => Promise<ContractTransaction>) => {
-            const newRate = fp(1.5);
+            const newRate = fp(4.5);
```
EndymionJkb, line 777: Why is this sensitive to the particular newRate value? It fails with 1.5, but succeeds with other values (such as 4.5).

Should the newRate be coordinated somehow (e.g., pulled out into a constant or something?)

EndymionJkb (commented):
I was thinking about this too, when working on the PR that to set the first 3 tokens

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
nventuro: Ah forgot to mention this in the description, sorry. Essentially each token is assigned an initial rate of `1.n`, where `n` is their index. This means the 5th token was assigned a rate of 1.5, and the test failed because the new rate was equal to the old rate. This will never happen with a rate of 4.5.

PR #1415, opened 2022-06-25
Migrate PhantomStable to BasePool

EndymionJkb:
Creates a temporary LegacyBaseGeneralPool.sol for StablePool, which will be deleted along with StablePool.


pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -968,237 +971,130 @@ describe('StablePhantomPool', () => {
...
 
-          const fee = amount.mul(swapFeePercentage).div(fp(1));
-          const protocolFee = fee.mul(protocolFeePercentage).div(fp(1));
-          return protocolFee;
-        }
-
-        context('on swaps given in', () => {
-          it('tracks fees when swapping tokens', async () => {
-            const previousDueFee = await pool.getDueProtocolFeeBptAmount();
+          function getApproxDueFee(amount: BigNumber, kind: AmountKind, recoveryMode: boolean): BigNumber {
```
TomAFrench, line 989:
```suggestion
          function getExpectedProtocolFee(amount: BigNumber, kind: AmountKind, recoveryMode: boolean): BigNumber {
```

I'd rename this function (along with the `approxFee` variables to `expectedProtocolFee`) so that the tests are of the format `expect(currentX).to.be..equalWithError(expectedX, 0.01)`

TomAFrench (approved):
Looks good. I'd rename some stuff in the tests for consistency with our other tests however.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -101,13 +101,6 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
     event AmpUpdateStarted(uint256 startValue, uint256 endValue, uint256 startTime, uint256 endTime);
     event AmpUpdateStopped(uint256 currentValue);
 
-    // Since this Pool is not joined or exited via the regular onJoinPool and onExitPool hooks, it lacks a way to
-    // continuously pay due protocol fees. Instead, it keeps track of those internally.
-    // Due protocol fees are expressed in BPT, which leads to reduced gas costs when compared to tracking due fees for
-    // each Pool token. This means that some of the BPT deposited in the Vault for the Pool is part of the 'virtual'
-    // supply, as it belongs to the protocol.
-    uint256 private _dueProtocolFeeBptAmount;
```
nventuro, line 109: It is true that we can now mint BPT and don't need to use a join to pay fees, but we might still want to keep this variable around. A proper fee payout involves updating total supply and the collector balances (2 sstores). If we keep this varaiable, we have one extra sload (for the variable), plus the overhead of requiring someone to periodically call here to collect fees.

Given that many operations will be joins and exits (causing fee payouts), the most sensible approach seems to be to keep this optimization and add a `collectDueProtocolFees` function (assuming that it is ok to require the manual collection call).

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -490,23 +478,20 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
 
-            _dueProtocolFeeBptAmount = _dueProtocolFeeBptAmount.add(protocolFeeAmount);
-
-            emit DueProtocolFeeIncreased(protocolFeeAmount);
+            _payProtocolFees(protocolFeeAmount);
         }
     }
 
     /**
      * @dev Tracks newly charged protocol fees after a swap where `bptAmount` was either sent or received (i.e. a
```
nventuro, line 486:
```suggestion
     * @dev Pays protocol fees charged after a swap where `bptAmount` was either sent or received (i.e. a
```

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -457,9 +445,9 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
     }
 
     /**
-     * @dev Tracks newly charged protocol fees after a swap where BPT was not involved (i.e. a regular swap).
+     * @dev Pay newly charged protocol fees after a swap where BPT was not involved (i.e. a regular swap).
```
nventuro, line 448:
```suggestion
     * @dev Pay protocol fees charged after a swap where BPT was not involved (i.e. a regular swap).
```

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -968,237 +971,130 @@ describe('StablePhantomPool', () => {
...
+          context('on swaps given in', () => {
+            it('pays any protocol fees due when swapping tokens', async () => {
+              const tokenIn = tokens.first;
+              const tokenOut = tokens.second;
+              await pool.swapGivenIn({ in: tokenIn, out: tokenOut, amount, from: lp, recipient });
 
-          it('tracks fees when swapping for BPT (join)', async () => {
-            const previousDueFee = await pool.getDueProtocolFeeBptAmount();
+              const currentFee = await pool.balanceOf(protocolFeesCollector.address);
+              const expectedFee = getExpectedProtocolFee(amount, AmountKind.WITH_FEE, inRecoveryMode);
```
nventuro, line 1013: This is a bit fragile as it relies on the initial fee collector balance to be zero. Can we change these so that they operate on deltas instead?

EndymionJkb: Will do

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
EndymionJkb: Missed one...

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -1057,237 +1060,136 @@ describe('StablePhantomPool', () => {
...
+              const tokenOut = tokens.second;
+              await pool.swapGivenIn({ in: tokenIn, out: tokenOut, amount, from: lp, recipient });
 
-            expect(currentDueFee).to.be.equalWithError(aproxFee, 0.01);
+              const currentFee = await pool.balanceOf(protocolFeesCollector.address);
+              const expectedFee = getExpectedProtocolFee(amount, AmountKind.WITH_FEE, inRecoveryMode);
 
-            expectEvent.inIndirectReceipt(receipt, pool.instance.interface, 'DueProtocolFeeIncreased', {
-              bptAmount: currentDueFee.sub(previousDueFee),
+              expect(currentFee).to.be.equalWithError(expectedFee.sub(previousBalance), 0.01);
```
nventuro, line 1106: These should be `currentFee.sub(previousBalance) equal expectedFee`. I'd suggest renaming `currentFee` to `currentBalance`.

nventuro (changes requested):

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
EndymionJkb: Yes; the naming made it confusing.

PR #1416, opened 2022-06-25
Add multi-token join to PhantomStable

EndymionJkb:


pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -627,17 +627,110 @@ describe('StablePhantomPool', () => {
...
+
+              const result = await pool.queryJoinGivenIn({ amountsIn, minimumBptOut });
+
+              expect(result.amountsIn).to.deep.equal(amountsIn);
+              expect(result.bptOut).to.be.equalWithError(expectedBptOut, 0.0001);
+            }
+
+            it('can tell how much BPT it will give in return', async () => {
+              await joinGivenIn();
+            });
```
TomAFrench, line 718: Seems like we only use this function here. Seems like we can inline it?

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -578,9 +578,60 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
 
         if (kind == StablePhantomPoolUserData.JoinKindPhantom.COLLECT_PROTOCOL_FEES) {
             return _collectProtocolFees();
+        } else if (kind == StablePhantomPoolUserData.JoinKindPhantom.EXACT_TOKENS_IN_FOR_BPT_OUT) {
+            return _joinExactTokensInForBPTOut(balances, scalingFactors, protocolSwapFeePercentage, userData);
+        } else {
```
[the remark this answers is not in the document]
EndymionJkb: Can do for this PR, though it's going away in a different one, so really doesn't matter.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -578,9 +578,60 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
+    {
+        (uint256[] memory amountsIn, uint256 minBPTAmountOut) = userData.exactTokensInForBptOut();
+        InputHelpers.ensureInputLengthMatch(_getTotalTokens() - 1, amountsIn.length);
+
+        uint256[] memory scaledAmountsInWithBpt = _addBptItem(amountsIn, 0);
+        _upscaleArray(scaledAmountsInWithBpt, scalingFactors);
+
+        // Balances come in from onJoinPool already upscaled
+        uint256 bptAmountOut = _computeBptAmountOut(balances, scaledAmountsInWithBpt);
+        _require(bptAmountOut >= minBPTAmountOut, Errors.BPT_OUT_MIN_AMOUNT);
```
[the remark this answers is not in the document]
EndymionJkb: Yes;  I tried a couple ways, too. The steps have to be performed somehow, and trying to optimize it adds complexity/reduces clarity, vs just communicating what's happening through the variable names.

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
EndymionJkb: Yes; this must have been called from multiple places at one time (maybe having to do with recovery mode)

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -578,9 +578,60 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
+
+        if (protocolSwapFeePercentage > 0) {
+            _trackDueProtocolFeeByBpt(bptAmountOut, protocolSwapFeePercentage);
+        }
+
+        return (bptAmountOut, _addBptItem(amountsIn, 0), new uint256[](balances.length));
```
TomAFrench, line 615:
```suggestion
        return (bptAmountOut, scaledAmountsInWithBpt, new uint256[](balances.length));
```
We're currently returning unscaled amounts back into `_onJoinPool` (which expects scaled values to be returned).

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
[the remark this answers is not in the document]
TomAFrench: Yeah, I'm not too bothered seeing as we're axing this.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -576,11 +576,62 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
+            uint256,
+            uint256[] memory,
+            uint256[] memory
+        )
+    {
+        (uint256[] memory amountsIn, uint256 minBPTAmountOut) = userData.exactTokensInForBptOut();
+        InputHelpers.ensureInputLengthMatch(_getTotalTokens() - 1, amountsIn.length);
+
+        uint256[] memory scaledAmountsInWithBpt = _addBptItem(amountsIn, 0);
+        _upscaleArray(scaledAmountsInWithBpt, scalingFactors);
```
nventuro, line 605:
```suggestion
        // Before we can upscale `amountsIn`, we must add BPT to it as the scaling factors must be applied to all 
        // registered tokens.
        uint256[] memory scaledAmountsInWithBpt = _addBptItem(amountsIn, 0);
        _upscaleArray(scaledAmountsInWithBpt, scalingFactors);
```

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -576,11 +576,62 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
+
+        // Balances come in from onJoinPool already upscaled
+        uint256 bptAmountOut = _computeBptAmountOut(balances, scaledAmountsInWithBpt);
```
nventuro, line 608: I'm not a fan of this 'compute' function - it's an odd abstranction, and doesn't do much other than hide details.

I propose an alternative refactor: we inline the 'compute' function, but create a function `_upscaleAmountsWithoutBPT` that does the whole insert-upscale-drop dance, explaining why that's required.

EndymionJkb: It's only there because I ran into stack-too-deep. I can try it another way

EndymionJkb: There seem to be just too many local variables need to inline it. This might be as clean as it gets; I added a lot of documentation.

PR #1417, opened 2022-06-25
Add regular exits to PhantomStable

EndymionJkb:
Support single and multi-token exits.


pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -641,17 +640,168 @@ describe('StablePhantomPool', () => {
...
+
+            const result = await pool.queryMultiExitGivenIn({ bptIn: previousBptBalance });
+
+            expect(result.bptIn).to.equal(previousBptBalance);
+            expect(result.amountsOut).to.be.lteWithError(expectedAmountsOut, 0.00001);
+          }
+
+          it('can tell how many tokens it will give in return', async () => {
+            await multiExitGivenIn();
+          });
```
TomAFrench, line 732:
```suggestion
          it('can tell how many tokens it will give in return', async () => {
                        const totalBPT = await pool.getVirtualSupply();
            const expectedAmountsOut = initialBalances.map((balance) =>
              bn(balance).mul(previousBptBalance).div(totalBPT)
            );

            const result = await pool.queryMultiExitGivenIn({ bptIn: previousBptBalance });

            expect(result.bptIn).to.equal(previousBptBalance);
            expect(result.amountsOut).to.be.lteWithError(expectedAmountsOut, 0.00001);
          });
```

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -641,17 +640,168 @@ describe('StablePhantomPool', () => {
...
+            const maximumBptIn = previousBptBalance.div(2);
+
+            await expect(pool.exitGivenOut({ from: lp, amountsOut, maximumBptIn })).to.be.revertedWith(
+              'BPT_IN_MAX_AMOUNT'
+            );
+          });
+
+          it('can tell how much BPT it will have to receive', async () => {
+            await exitGivenOut();
+          });
```
TomAFrench, line 796:
```suggestion
          it('fails if more BTP needed', async () => {
            // Call should fail because we are requesting a max amount lower than the actual needed
            const amountsOut = initialBalances;
            const maximumBptIn = previousBptBalance.div(2);

            await expect(pool.exitGivenOut({ from: lp, amountsOut, maximumBptIn })).to.be.revertedWith(
              'BPT_IN_MAX_AMOUNT'
            );
          });

          it('can tell how much BPT it will have to receive', async () => {
            const amountsOut = initialBalances.map((balance) => bn(balance).div(2));
            const expectedBptIn = previousBptBalance.div(2);
            const maximumBptIn = pct(expectedBptIn, 1.01);

            const result = await pool.queryExitGivenOut({ amountsOut, maximumBptIn });

            expect(result.amountsOut).to.deep.equal(amountsOut);
            expect(result.bptIn).to.be.equalWithError(previousBptBalance.div(2), 0.001);
          });
```

TomAFrench: ditto

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
EndymionJkb: Same as in the other one :)

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -584,29 +584,96 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
+
+        (uint256 virtualSupply, uint256[] memory balancesWithoutBpt) = _dropBptItem(balances);
+
+        uint256[] memory scaledAmountsOut = StableMath._calcTokensOutGivenExactBptIn(
+            balancesWithoutBpt,
+            bptAmountIn,
+            virtualSupply
+        );
+
+        return (bptAmountIn, _addBptItem(scaledAmountsOut, 0));
```
nventuro, line 644:
```suggestion
        uint256[] memory amountsOutWithoutBpt = StableMath._calcTokensOutGivenExactBptIn(
            balancesWithoutBpt,
            bptAmountIn,
            virtualSupply
        );

        return (bptAmountIn, _addBptItem(amountsOutWithoutBpt, 0));
```

The fact that these amounts don't have BPT is more important than that they are scaled (which all balances are by unless otherwise stated).

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -861,6 +928,22 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
+     * If the amounts given are token balances (e.g., from join/exit hooks), then the
+     * `virtualSupply` will be accurate, and is returned as a convenience, since often both
+     * values are needed together.
+     *
+     * It can be called with other kinds of amounts, but in that case the value returned in
+     * virtualSupply will not be meaningful.
+     *
+     * You can always call `getVirtualSupply` directly, if that is what you need.
+     */
     function _dropBptItem(uint256[] memory amounts)
```
nventuro, line 947: These docs are great, thanks. I'd suggest on a future PR to split this function into two, one that computes virtual supply and one that does not (since the virtual supply computation is not cheap, and we often don't use it).

Unfortunately the language does not support overloads that differ solely on the return type, so we'll need to have two different names (e.g. `_dropBptItem` and `_dropBptItemWithSupply` or w/e).

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -641,17 +640,160 @@ describe('StablePhantomPool', () => {
...
+            const bptIn = previousBptBalance.div(2);
+            const expectedAmountsOut = initialBalances.map((balance) => bn(balance).div(2));
+
+            const result = await pool.multiExitGivenIn({ from: lp, bptIn });
+
+            // Balances are reduced by half because we are returning half of the BPT supply
+            expect(result.amountsOut).to.be.equalWithError(expectedAmountsOut, 0.001);
+
+            // Current BPT balance should have been reduced by half
+            expect(await pool.balanceOf(lp)).to.be.equalWithError(bptIn, 0.001);
```
nventuro, line 698:
```suggestion
            // Exit with a third of the BPT balance
            const bptIn = previousBptBalance.div(3);
            // The pool should return about a third of its holdings
            const expectedAmountsOut = initialBalances.map((balance) => bn(balance).div(3));

            const result = await pool.multiExitGivenIn({ from: lp, bptIn });

            expect(result.amountsOut).to.be.equalWithError(expectedAmountsOut, 0.001);

            // Current BPT balance should have been reduced to two thirds
            expect(await pool.balanceOf(lp)).to.equa(previousBptBalance.sub(bptIn));
```

Avoid using 50% amounts - also changed a test to check exact amounts since there should be no error in that one.

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -641,17 +640,160 @@ describe('StablePhantomPool', () => {
...
+        });
+
+        function itExitsBptInForExactTokensOutProperly() {
+          it('grants exact tokens for bpt', async () => {
+            // Request half of the token balances
+            const amountsOut = initialBalances.map((balance) => bn(balance).div(2));
+
+            const expectedBptIn = previousBptBalance.div(2);
+            const maximumBptIn = pct(expectedBptIn, 1.01);
+
```
nventuro, line 759: It'd similarly be good to avoid using 50% here.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -632,29 +632,77 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
     }
 
     /**
-     * @dev dev Revert on all exit kinds. Note that this does not prevent recovery mode exits, as that one does not
-     * call`_onExitPool`.
+     * @dev Support multi-token exits, including proportional exits. Note that recovery mode exits
+     * do not call`_onExitPool`.
```
nventuro, line 636:
```suggestion
     * @dev Support multi-token exits. Note that recovery mode exits do not call`_onExitPool`.
```

nventuro (approved):

PR #1419, opened 2022-06-28
PhantomStable: clarify dropBptItem

EndymionJkb:
In _dropBptItem, the virtualSupply can only be calculated when the inputs are balances. This tripped me up, so for clarity I've made that explicit, and provided a generic dropBptItem that does only that (e.g., for amountsIn).

### DIFF
--- a/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
@@ -272,7 +272,7 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
         uint256 protocolSwapFeePercentage = _getProtocolSwapFeePercentage();
 
         // Compute virtual BPT supply and token balances (sans BPT).
-        (uint256 virtualSupply, uint256[] memory balances) = _dropBptItem(balancesIncludingBpt);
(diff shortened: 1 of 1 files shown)


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -861,16 +861,33 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
     {
-        virtualSupply = _getVirtualSupply(amounts[_bptIndex]);
+        virtualSupply = _getVirtualSupply(balances[_bptIndex]);
 
-        amountsWithoutBpt = new uint256[](amounts.length - 1);
-        for (uint256 i = 0; i < amountsWithoutBpt.length; i++) {
-            amountsWithoutBpt[i] = amounts[i < _bptIndex ? i : i + 1];
+        balancesWithoutBpt = new uint256[](balances.length - 1);
+        for (uint256 i = 0; i < balancesWithoutBpt.length; i++) {
+            balancesWithoutBpt[i] = balances[i < _bptIndex ? i : i + 1];
```
nventuro, line 890:
```suggestion
        balancesWithoutBpt = _dropBptItem(balances);
```

EndymionJkb: I think I am not getting enough sleep :)

PR #1420, opened 2022-06-28
Update GaugeAdder to support new networks.

TomAFrench:
This PR updates the Gauge Adder to have functions corresponding to Optimism, Gnosis Chain and ZKSync. 

To avoid the potential for duplicate L1 gauges, we now query from the previous gauge adder if we can't find an L1 gauge for a particular pool in the contract's local storage.


nventuro (approved):
Nice catch on getPoolGauge!

I checked that the build info from commit 07564c9ed03a0a98700d449eb23bdee5fc243bdd matches the one added here.

PR #1421, opened 2022-06-28
Deploy GaugeAdderV2

nventuro:
Once this is is live, we can deprecate the old deployment.


TomAFrench (approved):
Build info is the same as what I committed in #1420 and CI ensures that the deployed contract matches this.

PR #1422, opened 2022-06-29
PhantomPool: add single token exit

EndymionJkb:
Support single token exit, verifying that it is equivalent to a BPT for token swap

### DIFF
--- a/pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
+++ b/pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
@@ -759,6 +759,75 @@ describe('StablePhantomPool', () => {
         await expect(pool.exit({ data: wrongUserData })).to.be.revertedWith('Transaction reverted without a reason');
       });
 
+      describe('exit BPT in for one token out', () => {
+        let tokenIndex: number;
+        let token: Token;
+
+        sharedBeforeEach('get token to exit with', async () => {
+          tokenIndex = pool.bptIndex == 0 ? 1 : 0;
+          token = tokens.get(tokenIndex);
+        });
+
+        context('not in recovery mode', () => {
+          itExitsExactBptInForOneTokenOutProperly();
+        });
+
+        context('in recovery mode', () => {
+          sharedBeforeEach('enable recovery mode', async () => {
+            await pool.enableRecoveryMode(admin);
+          });
+
+          itExitsExactBptInForOneTokenOutProperly();
+        });
+
+        function itExitsExactBptInForOneTokenOutProperly() {
+          it('grants one token for exact bpt', async () => {
+            // 20% of previous balance
+            const previousBptBalance = await pool.balanceOf(lp);
+            const bptIn = pct(previousBptBalance, 0.2);
+            const expectedTokenOut = await pool.estimateTokenOutGivenBptIn(token, bptIn);
+
+            const result = await pool.singleExitGivenIn({ from: lp, bptIn, token });
+
+            // Only token out should be the one transferred
+            expect(result.amountsOut[tokenIndex]).to.be.equalWithError(expectedTokenOut, 0.0001);
+            expect(result.amountsOut.filter((_, i) => i != tokenIndex)).to.be.zeros;
+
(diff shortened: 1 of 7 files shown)


pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -759,6 +759,75 @@ describe('StablePhantomPool', () => {
...
+
+          it('can tell token amount returned (for either exit or exitSwap)', async () => {
+            const bptIn = pct(await pool.balanceOf(lp), 0.2);
+            const expectedTokenOut = await pool.estimateTokenOutGivenBptIn(token, bptIn);
+
+            const result = await pool.querySingleExitGivenIn({ bptIn, token });
+
+            expect(result.bptIn).to.equal(bptIn);
+            expect(result.amountsOut.filter((_, i) => i != tokenIndex)).to.be.zeros;
+            expect(result.amountsOut[tokenIndex]).to.be.equalWithError(expectedTokenOut, 0.0001);
```
nventuro, line 810: This test doesn't really do much, in that it is the same test as before (comparing the solidity implementation with the TS one), except using queries instead of actual joins. For completeness we could test that the query returns the same result as the actual operation, but that'd be a different test (i.e. do the query, then do the thing, and check for exact equality).

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -759,6 +759,75 @@ describe('StablePhantomPool', () => {
...
+
+            // Exit swap should give the same number of tokens
+            const { amountOut } = await pool.swapGivenIn({
+              from: lp,
+              in: pool.bpt,
+              out: token,
+              amount: bptIn,
+              recipient: lp,
+            });
+            expect(amountOut).to.be.equalWithError(expectedTokenOut, 0.00001);
```
nventuro, line 820: This is an interesting test, but a) it should be its own test, not bundled with the other one, and b) we would ideally compare the actual swap with the actual join, not with the expected amounts. This cannot be done as the swap/join will alter the state, but we _can_ do it via queries (i.e. check that the swap query exactly equals the join query).

Note that these tests will fail for now, as the swap fees are not handled the same way in joins and swaps, but they should eventually pass.

nventuro: We should also test protocol fees here, but those cannot be done via queries. Maybe we can just create an issue for now and tackle that later on.

nventuro (commented):
Lovely :clap:

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
EndymionJkb: How could it not, though? By definition they run the same code. And if A=B and A=C, then B=C.

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
EndymionJkb: Created the issue and updated the tests.

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -49,7 +49,7 @@ describe('StablePhantomPool', () => {
     itBehavesAsStablePhantomPool(2);
   });
 
-  context('for a 3 token pool', () => {
+  /*context('for a 3 token pool', () => {
```
nventuro, line 52:
```suggestion
  context('for a 3 token pool', () => {
```

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -59,7 +59,7 @@ describe('StablePhantomPool', () => {
 
   context('for a 5 token pool', () => {
     itBehavesAsStablePhantomPool(5);
-  });
+  });*/
```
nventuro, line 62:
```suggestion
  });
```

nventuro (approved):

PR #1423, opened 2022-06-29
PhantomPool: add single token join

EndymionJkb:
Support single token join, verifying that it is equivalent to a token for BPT swap


pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -727,6 +731,96 @@ describe('StablePhantomPool', () => {
...
+
+              const amountIn = await pool.querySwapGivenOut({
+                from: recipient,
+                in: token,
+                out: pool.bpt,
+                amount: bptOut,
+                recipient: lp,
+              });
+
+              expect(amountIn).to.be.equalWithError(queryResult.amountsIn[tokenIndex], 0.00001);
```
nventuro, line 813: Does this test fail with strict equality?

nventuro: Ah must be because of the swap fees. Can you test locally that if the swap fees are zero these two match exactly (just a temporary local change, don't add it to the suite)? If that happens we can then merge.

nventuro: They do indeed match.

Issue #1370, opened 2022-06-08
Make ProtocolFeeCache aware of recovery mode


nventuro:
https://github.com/balancer-labs/balancer-v2-monorepo/pull/1369 made recovery mode disable protocol fees. The ProtocolFeeCache module should also be aware of recovery mode and similarly return a value of 0 for the protocol fee percentage while in recovery mode.

nventuro:
In https://github.com/balancer-labs/balancer-v2-monorepo/pull/1397, we made it so PhantomPool overrides what the cache returns and instead evaluates protocol fees to zero while in recovery mode.

PR #1426, opened 2022-06-29
Remove dependency on FeeCache from AumFeeCache

nventuro:
Required for https://github.com/balancer-labs/balancer-v2-monorepo/issues/1370, but also nice to have in general.

AumProtocolFeeCache inherited from ProtocolFeeCache for no reason at all - I removed this, and in the process fixed a test that was wrong,

### DIFF
--- a/pkg/pool-utils/test/AumProtocolFeeCache.test.ts
+++ b/pkg/pool-utils/test/AumProtocolFeeCache.test.ts
@@ -11,12 +11,10 @@ import { ANY_ADDRESS, ZERO_ADDRESS } from '@balancer-labs/v2-helpers/src/constan
 import { MONTH } from '@balancer-labs/v2-helpers/src/time';
 
 describe('AumProtocolFeeCache', () => {
-  const FIXED_PROTOCOL_FEE = fp(0.1); // 10%
-
   const AUM_PROTOCOL_FEE = fp(0.05); // 5%
   const NEW_AUM_PROTOCOL_FEE = fp(0.1); // 10%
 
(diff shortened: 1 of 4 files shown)


pkg/pool-utils/test/AumProtocolFeeCache.test.ts
```diff
@@ -56,17 +54,17 @@ describe('AumProtocolFeeCache', () => {
...
-        expect(await protocolFeeCache.getProtocolAumFeePercentageCache()).to.equal(AUM_PROTOCOL_FEE);
+        expect(await feeCache.getProtocolAumFeePercentageCache()).to.equal(AUM_PROTOCOL_FEE);
       });
 
       it('updates the cached value', async () => {
-        await protocolFeeCache.updateProtocolAumFeePercentageCache();
+        await feeCache.updateProtocolAumFeePercentageCache();
 
-        expect(await protocolFeeCache.getProtocolSwapFeePercentageCache()).to.equal(NEW_AUM_PROTOCOL_FEE);
+        expect(await feeCache.getProtocolAumFeePercentageCache()).to.equal(NEW_AUM_PROTOCOL_FEE);
```
nventuro, line 63: We were calling the wrong function here, and it just so happened that `NEW_AUM_PROTOCOL_FEE` was equal to `FIXED_PROTOCOL_FEE`. This is an example of why we should avoid reusing values.

EndymionJkb (approved):
I don't remember now why it inherited, but agree that inheritance doesn't add anything. (I believe it was simpler in the beginning, without the fixed fee part.)

PR #1428, opened 2022-06-29
Remove last dependency on LegacyBasePool

EndymionJkb:
RelayedBasePool is the last thing other than StablePool that depends on LegacyBasePool.


nventuro:
Wow, I forgot this thing even existed. We can get rid of it as well. Thanks!

PR #1429, opened 2022-06-30
Make fee cache aware of recovery mode

### LINKED ISSUE #1370: Make ProtocolFeeCache aware of recovery mode

nventuro:
This makes the module replicate the `BasePool` behavior (replace protocol fees with zero, support recovery mode), meaning it can be used by Pools without requiring any further changes (such as the fee override we used to have).

Fixes #1370.

### DIFF
--- a/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
@@ -257,7 +257,7 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
     ) internal virtual override whenNotPaused returns (uint256 amountOut) {
         _cacheTokenRatesIfNecessary();
 
-        uint256 protocolSwapFeePercentage = _getProtocolSwapFeePercentage();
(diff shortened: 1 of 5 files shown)


EndymionJkb (commented):
I didn't do this because the multiple inheritance handling is messy and distasteful, but maybe the reuse is worth it.

nventuro:
It's not so much about reuse, but rather being able to plugin the module and it working out of the box without having to keep in mind that in order to use it correctly the fee thing needs to be adjusted. I'd say the override thing is a small price to pay.

EndymionJkb:
> It's not so much about reuse, but rather being able to plugin the module and it working out of the box without having to keep in mind that in order to use it correctly the fee thing needs to be adjusted. I'd say the override thing is a small price to pay.

Maybe reuse was the wrong word, but that's what I meant: that we wouldn't have to remember to add a function that returned 0 or the cached value in every descendant. (We do have to fix the override for derived classes, such as ManagedPool, but we can't forget that, since it won't build otherwise.)

PR #1430, opened 2022-06-30
Improve Recovery Mode interfaces

nventuro:
This started as a simple fix, since the interface file was in the wrong direction, but I also took the liberty to do a small wording update.


pkg/interfaces/contracts/pool-utils/IRecoveryMode.sol
```diff
@@ -15,16 +15,27 @@
 pragma solidity ^0.7.0;
 
 /**
- * @dev Interface for the RecoveryMode helper.
+ * @dev Interface for the RecoveryMode module.
  */
 interface IRecoveryMode {
     /**
-     * @dev Emitted every time the recovery mode changes through `_setRecoveryMode`.
+     * @dev Emitted when the Mecovery Mode status changes.
```
EndymionJkb, line 22:
```suggestion
     * @dev Emitted when the Recovery Mode status changes.
```

nventuro: :sweat_smile:

PR #1431, opened 2022-06-30
Make AumProtocolFeeCache aware of Recovery mode

EndymionJkb:


nventuro:
This feels a bit premature, no? I'm not sure we have a clear understanding of how recovery mode would affect managed pools. Would we really turn off the AUM fee? That feels like a bit of an outreach.

EndymionJkb:
> This feels a bit premature, no? I'm not sure we have a clear understanding of how recovery mode would affect managed pools. Would we really turn off the AUM fee? That feels like a bit of an outreach.

It's the AUM *protocol* fee, not the AUM management fee. If you think it's premature, we could label it with "needs discussion" or something like that. (I could make it a draft.)

nventuro:
Sure, but the reason why we turn off protocol fees in recovery mode is a bit indirect. Let's tag this for later discussion.

EndymionJkb:
Superceded by #1539

PR #1435, opened 2022-06-30
Remove StablePool

EndymionJkb:
Remove StablePool entirely; adjust any tests (and benchmarks) using regular StablePool to use StablePhantomPool instead


pvt/helpers/src/models/tokens/Token.ts
```diff
@@ -49,7 +49,8 @@ export default class Token {
     if (this.symbol === 'WETH') {
       await token.deposit({ value: amount });
       await token.transfer(TypesConverter.toAddress(to), amount);
-    } else {
+    } else if (this.symbol !== 'wstETH' && this.symbol != 'wampl') {
+      // wstETH has the permissioned wrap()
```
nventuro, line 53: Not sure what this change is about. It seems odd that the call to `mint` would do nothing in this case.

nventuro (commented):
Away it goes :smiling_face_with_tear:

pvt/helpers/src/models/tokens/Token.ts
EndymionJkb: Because wampl has a privileged mint that has to be called by Admin, and in the test is done before "init," which results in minting all the tokens. (Otherwise the "Unbutton" tests fail.)

nventuro: Can't we remove the `mint` call then there instead of doing this?

EndymionJkb: No, because it has to be done by the admin

nventuro: Ah that's annoying. Let's at least add a small comment here explaining that the unbutton tests do this, so we can reference it in the future.

EndymionJkb: Will do -  I did try to avoid this, but every other configuration added complexity and was even worse

nventuro (approved):

PR #1438, opened 2022-07-04
Add smarter timelock delay mechanism

### LINKED ISSUE #1436: Be smarter when changing permission delay
The TimelockAuthorizer currently waits for the full current delay when changing it to some other value. This is unnecessarily slow: we can not wait at all if the delay is increasing, and only wait for the difference of the two delays if it is decreasing (e.g. if the original delay is 5 and we're decreasing to 3, we only need to wait for 2).

We may want to have a minimum delay though (e.g. 3 days) for any delay change, as an immediate accidental increase from 2 days to 2 years wouldn't be fun.

TomAFrench:
This PR uses a more flexible delay mechanism for changing the delays of actions. Instead of always waiting for the entirety of the current delay before the new delay can be introduced it now only requires:

```
max(old_delay - new_delay, min_delay)
```

This ensures that we always have to wait at least `old_delay` before an action may be executed when reducing its delay, however when increasing the delay we don't need to wait for a long time unnecessarily. We introduce a minimum delay to allow for scrutiny of any pending delay changes to avoid potential mistakes.

closes #1436

### DIFF
--- a/pvt/helpers/src/models/authorizer/TimelockAuthorizer.ts
+++ b/pvt/helpers/src/models/authorizer/TimelockAuthorizer.ts
@@ -11,6 +11,7 @@ import { ANY_ADDRESS, ONES_BYTES32 } from '../../constants';
 import TimelockAuthorizerDeployer from './TimelockAuthorizerDeployer';
 import { TimelockAuthorizerDeployment } from './types';
 import { Account, NAry, TxParams } from '../types/types';
+import { advanceToTimestamp } from '../../time';
 
 export default class TimelockAuthorizer {
   static WHATEVER = ONES_BYTES32;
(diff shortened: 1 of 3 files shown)


pvt/helpers/src/models/authorizer/TimelockAuthorizer.ts
```diff
@@ -248,6 +249,7 @@ export default class TimelockAuthorizer {
     const setDelayAction = ethers.utils.solidityKeccak256(['bytes32', 'bytes32'], [SCHEDULE_DELAY_ACTION_ID, action]);
     await this.grantPermissions(setDelayAction, this.toAddress(from), this, params);
     const id = await this.scheduleDelayChange(action, delay, [], params);
+    await advanceToTimestamp((await this.getScheduledExecution(id)).executableAt);
```
TomAFrench, line 252: Silently incrementing the timestamp like these feels a little yucky but not doing so would push a bunch of boilerplate into the tests.

nventuro: Hm, unfortunate. I doubt this will introduce issues though, given that we often create permissions when setuping tests. Also, we don't do delays.

nventuro (approved):
Fantastic :raised_hands:

PR #1439, opened 2022-07-04
Set up initial TimelockAuthorizer delays in Migrator

TomAFrench:
This PR adds a new `_delaysData` argument to the TimelockAuthorizerMigrator which allows us to set an initial set of delays for various actions.

~~Thanks to #1438, this doesn't result in us extending the amount of time necessary for the migrator to execute out to the longest delay being set.~~ Delay mechanism smart, Tom dumb.

I've got an optional commit on here which reduces the number of SSTOREs we do to keep track of the execution IDs to execute but we could equally remove that.

### DIFF
--- a/pkg/vault/test/authorizer/TimelockAuthorizerMigrator.test.ts
+++ b/pkg/vault/test/authorizer/TimelockAuthorizerMigrator.test.ts
@@ -33,10 +33,11 @@ describe('TimelockAuthorizerMigrator', () => {
   let rolesData: RoleData[];
   let grantersData: RoleData[];
   let revokersData: RoleData[];
+  let executeDelaysData: DelayData[];
+  let grantDelaysData: DelayData[];
   const ROLE_1 = '0x0000000000000000000000000000000000000000000000000000000000000001';
   const ROLE_2 = '0x0000000000000000000000000000000000000000000000000000000000000002';
   const ROLE_3 = '0x0000000000000000000000000000000000000000000000000000000000000003';
-  let delaysData: DelayData[];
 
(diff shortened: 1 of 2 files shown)


TomAFrench:
Converting to draft as I noticed this only works for the delays on action execution, not delays on granting permissions

TomAFrench:
I've added an extra argument for the delays on granting permissions we want to set up.

We *could* include another for delays on revoking permissions but while we technically should set them up (we don't want the BalancerMinter's permission to mint BAL revoked) but considering the large amount of args appearing on the migrator and the benefit to be able to shut things down quickly in case there's an issue in the period after we migrate, I don't think we need it.

pkg/vault/test/authorizer/TimelockAuthorizerMigrator.test.ts
```diff
@@ -114,6 +150,23 @@ describe('TimelockAuthorizerMigrator', () => {
...
+
+        for (const delayData of executeDelaysData) {
+          expect(await newAuthorizer.getActionIdDelay(delayData.actionId)).to.be.eq(delayData.newDelay);
+        }
+
+        const GRANT_ACTION_ID = await newAuthorizer.GRANT_ACTION_ID();
+        for (const delayData of grantDelaysData) {
+          const grantActionId = await newAuthorizer['getActionId(bytes32,bytes32)'](
+            GRANT_ACTION_ID,
+            delayData.actionId
```
nventuro, line 164: We might want ot review this interface before deploying.

nventuro (approved):
Looks great! I didn't get the need for a non-empty array at first, but it makes sense now.

PR #1440, opened 2022-07-04
[Goerli] Adding deployment outputs for multiple tasks.

jubeira:
New tasks:

- batch-relayer-v2
- erc4626-linear-pool-v2
- child-chain-gauge-factory
- smart-wallet-checker
- distribution-scheduler
- unbutton-aave-linear-pool
- double-entrypoint-fix-relayer
- protocol-fee-withdrawer
- child-chain-gauge-token-adder
- preseeded-voting-escrow-delegation
- stable-pool-v2
- gauge-adder-v2
- smart-wallet-checker-coordinator


nventuro (approved):

nventuro:
Not sure why CI is failing here, might be due to an issue with secret injection since @jubeira was not part of the org when the PR was created. Merging - if the problem subsists we'll address it then.

PR #1441, opened 2022-07-04
Prefer deploy to deployAndVerify for Vyper contracts

nventuro:
This also makes `deploy` much more similar to `deployAndVerify`.

### DIFF
--- a/pkg/deployments/src/task.ts
+++ b/pkg/deployments/src/task.ts
@@ -91,19 +91,19 @@ export default class Task {
       return await this.check(name, args, libs);
     }
 
-    const output = this.output({ ensure: false });
-    if (force || !output[name]) {
-      const instance = await this.deploy(name, args, from, libs);
-      await this.verify(name, instance.address, args, libs);
-      return instance;
-    } else {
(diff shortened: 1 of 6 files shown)


pkg/deployments/src/task.ts
```diff
@@ -91,19 +91,19 @@ export default class Task {
       return await this.check(name, args, libs);
```
TomAFrench, line 91: This check on `TaskMode.CHECK` is redundant as we do another one right inside `deploy`. We *could* remove it if we're confident that we're not going to have any logic happening before the deploy in `deployAndVerify` (which seems reasonable).

TomAFrench (approved):
Looks good. We're double checking for if we're in `CHECK` mode but you could argue that's a safety check for if we modify `deployAndVerify` in future.

pkg/deployments/src/task.ts
nventuro: Yes, but that'd cause for CHECK executions to attempt verification, which we don't want to.

TomAFrench: Ah true, I forgot how functions work for a bit.

PR #1444, opened 2022-07-05
Always allow root to cancel a scheduled action

TomAFrench:
I'm open to being convinced that this is unnecessary but it feels like a risk free way of making the system safer as well as improving UX for the root.

Root has a global permission for `GRANT_ACTION_ID` for `WHATEVER`. If we assume that it create any granter of the same scope then it should have a right to cancel all scheduled granting of permissions which cannot be removed by anyone else (similarly for revoking of permissions).

Root doesn't however have a global permission to perform all actions. Should there be an action for which there is a delay for both granting and performing then if an address with that permission goes rogue then it's conceivable that the root will not be able to grant itself that permission in time to cancel any malicious actions.

Concrete (if pretty contrived) example: The `killBalancerProtocol()` function has a delay of 1 day to execute and can be granted to a new address with a delay of 2 days. There's a single address which has this permission and it schedules a call for this function to be executed. It's now impossible for root to cancel this action in time before it can be executed.

Granted, the conditions for root to not be able to safely cancel a scheduled execution are edge-casey but given that root can always grant itself the permissions necessary for cancelling a particular scheduled execution, we might as well allow them to do it without jumping through hoops.

### DIFF
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -493,10 +493,14 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication {
         require(!scheduledExecution.executed, "ACTION_ALREADY_EXECUTED");
         require(!scheduledExecution.cancelled, "ACTION_ALREADY_CANCELLED");
 
-        // The permission to cancel a scheduled action is the same one used to schedule it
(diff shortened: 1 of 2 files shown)


nventuro (approved):
This makes a lot of sense, yes. It draws a distinction from 'having a permission' and 'cancelling an action' that is a perfect fit for root. Well spotted!

PR #1446, opened 2022-07-05
Remove RelayedBasePool

EndymionJkb:
Continuing the swath of destruction, get rid of the. unused RelayedBasePool.

IRelayedBasePool is still used in RewardsAssetManager.


nventuro (approved):
Let ir burn :fire: 

RewardsAssetManager can also be removed.

EndymionJkb:
> Let it burn 🔥
> 
> RewardsAssetManager can also be removed.

That's coming out in the general asset-manager-utils PR

PR #1456, opened 2022-07-06
Remove overload from  `TimelockAuthorizer.getActionId`

### LINKED ISSUE #1449: Rename function for calculating extended action IDs
We've got an slightly awkward name for the function to calculate "extended" actionIds such that it's overloading the `getActionId` function. We should potentially rename this to be more explicit.

https://github.com/balancer-labs/balancer-v2-monorepo/pull/1439#discussion_r913257461

TomAFrench:
To remove the overload from the `getActionId` function I've introduced the concept of "extended action IDs". Whenever we calculate something of the form `keccak(actionId, how)`, the output is now an extended action ID.

Renaming the two argument version of `getActionId` to `getExtendedActionId` then removes the overloading from the interface.

Separately to this, the main usage of extended action IDs are going to be for the 4 actions of the authorizer (grant, revoke, execute, set delay). It's then probably worth creating specialised getters for these action as it saves users from having to make two calls into the authorizer (one to get the base action ID) and knowing to cast the execution ID to a bytes32.

I've then broken it out into `getGrantPermissionActionId`, `getRevokePermissionActionId`, `getExecuteExecutionActionId`, `getScheduleDelayActionId`. `getExtendedActionId` has been left public as it's possible that other contracts may use some extended action ID scheme so it could be helpful to have a generic version of this function available for consistency.

closes #1449

### DIFF
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -51,7 +51,7 @@ import "./TimelockExecutor.sol";
  *   the permission to set action delays, it is desirable to delineate whether an account can set delays for all
  *   actions indiscriminately or only for a specific action ID. In this case, the permission's "what" is the action
  *   ID for scheduling a delay change, and the "how" is the action ID for which the delay will be changed. The "what"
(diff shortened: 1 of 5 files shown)


TomAFrench:
Reviewed via slack due to login issues.
image

PR #1462, opened 2022-07-07
Add action IDs for single recipient gauges

TomAFrench:
I'll force merge this once CI is complete as that will verify this data is correct.


TomAFrench:
These new collisions are expected. `initialize()` isn't a protected function and `checkpoint` is expected to be called through the AuthorizerAdaptor.

PR #1465, opened 2022-07-07
PhantomStable: make join/exitSwap consistent with regular swaps

EndymionJkb:
Override higher level swap hooks from BaseGeneralPool to handle swaps with BPT separately. Override the swap fee calculations in the base class so that join/exit Swaps are exactly the same as regular swaps with BPT (and update tests to reflect this).

### DIFF
--- a/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
@@ -235,73 +235,57 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
     }
 
     /**
-     * @notice Top-level Vault hook to process all swaps (including those involving preminted BPT).
-     * @dev Override the top-level swap to check whether we are doing a regular swap (not involving BPT),
+     * @dev Override this hook called by the base class `onSwap`, to check whether we are doing a regular swap,
      * or a swap involving BPT, which is equivalent to a single token join or exit. Since one of the Pool's
      * tokens is the preminted BPT, we need to a) handle swaps where BPT is involved separately, and
      * b) remove BPT from the balances array when processing regular swaps, before calling the StableMath functions.
      *
-     * If this is a regular swap, call the BaseGeneralPool `onSwap`, which performs the default swap handling,
-     * including decimal scaling and charging swap fees, delegating the price quoting to this contract's
-     * `_onSwapGivenIn` and `_onSwapGivenOut` handlers.
+     * At this point, the balances are unscaled.
      *
-     * If this is a swap involving BPT, call `_onSwapBptGivenIn` or `_onSwapBptGivenOut`, which compute the amountIn/Out
-     * using the swapFeePercentage in the same manner as single token join/exits, and charge protocol fees on the
-     * corresponding bptAmount.
+     * If this is a swap involving BPT, call `_onSwapBpt`, which computes the amountOut using the swapFeePercentage,
+     * in the same manner as single token join/exits, and charges protocol fees on the corresponding bptAmount.
+     * Otherwise, perform the default processing for a regular swap.
      */
-    function onSwap(
+    function _swapGivenIn(
         SwapRequest memory swapRequest,
         uint256[] memory balances,
         uint256 indexIn,
-        uint256 indexOut
-    ) public virtual override whenNotPaused returns (uint256) {
+        uint256 indexOut,
+        uint256[] memory scalingFactors
+    ) internal virtual override whenNotPaused returns (uint256) {
         _cacheTokenRatesIfNecessary();
 
-        if (swapRequest.tokenIn == IERC20(this) || swapRequest.tokenOut == IERC20(this)) {
-            // This is a swap involving BPT
-            _validateIndexes(indexIn, indexOut, _getTotalTokens());
+        return
+            (swapRequest.tokenIn == IERC20(this) || swapRequest.tokenOut == IERC20(this))
+                ? _onSwapWithBpt(swapRequest, balances, indexIn, indexOut, scalingFactors)
+                : super._swapGivenIn(swapRequest, balances, indexIn, indexOut, scalingFactors);
+    }
 
-            uint256[] memory scalingFactors = _scalingFactors();
-            _upscaleArray(balances, scalingFactors);
+    /**
+     * @dev Override this hook called by the base class `onSwap`, to check whether we are doing a regular swap,
+     * or a swap involving BPT, which is equivalent to a single token join or exit. Since one of the Pool's
+     * tokens is the preminted BPT, we need to a) handle swaps where BPT is involved separately, and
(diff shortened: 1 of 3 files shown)


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -234,204 +234,283 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
-        } else if (request.tokenOut == IERC20(this)) {
-            amountIn = _onSwapTokenGivenBptOut(request.amount, _skipBptIndex(indexIn), virtualSupply, balances);
-
-            // For given out swaps, request.amount holds the amount out.
-            if (protocolSwapFeePercentage > 0) {
-                _payDueProtocolFeeByBpt(request.amount, protocolSwapFeePercentage);
-            }
+        bool bptIsTokenIn = swapRequest.tokenIn == IERC20(this);
+
+        // Would like to use a function selector here, but it causes stack issues
```
nventuro, line 390: What do you mean by function selector?

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -234,204 +234,283 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
-                uint256 amountInWithFee = _addSwapFeeAmount(amountIn);
-                balances[newIndexIn] = balances[newIndexIn].add(amountInWithFee);
-                balances[newIndexOut] = balances[newIndexOut].sub(request.amount);
-
-                _payDueProtocolFeeByInvariantIncrement(
-                    invariant,
-                    currentAmp,
-                    balances,
+            return
+                _onSwapBptGivenOut(
```
nventuro, line 406: We should downscale the result here, not in the inner function (i.e. keep upscaling and downscaling in the same place).

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -234,204 +234,283 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
-        uint256 virtualSupply,
-        uint256[] memory balances
-    ) internal view returns (uint256 bptIn) {
-        // Avoid BPT balance for stable pool math. Use virtual total supply and zero swap fees for exits.
-        (uint256 amp, ) = _getAmplificationParameter();
-        uint256[] memory amountsOut = new uint256[](_getTotalTokens() - 1);
-        amountsOut[tokenIndex] = amountOut;
-        bptIn = StableMath._calcBptInGivenExactTokensOut(amp, balances, amountsOut, virtualSupply, 0);
+        if (protocolSwapFeePercentage > 0) {
+            _payDueProtocolFeeByBpt(bptIsTokenIn ? amount : amountOut, protocolSwapFeePercentage);
```
nventuro, line 462: also unscaled here - see the suggestion on how to fix this above

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -810,7 +846,7 @@ describe('StablePhantomPool', () => {
                 recipient: lp,
               });
 
-              expect(amountIn).to.be.equalWithError(queryResult.amountsIn[tokenIndex], 0.00001);
+              expect(amountIn).to.be.equal(queryResult.amountsIn[tokenIndex]);
```
nventuro, line 849: Were these failing prior to the change?

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -1388,7 +1422,7 @@ describe('StablePhantomPool', () => {
               await pool.swapGivenIn({ in: pool.bpt, out: tokens.first, amount, from: lp, recipient });
 
               const currentBalance = await pool.balanceOf(protocolFeesCollector.address);
-              const expectedFee = getExpectedProtocolFee(amount, AmountKind.WITH_FEE, inRecoveryMode);
+              const expectedFee = getExpectedProtocolFee(amount, AmountKind.WITHOUT_FEE, inRecoveryMode);
```
nventuro, line 1425: I'm surprised that we had to change this, given the test bellow is quite lax when it comes to how much error it tolerates

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
EndymionJkb: Facu's trick: (boolVar ? funcA : funcB)(arg1, arg2, arg3)

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
EndymionJkb: It only scales when it needs to, taking advantage of knowing when it's BPT or not

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
EndymionJkb: Yes; if you switch them to "equal" in current master, they fail.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
nventuro: Ohh you mean function pointers. Yes, that'd be handy

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -390,29 +390,33 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
-                    swapRequest.amount,
-                    indexIn,
-                    indexOut,
-                    bptIsTokenIn,
-                    amp,
-                    protocolSwapFeePercentage,
-                    virtualSupply,
-                    balancesWithoutBpt,
-                    scalingFactors
+                _downscaleDown(
```
nventuro, line 393: I'd add a comment mentioning that since we're returning an amount out, we downscale down (and the opposite in the other case)

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -421,33 +425,31 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
      * @dev Process a GivenIn swap involving BPT.
      */
     function _onSwapBptGivenIn(
-        uint256 amount,
+        uint256 upscaledAmount,
```
nventuro, line 428: I'd just call this `amount`, everything else is also upscaled.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
EndymionJkb: Well now I can't anyway because I'm also doing the scaling up here.

nventuro (approved):

PR #1466, opened 2022-07-07
Rename some concepts in TimelockAuthorizer to help reduce potential confusion.

TomAFrench:
This PR aims to address two sources of potential confusion in the new Authorizer:
1. An overloading of the concept of "global" which can mean either a) a permission to perform an action anywhere or b) a permission to perform any form of a particular action in different places in the code.
2. The concept of "WHATEVER" which doesn't really have any easily understood meaning except by examining how it's used.
---

We have the concept of an "extended" action ID in the new Authorizer which was defined as `keccak(actionId, how)`. To give a permission to perform any `how` of an action, we would grant a permission for the action ID for when `how = WHATEVER`.

These terms are now renamed so that an extended action ID is now defined as `keccak(baseActionId, specifier)`. `WHATEVER` is then renamed to be the `GENERAL_PERMISSION_SPECIFIER` which (hopefully) gives a more descriptive idea of it's meaning.

In the code we've been calling `keccak(baseActionId, specifier)` the `granularActionId` and `keccak(baseActionId, GENERAL_PERMISSION_SPECIFIER)` the `globalActionId`. Under this renaming, it makes more sense to call these the `specificActionId` and the `generalActionId`.

Now we no longer have a concept of `WHATEVER`, we need to rename `_hasPermissionOrWhatever` and `_canPerformOrWhatever`. A sensible choice is `_hasPermissionSpecificallyOrGenerally` and `_canPerformSpecificallyOrGenerally` as these functions check that the user has either a specific or general permission to perform an action.

### DIFF
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -54,7 +54,7 @@ import "./TimelockExecutor.sol";
  *   actions indiscriminately or only for a specific action ID. In this case, the permission's "baseActionId" is the
  *   action ID for scheduling a delay change, and the "specifier" is the action ID for which the delay will be changed.
  *   The "baseActionId" and "specifier" of a permission are combined into a single "extended" `actionId`
- *   by computing `keccak256(baseActionId, specifier)`.
+ *   by calling `getExtendedActionId(baseActionId, specifier)`.
  */
 contract TimelockAuthorizer is IAuthorizer, IAuthentication {
     using Address for address;
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -47,18 +47,23 @@ import "./TimelockExecutor.sol";
  *   `keccak256(actionId, account, where)`.
  *
  * Permission granularity:
- *   In addition to the who/what/where of a permission, an extra notion of "how" is introduced to enable more granular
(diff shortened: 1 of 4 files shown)


pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -47,18 +47,23 @@ import "./TimelockExecutor.sol";
...
+ *   granular configuration. This concept is used within the Authorizer to provide clarity among four ambiguous actions:
  *   granting/revoking permissions, executing scheduled actions, and setting action delays. For example, in managing
  *   the permission to set action delays, it is desirable to delineate whether an account can set delays for all
- *   actions indiscriminately or only for a specific action ID. In this case, the permission's "what" is the action
- *   ID for scheduling a delay change, and the "how" is the action ID for which the delay will be changed. The "what"
- *   and "how" of a permission are combined into a single "extended" `actionId` by computing `keccak256(what, how)`.
+ *   actions indiscriminately or only for a specific action ID. In this case, the permission's "baseActionId" is the
+ *   action ID for scheduling a delay change, and the "specifier" is the action ID for which the delay will be changed.
+ *   The "baseActionId" and "specifier" of a permission are combined into a single "extended" `actionId`
+ *   by computing `keccak256(baseActionId, specifier)`.
```
nventuro, line 57:
```suggestion
 *   by calling `getExtendedActionId(baseActionId, specifier)`.
```

nventuro (approved):
This reads *much* better. I was originally skeptic on `_canPerformSpecificallyOrGenerally`, but after going through the source code it's great.

PR #1468, opened 2022-07-07
Move logic to revert on swap when paused to base contracts

EndymionJkb:
Add check to top-level onSwap, remove from derived contracts. Linear pools still need the modifier, as they implement the IGeneralPool interface, vs inheriting from it.

### DIFF
--- a/pkg/pool-linear/contracts/LinearPool.sol
+++ b/pkg/pool-linear/contracts/LinearPool.sol
@@ -395,7 +395,7 @@ abstract contract LinearPool is ILinearPool, IGeneralPool, IRateProvider, BasePo
         address recipient,
         uint256[] memory,
         bytes memory
-    ) internal view override whenNotPaused returns (uint256, uint256[] memory) {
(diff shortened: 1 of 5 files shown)


pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -395,7 +395,7 @@ abstract contract LinearPool is ILinearPool, IGeneralPool, IRateProvider, BasePo
         address recipient,
         uint256[] memory,
         bytes memory
-    ) internal view override whenNotPaused returns (uint256, uint256[] memory) {
```
nventuro, line 398: Linear Pools also pause on swaps

EndymionJkb: Yes; is something wrong here? onSwap still has the whenNotPaused modifier, because LinearPool doesn't inherit from BaseGeneralPool (only implements the interface). But it does in inherit from BasePool, and BasePool.onJoinPool blocks all joins on pause (including initialization), so we don't need it here

PR #1472, opened 2022-07-08
Ensure token cache rates updated before operations

EndymionJkb:
ensure the token rate cache is updated *before* calling `_scalingFactors`, which happens in the base contract Vault hooks during initial upscaling of balances, etc. Otherwise, the first transaction (swap/join/exit) after the cache period expired would still use the old rates.

### DIFF
--- a/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
@@ -234,6 +234,12 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
         return _bptIndex;
     }
 
+    /**
+     * @notice Top-level Vault hook for swaps.
+     * @dev Overriden here to ensure the token rate cache is updated *before* calling `_scalingFactors`, which happens
+     * in the base contract during upscaling of balances. Otherwise, the first transaction after the cache period
+     * expired would still use the old rates.
+     */
     function onSwap(
         SwapRequest memory swapRequest,
         uint256[] memory balances,
@@ -491,6 +497,12 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
         _payProtocolFees(protocolFeeAmount);
     }
 
+    /**
+     * @notice Top-level Vault hook for joins.
+     * @dev Overriden here to ensure the token rate cache is updated *before* calling `_scalingFactors`, which happens
+     * in the base contract during upscaling of balances. Otherwise, the first transaction after the cache period
+     * expired would still use the old rates.
+     */
     function onJoinPool(
         bytes32 poolId,
         address sender,
@@ -653,6 +665,12 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
         return (bptAmountOut, amountsIn);
     }
 
+    /**
+     * @notice Top-level Vault hook for exits.
+     * @dev Overriden here to ensure the token rate cache is updated *before* calling `_scalingFactors`, which happens
+     * in the base contract during upscaling of balances. Otherwise, the first transaction after the cache period
+     * expired would still use the old rates.
+     */
     function onExitPool(
         bytes32 poolId,
         address sender,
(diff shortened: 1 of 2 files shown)


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -699,6 +733,30 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
+        address recipient,
+        uint256[] memory balances,
+        uint256 lastChangeBlock,
+        uint256 protocolSwapFeePercentage,
+        bytes memory userData
+    ) public virtual override returns (uint256[] memory, uint256[] memory) {
+        // If this is a recovery mode exit, do not update the token rate cache: external calls might fail
+        if (!userData.isRecoveryModeExitKind()) {
+            _cacheTokenRatesIfNecessary();
+        }
```
nventuro, line 754: Lovely

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -1082,6 +1086,13 @@ describe('StablePhantomPool', () => {
         });
       });
 
+      const scaleRate = async (token: Token): Promise<BigNumber> => {
+        const index = tokens.indexOf(token);
+        const rateProvider = rateProviders[index];
+        const rate = await rateProvider.getRate();
+        return rate.mul(bn(10).pow(18 - token.decimals));
+      };
```
nventuro, line 1094: Why not call this `scalingFactor`?

nventuro: Ah I see it already existed. Still, we could rename this.

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -1322,6 +1326,166 @@ describe('StablePhantomPool', () => {
...
+            // This must be a real exit - otherwise the rate changes would revert along with the query
+            const newExitResult = await pool.singleExitGivenIn({ from: lp, bptIn, token });
+
+            // Verify the new rates are reflected in the scaling factors
+            await verifyScalingFactors(await pool.getScalingFactors());
+
+            expect(newExitResult.amountsOut[tokenIndex]).to.not.equal(previousExitResult.amountsOut[tokenIndex]);
+          });
+
+          it('recovery mode exits do not update the cache', async () => {
```
nventuro, line 1464: We should also have a test in which one of the rate providers reverts, but recovery mode exits still work.

nventuro (commented):
Fantastic work! Both the code changes and tests are spot-on :tada:

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -1322,6 +1326,166 @@ describe('StablePhantomPool', () => {
...
+            expect(await pool.getScalingFactors()).to.deep.equal(previousScalingFactors);
+
+            // And join again - amountIn should be different
+            // This must be a real join - otherwise the rate changes would revert along with the query
+            const newJoinResult = await pool.joinGivenOut({ from: lp, recipient: lp, bptOut, token });
+
+            // Verify the new rates are reflected in the scaling factors
+            await verifyScalingFactors(await pool.getScalingFactors());
+
+            expect(newJoinResult.amountsIn[tokenIndex]).to.not.equal(previousJoinResult.amountsIn[tokenIndex]);
```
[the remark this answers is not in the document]
nventuro: Did this in [ca7408c](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1472/commits/ca7408c1a806a574f9d8157e4203c31914043ed4)

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
EndymionJkb: We do the "with broken pool" test. It doesn't actually call the rate update and check that it reverts, though. Can add that.

nventuro: Ah that is true - I miss that we add that modifier to `_cacheTokenRateIfNecessary`. Nice.

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -1322,6 +1326,166 @@ describe('StablePhantomPool', () => {
...
+            // Verify the new rates are not yet loaded
+            const preOpScalingFactors = await pool.getScalingFactors();
+            for (let i = 0; i < preOpScalingFactors.length; i++) {
+              if (i != pool.bptIndex) {
+                expect(preOpScalingFactors[i]).to.equal(previousScalingFactors[i]);
+              }
+            }
+
+            // And swap again - amountOut should be different
+            // This must be a real swap - otherwise the rate changes would revert along with the query
```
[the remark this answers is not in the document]
EndymionJkb: I mean I can't check for the scaling factors afterward, yes

nventuro (approved):
Great work! :tada:

PR #1473, opened 2022-07-08
Add LinearPoolRebalancer

nventuro:
Marking as WIP as some things are still missing, mostly tests.

In terms of features, we might want to do an early return if the main balance is between the targets, unless a `force` value is passed. 

Also, it'd likely be a good idea to make it so the reward is always either the main token or the wrapped token, though this will require interacting with the wrapping entity to query the rate (so that we don't e.g. wrap the full amount).

It's also fairly simple to make this abstract and then specialize it for Aave the same way we did LinearPools, making this more reusable.

### DIFF
--- a/pkg/pool-linear/contracts/LinearPoolRebalancer.sol
+++ b/pkg/pool-linear/contracts/LinearPoolRebalancer.sol
@@ -56,11 +56,11 @@ contract LinearPoolRebalancer {
     }
 
     /**
-      * @notice Rebalance a Linear Pool from an asset manager, to maintain optimal operating conditions.
-      * @dev Use the asset manager mechanism to wrap/unwrap tokens as necessary to keep the main token
-      * balance as close as possible to the midpoint between the upper and lower targets: the fee-free zone
-      * where trading volume is highest.
-      */
+     * @notice Rebalance a Linear Pool from an asset manager, to maintain optimal operating conditions.
+     * @dev Use the asset manager mechanism to wrap/unwrap tokens as necessary to keep the main token
+     * balance as close as possible to the midpoint between the upper and lower targets: the fee-free zone
+     * where trading volume is highest.
+     */
     function rebalance() public {
         // The first thing we need to test is whether the Pool is below or above the target level, which will
         // determine whether we need to deposit or withdraw main tokens.
@@ -118,7 +118,7 @@ contract LinearPoolRebalancer {
     function _rebalanceExcessOfMainToken(uint256 excessMainAmount) private {
         // The Pool needs to reduce its main token balance, so we do a swap where we take the excess main token amount and send wrapped
         // tokens in exchange, that is, the main token is the token out. Since we know this amount, this is a 'given
-// out' swap.
+        // out' swap.
         IVault.SingleSwap memory swap = IVault.SingleSwap({
             poolId: _poolId,
             kind: IVault.SwapKind.GIVEN_OUT,
@@ -142,11 +142,14 @@ contract LinearPoolRebalancer {
         // Pool's `onSwap` function is `view`, this is irrelevant.
 
         _withdrawFromPool(_mainToken, excessMainAmount);
-        _wrapTokens(excessMainAmount);
+        // We're not going to wrap the full amount, only what is required to get `wrappedAmountIn` back. Any remaining
+        // main tokens will be transferred to the sender to refund the gas cost.
+        _wrapTokens(_getRequiredTokensToWrap(wrappedAmountIn));
         _depositToPool(_wrappedToken, wrappedAmountIn);
 
-        // Transfer any excess wrapped tokens to the sender
-        _wrappedToken.safeTransfer(msg.sender, _wrappedToken.balanceOf(address(this)));
+        // This contract will now hold excess main token, since we didn't wrap all that was withdrawn. These are sent to
+        // the caller to refund the gas cost.
+        _mainToken.safeTransfer(msg.sender, _mainToken.balanceOf(address(this)));
     }
 
     function _withdrawFromPool(IERC20 token, uint256 amount) private {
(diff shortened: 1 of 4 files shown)


pkg/pool-linear/contracts/LinearPoolRebalancer.sol
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    function _withdrawFromPool(IERC20 token, uint256 amount) private {
+        // Tokens can be withdrawn from the Vault with a 'withdraw' operation, but that will create 'managed' balance
+        // and leave the 'total' balance unchanged. We therefore have to perform two operations: one to withdraw, and
+        // another to clear the 'managed' balance (as the tokens withdrawn are about to be wrapped or unwrapped, and
+        // therefore lost to the Pool in their current format).
+        IVault.PoolBalanceOp[] memory withdrawal = new IVault.PoolBalanceOp[](2);
+
+        // First, we withdraw the tokens, creating 'managed' balance in the Pool.
```
EndymionJkb, line 154:
```suggestion
        // First, we withdraw the tokens, creating a non-zero 'managed' balance in the Pool.
```

pkg/pool-linear/contracts/LinearPoolRebalancer.sol
```diff
@@ -0,0 +1,209 @@
...
+        withdrawal[1].kind = IVault.PoolBalanceOpKind.UPDATE;
+        withdrawal[1].poolId = _poolId;
+        withdrawal[1].amount = 0;
+        withdrawal[1].token = token;
+
+        _vault.managePoolBalance(withdrawal);
+    }
+
+    function _depositToPool(IERC20 token, uint256 amount) private {
+        // Tokens can be deposited to the Vault with a 'deposit' operation, but that requires for prior 'managed'
```
EndymionJkb, line 170:
```suggestion
        // Tokens can be deposited to the Vault with a 'deposit' operation, but that requires a prior 'managed'
```

pkg/pool-linear/contracts/LinearPoolRebalancer.sol
```diff
@@ -0,0 +1,209 @@
...
+        deposit[1].poolId = _poolId;
+        deposit[1].amount = amount;
+        deposit[1].token = token;
+
+        _vault.managePoolBalance(deposit);
+    }
+
+    function _getDesiredMainTokenBalance() private view returns (uint256) {
+        // The desired main token balance is the midpoint of the lower and upper targets, which aims to maximize
+        // Pool swap volume by allowing for swaps in both directions with no rebalancing fees.
```
EndymionJkb, line 193:
```suggestion
        // The desired main token balance is the midpoint of the lower and upper targets. Keeping the balance
        // close to that value maximizes Pool swap volume by allowing zero-fee swaps in either direction.
```

pkg/pool-linear/contracts/LinearPoolRebalancer.sol
```diff
@@ -0,0 +1,209 @@
...
+
+        _pool = pool;
+        _poolId = pool.getPoolId();
+        _mainToken = pool.getMainToken();
+        _wrappedToken = pool.getWrappedToken();
+        _vault = vault;
+        _queries = queries;
+    }
+
+    function rebalance() public {
```
EndymionJkb, line 58:
```suggestion
    /**
      * @notice Rebalance a Linear Pool from an asset manager, to maintain optimal operating conditions.
      * @dev Use the asset manager mechanism to wrap/unwrap tokens as necessary to keep the main token
      * balance as close as possible to the midpoint between the upper and lower targets: the fee-free zone
      * where trading volume is highest.
      */
    function rebalance() public {
```

EndymionJkb:
Force rebalance (+ some renaming) would be something like:

```
    function rebalance(bool forceRebalance) public {
        // For a 3 token General Pool, it is cheaper to query the balance for a single token than to read all balances,
        // as getPoolTokenInfo will check for token existence, token balance and Asset Manager (3 reads), while
        // getPoolTokens will read the number of tokens, their addresses and balances (7 reads).
        // We can assume that the managed balance is zero (since we're the Pool's Asset Manager and we always set it to
        // zero), and work with the cash directly as if it were the total balance.
        (uint256 mainTokenBalance, , , ) = _vault.getPoolTokenInfo(_poolId, _mainToken);

        // The first thing we need to determine if whether the Pool is below or above the target level, which will
        // inform whether we need to deposit or withdraw main tokens.
        (uint256 targetMainTokenBalance, bool needsRebalancing) = _getTargetMainTokenBalance(mainTokenBalance);

        if (needsRebalancing || forceRebalance) {
            if (mainTokenBalance < targetMainTokenBalance) {
                _rebalanceMainTokenShortfall(targetMainTokenBalance - mainTokenBalance);
            } else {
                _rebalanceMainTokenExcess(mainTokenBalance - targetMainTokenBalance);
            }
        }
    }
...
    function _getTargetMainTokenBalance(uint256 mainTokenBalance)
        private
        view
        returns (uint256 targetMainTokenBalance, bool needsRebalancing)
    {
        // The desired main token balance is the midpoint of the lower and upper targets, which aims to maximize
        // Pool swap volume by allowing for swaps in both directions with no rebalancing fees.
        (uint256 lowerTarget, uint256 upperTarget) = _pool.getTargets();
        uint256 midpoint = (lowerTarget + upperTarget) / 2;

        // The targets are upscaled by the main token's scaling factor, so we undo that. Note that we're assuming that
        // the main token's scaling factor is constant.
        targetMainTokenBalance = FixedPoint.divDown(midpoint, _mainTokenScalingFactor);

        uint256 upscaledMainTokenBalance = FixedPoint.mulUp(mainTokenBalance, _mainTokenScalingFactor);
        needsRebalancing = upscaledMainTokenBalance > upperTarget || upscaledMainTokenBalance < lowerTarget;
    }
```

EndymionJkb:
Abstraction would be:

```
abstract contract LinearPoolRebalancer {
    ... 
    // No harm making all the immutables internal (need _wrappedToken at least)
    ILinearPool internal immutable _pool;
    bytes32 internal immutable _poolId;
    IERC20 internal immutable _mainToken;
    IERC20 internal immutable _wrappedToken;
    uint256 internal immutable _mainTokenScalingFactor;
    IVault internal immutable _vault;
    IBalancerQueries internal immutable _queries;
    ...
    // Override these in derived contracts to adapt to a specific protocol
    function _wrapTokens(uint256 amount) internal virtual;
    function _unwrapTokens(uint256 amount) internal virtual;
}

contract AAVELinearPoolRebalancer is LinearPoolRebalancer {
    constructor(
        ILinearPool pool,
        IVault vault,
        IBalancerQueries queries
    ) LinearPoolRebalancer(pool, vault, queries) {
    }

    function _wrapTokens(uint256 amount) internal override {
        IStaticAToken(address(_wrappedToken)).deposit(address(this), amount, 0, true);
    }

    function _unwrapTokens(uint256 amount) internal override {
        IStaticAToken(address(_wrappedToken)).withdraw(address(this), amount, true);
    }
}
```

TomAFrench:
> In terms of features, we might want to do an early return if the main balance is between the targets, unless a force value is passed.

I think that would be useful. We may want to encode this as a `minReward` argument as we should be able to calculate this relatively early in the transaction.

PR #1475, opened 2022-07-11
Add signature helper

nventuro:
This PR simply refactors existing code so that it all uses the same underlying helper. I only did minimal changes on the tests, to guarantee correctness. This PR will be followed up by one where the tests are migrated to the helper.

There's some changes:
 - the Vault is now slightly bigger, and requires a lower `runs` value to fit in 24kB
 - ERC20Permit (BPT) now has two extra functions (`getNextNonce` and `getDomainSeparator`), like all signature based contracts have (on top of the `nonce` and `DOMAIN_SEPARATOR` functions required by the EIP). A revert reason changed from `EXPIRED_PERMIT` to `EXPIRED_SIGNATURE`
 - a revert reason in BalancerMinter changed from `Invalid signature` to the `INVALID_SIGNATURE` error code

### DIFF
--- a/pkg/liquidity-mining/contracts/BalancerMinter.sol
+++ b/pkg/liquidity-mining/contracts/BalancerMinter.sol
@@ -22,8 +22,9 @@ import "@balancer-labs/v2-interfaces/contracts/liquidity-mining/ILiquidityGauge.
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/ReentrancyGuard.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/SafeMath.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/EIP712.sol";
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/EOASignaturesValidator.sol";
 
-contract BalancerMinter is IBalancerMinter, ReentrancyGuard, EIP712 {
+contract BalancerMinter is IBalancerMinter, ReentrancyGuard, EOASignaturesValidator {
     using SafeMath for uint256;
 
     IERC20 private immutable _token;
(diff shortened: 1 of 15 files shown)


pkg/liquidity-mining/contracts/BalancerMinter.sol
```diff
@@ -35,9 +36,6 @@ contract BalancerMinter is IBalancerMinter, ReentrancyGuard, EIP712 {
     // minter -> user -> can mint?
     mapping(address => mapping(address => bool)) private _allowedMinter;
 
-    // Signature replay attack prevention for each user.
-    mapping(address => uint256) internal _nextNonce;
-
     // solhint-disable-next-line var-name-mixedcase
     bytes32 private immutable _SET_MINTER_APPROVAL_TYPEHASH = keccak256(
```
TomAFrench, line 40:
```suggestion
    bytes32 private constant _SET_MINTER_APPROVAL_TYPEHASH = keccak256(
```
While we're touching these files, we can switch these to constant as immutable doesn't save us any gas here (which was the original justification afaik).

pkg/solidity-utils/contracts/openzeppelin/ERC20Permit.sol
```diff
@@ -17,9 +17,7 @@ import "./EIP712.sol";
  *
  * _Available since v3.4._
  */
-abstract contract ERC20Permit is ERC20, IERC20Permit, EIP712 {
-    mapping(address => uint256) private _nonces;
-
+abstract contract ERC20Permit is ERC20, IERC20Permit, EOASignaturesValidator {
     // solhint-disable-next-line var-name-mixedcase
     bytes32 private immutable _PERMIT_TYPEHASH = keccak256(
```
TomAFrench, line 22:
```suggestion
    bytes32 private constant _PERMIT_TYPEHASH = keccak256(
```

pkg/solidity-utils/contracts/test/ExtraCalldataEOASignaturesValidatorMock.sol
```diff
@@ -2,17 +2,17 @@
 
 pragma solidity ^0.7.0;
 
-import "../helpers/SignaturesValidator.sol";
+import "../helpers/ExtraCalldataEOASignaturesValidator.sol";
 
-contract SignaturesValidatorMock is SignaturesValidator {
+contract ExtraCalldataEOASignaturesValidatorMock is ExtraCalldataEOASignaturesValidator {
     bytes32 internal immutable AUTH_TYPE_HASH = keccak256(
```
TomAFrench, line 8:
```suggestion
    bytes32 internal constant AUTH_TYPE_HASH = keccak256(
```

TomAFrench (approved):
Looks good.

PR #1476, opened 2022-07-12
Proportional join should collect no protocol fee - test case

danielmkm:
Assuming a proportional join to a balanced stable phantom pool, the expectation would be that no protocol fee would be charged.

This test case shows that the ProtocolFeeCollector contains a BPT balance after a proportional join on a balanced StablePhantomPool, which is at odds with expectation.


nventuro:
Thanks a lot for this suggestion! We'll update the tests to include this scenario once we properly implement the new protocol fee handling.

EndymionJkb:
Incorporated into #1496. Thanks!

PR #1477, opened 2022-07-12
PhantomStable: protocol fee prep

EndymionJkb:
Introduce flags indicating whether a token's yield should be exempt from protocol fees. Handled the same way as rate providers, which seems to make the most sense, since the mapping key for the cache (which will be needed later) is a token, and it's parallel to the way everything else is handled.


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -116,6 +116,15 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
     IRateProvider internal immutable _rateProvider4;
     IRateProvider internal immutable _rateProvider5;
 
+    // Set true if the corresponding token should have its yield exempted from protocol fees.
+    // For example, the BPT of another PhantomStable Pool containing yield tokens.
+    bool internal immutable _protocolFeeExemptYieldToken0;
```
nventuro, line 121:
```suggestion
    bool internal immutable _exemptFromYieldProtocolFeeToken0;
```

nventuro, line 121: Since BPT will never be here, and this will always be used in BPT-less arrays (e.g. balances after dropBptItem), I'd make this array 5 tokens long and using the same indices as the balances array (i.e. without BPT).

Otherwise, we'll always have to call `dropBptItem` on it, leading to extra complexity and gas.

EndymionJkb: Went back and forth on that :)

EndymionJkb: I'd also considered that, but thought it was also confusing to have it use a different scheme than the rest of the variables (tradeoff with clarity). I will try it the other way and we'll see how it looks.

nventuro (approved):

PR #1480, opened 2022-07-13
StablePhantom: clarify tokenIndex on join/exit

EndymionJkb:
A case of pulling over tests and code from Stable without fully accounting for the new environment.

Swaps, with token in/out index values coming from the Vault, are always relative to the registered tokens. However, single token joins and exits have indices supplied by the user: these are generally relative to the token list *without* BPT, but it was a bit ambiguous. Clarify that by renaming variables.

The arrays being returned to the Vault must again be relative to the registered tokens (including BPT); the index must be adjusted for this as well.

Change tests to use the "last" token instead of the first.


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -424,7 +424,8 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
     }
 
     /**
-     * @dev Process a GivenIn swap involving BPT.
+     * @dev Process a GivenIn swap involving BPT. At this point, amount has been upscaled, and the BPT token
+     * has been removed from balances. `indexIn` and `indexOut` include BPT.
```
nventuro, line 428: Could we drop bpt from these? It is slightly wasteful as we only ever use only one of the two indices, but it's a cheap operation and will make this function more consistent (i.e. not have to be aware of bpt shenanigans).

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -634,7 +639,7 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
         bytes memory userData
     ) private returns (uint256, uint256[] memory) {
         (uint256[] memory amountsIn, uint256 minBPTAmountOut) = userData.exactTokensInForBptOut();
-        InputHelpers.ensureInputLengthMatch(_getTotalTokens() - 1, amountsIn.length);
+        InputHelpers.ensureInputLengthMatch(balances.length - 1, amountsIn.length);
```
nventuro, line 642: Let's add comments to these functions explainign that `balances` includes BPT.

nventuro, line 642: Why the removal of `_getTotalTokens()` throughout this PR?

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -1034,10 +1046,21 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
+    // into an array excluding BPT (usually from user input, such as amountsIn/Out).
+    // `index` should not be the BPT token index itself, or it would duplicate the preceeding index value.
     function _skipBptIndex(uint256 index) internal view returns (uint256) {
         return index < _bptIndex ? index : index.sub(1);
     }
 
+    // Convert from an index into an array excluding BPT (usually from user input, such as amountsIn/Out),
+    // to an index into an array excluding BPT (the Vault's registered token list).
+    // `index` should not be the BPT token index itself, if it is the last element, or this would return an
+    // invalid index.
```
nventuro, line 1059:
```suggestion
    // `index` must not be the BPT token index itself.
```
We could add a relatively cheap check for this.

nventuro (commented):
Great initiative! :raised_hands: Looking much better already

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
EndymionJkb: I'd thought we were trying to deprecate calling that: but maybe that was just for managed pools? I don't feel strongly. Can put them back if you prefer. (It does save a little bytecode, but we're not in trouble there yet.)

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
EndymionJkb: Added, along with tests. At least or the new "add" one, there is a case where we're passing it in from user input, so it could be sent in wrong.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
EndymionJkb: The trouble is we don't know which one we need until pretty late in the process. It makes it asymmetric with swaps (those call the base class, so we can't affect those), but I can do it on the swapWithBpt top level (which is symmetric with onRegularSwap)

EndymionJkb: Ok - next problem is this conflicts with the range checking (that skipBpt can't be given a BPT value). The code works now because we only call _skipBptIndex when it's not BPT. But if we call it in the top-level function, we know one of them is going to be BPT so it will fail.

Explaining and pre-calculating all that is probably worse than leaving it as is. (Or fixing it and not having the input range check on _skipBptIndex.)

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
nventuro: Hm I don't recall trying to remove that. I'd use `_getTotalTokens()` - it'll be cheaper and make intent more obvious.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
nventuro: Hm I think you're right. Let's keep as is then.

PR #1481, opened 2022-07-13
PhantomStable: Update stale comments

EndymionJkb:
Address some stale comments, mostly referring to not minting/burning BPT in PhantomStable pools, which is no longer the case.

(I swear I fixed some of these in other PRs.)

### DIFF
--- a/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
@@ -36,13 +36,13 @@ import "./StableMath.sol";
  * price ratio, such as Compound's cTokens.
  *
  * BPT is preminted on Pool initialization and registered as one of the Pool's tokens, allowing for swaps to behave as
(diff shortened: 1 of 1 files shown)


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -1088,8 +1087,11 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
     /**
      * @dev Returns the number of tokens in circulation.
      *
-     * In other pools, this would be the same as `totalSupply`, but since this pool pre-mints all BPT, `totalSupply`
-     * remains constant, whereas `getVirtualSupply` increases as users join the pool and decreases as they exit it.
+     * In other pools, this would be the same as `totalSupply`, but since this pool pre-mints BPT and holds it in the
+     * Vault as a token, we need to subtract the Vault's balance to get the total "circulating supply." Both the
```
nventuro, line 1091:
```suggestion
     * Vault as a token, we need to subtract the Vault's balance to get the total "circulating supply". Both the
```

Issue #1345, opened 2022-05-31
Only collect protocol fees on yield gains of some tokens


EndymionJkb:
Various issues arise with protocol fees and yield-bearing tokens. In general, fees are charged on the invariant increase - which arises both from swap fees *and* the intrinsic yield of the token. So, the StablePhantomPool bb-a-USD, for instance, would already be charging protocol fees on the AAVE yields associated with the underlying Linear Pools.

If we were then to pair bb-a-USD with another token (e.g., in a MetaStable or Weighted pool), and again charged fees based on the getRate() yield of the BPT, we could be charging fees twice.

Perhaps we could recognize this case somehow and use a rate of 1 for component pools? There also may be other issues with getRate() (e.g., sometimes returning 0), which should be explored and addressed as part of resolving this issue.

PR #1482, opened 2022-07-13
PhantomStable: Extend PriceRateCache

EndymionJkb:
This is the first PR on the road to correct protocol fee handling of nested yield tokens. We will need to store an extra rate value, so add that to the cache (adjusting sizes to fit), update all uses of it, and add tests.

I don't think it makes sense to try to use it yet, since other PRs will rearrange the cache calls (especially #1472), and it will be clearer then. Partially addresses #1345


pkg/pool-utils/contracts/rates/PriceRateCache.sol
```diff
@@ -23,55 +23,71 @@ import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
...
+    uint256 private constant _CURRENT_PRICE_RATE_OFFSET = 0;
+    uint256 private constant _POST_JOINEXIT_RATE_OFFSET = 96;
+    uint256 private constant _PRICE_RATE_CACHE_DURATION_OFFSET = 192;
+    uint256 private constant _PRICE_RATE_CACHE_EXPIRES_OFFSET = 224;
 
     /**
-     * @dev Returns the rate of a price rate cache.
+     * @dev Returns the current rate of a price rate cache.
      */
     function getRate(bytes32 cache) internal pure returns (uint256) {
```
nventuro, line 44: I'd rename this to `getCurrentRate`.

pkg/pool-utils/contracts/rates/PriceRateCache.sol
```diff
@@ -23,55 +23,71 @@ import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
  * useful for slow changing rates, such as those that arise from interest-bearing tokens (e.g. waDAI into DAI).
  *
  * The cache data is packed into a single bytes32 value with the following structure:
- * [   expires   | duration | price rate value ]
- * [   uint64    |  uint64  |      uint128     ]
- * [ MSB                                   LSB ]
+ * [ 32 bits |  32 bits  |   96 bits  |    96 bits    ]
+ * [ expires | duration  | last rate  | current rate  ]
```
nventuro, line 27:
```suggestion
 * [ expires | duration  | old rate  | current rate  ]
```

I'd rename this to 'old', and not make any assumptions here about how we're going to use it. We just have a function to update the current rate, and another to set the old rate to the current rate.

pkg/pool-utils/contracts/rates/PriceRateCache.sol
```diff
@@ -23,55 +23,71 @@ import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
...
      */
     function getRate(bytes32 cache) internal pure returns (uint256) {
-        return cache.decodeUint(_PRICE_RATE_CACHE_VALUE_OFFSET, 128);
+        return cache.decodeUint(_CURRENT_PRICE_RATE_OFFSET, 96);
+    }
+
+    /**
+     * @dev Returns the current rate of a price rate cache.
+     */
+    function getPostJoinExitRate(bytes32 cache) internal pure returns (uint256) {
```
nventuro, line 51: This would then be called `getOldRate`.

pkg/pool-utils/contracts/rates/PriceRateCache.sol
```diff
@@ -23,55 +23,71 @@ import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
...
+        return cache.decodeUint(_POST_JOINEXIT_RATE_OFFSET, 96);
+    }
+
+    /**
+     * @dev Sets the post join/exit rate in an existing cache record, and returns it.
+     */
+    function setPostJoinExitRate(bytes32 cache, uint256 rate) internal pure returns (bytes32) {
```
nventuro, line 58: I'd change this to `updateOldRate`, and have it equal the current rate.

pkg/pool-utils/contracts/rates/PriceRateCache.sol
nventuro: Essentially, one rate that is updated according to the cache expiry logic, and another that is updated according to arbitrary business logic.

pkg/pool-utils/contracts/rates/PriceRateCache.sol
```diff
@@ -23,55 +23,71 @@ import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
...
+
+        return cache.insertUint(rate, _POST_JOINEXIT_RATE_OFFSET, 96);
     }
 
     /**
      * @dev Returns the duration of a price rate cache.
      */
     function getDuration(bytes32 cache) internal pure returns (uint256) {
-        return cache.decodeUint(_PRICE_RATE_CACHE_DURATION_OFFSET, 64);
+        return cache.decodeUint(_PRICE_RATE_CACHE_DURATION_OFFSET, 32);
```
nventuro, line 68: Let's make these numbers be constants (e.g. DURATION_SIZE), and then build the offsets out of them (RATE_OFFSET = 0, OLD_RATE_OFFSET = RATE_OFFSET + RATE_DURATION, etc).

pkg/pool-utils/contracts/rates/PriceRateCache.sol
```diff
@@ -23,55 +23,71 @@ import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
...
+        _require(rate >> 96 == 0, Errors.PRICE_RATE_OVERFLOW);
 
         // solhint-disable not-rely-on-time
         return
-            WordCodec.encodeUint(rate, _PRICE_RATE_CACHE_VALUE_OFFSET, 128) |
-            WordCodec.encodeUint(duration, _PRICE_RATE_CACHE_DURATION_OFFSET, 64) |
-            WordCodec.encodeUint(block.timestamp + duration, _PRICE_RATE_CACHE_EXPIRES_OFFSET, 64);
+            WordCodec.encodeUint(rate, _CURRENT_PRICE_RATE_OFFSET, 96) |
+            WordCodec.encodeUint(duration, _PRICE_RATE_CACHE_DURATION_OFFSET, 32) |
+            WordCodec.encodeUint(block.timestamp + duration, _PRICE_RATE_CACHE_EXPIRES_OFFSET, 32);
```
nventuro, line 90: This would clear the 'old rate' field. Maybe we want to change this to `updateRate`, and instead of encoding these three we instead insert the expiration (now + duration) and the new rate.

nventuro (commented):
Great idea to start here. I made a couple suggestions to the PriceRateCache API - since that will end up affecting the tests I didn't pay much attention to those.

pkg/pool-utils/contracts/rates/PriceRateCache.sol
EndymionJkb: I saw the old code was like this, but we don't do this anywhere else we have these offset constants (e.g., ManagedPools). To me doing it that way seems brittle. If you change the order, you have to go through and fix all the arithmetic everywhere, vs just changing the offsets in one place and everything works.

I could see making the bitLengths constants.

nventuro: It'd be a change in a single place though, wouldn't it? Each field simply specifies the preceding one. But yeah, let's at least make the lenghts constant.

pkg/pool-utils/contracts/rates/PriceRateCache.sol
```diff
@@ -23,9 +23,9 @@ import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
  * useful for slow changing rates, such as those that arise from interest-bearing tokens (e.g. waDAI into DAI).
  *
  * The cache data is packed into a single bytes32 value with the following structure:
- * [ 32 bits |  32 bits  |   96 bits  |    96 bits    ]
- * [ expires | duration  | last rate  | current rate  ]
- *  |MSB                                           LSB|
+ * [ 32 bits |  32 bits  |  96 bits  |    96 bits    ]
+ * [ expires | duration  | old rate  | current rate  ]
+ *  |MSB                                          LSB|
```
nventuro, line 28: This should also be updated in StablePhantomPool, along with a (basic for now) explanation of what we use `old rate` for.

pkg/pool-utils/contracts/rates/PriceRateCache.sol
```diff
@@ -34,60 +34,66 @@ library PriceRateCache {
...
+        _require(rate >> _RATE_BIT_LENGTH == 0, Errors.PRICE_RATE_OVERFLOW);
 
         // solhint-disable not-rely-on-time
         return
-            WordCodec.encodeUint(rate, _CURRENT_PRICE_RATE_OFFSET, 96) |
-            WordCodec.encodeUint(duration, _PRICE_RATE_CACHE_DURATION_OFFSET, 32) |
-            WordCodec.encodeUint(block.timestamp + duration, _PRICE_RATE_CACHE_EXPIRES_OFFSET, 32);
+            cache
+                .insertUint(rate, _CURRENT_PRICE_RATE_OFFSET, _RATE_BIT_LENGTH)
+                .insertUint(duration, _PRICE_RATE_CACHE_DURATION_OFFSET, _DURATION_BIT_LENGTH)
```
nventuro, line 95: Since we're now inserting, we could not update the duration here, and have a separate `setDuration` function.

pkg/pool-utils/contracts/rates/PriceRateCache.sol
EndymionJkb: Ok; already did that.

pkg/pool-utils/contracts/rates/PriceRateCache.sol
EndymionJkb: We could, but the only place it's used sets both at the same time. I renamed to setRateAndDuration, so that it's accurate (especially since we have an update for oldRate, and it could be confusing that something called `setRate` had a side effect).

I also added separate updateCurrentRate and updateDuration functions, even though they're not used. (It's in a library, so no extra bytecode.) We seem to be using `update` rather than `set` in this library, so I kept that convention.

nventuro: Both are set at the same time, but it you look at the callsites, the duration is either read right before the call, or we're updating the duration and not the rate.

e.g.:

```
function updateTokenRateCache(IERC20 token) public virtual {
    IRateProvider provider = _getRateProvider(token);
    _require(address(provider) != address(0), Errors.TOKEN_DOES_NOT_HAVE_RATE_PROVIDER);
    uint256 duration = _tokenRateCaches[token].getDuration(); // <----
    _updateTokenRateCache(token, provider, duration);
}
```

```
function setTokenRateCacheDuration(IERC20 token, uint256 duration) external authenticate {
    IRateProvider provider = _getRateProvider(token);
    _require(address(provider) != address(0), Errors.TOKEN_DOES_NOT_HAVE_RATE_PROVIDER);
    _updateTokenRateCache(token, provider, duration); // <----- here we don't care about rate
    emit TokenRateProviderSet(token, provider, duration);
}
```

EndymionJkb: Yes, I could do this, using the new functions:

```
    function updateTokenRateCache(IERC20 token) public virtual {
        IRateProvider provider = _getRateProvider(token);
        _require(address(provider) != address(0), Errors.TOKEN_DOES_NOT_HAVE_RATE_PROVIDER);
        bytes32 cache = _tokenRateCaches[token];
        uint256 rate = provider.getRate();

        _tokenRateCaches[token] = cache.updateCurrentRate(rate);
        emit TokenRateCacheUpdated(token, rate);
    }

    function setTokenRateCacheDuration(IERC20 token, uint256 duration) external authenticate {
        IRateProvider provider = _getRateProvider(token);
        _require(address(provider) != address(0), Errors.TOKEN_DOES_NOT_HAVE_RATE_PROVIDER);
        bytes32 cache = _tokenRateCaches[token];

        _tokenRateCaches[token] = cache.updateDuration(duration);
        emit TokenRateProviderSet(token, provider, duration);
    }
```
But the tests fail then, because they expect `expires` to get updated based on the current timestamp of the update. `expires` still changes, even if duration stays the same.

The `itUpdatesTheCacheDuration` test also checks the rate, even though it supposedly only updates the duration. I'm not sure what the intent was here: are the tests wrong? *Should* it always update `expires` on every update, even if only the rate is changing?

nventuro: tbh this is a bit beside the point. Let's leave as-is for now and maybe revisit this later.

PR #1486, opened 2022-07-14
Let LinearPools have AssetManagers

nventuro:
Part of the work required to use and test https://github.com/balancer-labs/balancer-v2-monorepo/pull/1473


pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -150,6 +151,31 @@ abstract contract LinearPool is ILinearPool, IGeneralPool, IRateProvider, BasePo
...
+
+        address[] memory extendedAssetManagers = new address[](assetManagers.length + 1);
+        for (uint256 i = 0; i < extendedAssetManagers.length; ++i) {
+            if (i < bptIndex) {
+                extendedAssetManagers[i] = assetManagers[i];
+            } else if (i > bptIndex) {
+                extendedAssetManagers[i] = assetManagers[i - 1];
+            } else {
+                extendedAssetManagers[i] = address(0);
+            }
```
TomAFrench, line 173:
```suggestion
            }
```

optimizoor

TomAFrench (approved):
It's mildly yucky that we're mixing the explicit token naming scheme (`mainToken`/ `wrappedToken`) with the array based mechanism (`assetManagers`) but I imagine we'd run into stack issues otherwise.

pkg/pool-linear/contracts/LinearPool.sol
nventuro: :joy: given this happens only once in the constructor, I'd rather be extra explicit

PR #1489, opened 2022-07-15
Yield Fees for Protocol in Weighted Pools

rabmarut:
This is a preliminary implementation of yield-based protocol fees for weighted pools. In addition to collecting protocol fees on swaps, weighted pools now have rate providers and collect additional protocol fees on token rate growth. The rate providers are used only to compute fees and do not inform any swap/join/exit math in any way.

This is still a draft because the mechanism requires peer verification and the tests needs updating. Notably, the existing weighted pool tests need to be fixed to include rate providers, and more explicit tests of the yield fee mechanism should be added.

Also, it seems we've hit the byte code size limit for weighted pools...


pkg/pool-weighted/contracts/WeightedMath.sol
```diff
@@ -73,6 +73,20 @@ library WeightedMath {
         _require(invariant > 0, Errors.ZERO_INVARIANT);
     }
 
+    // Computes a weighted product (same math as invariant) while ignoring ones.
+    function _calculateWeightedProduct(uint256[] memory normalizedWeights, uint256[] memory balances)
```
gerrrg, line 77: Was it a deliberate decision to call the 2nd argument `balances`? If so, why? If not, can we change this to `rates`?

rabmarut: Although it's only used in one place for now, this is meant to be a generic function. That's why it's called `_calculateWeightedProduct` instead of, for example, `_calculateYield`.

But being generic, `balances` also no longer makes sense (it was blindly copied from the signature of `_calculateInvariant`). What's a good generic name?

pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
```diff
@@ -25,6 +27,57 @@ abstract contract InvariantGrowthProtocolFees is BaseWeightedPool {
...
+    IRateProvider internal immutable _rateProvider13;
+    IRateProvider internal immutable _rateProvider14;
+    IRateProvider internal immutable _rateProvider15;
+    IRateProvider internal immutable _rateProvider16;
+    IRateProvider internal immutable _rateProvider17;
+    IRateProvider internal immutable _rateProvider18;
+    IRateProvider internal immutable _rateProvider19;
+
+    constructor(IRateProvider[] memory rateProviders) {
+        uint256 numTokens = rateProviders.length;
```
nventuro, line 55: This is insufficient: we much check that `rateProviders` has the same lenght as the `tokens` array. Perhaps we could pass a `totalTokens` value to this constructor, and call it with e.g. `tokens.length`.

nventuro: (note that we cannot access `_getTotalTokens()` during construction)

pkg/pool-weighted/contracts/WeightedMath.sol
nventuro: `amounts`

pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
```diff
@@ -51,14 +241,35 @@ abstract contract InvariantGrowthProtocolFees is BaseWeightedPool {
...
             protocolSwapFeePercentage
         );
 
-        _payProtocolFees(toMint);
+        uint256 yieldFees = 0;
+        if (_lastRateGrowthProduct > 0) {
+            uint256 rateGrowthProduct = WeightedMath._calculateWeightedProduct(normalizedWeights, _getRates());
+
+            // Only collect protocol fees when yield growth exceeds previous all-time high.
+            if (rateGrowthProduct > _lastRateGrowthProduct) {
```
nventuro, line 257: Let's explicitly call it the ATH, instead of calling it 'the last one', but only updating it if we ever find a larger value

pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
```diff
@@ -51,14 +241,35 @@ abstract contract InvariantGrowthProtocolFees is BaseWeightedPool {
...
+                    rateGrowthProduct,
+                    // TODO: This fee pct should come from a different source.
+                    protocolSwapFeePercentage
+                );
+
+                _lastRateGrowthProduct = rateGrowthProduct;
+            }
+        } else {
+            _lastRateGrowthProduct = WeightedMath._calculateWeightedProduct(normalizedWeights, _getRates());
+        }
```
nventuro, line 270: It this because we need to measure growth relative to it? I'd add a comment here explaining that, and also mentioning that this will happen during pool initialization, so it's not like there's any unpaid fees.

nventuro: I'd also swap the order of the `if` and `else` bodies, for clarity.

pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
```diff
@@ -51,14 +241,35 @@ abstract contract InvariantGrowthProtocolFees is BaseWeightedPool {
...
 
-        _payProtocolFees(toMint);
+        uint256 yieldFees = 0;
+        if (_lastRateGrowthProduct > 0) {
+            uint256 rateGrowthProduct = WeightedMath._calculateWeightedProduct(normalizedWeights, _getRates());
+
+            // Only collect protocol fees when yield growth exceeds previous all-time high.
+            if (rateGrowthProduct > _lastRateGrowthProduct) {
+                yieldFees = WeightedMath._calcDueProtocolSwapFeeBptAmount(
+                    supply.add(swapFees),
```
nventuro, line 259: I'd similarly like to see some explanation as to why it's correct to charge fees on the total supply including unminted BPT.

pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
nventuro, line 257: Reading the variable to the stack would likely save an sload.

pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
rabmarut: Yeah, I included that in the `WeightedPool` constructor because of the access issues

pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
```diff
@@ -33,6 +86,143 @@ abstract contract InvariantGrowthProtocolFees is BaseWeightedPool {
...
+     * @dev Returns the token rates for all tokens. These are fixed-point values with 18 decimals.
+     * In case there is no rate provider for the provided token it returns FixedPoint.ONE.
+     */
+    function _getRates() internal view returns (uint256[] memory rates) {
+        uint256 totalTokens = _getTotalTokens();
+        rates = new uint256[](totalTokens);
+
+        // prettier-ignore
+        {
+            rates[0] = _rateProvider0 == IRateProvider(0) ? FixedPoint.ONE : _rateProvider0.getRate();
```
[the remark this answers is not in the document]
rabmarut: Good call. Still over bytecode limit for now, but I made this change

pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
rabmarut: Would you prefer passing the length into this constructor and performing the check here?

pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
rabmarut: Correct. This variable is used to measure relative growth, so the check ensures that we don't take a fee before initialization. As you point out, this will be initialized during the INIT join, but the check is wasteful on every subsequent loop.

I marked a TODO in the constructor to consider initializing there instead, but it would be very messy given that we can't call `_getRates` and `_getNormalizedWeights` in the constructor.

Do you think it's worth exploring that option or should we just leave this here, add the comment, and flip the if/else order?

pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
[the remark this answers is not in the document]
nventuro: Did my suggestion help with the overall size?

pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
nventuro: Yes, I think that'd be less error prone. You can't forget to pass a constructor argument.

pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
rabmarut: After further analysis, it seems that it is not correct, so we will simply remove this.

TomAFrench:
Closing this PR as it's been superceded by #1640

PR #1496, opened 2022-07-19
PhantomStable: incorporate yield exemptions

### LINKED ISSUE #1345: Only collect protocol fees on yield gains of some tokens

EndymionJkb:
Start using the previously added _exemptFromYieldProtocolFeeToken flags.

1) Initialize the OldRate for relevant tokens on construction (OldRate = CurrentRate)
    Even though they're used with arrays not including BPT, it seemed a lot easier/cleaner to calculate them including BPT and _skipBptItem before use, rather than have a lot of index calculations (mainly because we can't have immutable arrays).

2) Before any join or exit, "rescale" the balances so that exempt tokens use the OldRate, and any other tokens use the CurrentRate, to compute the "preJoinOrExit" invariant. Protocol fees are then due on the growth between the post- and pre- invariants.

3) Update OldRates for exempt tokens after any join/exit: basically anything but a regular swap. Also compute and store the postJoinExit invariant (and amp), using the post join/exit balances.

It's a draft because the protocol fee tests are just skipped for now, since they will need to be rewritten.

The diff is horrendous, and not really reviewable that way. Accordingly, here's a guide:

1) _exemptFromYieldProtocolFeeTokens now include BPT (though the one corresponding to the BPT will be false), because it was much easier than converting between hard-coded names and array indices (without an immutable array).

2) In the constructor, we set the _exempt... immutables, and as part of initializing the rate provider cache, also set the OldRate for exempt tokens (just in case they are accessed before the first join)

3) _onInitialize already calculates the "postJoin" invariant, so it just directly sets the _postJoinExitInvariant. It also updates the OldRates, in case there was a delay between creation and initialization.

4) Swaps: _onRegularSwap (between non-BPT tokens) does nothing with protocol fees, saving lots of gas.
     _onSwapWithBpt calls _payProtocolFeesBeforeJoinExit() with the pre-swap balances (which get rescaled, as described above). 
     _onSwapBptGivenIn and _onSwapBptGivenOut compute the post-join/exit balances according to what's happening, and then call _updateInvariantAfterJoinExit(), to compute and set the _postJoinExitInvariant (and amp).

5) Joins: _onJoinPool calls _payProtocolFeesBeforeJoinExit() with the preJoin balances, and the two join handlers compute the postJoin balances and call _updateInvariantAfterJoinExit().

6) Exits: Same as joins. _onExitPool calls _payProtocolFeesBeforeJoinExit() with the preExit balances, and the two exit handlers compute the postExit balances and call _updateInvariantAfterJoinExit().

Closes #1345


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -984,6 +1059,60 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
+
+        return rateRatios;
+    }
+
+    /**
+     * @dev Apply the token ratios to a set of balances (without BPT), to adjust for any exempt yield tokens.
+     * Mutates the balances in place.
+     */
+    function _adjustBalancesByTokenRatios(uint256[] memory balancesWithoutBpt) internal view {
+        uint256[] memory ratios = _dropBptItem(_tokenRateRatios());
```
TomAFrench, line 1108:
```suggestion
        // `_tokenRateRatios()` includes an item for BPT so to match the balances array we must drop that item.
        uint256[] memory ratios = _dropBptItem(_tokenRateRatios());
```

It took me a moment to process we're not including an off-by-one error here.

TomAFrench (commented):
Agreed that adding a dummy BPT entry for the exemption flags seems like the cleanest way to go.

I've made a couple of notes on things I noticed unrelated to the reasons for the PR being a draft.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -984,6 +1059,60 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
+        for (uint256 i = 0; i < balancesWithoutBpt.length; ++i) {
+            if (ratios[i] != FixedPoint.ONE) {
+                balancesWithoutBpt[i] = FixedPoint.mulDown(balancesWithoutBpt[i], ratios[i]);
+            }
```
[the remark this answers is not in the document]
EndymionJkb: Went back and forth on that. It depends on how many tokens are in the pool, and how many are yield exempt. There can't be more than 5 tokens, but I don't have a feel for how many would be exempt. Would most pools that use this be "pools of pools" where all of them are exempt?

Anyway we have bytecode issues and it saves 40 bytes, so for now at least, I'll remove the check.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -1017,6 +1146,54 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
+            else {
+                _revert(Errors.INVALID_TOKEN);
+            }
+        }
+    }
+
+    function _updateOldRatesAfterJoinExit() internal {
+        (IERC20[] memory tokens, , ) = getVault().getPoolTokens(getPoolId());
+
+        // Since we are getting the tokens from the Vault, we know they are all valid.
```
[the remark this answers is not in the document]
EndymionJkb: Yes - no benefit: I somehow forgot we also had immutable tokens lol.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -259,15 +263,16 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
         _exemptFromYieldProtocolFeeToken0 = exemptFromYieldProtocolFeeFlags[0];
         _exemptFromYieldProtocolFeeToken1 = exemptFromYieldProtocolFeeFlags[1];
-        _exemptFromYieldProtocolFeeToken2 = (exemptFromYieldProtocolFeeFlags.length > 2)
-            ? exemptFromYieldProtocolFeeFlags[2]
-            : false;
-        _exemptFromYieldProtocolFeeToken3 = (exemptFromYieldProtocolFeeFlags.length > 3)
+        _exemptFromYieldProtocolFeeToken2 = exemptFromYieldProtocolFeeFlags[2];
+        _exemptFromYieldProtocolFeeToken3 = (tokensAndBPTRateProviders.length > 3)
             ? exemptFromYieldProtocolFeeFlags[3]
             : false;
```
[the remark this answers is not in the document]
EndymionJkb: Take it up with lint! :) Yes, I shortened the variable names: they're a little excessive; already descriptive enough.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
EndymionJkb: I also renamed to ratiosWithoutBpt for extra clarity.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -232,51 +247,39 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
         // reference them by token index in the full base tokens plus BPT set (i.e. the tokens the Pool registers). Due
         // to immutable variables requiring an explicit assignment instead of defaulting to an empty value, it is
         // simpler to create a new memory array with the values we want to assign to the immutable state variables.
-        IRateProvider[] memory tokensAndBPTRateProviders = new IRateProvider[](params.tokens.length + 1);
-        for (uint256 i = 0; i < tokensAndBPTRateProviders.length; ++i) {
+        IRateProvider[] memory tokensAndRateProviders = new IRateProvider[](params.tokens.length + 1);
```
TomAFrench, line 250:
```suggestion
        IRateProvider[] memory allTokensRateProviders = new IRateProvider[](params.tokens.length + 1);
```

The current name is a bit weird like the array contains both token and rate provider addresses.

EndymionJkb: It's been changing. The really long names were causing lint/readability issues.

I think the "tokens" part of it was referring to the length: like it was the size of both the tokens and rate providers. Now that we have three things, it's more confusing. Maybe just rateProviders.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -232,42 +247,36 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
-        for (uint256 i = 0; i < tokensAndBPTRateProviders.length; ++i) {
+        IRateProvider[] memory rateProviders = new IRateProvider[](params.tokens.length + 1);
+        // Do the same with exemptFromYieldProtocolFeeFlags
+        bool[] memory exemptFromYieldFlags = new bool[](rateProviders.length);
+
+        for (uint256 i = 0; i < rateProviders.length; ++i) {
             if (i < bptIndex) {
-                tokensAndBPTRateProviders[i] = params.rateProviders[i];
+                rateProviders[i] = params.rateProviders[i];
+                exemptFromYieldFlags[i] = params.exemptFromYieldProtocolFeeFlags[i];
```
TomAFrench, line 257:
```suggestion
                rateProviders[i] = params.rateProviders[i];
                exemptFromYieldFlags[i] = params.exemptFromYieldProtocolFeeFlags[i] && rateProviders[i] != IRateProvider(0) ;
```

I think it's possible to lock funds if you pass a zero rate provider but say it's exempt from yield protocol fees (a not totally unreasonable thing to do).

If you did this then `_getTokenRateRatios` will attempt to compute the rate ratio for an uninitialised price rate cache. We'll then try to calculate `0 / 0` and everything breaks.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -232,42 +247,36 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
             } else if (i == bptIndex) {
-                tokensAndBPTRateProviders[i] = IRateProvider(0);
+                rateProviders[i] = IRateProvider(0);
             } else {
-                tokensAndBPTRateProviders[i] = params.rateProviders[i - 1];
+                rateProviders[i] = params.rateProviders[i - 1];
+                exemptFromYieldFlags[i] = params.exemptFromYieldProtocolFeeFlags[i - 1];
```
TomAFrench, line 262:
```suggestion
                rateProviders[i] = params.rateProviders[i - 1];
                exemptFromYieldFlags[i] = params.exemptFromYieldProtocolFeeFlags[i - 1] && rateProviders[i-1] != IRateProvider(0);
```

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
EndymionJkb: Good catch! We'll have to test every combination of everything, and every ordering of operations.

EndymionJkb: Maybe even disallow this case? Aren't the exempt tokens normally composed BPT of other pools, which would by definition have a rate function? If it didn't have a rate function, the rate would always be 1 anyway, so there couldn't be a problem, and there'd be no reason to set the exempt flag.

If it gets passed in this way, it seems more likely to be a mistake. If they mixed up the order of the flags - but it accidentally worked - it would just be silently wrong.

TomAFrench: On second thoughts this wouldn't have resulted in loss of funds due to recovery mode (proving how good of an idea it is).

> If it gets passed in this way, it seems more likely to be a mistake. If they mixed up the order of the flags - but it accidentally worked - it would just be silently wrong.

True, we should err on the side of being more restrictive when it comes to constructor arguments. It's better to reject them than try and "correct" them but fail.

pkg/pool-stable-phantom/test/StablePhantomPool.test.ts
```diff
@@ -76,6 +76,8 @@ describe('StablePhantomPool', () => {
     const tokenRateCacheDurations: number[] = [];
     const exemptFromYieldProtocolFeeFlags: boolean[] = [];
 
+    const ZEROS = Array(numberOfTokens + 1).fill(bn(0));
+
     async function deployPool(
```
TomAFrench, line 81: Something I noticed here is that we're deploying a rate for each token so we never test the case where a token doesn't have a rate provider to call.

EndymionJkb: Yes; addressing as part of the test I'm adding for the 0 rate provider with exempt flag set

PR #1499, opened 2022-07-20
StableMath: remove unused method for token protocol fees

EndymionJkb:
Don't think we need this anymore.


TomAFrench (approved):
We can fish it out of the git history if not.

PR #1500, opened 2022-07-20
Add batch relayer v3 fork tests

nventuro:
Not as good as full unit tests, but still very useful to reference, plus also a sanity check of correctness on live Pools.


nventuro:
@TomAFrench I think you'll enjoy 20a550b07a1e34929d67170f31ae4ab16def1450

TomAFrench:
> @TomAFrench I think you'll enjoy [20a550b](https://github.com/balancer-labs/balancer-v2-monorepo/commit/20a550b07a1e34929d67170f31ae4ab16def1450)

:brain:

PR #1510, opened 2022-07-21
Deploy gauge adder coordinator

nventuro:


pkg/deployments/action-ids/mainnet/action-ids.json
```diff
@@ -664,5 +664,14 @@
...
+  },
+  "20220721-gauge-adder-migration-coordinator": {
+    "GaugeAdderMigrationCoordinator": {
+      "useAdaptor": false,
+      "actionIds": {
+        "performNextStage()": "0x945b1637627fa36f3bb618b74050045ffa67875ffd8acd04f5135f3463ab9880",
+        "registerStages()": "0x08900d9a13d156ace75a5694772406e876e8f8f25d35af7a7639c1a95a2c2689"
+      }
+    }
   }
```
TomAFrench, line 676: Should we just omit all the coordinators from here? I don't see any real benefit to having them included.

TomAFrench (approved):
Looks good.

pkg/deployments/action-ids/mainnet/action-ids.json
nventuro: Eh that's a good point. I just run the task and didn't think about it.

nventuro: Tbh a better solution would be to have non `getActionId` contracts (i.e. those that go through the authorizer) only be added to the json if they're in some list of pre-approved contracts (gauges, etc.), as the vast majority are false positives.

PR #1514, opened 2022-07-21
Reorder functions in PhantomStablePool

TomAFrench:
This PR should be a noop as it's just moving functions around within the PhantomStablePool contract.

I'd recommend going through each commit one by one as I've moved all the functions on a particular topic together in each commit. Once that's done it would be good to take a look at the contract as a whole to get a feel for the flow.

Generally in each section we should have all getter functions sitting above the relevant setter functions (see Amplification section for an example)

### DIFF
--- a/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
@@ -278,6 +278,17 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
         return _bptIndex;
     }
 
+    function _getMaxTokens() internal pure override returns (uint256) {
+        // The BPT will be one of the Pool tokens, but it is unaffected by the Stable 5 token limit.
+        return StableMath._MAX_STABLE_TOKENS + 1;
+    }
+
+    function _getTotalTokens() internal view virtual override returns (uint256) {
+        return _totalTokens;
+    }
+
+    // Swap Hooks
+
     /**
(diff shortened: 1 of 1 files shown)


EndymionJkb (commented):
I didn't look at this in detail, but I pretty much did this already in #1496. Maybe this is redundant? Or we just have dueling reorderings :)

TomAFrench:
> I didn't look at this in detail, but I pretty much did this already in #1496. Maybe this is redundant? Or we just have dueling reorderings :)

I'm not sure, I'll need to have another run through. I'd be inclined to say that if #1496 performs a similar reordering then we should merge this first so that we can minimise the diff over there to help review (as we'll just be looking at functionality changes rather than layout).

nventuro (approved):
Big fan of the new order! The only thing I'd consider changing is having joins -> exits -> swaps, as swaps are rather involved, can be pseudo joins, etc., and it'd be useful to have the initialization code (preminting etc.) at the beginning.

Not sure when we should merge due to likely conflicts with #1496

TomAFrench:
>The only thing I'd consider changing is having joins -> exits -> swaps, as swaps are rather involved, can be pseudo joins, etc., and it'd be useful to have the initialization code (preminting etc.) at the beginning.

Yeah, I'd be happy with that as well. It felt weird to have protocol fees code in the middle of the swap/join/exit hooks but I also didn't want to move them too far from where they're used. This solves both of these issues.

TomAFrench:
Reverting the last commit to simplify merging with #1496

PR #1522, opened 2022-07-22
Abstract away boilerplate in CI into a custom action

TomAFrench:
This PR combines installing node and cached installs of dependencies into a single "setup" action so we don't need have these steps repeated in each job.


nventuro (approved):
Lovely

PR #1525, opened 2022-07-23
Add ProtocolFeePercentagesProvider

nventuro:
This new extensible component will hold all protocol fees, and should be the preferred solution when querying and altering them. Swap and flash loan fees can still be queried and set via the ProtocolFeesCollector (as this contract fallbacks to it for those), but ideally we'd just work with this new one for consistency and simplicity.

By making fee types immutable (except for their current value), I believe this will suffice for all expected use cases.

Note that there's a single function to set fee values, which means a single permission will allow setting fees for any type. In the future, we expect to expand our action id system to be more granular when it comes to action parameters, and make a contract with more fine-tuned access control the only account that can call `setFeeTypePercentage`.

### DIFF
--- a/pkg/standalone-utils/contracts/ProtocolFeePercentagesProvider.sol
+++ b/pkg/standalone-utils/contracts/ProtocolFeePercentagesProvider.sol
@@ -86,7 +86,7 @@ contract ProtocolFeePercentagesProvider is IProtocolFeePercentagesProvider, Sing
         uint256 maximumValue,
         uint256 initialValue
     ) private {
-        require(maximumValue <= _MAX_PROTOCOL_FEE_PERCENTAGE, "Invalid maximum fee percentage");
+        require((maximumValue > 0) && (maximumValue <= _MAX_PROTOCOL_FEE_PERCENTAGE), "Invalid maximum fee percentage");
         require(initialValue <= maximumValue, "Invalid initial percentage");
 
         _feeTypeData[feeType] = FeeTypeData({
--- a/pkg/standalone-utils/contracts/ProtocolFeePercentagesProvider.sol
+++ b/pkg/standalone-utils/contracts/ProtocolFeePercentagesProvider.sol
@@ -1,0 +1,161 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
+
+import "@balancer-labs/v2-interfaces/contracts/vault/IProtocolFeesCollector.sol";
+import "@balancer-labs/v2-interfaces/contracts/standalone-utils/IProtocolFeePercentagesProvider.sol";
+
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/SingletonAuthentication.sol";
+
+contract ProtocolFeePercentagesProvider is IProtocolFeePercentagesProvider, SingletonAuthentication {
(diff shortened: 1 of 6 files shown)


pkg/standalone-utils/contracts/ProtocolFeePercentagesProvider.sol
```diff
@@ -0,0 +1,161 @@
...
+
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/SingletonAuthentication.sol";
+
+contract ProtocolFeePercentagesProvider is IProtocolFeePercentagesProvider, SingletonAuthentication {
+    IProtocolFeesCollector private immutable _protocolFeesCollector;
+
+    struct FeeTypeData {
+        uint256 value;
+        uint256 maximum;
+        bool registered;
```
TomAFrench, line 29:
```suggestion
```

I'd remove this and determine fee type existence from from `feeType.maximum > 0` (We'd need to enforce this on `registerFeeType`). There's not valid reason to create a fee type with a zero maximum value so this doesn't limit us.

pkg/standalone-utils/contracts/ProtocolFeePercentagesProvider.sol
```diff
@@ -0,0 +1,161 @@
...
+import "@balancer-labs/v2-interfaces/contracts/standalone-utils/IProtocolFeePercentagesProvider.sol";
+
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/SingletonAuthentication.sol";
+
+contract ProtocolFeePercentagesProvider is IProtocolFeePercentagesProvider, SingletonAuthentication {
+    IProtocolFeesCollector private immutable _protocolFeesCollector;
+
+    struct FeeTypeData {
+        uint256 value;
+        uint256 maximum;
```
TomAFrench, line 28:
```suggestion
        uint128 value;
        uint128 maximum;
```

The struct packoooor

(we technically only need a `uint64` for this.)

pkg/standalone-utils/test/ProtocolFeePercentagesProvider.test.ts
```diff
@@ -0,0 +1,470 @@
...
+    it('reverts if the maximum aum value is too high', async () => {
+      await expect(
+        deploy('ProtocolFeePercentagesProvider', {
+          args: [vault.address, 0, fp(1).add(1)],
+        })
+      ).to.be.revertedWith('Invalid maximum fee percentage');
+    });
+
+    it('emits ProtocolFeeTypeRegistered events for custom types', async () => {
+      // We deploy manually instead of using our helper to get the transaction receipt
```
TomAFrench, line 61: You can read this off from the returned contract using the `deployTransaction` field. We use this in the `deployments` package:

https://github.com/balancer-labs/balancer-v2-monorepo/blob/77efc90be71c5e44434287e67b1b58d037241f01/pkg/deployments/src/task.ts#L123

nventuro: Ahh nice, thank you!

nventuro: Note that we have two `deploy` helpers, which seem to work slightly differently (in one we return the object created by ethers, in the other we discard it and just keep the address to then create a new contract). I'll keep the tests as-is - we can merge the helpers later down the line.

pkg/standalone-utils/contracts/ProtocolFeePercentagesProvider.sol
nventuro: I added a check for maximum > 0 (as that'd be a mistake), but instead of removing `registered` I packed all three values into a single slot. I prefer that, as we don't need to have too clever logic, nor worry about native fee types (which don't have max or value entries).

TomAFrench (commented):
We should add an `isValidFeeType` which returns `_feeTypeData[feeType].registered`. Currently there's no way to know if a fee type exists without potential for a revert.

It also wouldn't be too bad to have an `EnumerableSet` for valid fee types so they're discoverable without having to set up a subgraph.

nventuro:
>We should add an isValidFeeType which returns _feeTypeData[feeType].registered. Currently there's no way to know if a fee type exists without potential for a revert.

Fair, I added this.

>It also wouldn't be too bad to have an EnumerableSet for valid fee types so they're discoverable without having to set up a subgraph.

I considered this but don't want to make the data storage too fancy, nor do I want to add another read when querying data (`.contains()` and the data read). Given we'll only have a couple, I think it's fine to leave this as-is.

nventuro:
(also I don't think this contract is worth spending more time into)

PR #1526, opened 2022-07-25
pull in BoostV2 contract from Curve

TomAFrench:
This PR imports the Boost2 contract from https://github.com/curvefi/curve-veBoost/blob/db3dec43b6e4fac0fca1f01509f9133563f43ebb/contracts/BoostV2.vy

I've got a couple of modifications I'd like to make but this PR aims to be just show verifiably that  we're starting from a sound starting point.

### DIFF
--- a/pkg/liquidity-mining/contracts/BoostV2.vy
+++ b/pkg/liquidity-mining/contracts/BoostV2.vy
@@ -1,0 +1,374 @@
+# @version 0.3.3
+"""
+@title Boost Delegation V2
+@author CurveFi
+"""
+
+
+event Approval:
+    _owner: indexed(address)
+    _spender: indexed(address)
+    _value: uint256
+
+event Transfer:
+    _from: indexed(address)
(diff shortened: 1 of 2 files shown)


nventuro (approved):
I've checked that the code matches exactly the one deployed at https://etherscan.io/address/0xd0921691c7debc698e6e372c6f74dc01fc9d3778#code, which was set as the delegation [here](https://etherscan.io/tx/0x2427765302e933e209805936877f5efe9f918d61bd7403052ef4923f8285ed78).

Bizarrely, another delegation change happened ~10 hours before that one, see [this transaction](https://etherscan.io/tx/0xcf3282eb7c5bb84af66d2848c511db55c96271db94061249deaa779357809480), to a different looking contract. I am not sure what's up with that.

TomAFrench:
Our version of veBoost2 | Mystery contract

image
image

The interpolatoooor

PR #1527, opened 2022-07-25
Allow veBoost2 to migrate multiple boosts in a single call

TomAFrench:
This adds a simple function to avoid having to migrate boosts one by one.

### DIFF
--- a/pkg/liquidity-mining/contracts/BoostV2.vy
+++ b/pkg/liquidity-mining/contracts/BoostV2.vy
@@ -231,6 +231,16 @@ def boost(_to: address, _amount: uint256, _endtime: uint256, _from: address = ms
 
 @external
 def migrate(_token_id: uint256):
+    self._migrate(_token_id)
+
+@external
+def migrate_many(_token_ids: uint256[16]):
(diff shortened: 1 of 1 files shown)


pkg/liquidity-mining/contracts/BoostV2.vy
```diff
@@ -231,6 +231,16 @@ def boost(_to: address, _amount: uint256, _endtime: uint256, _from: address = ms
 
 @external
 def migrate(_token_id: uint256):
+    self._migrate(_token_id)
+
+@external
+def migrate_many(_token_ids: uint256[16]):
+    for i in range(16):
+        if (_token_ids[i] == 0) break
```
nventuro, line 239: The language of the future

nventuro: Also, this works because token ids can never be zero as they include the delegator address. See [line 609 of the preseeded delegation here](https://etherscan.io/address/0xB496FF44746A8693A060FafD984Da41B253f6790#code#L609).

nventuro (approved):

PR #1530, opened 2022-07-25
Add an explicit yield protocol fee exemption flag for BPT

TomAFrench:
#1496 currently include a switch from 4 to 5 protocol fee exemption flags as well as implementing the actual exemption from fees.

This PR splits off the first feature as this can be easily tested and merged in isolation.

### DIFF
--- a/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
@@ -251,7 +251,7 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
             // The exemptFromYieldFlag should never be set on a token without a rate provider.
             // This would cause division by zero errors downstream.
             _require(
-                !exemptFromYieldFlags[i] || rateProviders[i] != IRateProvider(0),
+                !(exemptFromYieldFlags[i] && rateProviders[i] == IRateProvider(0)),
                 Errors.TOKEN_DOES_NOT_HAVE_RATE_PROVIDER
             );
         }
--- a/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
@@ -911,28 +911,20 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
     // Protocol Fee Exemption
 
     /**
-     * @dev Returns the protocolFeeExemptTokenFlags flags. Note that this token list *excludes* BPT.
-     * Its length will be one less than the registered pool tokens, and it will correspond to the token
-     * list after removing the BPT token.
+     * @dev Returns whether the token is exempt from protocol fees on the yield.
+     * If the BPT token is passed in (which doesn't make much sense, but shouldn't fail,
+     * since it is a valid pool token), the corresponding flag will be false.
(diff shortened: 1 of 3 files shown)


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -232,15 +232,28 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
+
         IRateProvider[] memory tokensAndBPTRateProviders = new IRateProvider[](params.tokens.length + 1);
         for (uint256 i = 0; i < tokensAndBPTRateProviders.length; ++i) {
             if (i < bptIndex) {
-                tokensAndBPTRateProviders[i] = params.rateProviders[i];
+                rateProviders[i] = params.rateProviders[i];
+                exemptFromYieldFlags[i] = params.exemptFromYieldProtocolFeeFlags[i];
             } else if (i == bptIndex) {
-                tokensAndBPTRateProviders[i] = IRateProvider(0);
+                rateProviders[i] = IRateProvider(0);
```
nventuro, line 245:
```suggestion
                rateProviders[i] = IRateProvider(0);
                exemptFromYieldFlags[i] = false;
```

This would make it consistent with the rate providers.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -232,15 +232,28 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
             } else {
-                tokensAndBPTRateProviders[i] = params.rateProviders[i - 1];
+                rateProviders[i] = params.rateProviders[i - 1];
+                exemptFromYieldFlags[i] = params.exemptFromYieldProtocolFeeFlags[i - 1];
             }
+
+            // The exemptFromYieldFlag should never be set on a token without a rate provider.
+            // This would cause division by zero errors downstream.
+            _require(
+                !exemptFromYieldFlags[i] || rateProviders[i] != IRateProvider(0),
```
nventuro, line 254:
```suggestion
                !(exemptFromYieldFlags[i] && rateProviders[i] == IRateProvider(0)),
```

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -232,15 +232,28 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
+                Errors.TOKEN_DOES_NOT_HAVE_RATE_PROVIDER
```
nventuro, line 255:
```suggestion
                Errors.EXEMPT_TOKEN_DOES_NOT_HAVE_RATE_PROVIDER
```

nventuro (commented):
Added some suggestions, but I'm not sure if those will cause merge issues with the other PR.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
nventuro: The test should also be ported

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
EndymionJkb: It's already like this in 1496.

PR #1532, opened 2022-07-25
Add ProtocolFeeProvider to Vault model deployment

nventuro:
This is a bit of a pre-requisite for us to create pools that use this provider. I thought the Vault model would be a good place to do this, since it is a singleton like the Vault, and it only depends on it.


EndymionJkb (approved):
Looks like a good start; can't go much farther without contract changes. I added fee provider deployment to ManagedPool tests so they'd compile.

PR #1533, opened 2022-07-25
[Deployments] Adding protocol fee percentages provider task.

jubeira:


pkg/deployments/tasks/20220725-protocol-fee-percentages-provider/test/task.fork.ts
```diff
@@ -0,0 +1,92 @@
...
+  before('setup admin', async () => {
+    const DEFAULT_ADMIN_ROLE = await authorizer.DEFAULT_ADMIN_ROLE();
+    admin = await impersonate(await authorizer.getRoleMember(DEFAULT_ADMIN_ROLE, 0), fp(100));
+  });
+
+  context('without permissions', () => {
+    itRevertsSettingFee(FeeType.Yield, fp(0.0157));
+
+    itRevertsSettingFee(FeeType.Swap, fp(0.0126));
+  });
```
nventuro, line 54: We don't need to remove this, but conceptually it is a bit overkill.

We know the contract does not allow setting these values without permissions, as it has been thoroughly unit-tested. What we want to do rather is perform a big integration check to make sure that the whole provider -> vault -> collector -> authorizer flow will work correctly and doesn't blow up (and also as a simple reminder of which permissions need to be granted).

pkg/deployments/tasks/20220725-protocol-fee-percentages-provider/test/task.fork.ts
```diff
@@ -0,0 +1,92 @@
...
+        .connect(admin)
+        .grantRole(await actionId(protocolFeePercentagesProvider, 'setFeeTypePercentage'), admin.address);
+    });
+
+    itSetsFeeCorrectly(FeeType.Yield, fp(0.1537));
+
+    itRevertsSettingFee(FeeType.Swap, fp(0.0857));
+
+    context('with swapFeePercentage permission', () => {
+      before('grant setSwapFeePercentage permission to fees collector', async () => {
```
nventuro, line 68:
```suggestion
      before('grant setSwapFeePercentage permission to fees provider', async () => {
```

nventuro (approved):
Very nice, thanks a lot!

pkg/deployments/tasks/20220725-protocol-fee-percentages-provider/test/task.fork.ts
jubeira: Thanks; I thought about it too, but I decided to include the tests for completion. I'll take it into account for the next fork test.

PR #1534, opened 2022-07-26
StablePhantom: first pass bytecode reduction

EndymionJkb:
This does two things: 1) low-hanging fruit bytecode reduction: using functions for immutable access; and 2) de-cluttering by moving all the immutable storage to a separate contract. There are no semantic changes.

Drops about 750 bytes, and the constructor is now 20 lines.


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -29,7 +29,7 @@ import "@balancer-labs/v2-pool-utils/contracts/BaseGeneralPool.sol";
 import "@balancer-labs/v2-pool-utils/contracts/rates/PriceRateCache.sol";
 import "@balancer-labs/v2-pool-utils/contracts/ProtocolFeeCache.sol";
 
-import "./StableMath.sol";
```
nventuro, line 32: Seems like we accidentally removed this

pkg/pool-stable-phantom/contracts/StablePoolStorage.sol
```diff
@@ -0,0 +1,365 @@
...
+    function _dropBptItem(uint256[] memory amounts) internal view returns (uint256[] memory) {
+        uint256[] memory amountsWithoutBpt = new uint256[](amounts.length - 1);
+        for (uint256 i = 0; i < amountsWithoutBpt.length; i++) {
+            amountsWithoutBpt[i] = amounts[i < getBptIndex() ? i : i + 1];
+        }
+
+        return amountsWithoutBpt;
+    }
+
+    // Helpers
```
TomAFrench, line 205: This line has been stranded a bit due to reordering.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
EndymionJkb: It's included by StablePoolStorage

pkg/pool-stable-phantom/contracts/StablePoolStorage.sol
EndymionJkb: Yes - I moved it up to the virtual supply section. It's similar to a helper function in StablePoolStorage, and I considered trying to put it there: but then I would have to move other virtual supply stuff, and as it has nothing to do with immutable storage, I think that should remain in the main contract.

This made me realize StablePoolStorage is a bit of a mess, so I've reordered that as well.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
nventuro: Eh I'd still have it here, since we rely on it heavily. There's no harm in doing a double include.

TomAFrench (approved):

PR #1537, opened 2022-07-26
Split BPT swaps by whether they're a join or exit before considering given in/out

TomAFrench:
This PR implements the idea which came up in the review call to organise the BPT swap functions by whether they perform a join or exit so that we can move the protocol fee payment further up in the call stack.

There's stuff to do like updating comments etc but this should show whether we want to move forward with this or optimise the old style.


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -442,151 +442,134 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
-                        bptIsTokenIn,
-                        amp,
-                        virtualSupply,
-                        balancesWithoutBpt
-                    ),
-                    scalingFactors[indexOut]
-                );
+        if (isJoinSwap) {
+            uint256 indexInNoBpt = _skipBptIndex(indexIn);
+            uint256 unknownAmount = _onSwapBptJoin(
```
EndymionJkb, line 459:
```suggestion
            uint256 calculatedAmount = _onSwapBptJoin(
```
The original name sounds better to me. It's "known" when it's returned - calculated - so it seems odd to name the return value "unknown." I believe we've used "calculated" elsewhere and in the docs (e.g., amountCalculated in the Vault swaps contract).

EndymionJkb:
I had it this way at one point in the beginning, and had stack issues with all the index manipulation and rounding. Apparently things have been refactored enough to where it fits now.

In general this is a cleaner division of labor: join/exit is more "important" distinction than given in/given out. Even if we basically do the same things before and after either one.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
nventuro: Yes, this is Vault terminology. For example, swaps deal with `amountGiven` and `amountCalculated`.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -442,151 +442,134 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
+    ) private returns (uint256 result) {
         _upscaleArray(balances, scalingFactors);
 
         (uint256 virtualSupply, uint256[] memory balancesWithoutBpt) = _payProtocolFeesBeforeJoinExit(balances);
 
-        bool bptIsTokenIn = swapRequest.tokenIn == IERC20(this);
+        bool isJoinSwap = swapRequest.tokenOut == IERC20(this);
+        bool isGivenIn = swapRequest.kind == IVault.SwapKind.GIVEN_IN;
+
+        uint256 upscaledAmount = _upscale(swapRequest.amount, scalingFactors[isGivenIn ? indexIn : indexOut]);
```
nventuro, line 453: Similarly, this could be called the 'amount given'.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -442,151 +442,134 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
+                isGivenIn,
+                amp,
+                virtualSupply,
+                balancesWithoutBpt
+            );
+
+            if (isGivenIn) {
+                balancesWithoutBpt[indexInNoBpt] += upscaledAmount;
+                // If join is "given in" then `unknownAmount` is the amount of BPT received, so we round down.
+                result = _downscaleDown(unknownAmount, scalingFactors[indexOut]);
```
nventuro, line 471: The comment is incorrect as we send BPT in a join, not receive. 

Regardless, the logic is sound: in a 'given in' join, the amount given (here called `upscaledAmount`) is an amount in (and hence added to the balances array), and the amount calculated (here called `unknown`) is an amount out (and hence we downscale down).

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -442,151 +442,134 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
+            } else {
+                balancesWithoutBpt[indexInNoBpt] += unknownAmount;
+                // If join is "given out" then `unknownAmount` is the amount of tokens sent to the Vault, so we round up.
+                result = _downscaleUp(unknownAmount, scalingFactors[indexIn]);
```
nventuro, line 475: In cleaner terms, in a given-out operation the calculated value is an amount in, so it must be rounded up.

nventuro (commented):
I much prefer this version - the actual join/exit logic is very clean, and we move all of the complexity to a single place, where we always deal with similar concerns (glorified data manipulation).

EndymionJkb:
We now have both amountCalculated and calculatedAmount (in RegularSwap). Change those too for consistency?

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -442,151 +442,140 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, ProtocolFeeCache {
...
+    ) private returns (uint256 downscaledAmountCalculated) {
         _upscaleArray(balances, scalingFactors);
 
         (uint256 virtualSupply, uint256[] memory balancesWithoutBpt) = _payProtocolFeesBeforeJoinExit(balances);
 
-        bool bptIsTokenIn = swapRequest.tokenIn == IERC20(this);
+        bool isJoinSwap = swapRequest.tokenOut == IERC20(this);
+        bool isGivenIn = swapRequest.kind == IVault.SwapKind.GIVEN_IN;
+
+        uint256 upscaledAmountGiven = _upscale(swapRequest.amount, scalingFactors[isGivenIn ? indexIn : indexOut]);
```
nventuro, line 453: Nitpicky, but I'd move this upscale up a few lines to have it next to the balances upscaling.

nventuro: (conceptually I think of this as two functions - one that deals with upscaling/downscaling, extracting data from `request`, etc., and one that handles the actual join/exit)

nventuro (approved):

PR #1538, opened 2022-07-26
Use AaveLinearPoolRebalancer in AaveLinearPools

nventuro:
This ended up being non-trivial due to a circular dependency in the LinearPool and its AssetManager. I found a nice way to solve said cycle at the factory, at the cost of having to add a `getLastCreatedPool()` getter (if we wanted to, we could always `delete` this slot, but that's overkill IMO).

The bigger downside is that it seems there's no way (outside of assembly memory shenanigans) to craft the full creation code of a contract (including abi-encoded constructor arguments) without making a full copy of the original creation code, due to how `abi.encodePacked` works. However, even taking this into account, the total cost of a Linear Pool + Rebalancer deployment is ~5.5M gas, which is way below block sizes (if a bit on the expensive side :sweat_smile:).

Given that the constructor abi-encoding is trivial (it simply receives two addresses), we could potentially do it using assembly by manually appending these values, but this seems not worth it at all.

### DIFF
--- a/pkg/interfaces/contracts/pool-utils/IBasePoolSplitCodeFactory.sol
+++ b/pkg/interfaces/contracts/pool-utils/IBasePoolSplitCodeFactory.sol
@@ -1,0 +1,36 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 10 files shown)


nventuro:
>Given that the constructor abi-encoding is trivial (it simply receives two addresses), we could potentially do it using assembly by manually appending these values, but this seems not worth it at all.

This is actually simpler than I thought, and seems to reduce gas cost by ~1M. I may introduce that later on.

TomAFrench (commented):
I don't think that we should have `getLastCreatedPool` on the base factory implementation, it needlessly makes deploying all pool types more expensive. Better instead to have ad hoc functions specific to the extra arguments which the contract needs.

PR #1541, opened 2022-07-27
Improve AaveRebalancer, add new linear deployment

nventuro:
This fixes some small issues in the rebalancer, adds a new mode which can rebalance even if there's missing tokens, and creates a deployment task that includes fork tests (since we lacked Rebalancer tests).


pkg/pool-linear/contracts/LinearPoolRebalancer.sol
```diff
@@ -63,8 +63,24 @@ abstract contract LinearPoolRebalancer {
...
-    function rebalance(address recipient) public {
+    function rebalance(address recipient) external returns (uint256) {
+        return _rebalance(recipient, 0);
+    }
+
+    /**
+     * @notice Same as `rebalance`, except this also works in scenarios where the Pool is in the no-fee zone by first
+     * taking `extraMain` tokens from the caller, to cover for rounding errors that are normally offset by acccumulated
+     * fees. Any extra tokens unused during the rebalance are sent to the recipient as usual.
+     */
```
TomAFrench, line 78:
```suggestion
    /**
     * @notice Rebalance a Linear Pool from an asset manager to maintain optimal operating conditions.
     * @dev This function performs the same action as `rebalance`, except this also works in scenarios where the Pool
     * is in the no-fee zone. This is done by first taking `extraMain` tokens from the caller, to cover for rounding errors
     * that are normally offset by acccumulated fees. Any extra tokens unused during the rebalance are sent to the recipient as usual.
     */
```

TomAFrench: My comma nazi-ing also applies to the `@notice` tag on `rebalance(address)`

pkg/pool-linear/contracts/LinearPoolRebalancer.sol
```diff
@@ -77,13 +93,55 @@ abstract contract LinearPoolRebalancer {
...
+
+    function _receiveExtraMainDuringLackOfMain(uint256 mainTokenBalance, uint256 extraMain) private {
+        // The Pool lacks main tokens, but if the main balance is above the lower target (i.e. it is still in the no-fee
+        // zone), then there's no accumulated fees. It is therefore possible that the amount of wrapped tokens the Pool
+        // will give us in exchange for main cannot be unwrapped into the amount of main tokens desired by the Pool, due
+        // to  the Pool choosing to round in its favor (resulting in an unrealistic wrap/unwrap rate).
+        //
+        // This is typically a non-issue as the Linear Pool accumulated fees more than cover for said rounding error,
+        // but in scenarios with no fees we require some extra main tokens to be able to make the Pool whole. Any excess
+        // will be sent at the end of the transaction to the recipient.
```
TomAFrench, line 118:
```suggestion
        // The Pool lacks main tokens, but if the main balance is above the lower target (i.e. it is still in the no-fee
        // zone), then there's no accumulated fees. It is therefore possible that the amount of wrapped tokens the Pool
        // will give us in exchange for main cannot be unwrapped into the amount of main tokens desired by the Pool, due
        // to the Pool choosing to round in its favor (resulting in an unrealistic wrap/unwrap rate).
        //
        // This is typically a non-issue as the Linear Pool's accumulated fees more than cover for said rounding error,
        // but in scenarios with no fees we require some extra main tokens to be able to make the Pool whole. Any excess
        // will be sent to the recipient at the end of the transaction.
```

pkg/pool-linear/contracts/LinearPoolRebalancer.sol
```diff
@@ -77,13 +93,55 @@ abstract contract LinearPoolRebalancer {
...
+
+    function _receiveExtraMainDuringExcessOfMain(uint256 mainTokenBalance, uint256 extraMain) private {
+        // The Pool has excess of main tokens, but if the main balance is below the upper target (i.e. it is still in
+        // the no-fee zone), then there's no accumulated fees. It is therefore possible that the amount of main tokens
+        // the Pool will give us in exchange for wrapped cannot be wrapped into the amount of wrapped tokens desired by
+        // the Pool, due to the Pool choosing to round in its favor (resulting in an unrealistic wrap/unwrap rate).
+        //
+        // This is typically a non-issue as the Linear Pool accumulated fees more than cover for said rounding error,
+        // but in scenarios with no fees we require some extra main tokens to be able to make the Pool whole. Any excess
+        // will be sent at the end of the transaction to the recipient.
```
TomAFrench, line 135:
```suggestion
        // The Pool has excess of main tokens, but if the main balance is below the upper target (i.e. it is still in
        // the no-fee zone), then there's no accumulated fees. It is therefore possible that the amount of main tokens
        // the Pool will give us in exchange for wrapped cannot be wrapped into the amount of wrapped tokens desired by
        // the Pool, due to the Pool choosing to round in its favor (resulting in an unrealistic wrap/unwrap rate).
        //
        // This is typically a non-issue as the Linear Pool's accumulated fees more than cover for said rounding error,
        // but in scenarios with no fees we require some extra main tokens to be able to make the Pool whole. Any excess
        // will be sent  to the recipient at the end of the transaction.
```

pkg/pool-linear/contracts/LinearPoolRebalancer.sol
```diff
@@ -77,13 +93,55 @@ abstract contract LinearPoolRebalancer {
...
+        // will give us in exchange for main cannot be unwrapped into the amount of main tokens desired by the Pool, due
+        // to  the Pool choosing to round in its favor (resulting in an unrealistic wrap/unwrap rate).
+        //
+        // This is typically a non-issue as the Linear Pool accumulated fees more than cover for said rounding error,
+        // but in scenarios with no fees we require some extra main tokens to be able to make the Pool whole. Any excess
+        // will be sent at the end of the transaction to the recipient.
+
+        (uint256 lowerTarget, ) = _pool.getTargets();
+        // The lower target is upscaled, so we undo that
+        if (mainTokenBalance > FixedPoint.divDown(lowerTarget, _mainTokenScalingFactor)) {
```
TomAFrench, line 122:
```suggestion
        if (mainTokenBalance >= FixedPoint.divDown(lowerTarget, _mainTokenScalingFactor)) {
```

If we're right on the edge of the no-fee zone there's still no fees to cover the rounding errors so we should be inclusive here.

pkg/pool-linear/contracts/LinearPoolRebalancer.sol
```diff
@@ -77,13 +93,55 @@ abstract contract LinearPoolRebalancer {
...
+        // the Pool will give us in exchange for wrapped cannot be wrapped into the amount of wrapped tokens desired by
+        // the Pool, due to the Pool choosing to round in its favor (resulting in an unrealistic wrap/unwrap rate).
+        //
+        // This is typically a non-issue as the Linear Pool accumulated fees more than cover for said rounding error,
+        // but in scenarios with no fees we require some extra main tokens to be able to make the Pool whole. Any excess
+        // will be sent at the end of the transaction to the recipient.
+
+        (, uint256 upperTarget) = _pool.getTargets();
+        // The upper target is upscaled, so we undo that
+        if (mainTokenBalance < FixedPoint.divDown(upperTarget, _mainTokenScalingFactor)) {
```
TomAFrench, line 139:
```suggestion
        if (mainTokenBalance <= FixedPoint.divDown(upperTarget, _mainTokenScalingFactor)) {
```
ditto

rabmarut (commented):
Looks good! This is my first time reviewing any of the rebalancer code, so I went a bit beyond the scope of this PR and examined both contracts in their entirety.

I left one comment about the convoluted nature of the rebalance functions, but it's pretty subjective. I would personally prefer to minimize that amount of code rather than adding all the logic to save gas on a weird edge case (if I've understood the edge case correctly).

Also, I'm just going to assume that the whole `getLastCreatedPool` thing actually works and has been implemented in the pool and factory because I don't have time to review that and it seems like it is out of the scope of this PR anyway.

pkg/pool-linear/contracts/LinearPoolRebalancer.sol
```diff
@@ -77,13 +95,55 @@ abstract contract LinearPoolRebalancer {
...
+        // will give us in exchange for main cannot be unwrapped into the amount of main tokens desired by the Pool, due
+        // to the Pool choosing to round in its favor (resulting in an unrealistic wrap/unwrap rate).
+        //
+        // This is typically a non-issue as the Linear Pool's accumulated fees more than cover for said rounding error,
+        // but in scenarios with no fees we require some extra main tokens to be able to make the Pool whole. Any excess
+        // will be sent to the recipient at the end of the transaction.
+
+        (uint256 lowerTarget, ) = _pool.getTargets();
+        // The lower target is upscaled, so we undo that
+        if (mainTokenBalance >= FixedPoint.divDown(lowerTarget, _mainTokenScalingFactor)) {
```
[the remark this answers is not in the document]
nventuro: Hm that might be a fair point, after all, while the check is cheap due to the storage being warm, the same is also true for the transfer. If anything we'd be writing to the caller's allowance, but that's not terrible. You may have a good point there.

nventuro:
Even though this is a WIP still, merging for the goerli deployment. We expect one more PR on linear pools, addressing recovery mode.

PR #1542, opened 2022-07-28
Stable: Apply yield protocol fee

EndymionJkb:
Separate the protocol fee calculation into growth due to (non-exempt) yield, and growth due to swap fees, and apply the corresponding protocol fee percentages to each.

This complicates the rate calculations, as previously we were only storing the oldRate for exempt tokens. To compute the swap fee growth, we need to store the oldRate for all tokens.


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -851,19 +856,36 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, StablePoolStorage,
         // However, a part of the invariant growth was due to non protocol swap fees (i.e. value accrued by the
         // LPs), so we only mint a percentage of this BPT amount: that which corresponds to protocol fees.
 
-        // We round down, favoring LP fees.
+        uint256 totalGrowthInvariant = StableMath._calculateInvariant(_getPostJoinExitAmp(), totalGrowthBalances);
+        uint256 swapGrowthInvariant = StableMath._calculateInvariant(_getPostJoinExitAmp(), swapGrowthBalances);
```
nventuro, line 860:
```suggestion
        uint256 postJoinAmp = _getPostJoinExitAmp();
        uint256 totalGrowthInvariant = StableMath._calculateInvariant(postJoinAmp, totalGrowthBalances);
        uint256 swapGrowthInvariant = StableMath._calculateInvariant(postJoinAmp, swapGrowthBalances);
```

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -851,19 +856,36 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, StablePoolStorage,
...
 
-        uint256 invariantRatio = preJoinInvariant.divDown(_postJoinExitInvariant);
-        uint256 protocolFeeAmount;
-        if (invariantRatio > FixedPoint.ONE) {
-            // This condition should always be met outside of rounding errors (for non-zero swap fees).
+        // Total Growth = Invariant with masked rates / last invariant: swap fees + non-exempt token yields
```
nventuro, line 862:
```suggestion
        // Total Growth = Invariant with old rates for exempt tokes / last invariant: swap fees + non-exempt token yields
```

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -851,19 +856,36 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, StablePoolStorage,
...
+        // Swap Fee Growth = Invariant with old rates / last invariant: swap fees alone
```
nventuro, line 863:
```suggestion
        // Swap Fee Growth = Invariant with old rates for all tokens / last invariant: swap fees alone
```

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -877,23 +899,24 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, StablePoolStorage,
         _updateOldRates();
     }
 
+    function _getPostJoinExitAmp() internal view returns (uint256) {
+        return _postJoinExitAmp;
+    }
+
+    function _getPostJoinExitInvariant() internal view returns (uint256) {
+        return _postJoinExitInvariant;
+    }
```
nventuro, line 908: These are not immutable variables, so we don't really gain much from a getter. Specially considering we use them in a single place.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -958,6 +988,15 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, StablePoolStorage,
         return cache.getOldRate().divUp(cache.getCurrentRate());
     }
 
+    function _hasCacheEntry(uint256 index) private view returns (bool) {
+        if (index == 0) return _getRateProvider0() != IRateProvider(0) && getBptIndex() != 0;
+        if (index == 1) return _getRateProvider1() != IRateProvider(0) && getBptIndex() != 1;
+        if (index == 2) return _getRateProvider2() != IRateProvider(0) && getBptIndex() != 2;
+        if (index == 3) return _getRateProvider3() != IRateProvider(0) && getBptIndex() != 3;
+        if (index == 4) return _getRateProvider4() != IRateProvider(0) && getBptIndex() != 4;
+        if (index == 5) return _getRateProvider5() != IRateProvider(0) && getBptIndex() != 5;
```
nventuro, line 997:
```suggestion
        uint256 bptIndex = getBptIndex();
        if (index == 0) return _getRateProvider0() != IRateProvider(0) && bptIndex != 0;
        if (index == 1) return _getRateProvider1() != IRateProvider(0) && bptIndex != 1;
        if (index == 2) return _getRateProvider2() != IRateProvider(0) && bptIndex != 2;
        if (index == 3) return _getRateProvider3() != IRateProvider(0) && bptIndex != 3;
        if (index == 4) return _getRateProvider4() != IRateProvider(0) && bptIndex != 4;
        if (index == 5) return _getRateProvider5() != IRateProvider(0) && bptIndex != 5;
```

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -851,19 +856,36 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, StablePoolStorage,
...
+        // Growth due to yield = Total Growth / Swap Fee Growth
+        //                     = Invariant with masked rates / Invariant with old rates.
 
-            protocolFeeAmount = getProtocolFeePercentageCache(ProtocolFeeType.SWAP).mulDown(
-                invariantRatio.sub(FixedPoint.ONE).mulDown(virtualSupply)
-            );
+        // If the "growth" is negative, set the ratio to ONE: multiplying by (ratio - 1) will then result in zero fees.
+        uint256 yieldGrowthRatio = totalGrowthInvariant > swapGrowthInvariant
+            ? totalGrowthInvariant.divDown(swapGrowthInvariant)
+            : FixedPoint.ONE;
```
nventuro, line 870: @rabmarut pointed out that this might mean charging extra fees in the future if the rate were to grow again after it fell. Not sure what his conclusion was

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -851,19 +856,36 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, StablePoolStorage,
...
+            .mulDown(yieldGrowthRatio.sub(FixedPoint.ONE))
+            .add(getProtocolFeePercentageCache(ProtocolFeeType.SWAP).mulDown(swapGrowthRatio.sub(FixedPoint.ONE)))
+            .mulDown(virtualSupply);
+
+        if (protocolFeeAmount > 0) {
             _payProtocolFees(protocolFeeAmount);
         }
+
         // For this addition to overflow, the actual total supply would have already overflowed.
         return (virtualSupply + protocolFeeAmount, balancesWithoutBpt);
```
nventuro, line 890: Let's add a comment explaining why we add these two

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
rabmarut: Hm, we have a very simple solution for `WeightedPool`, but it won't work here. For reference, the weighted invariant can be trivially broken into two components: swaps and yield. So in `WeightedPool`, we calculate those two components completely independently and then collect yield fees only when the yield component exceeds its previous all-time high. But that independence is lost here with the yield portion being extracted from total growth. It won't be possible to track an all-time high yield.

I'm not sure if there is a solution to this for `StablePhantomPool`; we might have to accept this risk. Tokens with exceptionally volatile rates could always be exempted.

nventuro: At least this won't be an issue in bb-a-usd since linear rates are monotonically increasing

rabmarut: It's probably also safe for the great majority of yield-bearing tokens, whose rates would only decrease due to a black swan event (which causes far more trouble than some excess protocol fees). So, it's an edge case, and there is a workaround for rates that are expected to be volatile: the pool creator can exempt those tokens. It's not perfect but it's probably the best we can do for now.

nventuro (approved):

PR #1545, opened 2022-07-28
Use token getters in Phantom Pool

nventuro:
This saves ~500 bytes, putting the total bytecode at 23.7kB, meaning we're below the limit!


TomAFrench (approved):

EndymionJkb (approved):
Thanks! Must have missed these.

PR #1546, opened 2022-07-28
Simplify calculation of adjustedBalances

TomAFrench:
Rather than creating two arrays which don't include BPT, then iterating through both multiplying together we can just have one and apply the token rate ratios directly and then drop the BPT element.

### DIFF
--- a/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
@@ -829,8 +829,8 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, StablePoolStorage,
         (uint256 virtualSupply, uint256[] memory balancesWithoutBpt) = _dropBptItemFromBalances(balances);
 
         // Apply the rate adjustment to exempt tokens: multiply by oldRate / currentRate to "undo" the current scaling,
-        // and apply the old rate. This function copies the values in `balancesWithoutBpt and so doesn't mutate it.
-        uint256[] memory adjustedBalances = _adjustBalancesByTokenRatios(balancesWithoutBpt);
+        // and apply the old rate. This function copies the values in `balances` and so doesn't mutate it.
+        uint256[] memory adjustedBalances = _dropBptItem(_getAdjustedBalances(balances));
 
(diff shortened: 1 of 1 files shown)


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -898,61 +898,43 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, StablePoolStorage,
...
+        if (totalTokens == 5) return adjustedBalances;
+        adjustedBalances[5] = _isTokenExemptFromYieldProtocolFee(5)
+            ? _adjustedBalance(balances[5], _tokenRateCaches[_getToken5()])
+            : balances[5];
     }
 
-    function _computeRateRatio(bytes32 cache) private pure returns (uint256) {
-        return cache.getOldRate().divUp(cache.getCurrentRate());
+    function _adjustedBalance(uint256 balance, bytes32 cache) private pure returns (uint256) {
+        return balance.mulDown(cache.getOldRate()).divUp(cache.getCurrentRate());
```
nventuro, line 937:
```suggestion
        return Math.div(Math.mul(balance, cache.getOldRate()), cache.getCurrentRate());
```

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -898,61 +898,43 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, StablePoolStorage,
...
+            : balances[0];
+        adjustedBalances[1] = _isTokenExemptFromYieldProtocolFee(1)
+            ? _adjustedBalance(balances[1], _tokenRateCaches[_getToken1()])
+            : balances[1];
+        adjustedBalances[2] = _isTokenExemptFromYieldProtocolFee(2)
+            ? _adjustedBalance(balances[2], _tokenRateCaches[_getToken2()])
+            : balances[2];
+
+        // Before we adjust the remaining balances we must check that the Pool contains enough tokens.
+        if (totalTokens == 3) return adjustedBalances;
```
nventuro, line 920: I'd also add an explicit return at the end for consistency. Also drop the named return argument if we're not using it.

nventuro (approved):

PR #1547, opened 2022-07-28
Make immutable state variables private

nventuro:
This will prevent accidentally using them outside of getters, at least on other files.


EndymionJkb (approved):
I wasn't worried about "exposing" immutables to internal scope at first (I believe the scope predated the getters, at least for some of them). Since all you can do with immutables is "get" them anyway - and we want to encourage using getters vs direct reference for bytecode - this is better.

PR #1552, opened 2022-07-28
Stable: self-contained top-level join/exit

EndymionJkb:
It is a bit messy to pay protocol fees at the beginning of _onJoinPool and _onExitPool, but then rely on the individual functions to update the invariant afterwards. Refactor with the pay/reset header/footer, like the BPT swap does.

### DIFF
--- a/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
@@ -695,7 +695,6 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, StablePoolStorage,
             _revert(Errors.UNHANDLED_EXIT_KIND);
         }
 
-
         // Subtract amountsOut to get post-exit balances
         _mutateAmounts(balancesWithoutBpt, amountsOut, FixedPoint.sub);
 
--- a/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
@@ -548,7 +548,6 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, StablePoolStorage,
         (uint256 virtualSupply, uint256[] memory balancesWithoutBpt) = _payProtocolFeesBeforeJoinExit(balances);
         (uint256 currentAmp, ) = _getAmplificationParameter();
 
-        // The individual join functions mutate balancesWithoutBpt so that they reflect the post-join balances
         if (kind == StablePhantomPoolUserData.JoinKindPhantom.EXACT_TOKENS_IN_FOR_BPT_OUT) {
             (bptAmountOut, amountsIn) = _joinExactTokensInForBPTOut(
                 virtualSupply,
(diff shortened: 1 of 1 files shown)


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -534,26 +535,43 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, StablePoolStorage,
...
+                currentAmp,
+                balancesWithoutBpt,
+                userData
+            );
         } else {
             _revert(Errors.UNHANDLED_JOIN_KIND);
         }
+
+        // Pass in the post-join balances to reset the protocol fee basis.
+        _updateInvariantAfterJoinExit(currentAmp, balancesWithoutBpt);
```
TomAFrench, line 565:
```suggestion
        // Add amountsIn to get post-join balances
        _mutateAmounts(balancesWithoutBpt, _dropBptItem(amountsIn), FixedPoint.add);
        
        // Pass in the post-join balances to reset the protocol fee basis.
        _updateInvariantAfterJoinExit(currentAmp, balancesWithoutBpt);
```

I would perform this mutation here rather than one layer down.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -658,26 +672,43 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, StablePoolStorage,
...
+                currentAmp,
+                balancesWithoutBpt,
+                userData
+            );
         } else {
             _revert(Errors.UNHANDLED_EXIT_KIND);
         }
+
+        // Pass in the post-exit balances to reset the protocol fee basis.
+        _updateInvariantAfterJoinExit(currentAmp, balancesWithoutBpt);
```
TomAFrench, line 702:
```suggestion
        // Subtract amountsOut to get post-exit balances
        _mutateAmounts(balancesWithoutBpt, scaledAmountsOutWithoutBpt, FixedPoint.sub);
        
        // Pass in the post-exit balances to reset the protocol fee basis.
        _updateInvariantAfterJoinExit(currentAmp, balancesWithoutBpt);
```

Similarly here.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
EndymionJkb: Yes; much more self-contained

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -541,29 +542,49 @@ contract StablePhantomPool is IRateProvider, BaseGeneralPool, StablePoolStorage,
...
         uint256[] memory scalingFactors,
         bytes memory userData
-    ) internal override returns (uint256, uint256[] memory) {
+    ) internal override returns (uint256 bptAmountOut, uint256[] memory amountsIn) {
         StablePhantomPoolUserData.JoinKindPhantom kind = userData.joinKind();
 
         (uint256 virtualSupply, uint256[] memory balancesWithoutBpt) = _payProtocolFeesBeforeJoinExit(balances);
+        (uint256 currentAmp, ) = _getAmplificationParameter();
 
+        // The individual join functions mutate balancesWithoutBpt so that they reflect the post-join balances
```
nventuro, line 551: Isn't this wrong? We do this ourselves later

nventuro (commented):
This reads _way_ better split this way

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
EndymionJkb: Ah - forgot to remove that comment (I did fix the comments below)

PR #1553, opened 2022-07-29
Stable: separate Amp functionality

EndymionJkb:
Reduce length of the main file by moving the Amplification factor storage and handling into another contract. I think that's the most separable thing. (Anything that touches immutable storage has a StablePoolStorage dependency.)

Mutually exclusive with #1554 (which splits out both Amp and token rate cache functionality).

### DIFF
--- a/pkg/pool-stable-phantom/contracts/StablePoolAmplification.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePoolAmplification.sol
@@ -14,12 +14,12 @@
 
 pragma solidity ^0.7.0;
 
-import "@balancer-labs/v2-pool-utils/contracts/BasePool.sol";
+import "@balancer-labs/v2-pool-utils/contracts/BasePoolAuthorization.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
 
 import "./StableMath.sol";
 
-abstract contract StablePoolAmplification is BasePool {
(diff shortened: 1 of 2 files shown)


pkg/pool-stable-phantom/contracts/StablePoolAmplification.sol
```diff
@@ -0,0 +1,187 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/Authentication.sol";
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
+
+import "./StableMath.sol";
+
+abstract contract StablePoolAmplification is Authentication {
```
nventuro, line 22: Make this inherit from `BasePoolAuthorization` instead of plain Authentication, and override `isOwnerOnly` for the two amp functions.

EndymionJkb: It seems to need BasePool (for the inheritance graph to work out)

EndymionJkb: Per discussion, the problem was super; removing the call to super lets me derive from BasePoolAuthorization

nventuro (approved):

PR #1554, opened 2022-07-29
Stable: aggressive contract splitting

EndymionJkb:
Figured out a way to avoid stack-too-deep in the constructor. Had to "stack" inheritance of the subcontracts, and pass the params struct instead of individual parameters.

This pulls all the amplification and token rate cache handling into a separate contract (StablePoolRates).

Mutually exclusive with #1553. (Amp and Rates aren't really related, but trying to split it into three sub-contracts: amp, rates, and storage, again leads to stack-too-deep.)


EndymionJkb:
Superseded by #1560

PR #1555, opened 2022-07-29
Stable: Drop from 5 to 3 tokens

EndymionJkb:
If we want to do this, here's how it would look. Saves ~1250 bytecode.


EndymionJkb:
A good exercise for reference, but not needed (and conflicts have accumulated).

PR #1556, opened 2022-07-29
Deploy protocol fee percentage provider

nventuro:
The Optimism etherscan verification is failing for some reason, but the contract seems to be there as I was able to get the action ids.


nventuro:
The optimism verification just succeeded.

TomAFrench:
I get the same build-info when building commit [e86a958](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1556/commits/e86a95829c7e6591571b21d66b836a6303cf4932) as is in this PR.

PR #1557, opened 2022-07-30
Rename StablePhantomPool to ComposableStablePool

EndymionJkb:


pkg/balancer-js/src/utils/errors.ts
```diff
@@ -66,7 +66,7 @@ const balancerErrorCodes: Record<string, string> = {
   '337': 'UNAUTHORIZED_EXIT',
   '338': 'MAX_MANAGEMENT_SWAP_FEE_PERCENTAGE',
   '339': 'UNHANDLED_BY_MANAGED_POOL',
-  '340': 'UNHANDLED_BY_PHANTOM_POOL',
+  '340': 'UNHANDLED_BY_STABLE_POOL',
```
nventuro, line 69: Let's keep the old name - this error existed in previous versions of the phantom pool.

pkg/pool-stable/contracts/ComposableStablePool.sol
```diff
@@ -15,7 +15,7 @@
 pragma solidity ^0.7.0;
 pragma experimental ABIEncoderV2;
 
-import "@balancer-labs/v2-interfaces/contracts/pool-stable-phantom/StablePhantomPoolUserData.sol";
+import "@balancer-labs/v2-interfaces/contracts/pool-stable-composable/ComposableStablePoolUserData.sol";
```
nventuro, line 18: The directory in the interfaces package should match this one, so it ought to be called `pool-stable`, without the `-composable` suffix

pkg/pool-stable/contracts/ComposableStablePool.sol
```diff
@@ -29,8 +29,8 @@ import "@balancer-labs/v2-pool-utils/contracts/BaseGeneralPool.sol";
 import "@balancer-labs/v2-pool-utils/contracts/rates/PriceRateCache.sol";
 import "@balancer-labs/v2-pool-utils/contracts/ProtocolFeeCache.sol";
 
-import "./StablePoolAmplification.sol";
-import "./StablePoolStorage.sol";
+import "./ComposableStablePoolAmplification.sol";
```
nventuro, line 32: The Amp module is BPT agnostic, so it doesn't need this prefix.

nventuro (commented):
Fantastic! I'll leave it up to you to decide when to pull the trigger with this.

pkg/pool-stable/contracts/ComposableStablePool.sol
EndymionJkb: I had it that way, but the problem is (as documented) there are still things using the original StablePool (including the proportional joins/exits that we removed). Maybe review those things to see if they can be deleted? Or make pool-stable-composable to pool-stable, and rename the current pool-stable to pool-stable-legacy?

nventuro (approved):
Up to you when to merge!

EndymionJkb:
> Up to you when to merge!

Probably last - I don't want to cause conflicts with Tom.

PR #1560, opened 2022-08-01
Stable: extract rate provider functionality

EndymionJkb:
Extract the last semantically distinct non-immutable storage and handling into a separate contract: had to introduce parameter munging functions and pass structs too avoid stack issues.

There is still a fair amount of rate provider processing in the main contract, but this is because these functions also call into StableStorage, and I didn't want the subcontracts to inherit from each other. (I suppose this could be factored out further using interfaces if we really wanted a complete separation.)


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -1172,6 +1122,29 @@ contract StablePhantomPool is
...
+        returns (StablePoolStorage.StorageParams memory)
+    {
+        return
+            StablePoolStorage.StorageParams(
+                _insertSorted(params.tokens, IERC20(this)),
+                params.rateProviders,
+                params.exemptFromYieldProtocolFeeFlags
+            );
+    }
+
```
nventuro, line 1147: I'd place this right below the constructor

pkg/pool-stable-phantom/contracts/StablePoolRates.sol
```diff
@@ -0,0 +1,88 @@
...
+
+pragma solidity ^0.7.0;
+
+import "@balancer-labs/v2-interfaces/contracts/solidity-utils/openzeppelin/IERC20.sol";
+import "@balancer-labs/v2-interfaces/contracts/pool-utils/IRateProvider.sol";
+
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/InputHelpers.sol";
+import "@balancer-labs/v2-pool-utils/contracts/rates/PriceRateCache.sol";
+
+abstract contract StablePoolRates {
```
nventuro, line 23: I think it's fine to have this inherit from storage to be honest. That'll let us add many other functions, such as `getTokenRateCache`, `getTokenRate`, `_getAdjustedBalances`, `setTokenRateCacheDuration`, etc. As is, this module is not too useful for testing responsability removal purposes.

Down the line, if say we had multiple stable pools, we could then abstract away storage functions into virtual ones (like we do in BaseWeightedPool), but we don't need to worry about that now.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
EndymionJkb: I had it there, then moved it to helpers

pkg/pool-stable-phantom/contracts/StablePoolRates.sol
EndymionJkb: Ok; can try it that way. (Getting very close to the original "aggressive" PR, though the helper trick is an improvement.)

nventuro (commented):
Looking good! What about the `_getAdjustedBalances` family of functions, and potentially `_scalingFactors`?

pkg/pool-stable-phantom/contracts/StablePoolRates.sol
```diff
@@ -0,0 +1,319 @@
...
+    // [ uint32  |  uint32  |  uint96  |   uint96     ]
+
+    mapping(IERC20 => bytes32) internal _tokenRateCaches;
+
+    event TokenRateCacheUpdated(IERC20 indexed token, uint256 rate);
+    event TokenRateProviderSet(IERC20 indexed token, IRateProvider indexed provider, uint256 cacheDuration);
+
+    constructor(StablePoolStorage.StorageParams memory storageParams, RatesParams memory rateParams)
+        StablePoolStorage(storageParams)
+    {
```
nventuro, line 53: Let's not call `StablePoolStorage`'s constructor here, and instead leave it up to the parent contract (like we do in e.g. `EOASignaturesValidator`, which doesn't initialize `EIP712`).

pkg/pool-stable-phantom/contracts/StablePoolRates.sol
```diff
@@ -0,0 +1,319 @@
...
+        returns (
+            uint256 rate,
+            uint256 duration,
+            uint256 expires
+        )
+    {
+        _require(_getRateProvider(token) != IRateProvider(0), Errors.TOKEN_DOES_NOT_HAVE_RATE_PROVIDER);
+
+        rate = _tokenRateCaches[token].getCurrentRate();
+        (duration, expires) = _tokenRateCaches[token].getTimestamps();
```
nventuro, line 120: We could expand this to also return the old rate.

nventuro: (this is an effort separate from this PR, but I just noticed this)

nventuro (commented):
Looking great! StablePool looks very reasonable after this change.

We should be able to extensively test this new rates module (in a separate PR) without having to make real Pool operatrions. Thanks a lot!

PR #1561, opened 2022-08-01
Deploy authorizer

nventuro:
It's finally done!

We'll need to extend our json action id files to also support grant role permissions etc.


TomAFrench:
The authorizooooor

TomAFrench (approved):
I get the same build-info when building commit 15db57a36820feabf64e0657328149f009c82d8b as is in this PR.

PR #1562, opened 2022-08-01
Split off tests for StablePoolStorage

TomAFrench:
This PR pulls out all the contract initialisation tests from `StablePhantomPool.test.ts` to do with the constructor setup in `StablePoolStorage`.

Something to note is that we need to provide an implementation of `_scalingFactors` in `MockStablePoolStorage`. This implementation is different to that in `StablePhantomPool` as we cannot access the token rate caches. We could pull "live" rates from the rate providers but we'd still need to have separate tests for the full implementation anyway so I've simplified to ignore rates for the returned scaling factors.

### DIFF
--- a/pkg/pool-stable-phantom/contracts/test/MockStablePoolStorage.sol
+++ b/pkg/pool-stable-phantom/contracts/test/MockStablePoolStorage.sol
@@ -1,0 +1,120 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
+
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/ERC20Helpers.sol";
+
+import "../StablePoolStorage.sol";
+
+contract MockStablePoolStorage is StablePoolStorage {
+    constructor(
+        IVault vault,
+        IERC20[] memory tokens,
+        IRateProvider[] memory tokenRateProviders,
+        bool[] memory exemptFromYieldProtocolFeeFlags
+    )
+        StablePoolStorage(_insertSorted(tokens, IERC20(this)), tokenRateProviders, exemptFromYieldProtocolFeeFlags)
+        BasePool(
+            vault,
+            IVault.PoolSpecialization.GENERAL,
+            "MockStablePoolStorage",
+            "MOCK_BPT",
+            _insertSorted(tokens, IERC20(this)),
+            new address[](tokens.length + 1),
+            1e12, // BasePool._MIN_SWAP_FEE_PERCENTAGE
+            0,
+            0,
+            address(0)
+        )
+    {
+        // solhint-disable-previous-line no-empty-blocks
+    }
+
+    /**
+     * @notice Return the scaling factor for a token. This includes both the token decimals and the rate.
+     */
+    function getScalingFactor(IERC20 token) external view returns (uint256) {
+        return _scalingFactor(token);
+    }
+
+    // Computed the total scaling factor as a product of the token decimal adjustment and token rate.
+    function _scalingFactor(IERC20 token) internal view virtual override returns (uint256) {
+        return _tokenScalingFactor(token);
+    }
+
+    /**
+     * @dev Overrides scaling factor getter to compute the tokens' rates.
+     */
+    function _scalingFactors() internal view virtual override returns (uint256[] memory) {
+        // There is no need to check the arrays length since both are based on `_getTotalTokens`
+        uint256 totalTokens = _getTotalTokens();
+        uint256[] memory scalingFactors = new uint256[](totalTokens);
+
+        // The Pool will always have at least 3 tokens so we always load these three scaling factors.
+        // Given there is no generic direction for this rounding, it follows the same strategy as the BasePool.
+        scalingFactors[0] = _getScalingFactor0();
+        scalingFactors[1] = _getScalingFactor1();
+        scalingFactors[2] = _getScalingFactor2();
+
+        // Before we load the remaining scaling factors we must check that the Pool contains enough tokens.
+        if (totalTokens == 3) return scalingFactors;
+        scalingFactors[3] = _getScalingFactor3();
+
+        if (totalTokens == 4) return scalingFactors;
+        scalingFactors[4] = _getScalingFactor4();
+
+        if (totalTokens == 5) return scalingFactors;
+        scalingFactors[5] = _getScalingFactor5();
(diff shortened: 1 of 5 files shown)


TomAFrench:
It might end up being sensible to hold off on merging this until #1560 is merged depending on how that split affects the token rate code.

EndymionJkb:
> It might end up being sensible to hold off on merging this until #1560 is merged depending on how that split affects the token rate code.

Yes, especially since I'm now pulling out many more functions by making it inherit from StableStorage

pkg/pool-stable-phantom/contracts/test/MockStablePoolStorage.sol
```diff
@@ -0,0 +1,120 @@
...
+        scalingFactors[3] = _getScalingFactor3();
+
+        if (totalTokens == 4) return scalingFactors;
+        scalingFactors[4] = _getScalingFactor4();
+
+        if (totalTokens == 5) return scalingFactors;
+        scalingFactors[5] = _getScalingFactor5();
+
+        return scalingFactors;
+    }
```
nventuro, line 83: I'd consider exposing `getScalingFactor0()` and friends instead.

nventuro: It'd also be a good idea to expose and test `_getRateProvider(token)`, `_getRateProvider0()`, `_getToken0`, and `_skipBptIndex` and friends.

pkg/pool-stable-phantom/test/StablePoolStorage.test.ts
```diff
@@ -0,0 +1,330 @@
...
+          expect(await pool.getToken3()).to.be.eq(expectedTokenAddresses[3]);
+          expect(await pool.getToken4()).to.be.eq(expectedTokenAddresses[4]);
+          expect(await pool.getToken5()).to.be.eq(expectedTokenAddresses[5]);
+        });
+
+        it('sets the fee exemption flags correctly', async () => {
+          const bpt = await Token.deployedAt(pool);
+          const allTokens = new TokenList([...tokens.tokens, bpt]).sort();
+
+          const expectedExemptFromYieldProtocolFeeFlags = exemptFromYieldProtocolFeeFlags.slice();
```
nventuro, line 129: Huh, interesting. I typically do `array[...]` for copies, didn't think to use slice this way.

pkg/pool-stable-phantom/test/StablePoolStorage.test.ts
```diff
@@ -0,0 +1,330 @@
...
+
+    describe('scaling factors', () => {
+      describe('getScalingFactorX', () => {
+        it('returns the correct scaling factor', async () => {
+          const expectedScalingFactors = tokens.map((token) => fp(1).mul(bn(10).pow(18 - token.decimals)));
+          expectedScalingFactors.splice(bptIndex, 0, fp(1));
+
+          // Also check the individual getters.
+          // There's always 6 getters however not all of them may be used. Unused getters return the zero address.
+          const paddedScalingFactors = Array.from({ length: 6 }, (_, i) => expectedScalingFactors[i] ?? ZERO_ADDRESS);
```
nventuro, line 252:
```suggestion
          const paddedScalingFactors = Array.from({ length: 6 }, (_, i) => expectedScalingFactors[i] ?? 0);
```

pkg/pool-stable-phantom/test/StablePoolStorage.test.ts
```diff
@@ -0,0 +1,330 @@
...
+    });
+
+    describe('scaling factors', () => {
+      describe('getScalingFactorX', () => {
+        it('returns the correct scaling factor', async () => {
+          const expectedScalingFactors = tokens.map((token) => fp(1).mul(bn(10).pow(18 - token.decimals)));
+          expectedScalingFactors.splice(bptIndex, 0, fp(1));
+
+          // Also check the individual getters.
+          // There's always 6 getters however not all of them may be used. Unused getters return the zero address.
```
nventuro, line 251:
```suggestion
          // There's always 6 getters however not all of them may be used. Unused getters return zero.
```

pkg/pool-stable-phantom/test/StablePoolStorage.test.ts
```diff
@@ -0,0 +1,330 @@
...
+          const paddedScalingFactors = Array.from({ length: 6 }, (_, i) => expectedScalingFactors[i] ?? ZERO_ADDRESS);
+          expect(await pool.getScalingFactor0()).to.be.eq(paddedScalingFactors[0]);
+          expect(await pool.getScalingFactor1()).to.be.eq(paddedScalingFactors[1]);
+          expect(await pool.getScalingFactor2()).to.be.eq(paddedScalingFactors[2]);
+          expect(await pool.getScalingFactor3()).to.be.eq(paddedScalingFactors[3]);
+          expect(await pool.getScalingFactor4()).to.be.eq(paddedScalingFactors[4]);
+          expect(await pool.getScalingFactor5()).to.be.eq(paddedScalingFactors[5]);
```
nventuro, line 258: I thought you'd do

```
await Promise.all(range(6).map(async i => 
  expect(await pool.[`getScalingFactor${i}`]()).to.be.eq(paddedScalingFactors[i]);
);
```

pkg/pool-stable-phantom/test/StablePoolStorage.test.ts
```diff
@@ -0,0 +1,330 @@
...
+          const insertedElement = bn(420);
+
+          const expectedArray = array.slice();
+          expectedArray.splice(bptIndex, 0, insertedElement);
+          expect(await pool.addBptItem(array, insertedElement)).to.be.deep.eq(expectedArray);
+        });
+      });
+    });
+
+    describe('scaling factors', () => {
```
nventuro, line 244: We're missing tests for `_tokenScalingFactor(token)`

pkg/pool-stable-phantom/test/StablePoolStorage.test.ts
```diff
@@ -0,0 +1,330 @@
...
+
+          const expectedExemptFromYieldProtocolFeeFlags = exemptFromYieldProtocolFeeFlags.slice();
+          expectedExemptFromYieldProtocolFeeFlags.splice(bptIndex, 0, false);
+
+          for (let i = 0; i < allTokens.length; i++) {
+            // Initialized to true for even tokens
+            const expectedFlag = expectedExemptFromYieldProtocolFeeFlags[i];
+            const token = allTokens.get(i);
+
+            expect(await pool.isTokenExemptFromYieldProtocolFee(token.address)).to.equal(expectedFlag);
```
nventuro, line 137: We have two flavors of this, one that receives a token address and one that receives an index. It'd be good to do both.

nventuro (commented):
The spliceoooor

Looks great! Left a couple comments on small variants of functions we're not testing here. We should also include a test for `getTotalTokens()`, and one for `_upscaleWithoutBpt`.

pkg/pool-stable-phantom/test/StablePoolStorage.test.ts
```diff
@@ -0,0 +1,330 @@
...
+    ): Promise<void> {
+      const newRateProviders = [];
+      for (let i = 0; i < numRateProviders; i++) {
+        newRateProviders[i] = (await deploy('v2-pool-utils/MockRateProvider')).address;
+      }
+
+      const newExemptFromYieldProtocolFeeFlags = [];
+      for (let i = 0; i < numExemptFlags; i++) {
+        newExemptFromYieldProtocolFeeFlags[i] = i % 2 == 0; // set true for even tokens
+      }
```
nventuro, line 89: We're not relying on eveness so might as well make this random

PR #1563, opened 2022-08-02
Stable: no comment

EndymionJkb:
Proof we need to finish the tests = this

### DIFF
--- a/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
@@ -542,7 +542,8 @@ contract StablePhantomPool is
         }
 
         // Add amountsIn to get post-join balances
-        _mutateAmounts(balancesWithoutBpt, amountsIn, FixedPoint.add);
+        // amountsIn returned from specific join functions has BPT added back in for the Vault
+        _mutateAmounts(balancesWithoutBpt, _dropBptItem(amountsIn), FixedPoint.add);
 
         // Pass in the post-join balances to reset the protocol fee basis.
         _updateInvariantAfterJoinExit(currentAmp, balancesWithoutBpt);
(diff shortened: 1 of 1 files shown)


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -1167,6 +1169,8 @@ contract StablePhantomPool is
         uint256[] memory arguments,
         function(uint256, uint256) pure returns (uint256) mutation
     ) private pure {
+        InputHelpers.ensureInputLengthMatch(toMutate.length, arguments.length);
```
TomAFrench, line 1172: This just increases runtime cost for no reason. We shouldn't ever be passing in arrays which we don't _know_ are the same length.

TomAFrench (commented):
oof, yeah this isn't nice.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
EndymionJkb: I agree we *shouldn't*, except we just were :) I put it in to make sure there wasn't another issue. I can take it out, but we should make sure the tests cover this (at least indirectly). The fee amounts should eventually be wrong if we try different token combinations, at least before and after the BPT token.

TomAFrench: Yeah, to be clear I'm in favour of us having stuff which makes noise if we hit unexpected behaviour, just that we should remove these from the final build before we deploy. We could just make this a TODO to remove once tests are more comprehensive.

EndymionJkb: Well, I already removed it, hoping we could quietly merge and never speak of this again :)

I hesitate to make notes to do things "later," knowing that all too often, later never comes. I'd definitely keep it in your local copy while we're writing the tests, though.

TomAFrench (approved):

PR #1564, opened 2022-08-02
Stable: Charge protocol fees on Join/Exit swaps

EndymionJkb:
Compare invariant growth (from the current operation) to the BPT growth (which should be less, due to swap fees) to calculate protocol fees due on join- and exit-swaps.


nventuro (commented):
I like the idea of returning the computed invariants from StableMath to avoid recomputation, but making that change at the same time we address the protocol fee feature makes this a bit too complicated.

Could we make this PR be just about adding support for the new feature, regardless of whether we need to compute new invariants, and look into making StableMath better later?

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -681,12 +696,13 @@ contract StablePhantomPool is
             _revert(Errors.UNHANDLED_EXIT_KIND);
         }
 
+        // Add amountsIn to get post-join balances
```
nventuro, line 699: This is an exit: we're subsctracting amounts out

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -681,12 +696,13 @@ contract StablePhantomPool is
...
         // Subtract amountsOut to get post-exit balances
         // amountsOut returned from specific exit functions has BPT added back in for the Vault
         _mutateAmounts(balancesWithoutBpt, _dropBptItem(amountsOut), FixedPoint.sub);
 
         // Pass in the post-exit balances to reset the protocol fee basis.
-        _updateInvariantAfterJoinExit(currentAmp, balancesWithoutBpt);
+        _updateInvariantAfterJoinExit(currentAmp, balancesWithoutBpt, preJoinExitInvariant, virtualSupply, bptAmountIn);
```
nventuro, line 705: `_updateInvariantAfterJoinExit` assumes `bptAmountIn` is an amount out (minted), but here we're burning it.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -908,9 +924,49 @@ contract StablePhantomPool is
...
+
+        uint256 protocolSwapFeePercentage = getProtocolFeePercentageCache(ProtocolFeeType.SWAP);
+
+        if (protocolSwapFeePercentage > 0) {
+            uint256 invariantGrowthRatio = postJoinExitInvariant > preJoinExitInvariant
+                ? postJoinExitInvariant.divDown(preJoinExitInvariant)
+                : preJoinExitInvariant.divDown(postJoinExitInvariant);
+
+            // Compute the bptIncrease = (1 + bptAmount/virtualSupply)
+            uint256 bptIncreaseRatio = FixedPoint.ONE.add(bptAmount.divDown(virtualSupply));
```
nventuro, line 948: This assumes we're minting BPT

nventuro: Also, we'd have less error by simply doing `(bpt + supply) / supply`. It'd consider changing this function arguments to instead receive prejoinexit supply and post join exit supply instead (which will also let us determine if this is a join or an exit, if needed).

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -908,9 +924,49 @@ contract StablePhantomPool is
...
+            // Protocol fees due = (invariant growth / bpt increase - 1) * virtual supply * protocol fee %
+            // For instance, if the invariant growth is 1.05, and the bpt increase is 1.0475, with 1000 supply,
+            // and a protocol fee of 50%, we would mint (1.05/1.0475 - 1) * 1000 * 0.5 = 1.193 BPT.
+
+            if (invariantGrowthRatio > bptIncreaseRatio) {
+                uint256 protocolFeeAmount = invariantGrowthRatio
+                    .divDown(bptIncreaseRatio)
+                    .sub(FixedPoint.ONE)
+                    .mulDown(virtualSupply)
+                    .mulDown(protocolSwapFeePercentage);
```
nventuro, line 961: I'm having some trouble with this, I'll try to derive it via an alternative method and get back to you.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -297,16 +297,21 @@ contract StablePhantomPool is
...
         _upscaleArray(balances, scalingFactors);
         (uint256 virtualSupply, uint256[] memory balancesWithoutBpt) = _payProtocolFeesBeforeJoinExit(balances);
+
         uint256 upscaledAmountGiven = _upscale(swapRequest.amount, scalingFactors[isGivenIn ? indexIn : indexOut]);
 
         (uint256 amp, ) = _getAmplificationParameter();
+        uint256 preJoinExitInvariant = StableMath._calculateInvariant(amp, balancesWithoutBpt);
+        // bptAmount is the amount swapped in or out
+        uint256 bptAmount = upscaledAmountGiven;
+        uint256 amountCalculated;
```
nventuro, line 307: We're not using this here so there's no need to declare it in this bigger scope

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -297,16 +297,21 @@ contract StablePhantomPool is
 
         _upscaleArray(balances, scalingFactors);
         (uint256 virtualSupply, uint256[] memory balancesWithoutBpt) = _payProtocolFeesBeforeJoinExit(balances);
+
         uint256 upscaledAmountGiven = _upscale(swapRequest.amount, scalingFactors[isGivenIn ? indexIn : indexOut]);
 
         (uint256 amp, ) = _getAmplificationParameter();
+        uint256 preJoinExitInvariant = StableMath._calculateInvariant(amp, balancesWithoutBpt);
+        // bptAmount is the amount swapped in or out
+        uint256 bptAmount = upscaledAmountGiven;
```
nventuro, line 306: This is a bit magical tbh, I'd rather we always set bptAmount to the correct value instead of setting it to a potentially wrong value and then fixing it if that's the case

EndymionJkb: Changing the arguments so that this goes away

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
nventuro: This can be addressed later once we have tests.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -296,10 +296,14 @@ contract StablePhantomPool is
         bool isGivenIn = swapRequest.kind == IVault.SwapKind.GIVEN_IN;
 
         _upscaleArray(balances, scalingFactors);
-        (uint256 virtualSupply, uint256[] memory balancesWithoutBpt) = _payProtocolFeesBeforeJoinExit(balances);
+        (uint256 preJoinExitSupply, uint256[] memory balancesWithoutBpt) = _payProtocolFeesBeforeJoinExit(balances);
+        // Initialize, then add or subtract to reflect BPT transferred in or out
+        uint256 postJoinExitSupply = preJoinExitSupply;
```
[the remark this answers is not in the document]
nventuro: We'lll adress this as part of the refactor.

PR #1566, opened 2022-08-02
Make LinearPool targets use fewer bits

nventuro:
Instead of storing the target in all of its full precision glory, we downscale them by 1e18 on writes and upscale on reads. This means they'll have no fractional part.

With this change, I was able to reduce the number of bits each target uses to 32 (since the largest target will still be ~4 billion for an 18 decimal token). The extra space in misc data will be used in a follow-up PR to store the total supply.

### DIFF
--- a/pkg/pool-linear/contracts/LinearPool.sol
+++ b/pkg/pool-linear/contracts/LinearPool.sol
@@ -610,7 +610,7 @@ abstract contract LinearPool is ILinearPool, IGeneralPool, IRateProvider, BasePo
         _require(lowerTarget <= upperTarget, Errors.LOWER_GREATER_THAN_UPPER_TARGET);
         _require(upperTarget <= _MAX_UPPER_TARGET, Errors.UPPER_TARGET_TOO_HIGH);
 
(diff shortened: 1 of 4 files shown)


pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -600,11 +610,15 @@ abstract contract LinearPool is ILinearPool, IGeneralPool, IRateProvider, BasePo
         _require(lowerTarget <= upperTarget, Errors.LOWER_GREATER_THAN_UPPER_TARGET);
         _require(upperTarget <= _MAX_UPPER_TARGET, Errors.UPPER_TARGET_TOO_HIGH);
 
-        // Pack targets as two uint96 values into a single storage slot. This results in targets being capped to 96
-        // bits, but that should be more than enough. Values are already checked for validity above.
+        // Targets are stored downscaled by _TARGET_SCALING to make them fit in _TARGET_BITS, at the cost of some
```
TomAFrench, line 613:
```suggestion
        // Targets are stored downscaled by _TARGET_SCALING to make them fit in _TARGET_BITS at the cost of some
```

TomAFrench (approved):

PR #1567, opened 2022-08-02
Add _beforeSwapJoinExit()

nventuro:
This helper doesn't do much for now, but it is a stepping stone towards fixing ComposableStablePools behaving incorrectly during queries, and also preventing these types of mistakes in the future.

It also lets us remove a bunch of lines from the stable pool, which is a nice bonus.

### DIFF
--- a/pkg/pool-utils/contracts/BaseGeneralPool.sol
+++ b/pkg/pool-utils/contracts/BaseGeneralPool.sol
@@ -34,9 +34,8 @@ abstract contract BaseGeneralPool is IGeneralPool, BasePool {
         uint256[] memory balances,
         uint256 indexIn,
         uint256 indexOut
-    ) public virtual override onlyVault(swapRequest.poolId) returns (uint256) {
-        // Block all swaps when paused
-        _ensureNotPaused();
+    ) public override onlyVault(swapRequest.poolId) returns (uint256) {
+        _beforeSwapJoinExit();
 
         _validateIndexes(indexIn, indexOut, _getTotalTokens());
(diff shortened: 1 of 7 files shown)


pkg/pool-utils/contracts/BaseGeneralPool.sol
```diff
@@ -34,9 +34,8 @@ abstract contract BaseGeneralPool is IGeneralPool, BasePool {
         uint256[] memory balances,
         uint256 indexIn,
         uint256 indexOut
-    ) public virtual override onlyVault(swapRequest.poolId) returns (uint256) {
-        // Block all swaps when paused
-        _ensureNotPaused();
+    ) public override onlyVault(swapRequest.poolId) returns (uint256) {
```
EndymionJkb, line 37: Maybe add a comment on the reason for removing virtual here (and in corresponding join/exit hooks). You're actually preventing derived contracts from overriding onSwap now, because doing so (and adding extra functions, as we did) would make the query functions and the live functions diverge.

EndymionJkb (approved):
Will make things easier going forward, and future-proofs against people doing things that would break the equivalence between query and live functions.

EndymionJkb (approved):
Good again; just had to fix a merge conflict

pkg/pool-utils/contracts/BaseGeneralPool.sol
nventuro: Yes, but I want to add that later once I fix queries + add tests.

PR #1570, opened 2022-08-03
Stakeless gauge controller

jubeira:
This PR adds the `StakelessGaugeController`, which helps automate the process of performing checkpoints for multiple stakeless gauges at once.

### DIFF
--- a/pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
+++ b/pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
@@ -1,0 +1,29 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+
+import "@balancer-labs/v2-interfaces/contracts/liquidity-mining/IStakelessGauge.sol";
+
+interface IStakelessGaugeController {
+    enum GaugeType { POLYGON, ARBITRUM, OPTIMISM }
+
+    function addGauges(GaugeType gaugeType, IStakelessGauge[] calldata gauges) external;
+    function removeGauges(GaugeType gaugeType, IStakelessGauge[] calldata gauges) external;
+    function hasGauge(GaugeType gaugeType, IStakelessGauge gauge) external view returns (bool);
+    function getTotalGauges(GaugeType gaugeType) external view returns (uint256);
+    function getGaugeAt(GaugeType gaugeType, uint256 index) external view returns (address);
+
+    function checkpointGaugesAboveRelativeWeight(uint256 minRelativeWeight) external payable;
+}
--- a/pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
+++ b/pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
@@ -16,14 +16,55 @@ pragma solidity ^0.7.0;
 
 import "@balancer-labs/v2-interfaces/contracts/liquidity-mining/IStakelessGauge.sol";
 
+/**
+ * @title Stakeless Gauge Controller interface
+ * @notice Manages checkpoints for registered stakeless gauges, allowing to perform mutiple checkpoints in a
+ * single call.
+ * @dev Supports Polygon, Arbitrum and Optimism stakeless root gauges. Gauges to be checkpointed need to be
+ * registered (added) to the controller beforehand.
+ */
 interface IStakelessGaugeController {
     enum GaugeType { POLYGON, ARBITRUM, OPTIMISM }
 
+    /**
+    * @dev Registers an array of gauges from the given type.
+    * @param gaugeType - Type of the gauge.
+    * @param gauges - Gauges to register.
+    */
     function addGauges(GaugeType gaugeType, IStakelessGauge[] calldata gauges) external;
+
+    /**
+    * @dev De-registers an array of gauges from the given type.
+    * @param gaugeType - Type of the gauge.
+    * @param gauges - Gauges to de-register.
+    */
     function removeGauges(GaugeType gaugeType, IStakelessGauge[] calldata gauges) external;
+
+    /**
+    * @dev Returns true if the given gauge was registered for the given type; false otherwise.
+    * @param gaugeType - Type of the gauge.
+    * @param gauge - Gauge to check.
+    */
     function hasGauge(GaugeType gaugeType, IStakelessGauge gauge) external view returns (bool);
+
+    /**
+    * @dev Returns the amount of registered gauges for a given type.
+    * @param gaugeType - Type of the gauge.
+    */
     function getTotalGauges(GaugeType gaugeType) external view returns (uint256);
+
+    /**
+    * @dev Returns the gauge of a given type at the given index.
+    * Reverts if the index is greater than or equal to the amount of registered gauges for the given type.
+    * @param gaugeType - Type of the gauge.
+    * @param index - Index of the registered gauge.
+    */
     function getGaugeAt(GaugeType gaugeType, uint256 index) external view returns (address);
 
+    /**
+    * @dev Performs a checkpoint for all registered gauges above the given relative weight threshold.
+    * Reverts if the ETH sent in the call is not enough to cover bridge costs.
+    * @param minRelativeWeight - Threshold to filter out gauges below it.
+    */
     function checkpointGaugesAboveRelativeWeight(uint256 minRelativeWeight) external payable;
 }
--- a/pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
+++ b/pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
@@ -14,44 +14,54 @@
 
 pragma solidity ^0.7.0;
 
+import "@balancer-labs/v2-interfaces/contracts/liquidity-mining/IGaugeAdder.sol";
 import "@balancer-labs/v2-interfaces/contracts/liquidity-mining/IStakelessGauge.sol";
 
 /**
  * @title Stakeless Gauge Controller interface
  * @notice Manages checkpoints for registered stakeless gauges, allowing to perform mutiple checkpoints in a
  * single call.
- * @dev Supports Polygon, Arbitrum and Optimism stakeless root gauges. Gauges to be checkpointed need to be
- * registered (added) to the controller beforehand.
+ * @dev Supports Polygon, Arbitrum, Optimism, Gnosis and ZKSync stakeless root gauges. Gauges to be checkpointed need
+ * to be registered (added) to the controller beforehand.
  */
 interface IStakelessGaugeController {
-    enum GaugeType { POLYGON, ARBITRUM, OPTIMISM }
+
+    /**
+    * @dev Emitted when stakeless gauges are added to the controller.
+    */
+    event GaugesAdded(IGaugeAdder.GaugeType gaugeType, IStakelessGauge[] gauges);
+
+    /**
+    * @dev Emitted when stakeless gauges are removed from the controller.
+    */
+    event GaugesRemoved(IGaugeAdder.GaugeType gaugeType, IStakelessGauge[] gauges);
 
     /**
     * @dev Registers an array of gauges from the given type.
     * @param gaugeType - Type of the gauge.
     * @param gauges - Gauges to register.
     */
-    function addGauges(GaugeType gaugeType, IStakelessGauge[] calldata gauges) external;
+    function addGauges(IGaugeAdder.GaugeType gaugeType, IStakelessGauge[] calldata gauges) external;
 
     /**
     * @dev De-registers an array of gauges from the given type.
     * @param gaugeType - Type of the gauge.
     * @param gauges - Gauges to de-register.
     */
-    function removeGauges(GaugeType gaugeType, IStakelessGauge[] calldata gauges) external;
+    function removeGauges(IGaugeAdder.GaugeType gaugeType, IStakelessGauge[] calldata gauges) external;
 
     /**
     * @dev Returns true if the given gauge was registered for the given type; false otherwise.
     * @param gaugeType - Type of the gauge.
     * @param gauge - Gauge to check.
     */
-    function hasGauge(GaugeType gaugeType, IStakelessGauge gauge) external view returns (bool);
+    function hasGauge(IGaugeAdder.GaugeType gaugeType, IStakelessGauge gauge) external view returns (bool);
 
     /**
     * @dev Returns the amount of registered gauges for a given type.
     * @param gaugeType - Type of the gauge.
     */
-    function getTotalGauges(GaugeType gaugeType) external view returns (uint256);
+    function getTotalGauges(IGaugeAdder.GaugeType gaugeType) external view returns (uint256);
 
     /**
     * @dev Returns the gauge of a given type at the given index.
@@ -59,7 +69,7 @@ interface IStakelessGaugeController {
     * @param gaugeType - Type of the gauge.
     * @param index - Index of the registered gauge.
     */
-    function getGaugeAt(GaugeType gaugeType, uint256 index) external view returns (address);
+    function getGaugeAt(IGaugeAdder.GaugeType gaugeType, uint256 index) external view returns (address);
 
     /**
     * @dev Performs a checkpoint for all registered gauges above the given relative weight threshold.
--- a/pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
+++ b/pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
@@ -56,10 +56,11 @@ contract StakelessGaugeController is IStakelessGaugeController, ReentrancyGuard 
     /**
      * @dev See IStakelessGaugeController#addGauges.
      */
-    function addGauges(
-        IGaugeAdder.GaugeType gaugeType,
-        IStakelessGauge[] calldata gauges
-    ) external override gaugeTypeChecked(gaugeType) {
+    function addGauges(IGaugeAdder.GaugeType gaugeType, IStakelessGauge[] calldata gauges)
+        external
+        override
+        gaugeTypeChecked(gaugeType)
+    {
         EnumerableSet.AddressSet storage gaugesForType = _gauges[gaugeType];
 
         for (uint256 i = 0; i < gauges.length; i++) {
@@ -78,10 +79,11 @@ contract StakelessGaugeController is IStakelessGaugeController, ReentrancyGuard 
     /**
      * @dev See IStakelessGaugeController#removeGauges.
      */
-    function removeGauges(
-        IGaugeAdder.GaugeType gaugeType,
-        IStakelessGauge[] calldata gauges
-    ) external override gaugeTypeChecked(gaugeType) {
+    function removeGauges(IGaugeAdder.GaugeType gaugeType, IStakelessGauge[] calldata gauges)
+        external
+        override
+        gaugeTypeChecked(gaugeType)
+    {
         EnumerableSet.AddressSet storage gaugesForType = _gauges[gaugeType];
 
         for (uint256 i = 0; i < gauges.length; i++) {
@@ -95,29 +97,39 @@ contract StakelessGaugeController is IStakelessGaugeController, ReentrancyGuard 
     /**
      * @dev See IStakelessGaugeController#hasGauge.
      */
-    function hasGauge(
-        IGaugeAdder.GaugeType gaugeType,
-        IStakelessGauge gauge
-    ) external view override gaugeTypeChecked(gaugeType) returns (bool) {
+    function hasGauge(IGaugeAdder.GaugeType gaugeType, IStakelessGauge gauge)
+        external
+        view
+        override
+        gaugeTypeChecked(gaugeType)
+        returns (bool)
+    {
         return _gauges[gaugeType].contains(address(gauge));
     }
 
     /**
      * @dev See IStakelessGaugeController#getTotalGauges.
      */
-    function getTotalGauges(
-        IGaugeAdder.GaugeType gaugeType
-    ) external view override gaugeTypeChecked(gaugeType) returns (uint256) {
+    function getTotalGauges(IGaugeAdder.GaugeType gaugeType)
+        external
+        view
+        override
+        gaugeTypeChecked(gaugeType)
+        returns (uint256)
+    {
         return _gauges[gaugeType].length();
     }
 
     /**
      * @dev See IStakelessGaugeController#getGaugeAt.
      */
-    function getGaugeAt(
-        IGaugeAdder.GaugeType gaugeType,
-        uint256 index
-    ) external view override gaugeTypeChecked(gaugeType) returns (address) {
+    function getGaugeAt(IGaugeAdder.GaugeType gaugeType, uint256 index)
+        external
+        view
+        override
+        gaugeTypeChecked(gaugeType)
+        returns (address)
+    {
         return _gauges[gaugeType].at(index);
     }
 
@@ -163,13 +175,13 @@ contract StakelessGaugeController is IStakelessGaugeController, ReentrancyGuard 
     }
 
     /**
-    * @dev Performs checkpoints for all gauges of the given type whose relative weight is at least the specified one.
-    * @param gaugeType - Type of the gauges to checkpoint.
-    * @param minRelativeWeight - Threshold to filter out gauges below it.
-    * @param currentPeriod - Current block time rounded down to the start of the week.
-    * @param costPerCheckpoint - Value in ETH to be spent for each gauge to checkpoint to cover bridging costs.
-    * This method doesn't check whether the caller transferred enough ETH to cover the whole operation.
-    */
+     * @dev Performs checkpoints for all gauges of the given type whose relative weight is at least the specified one.
+     * @param gaugeType - Type of the gauges to checkpoint.
+     * @param minRelativeWeight - Threshold to filter out gauges below it.
+     * @param currentPeriod - Current block time rounded down to the start of the week.
+     * @param costPerCheckpoint - Value in ETH to be spent for each gauge to checkpoint to cover bridging costs.
+     * This method doesn't check whether the caller transferred enough ETH to cover the whole operation.
+     */
     function _checkpointGauges(
         IGaugeAdder.GaugeType gaugeType,
         uint256 minRelativeWeight,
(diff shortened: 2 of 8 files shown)


pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
```diff
@@ -0,0 +1,80 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+
+import "@balancer-labs/v2-interfaces/contracts/liquidity-mining/IGaugeAdder.sol";
+import "@balancer-labs/v2-interfaces/contracts/liquidity-mining/IStakelessGauge.sol";
+
+/**
+ * @title Stakeless Gauge Controller interface
```
nventuro, line 21: A better name would be the 'GaugeCheckpointer', as there's already a GaugeController (which does something very different from this).

nventuro: Maybe even L2GaugeCheckpointer, since we don't checkpoint all stakeless gauges but only the L2 root ones.

pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
```diff
@@ -0,0 +1,80 @@
...
+    * @param index - Index of the registered gauge.
+    */
+    function getGaugeAt(IGaugeAdder.GaugeType gaugeType, uint256 index) external view returns (address);
+
+    /**
+    * @dev Performs a checkpoint for all registered gauges above the given relative weight threshold.
+    * Reverts if the ETH sent in the call is not enough to cover bridge costs.
+    * @param minRelativeWeight - Threshold to filter out gauges below it.
+    */
+    function checkpointGaugesAboveRelativeWeight(uint256 minRelativeWeight) external payable;
```
nventuro, line 78: I'd add `isValidGaugeType`

pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
```diff
@@ -0,0 +1,80 @@
...
+    * @dev Returns the gauge of a given type at the given index.
+    * Reverts if the index is greater than or equal to the amount of registered gauges for the given type.
+    * @param gaugeType - Type of the gauge.
+    * @param index - Index of the registered gauge.
+    */
+    function getGaugeAt(IGaugeAdder.GaugeType gaugeType, uint256 index) external view returns (address);
+
+    /**
+    * @dev Performs a checkpoint for all registered gauges above the given relative weight threshold.
+    * Reverts if the ETH sent in the call is not enough to cover bridge costs.
```
nventuro: Let's then have a getter that returns what these costs are for a given `minRelativeWeight`

pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+    * @dev Emitted when stakeless gauges are added to the controller.
+    */
+    event GaugesAdded(IGaugeAdder.GaugeType gaugeType, IStakelessGauge[] gauges);
+
+    /**
+    * @dev Emitted when stakeless gauges are removed from the controller.
+    */
+    event GaugesRemoved(IGaugeAdder.GaugeType gaugeType, IStakelessGauge[] gauges);
```
nventuro: It's better to a emit one event per gauge as opposed to including the full array. We should also make both the gauge type and the gauge itself `indexed`.

pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+    * @dev Registers an array of gauges from the given type.
+    * @param gaugeType - Type of the gauge.
+    * @param gauges - Gauges to register.
+    */
+    function addGauges(IGaugeAdder.GaugeType gaugeType, IStakelessGauge[] calldata gauges) external;
```
nventuro: I'd try to use consistent language - either call this 'add' or 'register', but not both.

nventuro: I'd also list the conditions a gauge must meet to be added, and explain _why_ they have to be added (to checkpoint them later)

pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+    * @dev De-registers an array of gauges from the given type.
```
nventuro: Same comment re. why and requirements

pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
```diff
@@ -0,0 +1,80 @@
...
+    */
+    function getTotalGauges(IGaugeAdder.GaugeType gaugeType) external view returns (uint256);
+
+    /**
+    * @dev Returns the gauge of a given type at the given index.
+    * Reverts if the index is greater than or equal to the amount of registered gauges for the given type.
+    * @param gaugeType - Type of the gauge.
+    * @param index - Index of the registered gauge.
+    */
+    function getGaugeAt(IGaugeAdder.GaugeType gaugeType, uint256 index) external view returns (address);
```
nventuro, line 71:
```suggestion
    function getGaugeAt(IGaugeAdder.GaugeType gaugeType, uint256 index) external view returns (IStakelessGauge);
```

pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
```diff
@@ -0,0 +1,223 @@
...
+
+    mapping(IGaugeAdder.GaugeType => EnumerableSet.AddressSet) private _gauges;
+    IAuthorizerAdaptor private immutable _authorizerAdaptor;
+    IGaugeController private immutable _gaugeController;
+    IGaugeAdder private immutable _gaugeAdder;
+
+    constructor(
+        IAuthorizerAdaptor authorizerAdaptor,
+        IGaugeController gaugeController,
+        IGaugeAdder gaugeAdder
```
nventuro, line 44: You actually only need the adder - you can get the adaptor and controller from it (`getGaugeController()`, `getAuthorizerAdaptor()`).

pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
```diff
@@ -0,0 +1,223 @@
...
+    ) {
+        _authorizerAdaptor = authorizerAdaptor;
+        _gaugeController = gaugeController;
+        _gaugeAdder = gaugeAdder;
+    }
+
+    modifier gaugeTypeChecked(IGaugeAdder.GaugeType gaugeType) {
+        require(_isValidGaugeType(gaugeType), "Unsupported gauge type");
```
nventuro, line 52: If you're calling this concept 'valid gauge type', then I'd call the modifier `withValidGaugeType(type)` instead. Similar to the comment in add/register, try to avoid introducing new terms ('type checked' vs 'valid type').

pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
```diff
@@ -0,0 +1,223 @@
...
+    function removeGauges(IGaugeAdder.GaugeType gaugeType, IStakelessGauge[] calldata gauges)
+        external
+        override
+        gaugeTypeChecked(gaugeType)
+    {
+        EnumerableSet.AddressSet storage gaugesForType = _gauges[gaugeType];
+
+        for (uint256 i = 0; i < gauges.length; i++) {
+            IStakelessGauge gauge = gauges[i];
+            require(gauge.is_killed(), "Gauge was not killed");
```
nventuro, line 91: I'd add a comment here mentioning we're not checking if the gauge comes from the factory or if it exists in the controller because we've already done that (as we're checking the gauge was added to our list), and gauges cannot be removed from either factory nor controller. Therefore, the only way to get a gauge to be removed here is to kill it.

pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
```diff
@@ -0,0 +1,223 @@
...
+     * @dev See IStakelessGaugeController#checkpointGaugesAboveRelativeWeight.
+     * Unspent ETH is sent back to sender.
+     */
+    function checkpointGaugesAboveRelativeWeight(uint256 minRelativeWeight) external payable override nonReentrant {
+        (uint256 singleArbGaugeBridgeETH, uint256 totalArbBridgeETH) = _getArbitrumBridgeCosts();
+
+        // solhint-disable-next-line not-rely-on-time
+        uint256 currentPeriod = _roundDownTimestamp(block.timestamp);
+
+        _checkpointGauges(IGaugeAdder.GaugeType.Polygon, minRelativeWeight, currentPeriod, 0);
```
nventuro, line 146: I'd consider computing the value _inside_ `_checkpointGauges` instead of here (i.e. 0 if non arbitrum, whatever the bridge returns otherwise). That'll simplify some things, such as making it obvious why it is correct to return 0 in the current `getBridgeCosts()` impl if we have no arbitrum gauges (which is right now a bit indirect).

pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
```diff
@@ -0,0 +1,223 @@
...
+        _checkpointGauges(IGaugeAdder.GaugeType.Arbitrum, minRelativeWeight, currentPeriod, singleArbGaugeBridgeETH);
+        _checkpointGauges(IGaugeAdder.GaugeType.Optimism, minRelativeWeight, currentPeriod, 0);
+        _checkpointGauges(IGaugeAdder.GaugeType.Gnosis, minRelativeWeight, currentPeriod, 0);
+        _checkpointGauges(IGaugeAdder.GaugeType.ZKSync, minRelativeWeight, currentPeriod, 0);
+
+        // transfer msg.value - total spent ETH back to sender.
+        // _getArbitrumBridgeCosts ensures the difference is always positive.
+        Address.sendValue(msg.sender, msg.value - totalArbBridgeETH);
```
nventuro, line 154:
```suggestion
        // Send back any leftover ETH to the caller
        Address.sendValue(msg.sender, address(this).balance);
```

This is both cheaper in terms of gas and less error-prone.

pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
```diff
@@ -0,0 +1,223 @@
...
+     * This method doesn't check whether the caller transferred enough ETH to cover the whole operation.
+     */
+    function _checkpointGauges(
+        IGaugeAdder.GaugeType gaugeType,
+        uint256 minRelativeWeight,
+        uint256 currentPeriod,
+        uint256 costPerCheckpoint
+    ) private {
+        uint256 totalGauges = _gauges[gaugeType].length();
+        EnumerableSet.AddressSet storage gaugeAddressSet = _gauges[gaugeType];
```
nventuro, line 192: I'd call this `typeGauges`, but that's just me. Including the type in the name feels a bit boilerplatey.

pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
```diff
@@ -0,0 +1,223 @@
...
+     * @param costPerCheckpoint - Value in ETH to be spent for each gauge to checkpoint to cover bridging costs.
+     * This method doesn't check whether the caller transferred enough ETH to cover the whole operation.
+     */
+    function _checkpointGauges(
+        IGaugeAdder.GaugeType gaugeType,
+        uint256 minRelativeWeight,
+        uint256 currentPeriod,
+        uint256 costPerCheckpoint
+    ) private {
+        uint256 totalGauges = _gauges[gaugeType].length();
```
nventuro, line 191: Consider `totalTypeGauges` instead - the current name is slightly misleading.

pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
```diff
@@ -0,0 +1,223 @@
...
+        return (timestamp / 1 weeks) * 1 weeks;
+    }
+
+    /**
+     * @dev Returns true if gauge type is Polygon, Arbitrum, Optimism, Gnosis or ZKSync; false otherwise.
+     */
+    function _isValidGaugeType(IGaugeAdder.GaugeType gaugeType) private pure returns (bool) {
+        return
+            uint8(gaugeType) >= uint8(IGaugeAdder.GaugeType.Polygon) &&
+            uint8(gaugeType) <= uint8(IGaugeAdder.GaugeType.ZKSync);
```
nventuro, line 221: :scream:

A big `type == enum.Value || ...` chain would be much more explicit :grin: The gas savings this would introduce are minimal compared to the rest of the execution.

pkg/liquidity-mining/contracts/test/MockGaugeController.sol
```diff
@@ -53,8 +53,12 @@ contract MockGaugeController is IGaugeController {
         emit NewGauge(gauge, gaugeType, 0);
     }
 
-    function add_type(string calldata, uint256) external override {
-        _numGaugeTypes += 1;
+    function add_type(string calldata, uint256 numTypes) external override {
```
nventuro, line 56: The second argument of `add_type` is not `numTypes` but rather the new type's initial weight. You can simply ignore it.

nventuro (commented):
Fantastic first approach, congratulations! :grin:

I left a bunch of comments and suggestions on the Solidity code, didn't get to review the tests yet.

pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
jubeira: Ack; done!

pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
jubeira: Done; register --> add.

pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
jubeira: Done.

pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
jubeira: Done.
As discussed offline, this getter shares some code with `checkpointGauges` without reusing it, but this is the best approach so as not to make the code more cumbersome unnecesarily.

pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
jubeira: This one doesn't receive the gauge type, since it just performs the checkpoint for all the ones that are valid.

Therefore, no check should be needed. Am I missing something?

pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
jubeira: Great, good catch.

Changed the parameter from `IGaugeAdder` to `GaugeAdder`, since one of the getters was not declared in the interface.

pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
jubeira: +1; done.

pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
jubeira: Done.

pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
jubeira: Done.

pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
jubeira: Ah, that's a good one. This doesn't need the total eth consumed so far, which is great.

pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
jubeira: +1

pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
jubeira: +1

pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
jubeira: done

pkg/liquidity-mining/contracts/test/MockGaugeController.sol
jubeira: Reverted (this was just to do a shortcut in the test, no real need for it).

jubeira (commented):
Thanks for the review! First pass of comments addressed; I've left one question.

pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
nventuro: Sorry, I mean that I'd add that getter to the interface.

pkg/interfaces/contracts/liquidity-mining/IStakelessGaugeController.sol
nventuro: We should also add it to the interface.

pkg/liquidity-mining/contracts/gauges/StakelessGaugeController.sol
nventuro: Add the getter to IGaugeAdder instead: if you import GaugeAdder, that causes the contract to depend on the entire source code, which means a) when uploading to ethercan we'll also upload all of gauge adder, and b) people who want to compile it will need the full GaugeAdder as well.

pkg/liquidity-mining/contracts/gauges/L2GaugeCheckpointer.sol
```diff
@@ -0,0 +1,239 @@
...
+    IGaugeAdder private immutable _gaugeAdder;
+
+    constructor(GaugeAdder gaugeAdder) {
+        _gaugeAdder = gaugeAdder;
+        _gaugeController = gaugeAdder.getGaugeController();
+        _authorizerAdaptor = gaugeAdder.getAuthorizerAdaptor();
+    }
+
+    modifier withValidGaugeType(IGaugeAdder.GaugeType gaugeType) {
+        require(_isValidGaugeType(gaugeType), "Unsupported gauge type");
```
nventuro, line 49: The error message is actually correct, this should be called `withSupportedGaugeType` - the other types are not invalid, we just don't support them.

pkg/liquidity-mining/contracts/gauges/L2GaugeCheckpointer.sol
```diff
@@ -0,0 +1,239 @@
...
+    {
+        EnumerableSet.AddressSet storage gaugesForType = _gauges[gaugeType];
+
+        for (uint256 i = 0; i < gauges.length; i++) {
+            IStakelessGauge gauge = gauges[i];
+            require(
+                _gaugeAdder.isGaugeFromValidFactory(address(gauge), gaugeType),
+                "Gauge does not come from valid factory"
+            );
+            require(_gaugeController.gauge_exists(address(gauge)), "Gauge does not exist in controller");
```
nventuro, line 69:
```suggestion
            require(_gaugeController.gauge_exists(address(gauge)), "Gauge was not added to the GaugeController");
```

This is more actionable

pkg/liquidity-mining/contracts/gauges/L2GaugeCheckpointer.sol
```diff
@@ -0,0 +1,239 @@
...
+        override
+        withValidGaugeType(gaugeType)
+    {
+        EnumerableSet.AddressSet storage gaugesForType = _gauges[gaugeType];
+
+        for (uint256 i = 0; i < gauges.length; i++) {
+            IStakelessGauge gauge = gauges[i];
+            require(
+                _gaugeAdder.isGaugeFromValidFactory(address(gauge), gaugeType),
+                "Gauge does not come from valid factory"
```
nventuro, line 67:
```suggestion
                "Gauge does not come from a valid factory"
```

pkg/liquidity-mining/contracts/gauges/L2GaugeCheckpointer.sol
```diff
@@ -0,0 +1,239 @@
...
+            );
+            require(_gaugeController.gauge_exists(address(gauge)), "Gauge does not exist in controller");
+            require(!gauge.is_killed(), "Gauge was killed");
+            require(gaugesForType.add(address(gauge)), "Gauge already present");
```
nventuro, line 71:
```suggestion
            require(gaugesForType.add(address(gauge)), "Gauge already added to the checkpointer");
```

pkg/liquidity-mining/contracts/gauges/L2GaugeCheckpointer.sol
```diff
@@ -0,0 +1,239 @@
...
+        withValidGaugeType(gaugeType)
+    {
+        EnumerableSet.AddressSet storage gaugesForType = _gauges[gaugeType];
+
+        for (uint256 i = 0; i < gauges.length; i++) {
+            // Gauges added must come from a valid factory and exist in the controller, and they can't be removed from
+            // them. Therefore, the only required check at this point is whether the gauge was killed.
+            IStakelessGauge gauge = gauges[i];
+            require(gauge.is_killed(), "Gauge was not killed");
+            require(gaugesForType.remove(address(gauge)), "Gauge not present");
```
nventuro, line 92:
```suggestion
            require(gaugesForType.remove(address(gauge)), "Gauge was not added to the checkpointer");
```

pkg/liquidity-mining/contracts/gauges/L2GaugeCheckpointer.sol
```diff
@@ -0,0 +1,239 @@
...
+        for (uint256 i = 0; i < totalArbitrumGauges; ++i) {
+            address gauge = arbitrumGauges.unchecked_at(i);
+            // Skip gauges that are below the threshold.
+            if (_gaugeController.gauge_relative_weight(gauge, currentPeriod) < minRelativeWeight) {
+                continue;
+            } else {
+                // Cost per gauge is always the same, but getting the cost every time makes the code simpler,
+                // and this function is only to be used off-chain.
+                totalCost += ArbitrumRootGauge(gauge).getTotalBridgeCost();
+            }
```
nventuro, line 174:
```suggestion
            }

            // Cost per gauge is always the same, but getting the cost every time makes the code simpler,
            // and this function is only to be used off-chain.
            totalCost += ArbitrumRootGauge(gauge).getTotalBridgeCost();
```

pkg/liquidity-mining/test/L2GaugeCheckpointer.test.ts
```diff
@@ -0,0 +1,223 @@
...
+    context('with correct factory and wrong controller setup', () => {
+      it('reverts', async () => {
+        await expect(L2GaugeCheckpointer.addGauges(testGaugeType, testGauges)).to.be.revertedWith(
+          'Gauge does not exist in controller'
+        );
+      });
+    });
+
+    context('with correct factory and controller setup', () => {
+      sharedBeforeEach('add gauges to regular controller', async () => {
```
nventuro, line 117:
```suggestion
      sharedBeforeEach('add gauges to controller', async () => {
```

pkg/liquidity-mining/test/L2GaugeCheckpointer.test.ts
```diff
@@ -0,0 +1,223 @@
...
+        }
+        await expectGaugesAdded(testGaugeType, testGauges);
+        // Check that all remaining types are empty.
+        await Promise.all(
+          GAUGE_TYPES.filter((gaugeType) => gaugeType != testGaugeType).map((gaugeType) =>
+            expectGaugesAdded(gaugeType, [])
+          )
+        );
+      });
+
```
nventuro, line 135: You can split this into three tests: one where you check the gauges have been added (i.e. the contract returns true in `hasGauge` for all of them), one where you test the events (`it('emits one event per added gauge')`), and one where you enumerate the gauges and find all of them. You'd start all three `it`s with `await checkpointer.adddGauges(...)`.

pkg/liquidity-mining/test/L2GaugeCheckpointer.test.ts
```diff
@@ -0,0 +1,223 @@
...
+      it('removes added stakeless gauges correctly ', async () => {
+        expect(await L2GaugeCheckpointer.getTotalGauges(testGaugeType)).to.be.eq(GAUGES_PER_TYPE);
+        const tx = await L2GaugeCheckpointer.removeGauges(testGaugeType, testGauges);
+        const receipt = await tx.wait();
+        for (const testGauge of testGauges) {
+          expectEvent.inReceipt(receipt, 'GaugeRemoved', { gaugeType: testGaugeType, gauge: testGauge });
+        }
+        expect(await L2GaugeCheckpointer.getTotalGauges(testGaugeType)).to.be.eq(0);
+      });
```
nventuro, line 189: Same comment as for `add`.

pkg/liquidity-mining/test/L2GaugeCheckpointer.test.ts
```diff
@@ -0,0 +1,223 @@
...
+        gauges.set(factory.type, await createGauges(factory.contract, factory.type, GAUGES_PER_TYPE))
+      )
+    );
+  });
+
+  before('get test gauges', () => {
+    testGaugeType = GaugeType.Polygon;
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    testGauges = gauges.get(testGaugeType)!;
+  });
```
nventuro, line 80: You can take all of the tests below and put them in a function that receives `gaugeType`, e.g.:

```
fn itAddsAndRemoves(...) {
  describe(`for type ${GaugeType[type]}`, () => {
    ...
  });
}
```

And then call it repeatedly with all types, getting tests for all of them. I'd then have tests outside of this function for invalid types.

jubeira (commented):
Comments on the Solidity side addressed; I'll do one more pass over the tests now.

pkg/liquidity-mining/contracts/gauges/L2GaugeCheckpointer.sol
jubeira: +1, changed to `withSupportedGaugeType` and `isSupportedGaugeType`.

pkg/liquidity-mining/test/L2GaugeCheckpointer.test.ts
jubeira: +1; done.

pkg/liquidity-mining/test/L2GaugeCheckpointer.test.ts
jubeira: +1; done.

pkg/liquidity-mining/test/L2GaugeCheckpointer.test.ts
jubeira: I started writing the tests with this in mind; at some point I convinced myself it was an overkill.

done :)

jubeira:
Comments on tests addressed; I'm pretty satisfied with the result.
Last step: add some tests for unsupported types.

jubeira:
I've just added the tests for unsupported gauge types. This is ready for one more pass.

pvt/helpers/src/address.ts
```diff
@@ -0,0 +1,20 @@
+import { ethers } from 'hardhat';
+import { ANY_ADDRESS } from './constants';
+
+/**
+ * Creates an array of pseudo random addresses.
+ * @param seed Number to start generating the pseudo random addresses. Use different inputs to get different outputs.
+ * @param amount Number of addresses to create.
+ */
+export const anyAddressArray = (seed: number, amount: number): string[] => {
```
nventuro, line 9: Isn't this just `range(n).map(randomAddress)`?

nventuro (approved):

pvt/helpers/src/address.ts
jubeira: well... 

yes :grimacing: this is my 3m old TS-foo

PR #1571, opened 2022-08-03
Refactor swaps, joins and exits in StablePhantom

nventuro:
This is very large - I'm not sure how I'd go about reviewing the diff. Perhaps the point is that the diff is hard to review because the original code was hard to follow. Instead, a better strategy might be to simply read the new code, which is (hopefully) much more straightforward.

Some documentation is missing, but the rough shape is there. We now work almost exclusively with BPT-less arrays and indices, and only adjust those at the borders. With these changes in, we can rework the storage layer to better match what we need.

In the process I also addressed the famous issue where upscaleArray would have an extra require check, in what I think is a nice way.

### DIFF
--- a/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
@@ -248,7 +248,7 @@ contract StablePhantomPool is
         uint256[] memory balancesIncludingBpt,
         uint256 indexIn,
         uint256 indexOut
-    ) private view returns (uint256 amountCalculated) {
+    ) private view returns (uint256) {
         uint256[] memory balances = _dropBptItem(balancesIncludingBpt);
         (uint256 currentAmp, ) = _getAmplificationParameter();
         uint256 invariant = StableMath._calculateInvariant(currentAmp, balances);
@@ -257,25 +257,10 @@ contract StablePhantomPool is
         indexIn = _skipBptIndex(indexIn);
         indexOut = _skipBptIndex(indexOut);
 
-        // Would like to use a function pointer here, but it causes stack issues
         if (kind == IVault.SwapKind.GIVEN_IN) {
-            amountCalculated = StableMath._calcOutGivenIn(
-                currentAmp,
-                balances,
-                indexIn,
-                indexOut,
-                amountGiven,
-                invariant
-            );
+            return StableMath._calcOutGivenIn(currentAmp, balances, indexIn, indexOut, amountGiven, invariant);
         } else {
-            amountCalculated = StableMath._calcInGivenOut(
-                currentAmp,
-                balances,
-                indexIn,
-                indexOut,
-                amountGiven,
-                invariant
-            );
+            return StableMath._calcInGivenOut(currentAmp, balances, indexIn, indexOut, amountGiven, invariant);
         }
     }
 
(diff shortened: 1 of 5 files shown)


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -286,162 +271,177 @@ contract StablePhantomPool is
...
+        balancesWithoutBpt[indexIn] += amountIn;
+        uint256 postJoinExitSupply = virtualSupply + bptOut;
+
+        return (amountIn, postJoinExitSupply);
     }
 
     /**
-     * @dev Process a swap from BPT into one of the Pool's tokens. At this point, amount has been upscaled, and the BPT
-     * token has been removed from balances. `indexOut` is the output token's position within `balancesWithoutBpt`.
+     * @dev This mutates balancesWithoutBpt so that they become the post-joinswap balances.
```
TomAFrench, line 383:
```suggestion
     * @dev This mutates balancesWithoutBpt so that they become the post-exitswap balances.
```

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -511,45 +511,92 @@ contract StablePhantomPool is
...
-                userData
-            );
-        } else {
-            _revert(Errors.UNHANDLED_JOIN_KIND);
-        }
 
-        // Add amountsIn to get post-join balances
-        // amountsIn returned from specific join functions has BPT added back in for the Vault
-        _mutateAmounts(balancesWithoutBpt, _dropBptItem(amountsIn), FixedPoint.add);
+            function(uint256[] memory, uint256, uint256, uint256[] memory, bytes memory)
```
TomAFrench, line 548: I hate linters sometimes.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -566,23 +613,20 @@ contract StablePhantomPool is
         // Balances are passed through from the Vault hook, and include BPT
```
TomAFrench, line 613:
```suggestion
```

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -724,12 +734,12 @@ contract StablePhantomPool is
         _require(tokenIndexWithoutBpt < balancesWithoutBpt.length, Errors.OUT_OF_BOUNDS);
 
         // We exit in a single token, so initialize amountsOut with zeros
-        uint256[] memory amountsOut = new uint256[](balancesWithoutBpt.length + 1);
+        uint256[] memory amountsOut = new uint256[](balancesWithoutBpt.length);
 
         // And then assign the result to the selected token.
         // The token index passed to the StableMath function must match the balances array (without BPT),
         // But the amountsOut array passed back to the Vault must include BPT.
```
TomAFrench, line 741:
```suggestion
```

TomAFrench (commented):
Looks great, It's much cleaner now. I need to do a second pass but I've found a few comments which are wrong/stale.

EndymionJkb (commented):
Did a pass over the comments. It's looking very "Vaulty" now, the way joins and exits are combined as long as possible, and only split at the very end.

pkg/pool-utils/contracts/BasePool.sol
```diff
@@ -633,8 +633,9 @@ abstract contract BasePool is IBasePool, BasePoolAuthorization, BalancerPoolToke
      * @dev Same as `_upscale`, but for an entire array. This function does not return anything, but instead *mutates*
      * the `amounts` array.
      */
-    function _upscaleArray(uint256[] memory amounts, uint256[] memory scalingFactors) internal view {
-        for (uint256 i = 0; i < _getTotalTokens(); ++i) {
+    function _upscaleArray(uint256[] memory amounts, uint256[] memory scalingFactors) internal pure {
+        uint256 length = Math.max(amounts.length, scalingFactors.length);
```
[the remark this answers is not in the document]
nventuro: Agreed - addressing this in a separate PR.

PR #1572, opened 2022-08-04
Move identification of setTokenRateCacheDuration as owner only action to StablePoolRates

TomAFrench:
This caused some mystery issues in testing `StablePoolRates` as we can't properly test non-delegated ownership.

### DIFF
--- a/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
@@ -974,8 +974,8 @@ contract StablePhantomPool is
         view
         virtual
         override(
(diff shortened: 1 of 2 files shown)


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -980,10 +977,11 @@ contract StablePhantomPool is
             // The ProtocolFeeCache module creates a small diamond that requires explicitly listing the parents here
```
nventuro, line 977:
```suggestion
            // Our inheritance pattern creates a small diamond that requires explicitly listing the parents here. Each parent
            // calls the `super` version, so linearization takes care of making sure all implementations are called.
```

nventuro (approved):

PR #1573, opened 2022-08-05
Stable: calculate preInvariant only once

EndymionJkb:
Calculate the preJoinExitInvariant externally, and pass it into StableMath, avoiding 1) computing it twice or 2) passing it *back* from StableMath.


pkg/pool-stable-phantom/contracts/StableMath.sol
```diff
@@ -243,7 +244,6 @@ library StableMath {
         }
 
         // Get current and new invariants, taking swap fees into account
```
nventuro, line 246: These comments are now stale

nventuro (approved):
Fantastic!

This also means we have no stable math tests though :scream:

PR #1575, opened 2022-08-05
Stable pool rates tests

TomAFrench:
Builds on #1574.

There's some jankiness in here, I'd preferably like to migrate the rate provider failure case testing across but some of the testing we can only perform in the `MockStablePhantomPool` (e.g. checking that we update rates on swaps/joins properly) so I don't think it's a huge deal to leave that there for now.


pkg/pool-stable-phantom/test/StablePoolRates.test.ts
```diff
@@ -0,0 +1,811 @@
...
+              const updatedAt = await currentTimestamp();
+
+              await action(token);
+
+              const currentCache = await pool.getTokenRateCache(token.address);
+              expect(currentCache.rate).to.be.equal(newRate);
+              expect(previousCache.rate).not.to.be.equal(newRate);
+
+              expect(currentCache.duration).to.be.equal(previousCache.duration);
+              expect(currentCache.expires).to.be.at.least(updatedAt.add(previousCache.duration));
```
nventuro, line 366: We should also test the old rates here (i.e. the old rates are not updated)

nventuro: Same in the other tests

pkg/pool-stable-phantom/test/StablePoolRates.test.ts
```diff
@@ -0,0 +1,811 @@
...
+            const inputArray = allTokens.map(() => fp(Math.random()));
+            const expectedOutputArray = inputArray.map((input, i) => input.mul(fp(1)).div(rates[i]));
+
+            expect(await pool.getAdjustedBalances(inputArray, true)).to.be.deep.eq(expectedOutputArray);
+          }
+        });
+      });
+
+      context('when not ignoring exempt flags', () => {
+        it('returns the array with elements scaled by the cached token rate if exempt', async () => {
```
nventuro, line 798:
```suggestion
        it('returns the array  with elements scaled by the old token rate', async () => {
          for (let i = 0; i < 5; i++) {
            const inputArray = allTokens.map(() => fp(Math.random()));
            const expectedOutputArray = inputArray.map((input, i) => input.mul(fp(1)).div(rates[i]));

            expect(await pool.getAdjustedBalances(inputArray, true)).to.be.deep.eq(expectedOutputArray);
          }
        });
      });

      context('when not ignoring exempt flags', () => {
        it('returns the array with elements scaled by the old token rate if exempt', async () => {
```

TomAFrench: I've update this to make clear that it's referring to a ratio.

TomAFrench:
Merging as Nico said this was good to go on a call.

PR #1577, opened 2022-08-05
Remove single token scaling factor access

nventuro:
The fact that we need to implement this function is one of the sources of confusion re. storage.

Turns out, we don't.

### DIFF
--- a/pkg/balancer-js/src/utils/errors.ts
+++ b/pkg/balancer-js/src/utils/errors.ts
@@ -157,6 +157,7 @@ const balancerErrorCodes: Record<string, string> = {
   '601': 'FLASH_LOAN_FEE_PERCENTAGE_TOO_HIGH',
   '602': 'INSUFFICIENT_FLASH_LOAN_FEE_AMOUNT',
   '603': 'AUM_FEE_PERCENTAGE_TOO_HIGH',
+  '998': 'UNIMPLEMENTED',
   '999': 'SHOULD_NOT_HAPPEN',
 };
 
--- a/pkg/pool-stable-phantom/contracts/StablePoolStorage.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePoolStorage.sol
@@ -221,8 +221,8 @@ abstract contract StablePoolStorage is BasePool {
     }
 
     function _scalingFactor(IERC20) internal view virtual override returns (uint256) {
-        // We never need a single token's scaling factor: we aways process the entire array at once. Therefore, we don't
-        // bother providing an implementation for this.
+        // We never use a single token's scaling factor by itself, we always process the entire array at once.
+        // Therefore we don't bother providing an implementation for this.
         _revert(Errors.UNIMPLEMENTED);
     }
 
--- a/pkg/pool-stable-phantom/contracts/StablePoolStorage.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePoolStorage.sol
@@ -54,12 +54,12 @@ abstract contract StablePoolStorage is BasePool {
     // not change throughout its lifetime, and store the corresponding scaling factor for each at construction time.
     // These factors are always greater than or equal to one: tokens with more than 18 decimals are not supported.
 
-    uint256 private immutable _scalingFactor0;
-    uint256 private immutable _scalingFactor1;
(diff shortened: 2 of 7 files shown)


pkg/pool-stable-phantom/contracts/StablePoolStorage.sol
```diff
@@ -220,16 +220,10 @@ abstract contract StablePoolStorage is BasePool {
...
-        if (token == _getToken3()) return _getScalingFactor3();
-        if (token == _getToken4()) return _getScalingFactor4();
-        if (token == _getToken5()) return _getScalingFactor5();
-        else {
-            _revert(Errors.INVALID_TOKEN);
-        }
+    function _scalingFactor(IERC20) internal view virtual override returns (uint256) {
+        // We never a single token's scaling factor: we aways process the entire array at once. Therefore, we don't
+        // bother providing an implementation for this.
+        _revert(Errors.UNIMPLEMENTED);
```
TomAFrench, line 226:
```suggestion
        // We never use a single token's scaling factor by itself, we always process the entire array at once.
        // Therefore we don't bother providing an implementation for this.
        _revert(Errors.UNIMPLEMENTED);
```

pkg/pool-stable-phantom/contracts/StablePoolStorage.sol
```diff
@@ -220,16 +220,10 @@ abstract contract StablePoolStorage is BasePool {
...
-        if (token == _getToken1()) return _getScalingFactor1();
-        if (token == _getToken2()) return _getScalingFactor2();
-        if (token == _getToken3()) return _getScalingFactor3();
-        if (token == _getToken4()) return _getScalingFactor4();
-        if (token == _getToken5()) return _getScalingFactor5();
-        else {
-            _revert(Errors.INVALID_TOKEN);
-        }
+    function _scalingFactor(IERC20) internal view virtual override returns (uint256) {
+        // We never a single token's scaling factor: we aways process the entire array at once. Therefore, we don't
```
EndymionJkb, line 224:
```suggestion
        // We never need a single token's scaling factor: we aways process the entire array at once. Therefore, we don't
```

pkg/balancer-js/src/utils/errors.ts
```diff
@@ -157,6 +157,7 @@ const balancerErrorCodes: Record<string, string> = {
   '601': 'FLASH_LOAN_FEE_PERCENTAGE_TOO_HIGH',
   '602': 'INSUFFICIENT_FLASH_LOAN_FEE_AMOUNT',
   '603': 'AUM_FEE_PERCENTAGE_TOO_HIGH',
+  '998': 'UNIMPLEMENTED',
```
EndymionJkb, line 160: We have an "INVALID_OPERATION". Maybe "UNSUPPORTED_OPERATION"?

TomAFrench (approved):
Looks good to me. Only one comment change.

pkg/pool-stable-phantom/contracts/StablePoolStorage.sol
```diff
@@ -220,16 +220,10 @@ abstract contract StablePoolStorage is BasePool {
...
-        if (token == _getToken3()) return _getScalingFactor3();
-        if (token == _getToken4()) return _getScalingFactor4();
-        if (token == _getToken5()) return _getScalingFactor5();
-        else {
-            _revert(Errors.INVALID_TOKEN);
-        }
+    function _scalingFactor(IERC20) internal view virtual override returns (uint256) {
+        // We never need a single token's scaling factor: we aways process the entire array at once. Therefore, we don't
+        // bother providing an implementation for this.
+        _revert(Errors.UNIMPLEMENTED);
```
nventuro, line 226:
```suggestion
        // We never use a single token's scaling factor by itself, we always process the entire array at once.
        // Therefore we don't bother providing an implementation for this.
        _revert(Errors.UNIMPLEMENTED);
```

pkg/balancer-js/src/utils/errors.ts
nventuro: Eh I'd say this is different, there's no 'operation'. It's just a piece from an abstract contract we choose to not implement.

PR #1578, opened 2022-08-05
Change token rate cache to be keyed by indices rather than addresses

TomAFrench:


pkg/pool-stable-phantom/contracts/StablePoolRates.sol
```diff
@@ -269,14 +266,12 @@ abstract contract StablePoolRates is StablePoolStorage {
     // Return true if the token at this index is a non-BPT token with a rate provider, so that it has
```
EndymionJkb, line 266:
```suggestion
    // Return true if the token at this index has a rate provider, so that it has
```
It looks like you're just skipping the BPT index when writing to it, so the rate provider value at the BPT index will have the default value of zero, so you don't care whether it's the BPT index: the check for 0 rate provider is sufficient.  Maybe document this a bit more?

nventuro (approved):

PR #1581, opened 2022-08-05
Change adjustBalances to take balances without BPT

TomAFrench:
In `_payProtocolFeesBeforeJoinExit` we create two arrays of adjusted balances which are only used in order to construct new arrays which drop the BPT element. This has memory allocation gas costs associated with it (along with the costs from just copying all the data around).

This PR adds behaviour to skip over the BPT's index so that we can pass `balancesWithoutBpt` in instead of `balances`

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -1,6 +1,6 @@
 {
   "extension": ["ts"],
-  "timeout": 20000,
+  "timeout": 50000,
   "recursive": "test",
   "slow": 500
 }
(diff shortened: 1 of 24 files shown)


TomAFrench:
@nventuro You mentioned getting rid of the dummy rate provider variable for BPT which got the indices to line up. I've got this on another branch but it means that we need to have index-juggling happen in way more places than this so I'm inclined to take this solution.

Can share the branch if you want to take a look.

nventuro (approved):

nventuro:
Merging this as-is - potentially there's a better way to handle indices so we barely do any skipping.

PR #1584, opened 2022-08-08
Add StableProtocolFees

nventuro:
This extracts fees from the main contract into its own module. No changes in behavior.


EndymionJkb (approved):
Looks good. Bonus that the add/drop BPT functions are now together.

PR #1587, opened 2022-08-09
Allow to pass custom prefix for error messages

danhper:
Instead of the error messages always being prefixed with `BAL#`, we allow to pass in the first three characters, so that custom pools can use their own set of error codes without risking any collision.
This replaces #1565.

For more context
image

### DIFF
--- a/pkg/interfaces/contracts/solidity-utils/helpers/BalancerErrors.sol
+++ b/pkg/interfaces/contracts/solidity-utils/helpers/BalancerErrors.sol
@@ -19,15 +19,33 @@ pragma solidity ^0.7.0;
 /**
  * @dev Reverts if `condition` is false, with a revert reason containing `errorCode`. Only codes up to 999 are
  * supported.
+ * Uses the default 'BAL' prefix for the error code
  */
 function _require(bool condition, uint256 errorCode) pure {
     if (!condition) _revert(errorCode);
 }
 
 /**
+ * @dev Reverts if `condition` is false, with a revert reason containing `errorCode`. Only codes up to 999 are
+ * supported.
+ */
+function _require(bool condition, uint256 errorCode, bytes3 prefix) pure {
+    if (!condition) _revert(errorCode, prefix);
+}
+
+/**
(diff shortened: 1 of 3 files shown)


pkg/solidity-utils/test/BalancerErrors.test.ts
```diff
@@ -17,4 +17,9 @@ describe('BalancerErrors', function () {
   it('translates the error code to its corresponding string if existent', async () => {
     await expect(errors.fail(102)).to.be.revertedWith('UNSORTED_TOKENS');
   });
+
+  it('encodes the prefix as expected', async () => {
+    // GYR = 0x475952
+    await expect(errors.failWithPrefix(123, '0x475952')).to.be.revertedWith('GYR#123');
```
nventuro, line 23: Can you also update the test above so we're looking for the entire `BAL#123` string?

pkg/interfaces/contracts/solidity-utils/helpers/BalancerErrors.sol
```diff
@@ -19,15 +19,33 @@ pragma solidity ^0.7.0;
...
+function _require(bool condition, uint256 errorCode, bytes3 prefix) pure {
+    if (!condition) _revert(errorCode, prefix);
+}
+
 /**
  * @dev Reverts with a revert reason containing `errorCode`. Only codes up to 999 are supported.
+ * Uses the default 'BAL' prefix for the error code
  */
 function _revert(uint256 errorCode) pure {
+    _revert(errorCode, 0x42414c);
```
nventuro, line 41:
```suggestion
    _revert(errorCode, 0x42414c); // This is the raw byte representation of "BAL"
```

nventuro (approved):
Fantastic, this looks much cleaner. Thanks!

pkg/solidity-utils/test/BalancerErrors.test.ts
danhper: Done!

nventuro:
(fork tests failed since this CI run does not have access to repository secrets, and therefore the archive node API keys)

PR #1588, opened 2022-08-09
Stable: standardize naming conventions

EndymionJkb:
We have lots of "withBpt" and "withoutBpt" names, which while helpful and self-documenting, aren't always consistent. (And are a bit awkward besides.)

Since nearly all "lower level" processing occurs with BPT removed, and generally only the outermost Vault interface functions (to and from hooks) require the full array of registered tokens, we'd like to standardize on the convention that "balances" and "index" refer to "withoutBpt" arrays. The "withBpt" arrays, as passed in, correspond to the registered tokens, so we call them registeredBalances and registeredIndexIn/Out.

scalingFactors normally do include all factors, and are "withBpt", so if we pass around scaling factors with Bpt removed, this is the exception, and we'll call them "scalingFactorsWithoutBpt".

(I tried to drop the BPT "higher" up, in `_onJoinExitPool`, since this would have been more consistent and saved a call to _dropBptItem - but was defeated by our nemesis, "stack-too-deep".)

### DIFF
--- a/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
@@ -158,7 +158,8 @@ contract StablePhantomPool is
      * or a swap involving BPT, which is equivalent to a single token join or exit. Since one of the Pool's
      * tokens is the preminted BPT, we need to handle swaps where BPT is involved separately.
      *
-     * At this point, the balances are unscaled.
+     * At this point, the balances are unscaled. The indices are coming from the Vault, so they are indices into
+     * the array of registered tokens (including BPT).
      *
      * If this is a swap involving BPT, call `_swapWithBpt`, which computes the amountOut using the swapFeePercentage
      * and charges protocol fees, in the same manner as single token join/exits. Otherwise, perform the default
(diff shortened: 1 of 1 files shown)


pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -657,19 +688,12 @@ contract StablePhantomPool is
                     preJoinExitSupply,
                     preJoinExitInvariant,
                     currentAmp,
-                    balancesWithoutBpt,
-                    scalingFactors,
+                    balances,
+                    _dropBptItem(scalingFactors), // drop BPT to match balances
```
nventuro, line 692: I think this is unnecessary - we can do it at the place where we use the scaling factors, instead of here. And that way scaling factors will always be for the full registered set.

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
```diff
@@ -682,19 +706,19 @@ contract StablePhantomPool is
...
+        uint256[] memory scalingFactorsWithoutBpt,
         bytes memory userData
     ) private view returns (uint256, uint256[] memory) {
         (uint256[] memory amountsIn, uint256 minBPTAmountOut) = userData.exactTokensInForBptOut();
-        InputHelpers.ensureInputLengthMatch(balancesWithoutBpt.length, amountsIn.length);
+        InputHelpers.ensureInputLengthMatch(balances.length, amountsIn.length);
 
         // The user-provided amountsIn is unscaled, so we address that.
-        _upscaleArray(amountsIn, _dropBptItem(scalingFactors));
+        _upscaleArray(amountsIn, scalingFactorsWithoutBpt);
```
nventuro, line 717: We'd then undo the change here as well.

nventuro (approved):

pkg/pool-stable-phantom/contracts/StablePhantomPool.sol
EndymionJkb: Yeah, I tried to do it higher, where it would have made more sense.

PR #1589, opened 2022-08-09
Pack values in storage

nventuro:
We don't need a full slot for these values.

### DIFF
--- a/pkg/pool-stable-phantom/contracts/StablePoolProtocolFees.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePoolProtocolFees.sol
@@ -16,6 +16,7 @@ pragma solidity ^0.7.0;
 pragma experimental ABIEncoderV2;
 
 import "@balancer-labs/v2-solidity-utils/contracts/math/FixedPoint.sol";
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
 import "@balancer-labs/v2-pool-utils/contracts/ProtocolFeeCache.sol";
 
 import "./StablePoolStorage.sol";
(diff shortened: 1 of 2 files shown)


pkg/pool-stable-phantom/contracts/StablePoolProtocolFees.sol
```diff
@@ -24,6 +25,24 @@ import "./StableMath.sol";
...
+    // capped at 112 bits.
+    // The data structure is as follows:
+    //
+    // [ last join-exit amplification  | last post join-exit invariant ]
+    // [           23 bits             |            233 bits           ]
+    bytes32 private _lastJoinExitData;
+
+    uint256 private constant _LAST_POST_JOIN_EXIT_INVARIANT_OFFSET = 0;
+    uint256 private constant _LAST_POST_JOIN_EXIT_INVARIANT_SIZE = 233;
+    uint256 private constant _LAST_JOIN_EXIT_AMPLIFICATION_OFFSET = _LAST_POST_JOIN_EXIT_INVARIANT_OFFSET +
```
EndymionJkb, line 42: I like having the offsets and sizes separated, and all hard numbers vs calculations (so that you can easily change the order, without redoing all the "equations"). But for only two things it doesn't really matter

pkg/pool-stable-phantom/contracts/StablePoolProtocolFees.sol
```diff
@@ -96,19 +115,20 @@ abstract contract StablePoolProtocolFees is StablePoolStorage, StablePoolRates,
         // amplification are not translated into changes to the invariant. Since amplification factor changes are both
         // infrequent and slow, they should have little effect in the pool balances, making this a very good
         // approximation.
-        uint256 lastPostJoinExitAmp = _lastPostJoinExitAmp;
+
+        (uint256 lastJoinExitAmp, uint256 lastPostJoinExitInvariant) = getLastJoinExitData();
 
         uint256 swapFeeGrowthInvariant = StableMath._calculateInvariant(
-            lastPostJoinExitAmp,
+            lastJoinExitAmp,
```
EndymionJkb, line 122: I was questioning the name change, but it makes sense. The invariant changes as a result of the join/exit, so it is a "PostJoinExit" thing, whereas the Amp is unrelated - you're just grabbing whatever it happens to be at that time, so it's the Amp "at" the last JoinExit.

pkg/pool-stable-phantom/contracts/StablePoolProtocolFees.sol
```diff
@@ -96,19 +115,20 @@ abstract contract StablePoolProtocolFees is StablePoolStorage, StablePoolRates,
...
         );
 
         uint256 totalNonExemptGrowthInvariant = StableMath._calculateInvariant(
-            lastPostJoinExitAmp,
+            lastJoinExitAmp,
             _getAdjustedBalances(balancesWithoutBpt, false) // Only adjust non-exempt balances
         );
 
-        uint256 totalGrowthInvariant = StableMath._calculateInvariant(lastPostJoinExitAmp, balancesWithoutBpt);
+        uint256 totalGrowthInvariant = StableMath._calculateInvariant(lastJoinExitAmp, balancesWithoutBpt);
```
EndymionJkb, line 131: The balances are all getting renamed (registeredBalances, balances). Maybe wait until... and you merged it before I finished writing :) Anyway, update the names here too.

pkg/pool-stable-phantom/contracts/StablePoolProtocolFees.sol
```diff
@@ -24,6 +25,24 @@ import "./StableMath.sol";
 
 abstract contract StablePoolProtocolFees is StablePoolStorage, StablePoolRates, ProtocolFeeCache {
     using FixedPoint for uint256;
+    using WordCodec for bytes32;
+
+    // We store the invariant after the last join-exit, along with the amplification factor used to compute it. The
```
EndymionJkb, line 30:
```suggestion
    // We store the invariant after the last join or exit, along with the amplification factor used to compute it. The
```
Nit, but sounds like there is something called a join-exit

EndymionJkb (approved):

PR #1592, opened 2022-08-10
BasePool: use _miscData for Recovery Mode

### LINKED ISSUE #1371: Pack RecoveryMode boolean into Pool misc data
The introduction of https://github.com/balancer-labs/balancer-v2-monorepo/pull/1369 caused for this value to be read in all joins and exits - we could save gas by storing this along the swap fee.

It is only a storage read during joins and exits however, so not critical.

EndymionJkb:
Every join and exit in the system needs to check the Recovery Mode flag, so it will save gas to pack it into miscData, vs loading it from a bool variable every time.

It means we lose a bit off the swap fee percentage, but 63 is more than enough.

Closes #1371

### DIFF
--- a/pkg/pool-utils/contracts/BasePool.sol
+++ b/pkg/pool-utils/contracts/BasePool.sol
@@ -64,12 +64,21 @@ abstract contract BasePool is IBasePool, BasePoolAuthorization, BalancerPoolToke
     uint256 private constant _MIN_SWAP_FEE_PERCENTAGE = 1e12; // 0.0001%
     uint256 private constant _MAX_SWAP_FEE_PERCENTAGE = 1e17; // 10% - this fits in 64 bits
 
-    // Storage slot that can be used to store unrelated pieces of information. In particular, by default is used
-    // to store only the swap fee percentage of a pool. But it can be extended to store some more pieces of information.
-    // The swap fee percentage is stored in the most-significant 64 bits, therefore the remaining 192 bits can be
-    // used to store any other piece of information.
+    // `_miscData` is a storage slot that can be used to store unrelated pieces of information. All pools store the
+    // recovery mode flag and swap fee percentage, but `miscData` can be extended to store more pieces of information.
+    // The most signficant bit is reserved for the recovery mode flag, and the swap fee percentage is stored in
+    // the next most significant 63 bits, leaving the remaining 192 bits free to store any other information derived
+    // pools might need.
+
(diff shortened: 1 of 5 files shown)


pkg/pool-utils/contracts/BasePool.sol
```diff
@@ -186,6 +200,25 @@ abstract contract BasePool is IBasePool, BasePoolAuthorization, BalancerPoolToke
...
+    function inRecoveryMode() public view override returns (bool) {
+        return _miscData.decodeBool(_RECOVERY_MODE_BIT_OFFSET);
+    }
+
+    /**
+     * @dev Sets the recoveryMode state, and emits the corresponding event.
+     *
+     * No complex code or external calls that could fail should be placed here, which could jeopardize
+     * the ability to enable and disable Recovery Mode.
+     */
```
nventuro, line 215: I'd place these docs in the RecoveryMode module

pkg/pool-utils/contracts/BasePool.sol
```diff
@@ -186,6 +200,25 @@ abstract contract BasePool is IBasePool, BasePoolAuthorization, BalancerPoolToke
...
+    function _setRecoveryMode(bool enabled) internal virtual override {
+        _miscData = _miscData.insertBool(enabled, _RECOVERY_MODE_BIT_OFFSET);
+
+        emit RecoveryModeStateChanged(enabled);
```
nventuro, line 219: Bit unfortunate that we have to emit this event here - ideally recovery mode would emit it and we'd just deal with storage. But this is fine for now.

nventuro (commented):
Love the approach you took by handling this via virtual functions!

pkg/pool-utils/contracts/BasePool.sol
EndymionJkb: Well, I actually had it that way. This routine called super into RecoveryMode, which only emitted the event. But that seemed awkward too,  since then if derived classes didn't call super, it would. change the storage without the event, or vice versa. So I put them together to make sure both happened. A bit messy in different ways either way.

nventuro: Yeah an alternative would be something like

```
fn _privateSetRecovery() private {
  _setRecovery();
  emit event
}

fn _setRecovery() internal virtual;
```

But then if you called `_setRecovery` directly that'd be a mistake.

PR #1594, opened 2022-08-10
Fix BasePool queries

nventuro:
This adds a call to `_beforeSwapJoinExit()` to the queries, which will make them more correct: they'll revert if the pool is paused, and more importantly do any work that is required before the action, such as updating the rate cache in a stable pool.

Adding tests for BasePool directly would be a bit cumbersome as we don't have a proper mock right now, so I simply added tests to the queries for now.


EndymionJkb (approved):
Added some tests to ensure we're updating the rates cache in join/exit queries

nventuro:
Thanks a lot!

PR #1595, opened 2022-08-10
Add global exemption flags

nventuro:
These will let us detect situations in which we can skip some invariants (e.g. the non-exempt invariant or the total growth invariant) since they will equal one of the other two.

Storing them as immutable values uses less bytecode than packing them into the bitmap (and also uses less gas).


EndymionJkb (commented):
Would it also help to check if there are no rate providers? In that case _allTokensExempt would be true, and we could additionally check for no rate providers with "_rateProviderInfoBitmap == 0"

We're also not checking for 0 fee percentages anywhere now (only that the final result > 0 before minting), so we could also short-circuit parts of it when the corresponding fee percentage is 0: and the entire thing if they're both 0.

PR #1597, opened 2022-08-10
Optimize invariant calculation for edge cases

nventuro:
This reduces the number of invariants we compute by one when either all tokens are exempt or none are (see #1595), as well as skips an unnecessary balance adjustment. It also begins the introduction of protocol fee tests.

### DIFF
--- a/pkg/pool-stable-phantom/contracts/StablePoolProtocolFees.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePoolProtocolFees.sol
@@ -189,9 +189,9 @@ abstract contract StablePoolProtocolFees is StablePoolStorage, StablePoolRates, 
         // tokens are exempt from yield, there's one fewer invariant to compute.
 
         if (_areNoTokensExempt()) {
-            // If there are no tokens exempt of yield fee, then the total non exempt growth will equal the total growth
-            // (because all yield growth is non exempt). There's also no point in adjusting balances to get the
-            // non-exempt ones, as none are exempt and as such this equals the current balances.
+            // If there are no tokens with fee-exempt yield, then the total non-exempt growth will equal the total
+            // growth: all yield growth is non-exempt. There's also no point in adjusting balances, since we
+            // already know none are exempt.
 
             totalNonExemptGrowthInvariant = StableMath._calculateInvariant(lastPostJoinExitAmp, balances);
             totalGrowthInvariant = totalNonExemptGrowthInvariant;
--- a/pkg/pool-stable-phantom/contracts/StablePoolProtocolFees.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePoolProtocolFees.sol
@@ -196,7 +196,7 @@ abstract contract StablePoolProtocolFees is StablePoolStorage, StablePoolRates, 
             totalNonExemptGrowthInvariant = StableMath._calculateInvariant(lastPostJoinExitAmp, balances);
             totalGrowthInvariant = totalNonExemptGrowthInvariant;
         } else if (_areAllTokensExempt()) {
-            // If all tokens are exempt of yield fee, then the non exempt growth is equal to the swap fee growth - no
+            // If no tokens are charged fees on yield, then the non-exempt growth is equal to the swap fee growth - no
             // yield fees will be collected.
 
             totalNonExemptGrowthInvariant = swapFeeGrowthInvariant;
--- a/pkg/pool-stable-phantom/contracts/StablePoolProtocolFees.sol
+++ b/pkg/pool-stable-phantom/contracts/StablePoolProtocolFees.sol
@@ -117,7 +117,7 @@ abstract contract StablePoolProtocolFees is StablePoolStorage, StablePoolRates, 
 
         (uint256 lastJoinExitAmp, uint256 lastPostJoinExitInvariant) = getLastJoinExitData();
 
-(
+        (
             uint256 swapFeeGrowthInvariant,
             uint256 totalNonExemptGrowthInvariant,
             uint256 totalGrowthInvariant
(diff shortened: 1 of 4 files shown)


pkg/pool-stable-phantom/contracts/StablePoolProtocolFees.sol
```diff
@@ -156,6 +150,52 @@ abstract contract StablePoolProtocolFees is StablePoolStorage, StablePoolRates,
...
+
+        swapFeeGrowthInvariant = StableMath._calculateInvariant(
+            lastPostJoinExitAmp,
+            _getAdjustedBalances(balances, true) // Adjust all balances
+        );
+
+        // For the other invariants, we can potentially skip some work. In the edge cases where none or all of the
+        // tokens are exempt from yield, there's one fewer invariant to compute.
+
+        if (_areNoTokensExempt()) {
```
EndymionJkb, line 191: I guess _areXXX is analogous to _isXXX, but _noTokensAreExempt() and _allTokensAreExempt() sound more natural to me. (Or even drop the "are": _allTokensExempt(), _noTokensExempt().)

pkg/pool-stable-phantom/contracts/StablePoolProtocolFees.sol
```diff
@@ -156,6 +150,52 @@ abstract contract StablePoolProtocolFees is StablePoolStorage, StablePoolRates,
...
+            // If there are no tokens exempt of yield fee, then the total non exempt growth will equal the total growth
+            // (because all yield growth is non exempt). There's also no point in adjusting balances to get the
+            // non-exempt ones, as none are exempt and as such this equals the current balances.
```
EndymionJkb, line 194:
```suggestion
            // If there are no tokens with fee-exempt yield, then the total non-exempt growth will equal the total
            // growth: all yield growth is non-exempt. There's also no point in adjusting balances, since we
            // already know none are exempt.
```

pkg/pool-stable-phantom/contracts/StablePoolProtocolFees.sol
```diff
@@ -156,6 +150,52 @@ abstract contract StablePoolProtocolFees is StablePoolStorage, StablePoolRates,
...
+
+            totalNonExemptGrowthInvariant = StableMath._calculateInvariant(lastPostJoinExitAmp, balances);
+            totalGrowthInvariant = totalNonExemptGrowthInvariant;
+        } else if (_areAllTokensExempt()) {
+            // If all tokens are exempt of yield fee, then the non exempt growth is equal to the swap fee growth - no
+            // yield fees will be collected.
```
EndymionJkb, line 200:
```suggestion
            // If no tokens are charged fees on yield, then the non-exempt growth is equal to the swap fee growth - no
            // yield fees will be collected.
```

pkg/pool-stable-phantom/contracts/test/MockStablePoolProtocolFees.sol
```diff
@@ -0,0 +1,141 @@
...
+    constructor(
+        IVault vault,
+        IProtocolFeePercentagesProvider protocolFeeProvider,
+        IERC20[] memory tokens,
+        IRateProvider[] memory tokenRateProviders,
+        uint256[] memory tokenRateCacheDurations,
+        bool[] memory exemptFromYieldProtocolFeeFlags
+    )
+        StablePoolStorage(
+            StorageParams(_insertSorted(tokens, IERC20(this)), tokenRateProviders, exemptFromYieldProtocolFeeFlags)
```
EndymionJkb, line 32: Use the same {} syntax here?

pkg/pool-stable-phantom/test/StablePoolProtocolFees.test.ts
```diff
@@ -0,0 +1,246 @@
...
+        it('computes the swap fee growth invariant correctly', async () => {
+          // The swap fee growth invariant is computed by using old rates for all tokens
+          const oldRateBalances = await Promise.all(
+            balances.map(async (balance, i) => {
+              const { rate, oldRate } = await pool.getTokenRateCache(tokens.get(i).address);
+              return balance.mul(oldRate).div(rate);
+            })
+          );
+
+          const expectedSwapFeeGrowhtInvariant = calculateInvariant(
```
EndymionJkb, line 141:
```suggestion
          const expectedSwapFeeGrowthInvariant = calculateInvariant(
```

EndymionJkb: A test to see if we're reading carefully :)

pkg/pool-stable-phantom/test/StablePoolProtocolFees.test.ts
```diff
@@ -0,0 +1,246 @@
...
+            oldRateBalances,
+            AMPLIFICATION_FACTOR.div(AMPLIFICATION_PRECISION)
+          );
+
+          const { swapFeeGrowthInvariant } = await pool.getGrowthInvariants(balances, AMPLIFICATION_FACTOR);
+          expect(swapFeeGrowthInvariant).to.almostEqual(expectedSwapFeeGrowhtInvariant, 1e-10);
```
EndymionJkb, line 147:
```suggestion
          expect(swapFeeGrowthInvariant).to.almostEqual(expectedSwapFeeGrowthInvariant, 1e-10);
```
Maybe pull the 1e-10 out to a MAX_RELATIVE_ERROR field like in other tests?

pkg/pool-stable-phantom/contracts/StablePoolProtocolFees.sol
nventuro: That'd make them collide with the variables though :( Can I keep my names?

pkg/pool-stable-phantom/test/StablePoolProtocolFees.test.ts
nventuro: This is only for the invariants though - other tests will have other errors. i'll look into this later.

pkg/pool-stable-phantom/contracts/StablePoolProtocolFees.sol
EndymionJkb: Yeah; I think _areNoTokensExempt / _noTokensExempt is clearer than 
                     _noTokensAreExempt / _noTokensExempt

PR #1598, opened 2022-08-11
Finalize pre join-exit protocol fees

nventuro:
This fixes the protocol fee computation, and adds lots of tests to check this behavior. The tests don't perform the same calculations as the contract, but instead attempt to derive the amount of due fees relatively indirectly, and arrive at the same value as the contract.


EndymionJkb (approved):
Pushed some doc changes; everything else looks good. Also removed some obsolete / unfinished tests, to not clutter the suite. The non-18 decimal tests are in a separate file, which I can add in a new draft PR. We should finish those at some point.

PR #1599, opened 2022-08-11
Test deployment of stable pools

nventuro:
WIP as this isn't even the final version.


nventuro:
Deleting this as the rename introduced too many conflicts.

PR #1602, opened 2022-08-11
Stable: Stub for non-18-decimal tests

EndymionJkb:


EndymionJkb:
Closed for now, since Stable was already deployed.

PR #1603, opened 2022-08-12
Stable: cleanup post join

EndymionJkb:
Clean up and add tests to ensure we're updating the state after each regular join or exit, and collecting protocol fees on non-proportional joins and exits.


pkg/pool-stable/test/ComposableStablePool.test.ts
```diff
@@ -686,11 +704,26 @@ describe('ComposableStablePool', () => {
...
+              const preJoinInvariant = await pool.estimateInvariant();
+              await pool.joinGivenIn({ amountsIn, minimumBptOut, recipient, from: recipient });
+              const expectedLastInvariant = await pool.estimateInvariant();
+              expect(expectedLastInvariant).gt(preJoinInvariant);
+
+              await changeAmplification();
+
+              const result = await pool.getLastJoinExitData();
+              expect(result.lastJoinExitAmplification).to.equal(AMPLIFICATION_PARAMETER.mul(AMP_PRECISION));
+              expect(result.lastPostJoinExitInvariant).to.equalWithError(expectedLastInvariant, 0.000001);
```
TomAFrench, line 724: This feels like a separate test for amp updates don't affect the stored values.

pkg/pool-stable/test/ComposableStablePool.test.ts
```diff
@@ -686,11 +704,26 @@ describe('ComposableStablePool', () => {
...
               expect(currentBptBalance.sub(previousBptBalance)).to.be.equalWithError(expectedBptOut, 0.0001);
             });
 
+            it('updates the invariant afterward', async () => {
+              const minimumBptOut = pct(expectedBptOut, 0.99);
+
+              const preJoinInvariant = await pool.estimateInvariant();
+              await pool.joinGivenIn({ amountsIn, minimumBptOut, recipient, from: recipient });
+              const expectedLastInvariant = await pool.estimateInvariant();
+              expect(expectedLastInvariant).gt(preJoinInvariant);
```
TomAFrench, line 718: As we're doing a totally proportional join on initialisation we can calculate the invariant exactly as it should be equal to the sum of all the tokens (up to a scaling factor) so we could tighten up the test using that.

Up to you on whether that's worth it.

pkg/pool-stable/contracts/ComposableStablePoolProtocolFees.sol
```diff
@@ -226,44 +231,59 @@ abstract contract ComposableStablePoolProtocolFees is
...
-                    .mulDown(protocolSwapFeePercentage);
-
-                _payProtocolFees(protocolFeeAmount);
-            }
+        // `_payProtocolFeesBeforeJoinExit` paid protocol fees accumulated between the previous and current
+        // join or exit. This code pays any protocol fees due on a `current` joinSwap or exitSwap.
+        // The amp and rates are constant during a single transaction, so it doesn't matter if there
+        // is an ongoing amp change, and we can ignore yield.
+
+        // This represents the total value exchanged in this particular join-exit, including swap fees.
```
TomAFrench, line 239:
```suggestion
        // This represents the total value exchanged in this particular join-exit, including swap fees.
        // Note that as swap fees are symmetric, we can swap the pre- and post-exit invariants and treat it like a join.
```

I think the naming of variables here is a bit of a hurdle to it being obvious that this is correct for exits. We could rename them to `*Change` but I think it might be best to just add a comment like the above maybe?

pkg/pool-stable/contracts/ComposableStablePoolProtocolFees.sol
```diff
@@ -226,44 +231,59 @@ abstract contract ComposableStablePoolProtocolFees is
...
+        )
+            .divDown(preJoinExitInvariant);
+
+        // The BPT increase represents the value exchanged for the user's tokens in the join or exit.
+        // The more unbalanced the join or exit, the greater the "taxable" amount subject to swap fees,
+        // and the lower the BPT amount in or out.
+        //
+        // In a proportional join or exit, the taxable amount will be 0, so the BPT growth would
+        // equal the total growth. Otherwise, the BPT would be minted based on fewer tokens, and
+        // the BPT growth will be less than the total.
```
TomAFrench, line 253:
```suggestion
        // the BPT growth will be less than the total.
        // Note that as swap fees are symmetric, we can swap the pre- and post-exit total supplies and treat it like a join.
```

TomAFrench (commented):
Looks good. I haven't looked at the tests for `ComposableStablePoolProtocolFees` yet but I've made a couple of notes on other things.

pkg/pool-stable/contracts/ComposableStablePoolProtocolFees.sol
EndymionJkb: Yeah; I went back and forth on that. It was "delta" for a while, but it's not really a delta in the sense that's used elsewhere (after dividing, anyway). I added a detailed comment, perhaps a bit long-winded.

TomAFrench: Ah yeah, I forgot about the division so my comment above isn't correct.

pkg/pool-stable/contracts/ComposableStablePoolProtocolFees.sol
```diff
@@ -226,44 +231,59 @@ abstract contract ComposableStablePoolProtocolFees is
...
+        // `_payProtocolFeesBeforeJoinExit` paid protocol fees accumulated between the previous and current
+        // join or exit. This code pays any protocol fees due on a `current` joinSwap or exitSwap.
+        // The amp and rates are constant during a single transaction, so it doesn't matter if there
+        // is an ongoing amp change, and we can ignore yield.
+
+        // This represents the total value exchanged in this particular join-exit, including swap fees.
+        uint256 invariantGrowth = (
+            postJoinExitInvariant > preJoinExitInvariant
+                ? postJoinExitInvariant.sub(preJoinExitInvariant)
+                : preJoinExitInvariant.sub(postJoinExitInvariant)
```
TomAFrench, line 243:
```suggestion
                ? postJoinExitInvariant - preJoinExitInvariant
                : preJoinExitInvariant - postJoinExitInvariant
```

We can optimise this as we know it can't underflow. Similar with the total supplies.

pkg/pool-stable/contracts/ComposableStablePoolProtocolFees.sol
```diff
@@ -226,44 +231,59 @@ abstract contract ComposableStablePoolProtocolFees is
...
+        )
+            .divDown(preJoinExitSupply);
+
+        // Therefore, the difference between the invariant growth and bpt increase rates represents the amount
+        // that was charged swap fees, on which protocol fees are also due.
+
+        if (invariantGrowth > bptGrowth) {
+            // Since the growth values are already normalized, multiply by the protocol swap fee to convert
+            // to a percentage owned to the protocol.
+            uint256 protocolSwapFeePercentage = invariantGrowth.sub(bptGrowth).mulDown(
```
TomAFrench, line 267:
```suggestion
            uint256 protocolSwapFeePercentage = (invariantGrowth - bptGrowth).mulDown(
```

Saves some gas and makes the order of operations clearer imo.

pkg/pool-stable/test/ComposableStablePool.test.ts
EndymionJkb: I don't know - people don't necessary have to initialize it proportionally.

If they do initialize disproportionately, we aren't charging them protocol fees (we just update the cache directly). We don't have a proportional join now, so we can't enforce that. Maybe that's another test? That we don't pay protocol fees on disproportionate initialization.

pkg/pool-stable/test/ComposableStablePoolProtocolFees.test.ts
```diff
@@ -598,6 +605,214 @@ describe('ComposableStablePoolProtocolFees', () => {
           }
         }
       });
+
+      describe('payProtocolFeesAfterJoinExit', () => {
```
TomAFrench, line 609: I like how you've set this up.

TomAFrench: Do you need any help with the skipped test cases?

EndymionJkb: Thank Nico :) Just copied his structure.

pkg/pool-stable/contracts/ComposableStablePoolProtocolFees.sol
```diff
@@ -227,44 +232,68 @@ abstract contract ComposableStablePoolProtocolFees is
...
+        // decrease. Since swap fees are symmetrical, it is numerically equivalent (and conceptually simpler)
+        // to just invert the sign in the exit case, and use the same "growth" terminology and calculations
+        // for both.
+        //
+        uint256 invariantGrowth = (
+            postJoinExitInvariant > preJoinExitInvariant
+                ? postJoinExitInvariant - preJoinExitInvariant
+                : preJoinExitInvariant - postJoinExitInvariant
+        )
+            .divDown(preJoinExitInvariant);
```
nventuro, line 252: I find this slightly concerning, because due to rounding errors we might have post join invariants that are actually smaller than the pre join. Could we add a boolean to this function that indicates whether to expect a join or an exit, and then do the expected subtraction depending on it (capping to 0 if the result would be negative)?

pkg/pool-stable/contracts/ComposableStablePoolProtocolFees.sol
```diff
@@ -227,44 +232,68 @@ abstract contract ComposableStablePoolProtocolFees is
...
+                getProtocolFeePercentageCache(ProtocolFeeType.SWAP)
+            );
+
+            // Since this will be minted as BPT, which increases the total supply, we need to mint
+            // slightly more so that it reflects this percentage of the total supply after minting.
+            uint256 protocolFeeAmount = preJoinExitSupply.mulDown(protocolSwapFeePercentage).divDown(
+                protocolSwapFeePercentage.complement()
+            );
+
+            _payProtocolFees(protocolFeeAmount);
```
nventuro, line 287: I think this is wrong. Here's how I think about this (assuming join):

a) the supply increased by some ratio `SR`. the invariant can increase by `SR` without fees
b) `after invariant - pre invariant * SR` is the invariant delta due to fees, `fee delta inv`
c) to convert that to a pool percentage, we need to divide `fee delta inv` not by `pre invariant`, but by `after invariant`. the protocol swap percentage of this then is the protocol ownership %
d) we apply `pct = m / (s + m)`

nventuro (changes requested):
Left two comments on the contract math - will do a second pass on the tests

pkg/pool-stable/contracts/ComposableStablePoolProtocolFees.sol
TomAFrench: Addressed in [1ce7bf8](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1603/commits/1ce7bf84a301c534494cb453fc1651c6adaa3ee6).

pkg/pool-stable/contracts/ComposableStablePoolProtocolFees.sol
TomAFrench: Looking into this now. I think we should also be using `postJoinExitSupply` on L283

pkg/pool-stable/contracts/ComposableStablePoolProtocolFees.sol
```diff
@@ -227,44 +232,53 @@ abstract contract ComposableStablePoolProtocolFees is
...
+
+        uint256 invariantDelta = (
+            isJoin
+                ? (postJoinExitInvariant > preJoinExitInvariant ? postJoinExitInvariant - preJoinExitInvariant : 0)
+                : (preJoinExitInvariant > postJoinExitInvariant ? preJoinExitInvariant - postJoinExitInvariant : 0)
+        );
+
+        uint256 bptDelta = isJoin
+            ? (postJoinExitSupply > preJoinExitSupply ? postJoinExitSupply - preJoinExitSupply : 0)
+            : (preJoinExitSupply > postJoinExitSupply ? preJoinExitSupply - postJoinExitSupply : 0);
```
TomAFrench, line 252:
```suggestion
            ? postJoinExitSupply - preJoinExitSupply 
            : preJoinExitSupply - postJoinExitSupply;
```

These conditions are true by definition based on how we've defined `isJoin`

TomAFrench (approved):
Things look good to me. There's a couple of minor changes that could improve consistency in argument ordering across different functions.

PR #1604, opened 2022-08-12
Deploy Child Chain Gauge Helper

nventuro:
A bit overdue, as this has been ready for a while. Needs checking build-info against `liquidity-mining`.


EndymionJkb:
The build-info I generated matches what's here

PR #1606, opened 2022-08-16
Final touches on LinearPool

TomAFrench:
I've pulled across the implementation of `getVirtualSupply` from `ComposableStablePool` as it's more efficient, the same optimisation is used in `_isMainBalanceWithinTargets`.

I've also updated some comments as they were saying that the total supply is fixed whereas it's isn't.


TomAFrench:
I've had to create a version of `getPoolTokenInfo` on the mocked Vault as otherwise we cannot set the targets in the LinearPool tests.

nventuro (approved):
Lovely

PR #1607, opened 2022-08-16
Avoid recalculation of invariant during joins/exits in constant amp scenario

TomAFrench:
We could leave this code in the original functions but imo it's much cleaner to move it out into a `_beforeJoinExit` as it gets rid of the variable declaration juggling while matching what we have in `BaseWeightedPool` (although with return values).

### DIFF
--- a/pkg/pool-stable/contracts/ComposableStablePool.sol
+++ b/pkg/pool-stable/contracts/ComposableStablePool.sol
@@ -313,10 +313,28 @@ contract ComposableStablePool is
             scalingFactors[isGivenIn ? registeredIndexIn : registeredIndexOut]
         );
 
-        (uint256 preJoinExitSupply, uint256[] memory balances) = _payProtocolFeesBeforeJoinExit(registeredBalances);
-        (uint256 currentAmp, ) = _getAmplificationParameter();
+        uint256[] memory balances;
+        uint256 currentAmp;
+        uint256 preJoinExitSupply;
+        uint256 preJoinExitInvariant;
+        {
+            // New scope to avoid stack-too-deep errors.
+            uint256 oldAmpInvariant;
 
-        uint256 preJoinExitInvariant = StableMath._calculateInvariant(currentAmp, balances);
+            (uint256 lastJoinExitAmp, uint256 lastPostJoinExitInvariant) = getLastJoinExitData();
+            (preJoinExitSupply, balances, oldAmpInvariant) = _payProtocolFeesBeforeJoinExit(
+                registeredBalances,
+                lastJoinExitAmp,
+                lastPostJoinExitInvariant
+            );
+            (currentAmp, ) = _getAmplificationParameter();
+
+            // If the amplification factor is the same as it was during the last join/exit then we can reuse the
+            // value calculated using the "old" amplification factor. If not, then we have to calculate this now.
+            preJoinExitInvariant = currentAmp != lastJoinExitAmp
+                ? StableMath._calculateInvariant(currentAmp, balances)
+                : oldAmpInvariant;
+        }
 
         // These calls mutate `balances` so that it holds the post join-exit balances.
         (uint256 amountCalculated, uint256 postJoinExitSupply) = registeredIndexOut == getBptIndex()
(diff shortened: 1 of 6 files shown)


nventuro (commented):
The logic seems sound! Can you add tests for this?

The protocol fee module tests can be made simpler as they now depend on less state (we don't need to set the last invariant anymore, since this is an argument now), while `_beforeJoinExit` could simply be exposed and tested. It might be a bit annoying however as we'd need to setup an amp change, etc.

pkg/pool-stable/test/ComposableStablePoolProtocolFees.test.ts
```diff
@@ -346,10 +346,6 @@ describe('ComposableStablePoolProtocolFees', () => {
 
         // The virtual supply is some factor of the invariant
         preVirtualSupply = preInvariant.mul(fp(random(1.5, 10))).div(FP_SCALING_FACTOR);
-
-        // This will store the amplification factor and invariant as the lastJoinExit values, as well as setup the
-        // old rates.
-        await pool.updatePostJoinExit(AMPLIFICATION_FACTOR, preInvariant);
```
nventuro, line 352: Note that we _do_ need for the old rates to be setup, but since they equal 1.0 and were set at construction, I think in this case it is unnecessary.

pkg/pool-stable/contracts/ComposableStablePool.sol
```diff
@@ -305,6 +305,13 @@ contract ComposableStablePool is
...
+        (
+            uint256 preJoinExitSupply,
+            uint256[] memory balances,
+            uint256 currentAmp,
+            uint256 preJoinExitInvariant
+        ) = _beforeJoinExit(registeredBalances);
+
         bool isGivenIn = swapRequest.kind == IVault.SwapKind.GIVEN_IN;
 
         _upscaleArray(registeredBalances, scalingFactors);
```
nventuro, line 317: This is wrong :( We need to upscale the `registeredBalances` _before_ calling `_beforeJoinExit`.

nventuro: This doesn't happen on the join/exit functions since we have a layer there that deals with scaling, but we're manually doing all of that here.

pkg/pool-stable/contracts/ComposableStablePool.sol
```diff
@@ -678,6 +682,35 @@ contract ComposableStablePool is
...
+        (
+            uint256 preJoinExitSupply,
+            uint256[] memory balances,
+            uint256 oldAmpInvariant
+        ) = _payProtocolFeesBeforeJoinExit(registeredBalances, lastJoinExitAmp, lastPostJoinExitInvariant);
+        (uint256 currentAmp, ) = _getAmplificationParameter();
+
+        // If the amplification factor is the same as it was during the last join/exit then we can reuse the
+        // value calculated using the "old" amplification factor. If not, then we have to calculate this now.
+        uint256 preJoinExitInvariant = currentAmp != lastJoinExitAmp
```
nventuro, line 707:
```suggestion

        (
            uint256 preJoinExitSupply,
            uint256[] memory balances,
            uint256 oldAmpInvariant
        ) = _payProtocolFeesBeforeJoinExit(registeredBalances, lastJoinExitAmp, lastPostJoinExitInvariant);

        // If the amplification factor is the same as it was during the last join/exit then we can reuse the
        // value calculated using the "old" amplification factor. If not, then we have to calculate this now.
        (uint256 currentAmp, ) = _getAmplificationParameter();
        uint256 preJoinExitInvariant = currentAmp != lastJoinExitAmp
```

Nitpicky - the original whitespace made it a bit hard for me to understand what was going on.

pkg/pool-stable/test/ComposableStablePool.test.ts
```diff
@@ -605,6 +607,124 @@ describe('ComposableStablePool', () => {
...
+
+              // We assume all tokens have similar value, and simply add all of the amounts together to represent how
+              // much value is being added to the Pool. This is equivalent to assuming the invariant is the sum of the
+              // tokens (which is a close approximation while the Pool is balanced).
+
+              const deltaSum = bnSum(deltas);
+              const currSum = bnSum(registeredBalancesWithFees);
+              const poolPercentageDueToDeltas = deltaSum.mul(FP_SCALING_FACTOR).div(currSum);
+
+              const protocolSwapFeesPercentage = fp(0.5);
```
nventuro, line 645: Don't you need for this value to be set globally and the pool to update its cache to get it to be used?

pkg/pool-stable/test/ComposableStablePool.test.ts
```diff
@@ -605,6 +607,124 @@ describe('ComposableStablePool', () => {
...
+      }
+
+      context('when the amplification factor is unchanged from the last join/exit', () => {
+        itPaysProtocolFeesAndReturnsNecessaryData();
+      });
+
+      context('when the amplification factor has changed from the last join/exit', () => {
+        context('when the amplification factor update is ongoing', () => {
+          sharedBeforeEach('perform an amp update', async () => {
+            await pool.startAmpChange(AMPLIFICATION_PARAMETER.mul(2));
```
nventuro, line 710: Let's pass the `duration` argument here - otherwise we need to remember that we default to two days.

pkg/pool-stable/test/ComposableStablePoolProtocolFees.test.ts
TomAFrench: Ah, I missed that they were being set up as well. I've re-added this line just to avoid debugging hell in future.

PR #1609, opened 2022-08-17
Improve post join swap exit tests

nventuro:
Best viewed by ignoring whitespace changes.


TomAFrench (approved):
Looks good to me. We may want to add a quick check that we don't store the amp value on token <> token swaps.

PR #1611, opened 2022-08-17
Extract artifacts for other contracts in task

nventuro:
This is a simple change so we can e.g. extract the artifacts for a pool in a deployment of its factory. I had to change one of the internal functions so that they received the build info from the outside instead of recomputing the same value, but other than that the change is quite minimal.

### DIFF
--- a/pkg/deployments/src/artifact.ts
+++ b/pkg/deployments/src/artifact.ts
@@ -8,12 +8,15 @@ import Task from './task';
 /**
  * Reads each of the task's build-info files and extract the ABI and bytecode for the matching contract.
  */
-export function extractArtifact(task: Task): void {
+export function extractArtifact(task: Task, contractName?: string): void {
   const buildInfoDirectory = path.resolve(task.dir(), 'build-info');
   if (existsSync(buildInfoDirectory) && statSync(buildInfoDirectory).isDirectory()) {
     for (const buildInfoFileName of readdirSync(buildInfoDirectory)) {
-      const contractName = path.parse(buildInfoFileName).name;
-      const artifact = extractContractArtifact(task, contractName);
+      if (contractName == undefined) {
+        contractName = path.parse(buildInfoFileName).name;
+      }
+
+      const artifact = extractContractArtifact(task, contractName, buildInfoFileName);
       writeContractArtifact(task, contractName, artifact);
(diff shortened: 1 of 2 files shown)


pkg/deployments/src/artifact.ts
```diff
@@ -8,12 +8,15 @@ import Task from './task';
...
     for (const buildInfoFileName of readdirSync(buildInfoDirectory)) {
-      const contractName = path.parse(buildInfoFileName).name;
-      const artifact = extractContractArtifact(task, contractName);
+      if (contractName == undefined) {
+        contractName = path.parse(buildInfoFileName).name;
+      }
+
+      const artifact = extractContractArtifact(task, contractName, buildInfoFileName);
       writeContractArtifact(task, contractName, artifact);
     }
```
TomAFrench, line 21: Here we'd looking for a particular contract in all of the buildinfo files.

`extractContractArtifact` throws if it can't find the artifact for the given contract name so `yarn hardhat extract-artifacts` throws as soon as we hit a task which has a build-info file which doesn't contain the all the contracts which the other build-info files are named after (i.e. they come from different packages).

This means that we can't run this task for `2021-04-18-vault` as the Vault (`vault`) and BalancerHelpers (`standalone-utils`) breaks this requirement.

TomAFrench (changes requested):
The way I've handled this in the past is to just create a duplicate build-info for the pool implementation.

pkg/deployments/src/artifact.ts
nventuro: Ah, true.

What if we made all tasks have a single build info?

TomAFrench: Eh, that would mean that we need to separate out tasks for `LiquidityGaugeV5` and `LiquidityGaugeFactory` (as the gauge is vyper so has a separate build-info) which seems ugly to me.

nventuro: I see. But surely having duplicated build infos is not a good idea either?

jubeira:
(see #1951 for an alternative)

nventuro:
Superceded by https://github.com/balancer-labs/balancer-v2-monorepo/pull/1951

PR #1613, opened 2022-08-17
Microcap gauges

jubeira:
This PR gives `LiquidityGaugeV5` and `StakelessGauge` the capability to set an internal maximum weight through a permissioned method.

When performing a checkpoint over these gauges, the internal maximum will be used as an upper bound to the relative weight returned by the `GaugeController`.

Tests cover both `LiquidityGaugeV5` and `StakelessGauge`, and therefore it applies to `SingleRecipientGauge`, `PolygonRootGauge`, `ArbitrumRootGauge` and `OptimismRootGauge`.


pkg/interfaces/contracts/liquidity-mining/ILiquidityGauge.sol
```diff
@@ -28,4 +32,12 @@ interface ILiquidityGauge {
     function killGauge() external;
 
     function unkillGauge() external;
+
+    function set_max_relative_weight(uint256 maxRelativeWeight) external;
+
+    function get_capped_relative_weight(uint256 time) external view returns (uint256);
+
+    function get_current_capped_relative_weight() external view returns (uint256);
```
nventuro, line 40:
```suggestion
    function get_relative_weight_cap() external view returns (uint256);
```

pkg/interfaces/contracts/liquidity-mining/ILiquidityGauge.sol
```diff
@@ -28,4 +32,12 @@ interface ILiquidityGauge {
...
+
+    function get_absolute_max_relative_weight() external pure returns (uint256);
```
nventuro, line 42: I'd consider getting rid of this tbh, we use 18 decimals everywhere

pkg/interfaces/contracts/liquidity-mining/ILiquidityGauge.sol
```diff
@@ -28,4 +32,12 @@ interface ILiquidityGauge {
     function killGauge() external;
 
     function unkillGauge() external;
+
+    function set_max_relative_weight(uint256 maxRelativeWeight) external;
```
nventuro, line 36: I'd consider something like `set_relative_weight_cap`, to use consistent language everywhere.

pkg/liquidity-mining/contracts/gauges/StakelessGauge.sol
```diff
@@ -173,4 +177,51 @@ abstract contract StakelessGauge is IStakelessGauge, ReentrancyGuard {
...
+     * Calling this method is equivalent to reading the public constant; the method is present just to be consistent
+     * with the Vyper gauge counterpart.
+     */
+    function get_absolute_max_relative_weight() external pure override returns (uint256) {
+        return ABSOLUTE_MAX_RELATIVE_WEIGHT;
+    }
+
+    function _min(uint256 a, uint256 b) internal pure returns (uint256) {
+        return a <= b ? a : b;
+    }
```
nventuro, line 226: This is included in the `Math` library

pkg/liquidity-mining/contracts/gauges/StakelessGauge.sol
```diff
@@ -76,6 +79,7 @@ abstract contract StakelessGauge is IStakelessGauge, ReentrancyGuard {
         _rate = rate;
         _period = _currentPeriod();
         _startEpochTime = _tokenAdmin.startEpochTimeWrite();
+        max_relative_weight = ABSOLUTE_MAX_RELATIVE_WEIGHT;
```
nventuro, line 82: I'd pass this as a constructor argument - otherwise it needs to be manually decreased later

pkg/liquidity-mining/contracts/gauges/StakelessGauge.sol
```diff
@@ -32,6 +34,7 @@ abstract contract StakelessGauge is IStakelessGauge, ReentrancyGuard {
     event Checkpoint(uint256 indexed periodTime, uint256 periodEmissions);
 
     // solhint-disable var-name-mixedcase
+    uint256 public max_relative_weight;
```
nventuro, line 37: Make this a private variable (e.g. `_relativeWeightCap`) and provide a getter to it instead

pkg/liquidity-mining/contracts/gauges/StakelessGauge.sol
```diff
@@ -173,4 +177,51 @@ abstract contract StakelessGauge is IStakelessGauge, ReentrancyGuard {
...
+     * @param maxRelativeWeight New maximum relative weight.
+     */
+    function set_max_relative_weight(uint256 maxRelativeWeight) external override {
+        require(msg.sender == address(_authorizerAdaptor), "SENDER_NOT_ALLOWED");
+        require(
+            maxRelativeWeight <= ABSOLUTE_MAX_RELATIVE_WEIGHT,
+            "Max relative weight exceeds allowed absolute maximum"
+        );
+        max_relative_weight = maxRelativeWeight;
+        emit ILiquidityGauge.MaxRelativeWeightChanged(maxRelativeWeight);
```
nventuro, line 193:
```suggestion
        emit MaxRelativeWeightChanged(maxRelativeWeight);
```

I think you don't need to declare the interface. I actually didn't know you _could_ declare it

pkg/interfaces/contracts/liquidity-mining/ILiquidityGauge.sol
```diff
@@ -17,8 +17,12 @@ pragma solidity ^0.7.0;
 // For compatibility, we're keeping the same function names as in the original Curve code, including the mixed-case
 // naming convention.
 // solhint-disable func-name-mixedcase
+// solhint-disable func-param-name-mixedcase
 
 interface ILiquidityGauge {
+    // solhint-disable-next-line var-name-mixedcase
+    event MaxRelativeWeightChanged(uint256 indexed new_max_relative_weight);
```
nventuro, line 24:
```suggestion
    event MaxRelativeWeightChanged(uint256 new_max_relative_weight);
```

There's no point in indexing this

pkg/liquidity-mining/contracts/gauges/ethereum/LiquidityGaugeV5.vy
```diff
@@ -68,6 +68,9 @@ event RewardDistributorUpdated:
     reward_token: indexed(address)
     distributor: address
 
+event MaxRelativeWeightChanged:
+    new_max_relative_weight: indexed(uint256)
```
nventuro, line 72:
```suggestion
    new_max_relative_weight: uint256
```

Same here

pkg/liquidity-mining/contracts/test/MockGaugeController.sol
```diff
@@ -65,11 +63,33 @@ contract MockGaugeController is IGaugeController {
...
+        uint256 weight
+    ) internal {
+        require(!_validGauge[gauge], "Gauge already exists on controller");
+        require(gaugeType >= 0 && gaugeType < _numGaugeTypes, "Invalid gauge type");
+        _validGauge[gauge] = true;
+        if (weight > 0) {
+            _weights[gauge] = weight;
+        }
+        emit NewGauge(gauge, gaugeType, weight);
+    }
```
nventuro, line 94: I'd keep the old `add_gauge`, and instead of having `_with_weight` have `setGaugeWeight` instead.

pkg/liquidity-mining/test/GaugeMaxRelativeWeight.test.ts
```diff
@@ -0,0 +1,213 @@
...
+  sharedBeforeEach('deploy gauge implementation and factory', async () => {
+    const balTokenAdmin = await deploy('MockBalancerTokenAdmin', { args: [vault.address, BAL.address] });
+    const balMinter = await deploy('BalancerMinter', { args: [balTokenAdmin.address, gaugeController.address] });
+
+    // We won't be using the code that requires the VotingEscrowDelegationProxy, so we just use any address.
+    liquidityGaugeImplementation = await deploy('LiquidityGaugeV5', {
+      args: [balMinter.address, ANY_ADDRESS, adaptor.address],
+    });
+    liquidityGaugeFactory = await deploy('LiquidityGaugeFactory', { args: [liquidityGaugeImplementation.address] });
+    stakelessGaugeFactory = await deploy('SingleRecipientGaugeFactory', { args: [balMinter.address] });
```
nventuro, line 53:
```suggestion
    // SingleRecipient is the simplest StakelessGauge, so we test with that instead of using e.g. a mock (which would be
    // identical to SingleRecipient)
    stakelessGaugeFactory = await deploy('SingleRecipientGaugeFactory', { args: [balMinter.address] });
```

pkg/liquidity-mining/test/GaugeMaxRelativeWeight.test.ts
```diff
@@ -0,0 +1,213 @@
...
+    [, admin] = await ethers.getSigners();
+  });
+
+  sharedBeforeEach('deploy authorizer', async () => {
+    vault = await Vault.create({ admin });
+
+    adaptor = await deploy('AuthorizerAdaptor', { args: [vault.address] });
+    gaugeController = await deploy('MockGaugeController', { args: [ZERO_ADDRESS, adaptor.address] });
+
+    await gaugeController.add_type('Ethereum', 0);
```
nventuro, line 36:
```suggestion
    await gaugeController.add_type('Ethereum', GAUGE_TYPE);
```

Make a constant out of this

pkg/liquidity-mining/test/GaugeMaxRelativeWeight.test.ts
```diff
@@ -0,0 +1,213 @@
...
+  sharedBeforeEach('deploy token mocks', async () => {
+    token0 = await deploy('TestBalancerToken', { args: [admin.address, 'Balancer-LP0', 'LP0'] });
+    BAL = await deploy('TestBalancerToken', { args: [admin.address, 'Balancer', 'BAL'] });
+  });
+
+  sharedBeforeEach('deploy gauge implementation and factory', async () => {
+    const balTokenAdmin = await deploy('MockBalancerTokenAdmin', { args: [vault.address, BAL.address] });
+    const balMinter = await deploy('BalancerMinter', { args: [balTokenAdmin.address, gaugeController.address] });
+
+    // We won't be using the code that requires the VotingEscrowDelegationProxy, so we just use any address.
```
nventuro, line 48:
```suggestion
    // We won't be using the code that requires the VotingEscrowDelegationProxy, so we just use any address, since we
    // must initialize to a non-zero value.
```

pkg/liquidity-mining/test/GaugeMaxRelativeWeight.test.ts
```diff
@@ -0,0 +1,213 @@
...
+    itTestsMaxRelativeWeightForGauge();
+  });
+
+  function itTestsMaxRelativeWeightForGauge() {
+    describe('set_maximum_relative_weight', () => {
+      const defaultMaxRelativeWeight = fp(1.0);
+      const newMaxRelativeWeight = fp(0.3);
+      const aboveAbsoluteMaxRelativeWeight = defaultMaxRelativeWeight.add(1);
+
+      context('without permissions', () => {
```
nventuro, line 104:
```suggestion
      context('when the caller is not authorized', () => {
```
It's better to be a bit more explicit about these things, even if it results in added verbosity

pkg/liquidity-mining/test/GaugeMaxRelativeWeight.test.ts
```diff
@@ -0,0 +1,213 @@
...
+        it('reverts', async () => {
+          await expect(gauge.set_max_relative_weight(newMaxRelativeWeight)).to.be.reverted;
```
nventuro, line 106: Consider doing `.connect(other)`, where `other` is just another signer, for even more explicitness.

pkg/liquidity-mining/test/GaugeMaxRelativeWeight.test.ts
```diff
@@ -0,0 +1,213 @@
...
+        });
+      });
+
+      context('with permissions', () => {
+        context('within the allowed range', () => {
+          it('sets max relative weight', async () => {
+            expect(await gauge.max_relative_weight()).to.be.eq(defaultMaxRelativeWeight);
```
nventuro, line 113:
```suggestion
```

pkg/liquidity-mining/test/GaugeMaxRelativeWeight.test.ts
```diff
@@ -0,0 +1,213 @@
...
+        });
+      });
+
+      context('when the gauge is added to the gauge controller', () => {
+        sharedBeforeEach('add gauge to GaugeController', async () => {
+          await gaugeController.add_gauge_with_weight(gauge.address, 0, gaugeControllerWeight);
+        });
+
+        const itChecksMaxRelativeWeightHasNoEffect = (maxRelativeWeightUnderTest: BigNumber, condition: string) => {
+          context(`when max relative weight ${condition} gauge controller relative weight for the gauge`, () => {
```
nventuro, line 161: This on the other hand is a bit _too_ verbose - I get crossed eyes trying to read it.

I'd suggest something like 'when the cap is [larger/smaller] than the the weight' - we don't need to have the word 'gauge' (twice) since it is obvious from the context, nor mention the controller etc.

pkg/liquidity-mining/test/GaugeMaxRelativeWeight.test.ts
```diff
@@ -0,0 +1,213 @@
...
+            let uncappedRelativeWeight: BigNumber;
+            sharedBeforeEach(
+              `set gauge max relative weight ${condition} gauge controller relative weight for the gauge`,
```
nventuro, line 164:
```suggestion
```

For simple preparation steps like this one i'd just not have a description - the code is self explanatory

pkg/liquidity-mining/test/GaugeMaxRelativeWeight.test.ts
```diff
@@ -0,0 +1,213 @@
...
+              async () => {
+                await setNewMaxRelativeWeight(maxRelativeWeightUnderTest);
+                uncappedRelativeWeight = await gaugeController.gauge_relative_weight(gauge.address, anyTimestamp);
```
nventuro, line 167:
```suggestion
```

Instead read this value on each `it` - it's much more direct, and it's not like this will affect performance in any real way

pkg/liquidity-mining/test/GaugeMaxRelativeWeight.test.ts
```diff
@@ -0,0 +1,213 @@
...
+              }
+            );
+
+            it('verifies that max relative weight >= gauge controller relative weight for the gauge', async () => {
+              expect(await gauge.max_relative_weight()).to.be.gte(uncappedRelativeWeight);
+            });
```
nventuro, line 173: This is not a test - it's a test on the tests themselves, to make sure you didn't screw up the setup, a sanity check. 

If you want to include it (which is not a bad idea), don't make an `it` out of it, but instead add it at the end of the `sharedBeforeEach` (the setup you're testing)

pkg/interfaces/contracts/liquidity-mining/ILiquidityGauge.sol
jubeira: about naming: I thought `max_relative_weight` and `capped_relative_weight` as two different things. The former refers to the limit / upper bound itself, and the latter refers to the upper bound _applied_ to the value you get from the controller.

We can use `set_relative_weight_cap` here, but `get_relative_weight_cap` would then be a getter for the upper bound, not the bound applied to the weight (i.e. `min(upper_bound, controller.gauge_relative_weight(...)`).

If we use `cap` everywhere, do you still think it's clear?

pkg/liquidity-mining/contracts/gauges/StakelessGauge.sol
jubeira: Sounds good.
Note that since gauges are actually proxies, that change would affect the `create` method for every factory type.

pkg/liquidity-mining/test/GaugeMaxRelativeWeight.test.ts
jubeira: actually I caught a bug with this line :grimacing: 

At first I had initialized the default in the gauge constructor, but since gauges are proxies that didn't work, and it made me realize I had to do it in the `initialize()` method.

pkg/interfaces/contracts/liquidity-mining/ILiquidityGauge.sol
nventuro: There are two concepts: the limit itself (which can be read and set), and the weight with the limit applied. We should use a similar term for both, e.g. get/set limit and get limited weight, or get/set cap and get capped weight.

You could go with get/set max weight and get maxed weight, but that's a bit odd. I personally like capped.

pkg/liquidity-mining/contracts/gauges/StakelessGauge.sol
nventuro: Yes, that makes sense. They should all receive this argument.

pkg/interfaces/contracts/liquidity-mining/ILiquidityGauge.sol
jubeira: yes, but given that the `set` method reverts if you pass a value above this one, shouldn't we keep it JIC?

pkg/liquidity-mining/test/GaugeMaxRelativeWeight.test.ts
nventuro: Ah good catch.

But that doesn't mean that the test belongs here though. You should have a set of tests for what the contract looks like after creation (e.g. that the constructor arguments were correctly applied). That'd let you spot another incorrect behavior, which is that the constructor is not emitting `MaxRelativeWeightChanged` events.

pkg/interfaces/contracts/liquidity-mining/ILiquidityGauge.sol
jubeira: Thanks; I was hesitant about it too (maybe cap values are somewhat standard?).

Removed.

pkg/interfaces/contracts/liquidity-mining/ILiquidityGauge.sol
jubeira: thanks; done.

pkg/liquidity-mining/contracts/gauges/StakelessGauge.sol
jubeira: Done.

pkg/liquidity-mining/contracts/gauges/StakelessGauge.sol
jubeira: Updated.
`create` was removed from the interface, since `RewardsOnlyGauge` doesn't need this.

pkg/liquidity-mining/contracts/gauges/StakelessGauge.sol
jubeira: Done.

pkg/liquidity-mining/contracts/test/MockGaugeController.sol
jubeira: Done.

pkg/liquidity-mining/test/GaugeMaxRelativeWeight.test.ts
jubeira: Done, although it's the type weight, and it's not actually being used in the mock.

pkg/liquidity-mining/test/GaugeMaxRelativeWeight.test.ts
jubeira: New test added at the beginning (deployment).

pkg/liquidity-mining/test/GaugeMaxRelativeWeight.test.ts
jubeira: +1; done!

pkg/liquidity-mining/test/GaugeMaxRelativeWeight.test.ts
jubeira: +1!

jubeira (commented):
Thanks for the review!
Comments addressed; only one question regarding the max value for the cap is left.

jubeira:
As discussed offline, I've done one more pass over the tests to improve readability.

https://github.com/balancer-labs/balancer-v2-monorepo/pull/1613/commits/2257fd3e427378352fa77a087ea64846404fd266 takes a stab at fixing the scripts, using a default initial cap that would be backwards compatible.

pkg/liquidity-mining/contracts/gauges/StakelessGauge.sol
```diff
@@ -173,4 +179,55 @@ abstract contract StakelessGauge is IStakelessGauge, ReentrancyGuard {
         require(msg.sender == address(_authorizerAdaptor), "SENDER_NOT_ALLOWED");
         _isKilled = false;
     }
+
+    /**
+     * @notice Sets a new relative weight cap for the gauge.
+     * The value shall be normalized to 1e18, and not greater than MAX_RELATIVE_WEIGHT_CAP.
+     * @param relativeWeightCap New relative weight cap.
+     */
```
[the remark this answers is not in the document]
jubeira: Done (moved existing doc as well)

pkg/liquidity-mining/test/GaugeRelativeWeightCap.test.ts
```diff
@@ -0,0 +1,246 @@
...
+  });
+
+  function itTestsRelativeWeightCapForGauge(contractName: string, creationEventName: string) {
+    let gauge: Contract;
+    const setRelativeWeightCap = async (relativeWeightCap: BigNumber): Promise<ContractTransaction> => {
+      const calldata = gauge.interface.encodeFunctionData('set_relative_weight_cap', [relativeWeightCap]);
+      return adaptor.connect(admin).performAction(gauge.address, calldata);
+    };
+
+    describe('gauge creation', () => {
```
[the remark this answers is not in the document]
jubeira: +1; added.

pkg/liquidity-mining/test/GaugeRelativeWeightCap.test.ts
```diff
@@ -0,0 +1,246 @@
...
+          it('sets relative weight cap', async () => {
+            await setRelativeWeightCap(newRelativeWeightCap);
+            expect(await gauge.get_relative_weight_cap()).to.be.eq(newRelativeWeightCap);
+          });
+
+          it('emits an event', async () => {
+            const gaugeRelativeWeightCapChangedInterface = new ethers.utils.Interface([
+              'event RelativeWeightCapChanged(uint256 new_relative_weight_cap)',
+            ]);
+            const tx = await setRelativeWeightCap(newRelativeWeightCap);
```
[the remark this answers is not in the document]
jubeira: this was already refactored :)

jubeira:
Thanks for the review! Comments addressed.

pkg/governance-scripts/contracts/20220415-veBAL-L2/veBALL2GaugeSetupCoordinator.sol
```diff
@@ -49,13 +50,15 @@ contract veBALL2GaugeSetupCoordinator is ReentrancyGuard {
     IVotingEscrow private immutable _votingEscrow;
     IGaugeController private immutable _gaugeController;
     IGaugeAdder private immutable _gaugeAdder;
-    ILiquidityGaugeFactory private immutable _ethereumGaugeFactory;
+    IBaseRootGaugeFactory private immutable _ethereumGaugeFactory;
```
nventuro, line 53: With https://github.com/balancer-labs/balancer-v2-monorepo/pull/1633 merged, we'll no longer need to worry about this.

pkg/liquidity-mining/contracts/gauges/StakelessGauge.sol
```diff
@@ -158,19 +164,44 @@ abstract contract StakelessGauge is IStakelessGauge, ReentrancyGuard {
...
+        return _relativeWeightCap;
+    }
+
+    function getCappedRelativeWeight(uint256 time) external view override returns (uint256) {
+        return _getCappedRelativeWeight(time);
+    }
+
+    function _getCappedRelativeWeight(uint256 time) internal view returns (uint256) {
+        return Math.min(_gaugeController.gauge_relative_weight(address(this), time), _relativeWeightCap);
+    }
```
nventuro, line 198: Let's just make the first version `public` and merge the implementations

pkg/liquidity-mining/contracts/gauges/StakelessGauge.sol
```diff
@@ -158,19 +164,44 @@ abstract contract StakelessGauge is IStakelessGauge, ReentrancyGuard {
...
+
+    function getCurrentCappedRelativeWeight() external view override returns (uint256) {
+        return _getCappedRelativeWeight(_currentPeriod());
+    }
+
+    function getMaxRelativeWeightCap() external pure override returns (uint256) {
+        return MAX_RELATIVE_WEIGHT_CAP;
+    }
```
nventuro, line 206: We can get rid of this

pkg/liquidity-mining/test/GaugeRelativeWeightCap.test.ts
```diff
@@ -0,0 +1,243 @@
...
+      });
+
+      context('when the caller is not authorized', () => {
+        it('reverts', async () => {
+          await expect(gauge.connect(other).setRelativeWeightCap(newCap)).to.be.reverted;
+        });
+      });
+
+      context('when the caller is authorized', () => {
+        context('when the cap value is valid', () => {
```
nventuro, line 133: I'd add one more test for the max cap value

pkg/liquidity-mining/test/GaugeRelativeWeightCap.test.ts
```diff
@@ -0,0 +1,243 @@
...
+          capHasNoEffect(weight.add(1));
+        });
+
+        context('when the cap is below the weight', () => {
+          capAffectsWeight(weight.sub(1));
+        });
+
+        context('when the cap is 0', () => {
+          capAffectsWeight(bn(0));
+        });
```
nventuro, line 206: Lovely

nventuro (approved):
Fantastic! Feel free to merge after addressing comments - we'll take care of the factory stuff once this is in

pkg/liquidity-mining/test/GaugeRelativeWeightCap.test.ts
jubeira: :heart:

PR #1614, opened 2022-08-17
Deployment of ComposableStablePools

nventuro:
We'll be ready to deploy once build info is verified and we are sure no more changes need to be introduced.


EndymionJkb:
Confirmed (on b566b19)

PR #1617, opened 2022-08-18
Cut token count in WeightedPool from 20 to 8

TomAFrench:
We've very likely going to have to cut this token count to fit inside the bytecode limit. It makes sense to do this first so it doesn't cause noise in a later diff.


nventuro (commented):
Ah very nice, I much prefer this alternate style.

I compared bytecode size with and without 0243fd0666ebabdcf74ccd0e61ac1596cd109b63 however and found the old version resulted in 100 fewer bytes. We don't need to revert that change necessarily, but it's interesting to see we can rework these and get different results.

TomAFrench:
hmm, that's annoying. I'm fine with reverting that commit.

PR #1618, opened 2022-08-18
Add ProtocolFeeCache to WeightedPool

TomAFrench:
This PR makes `InvariantGrowthProtocolFees` inherit from `ProtocolFeeCache` and updates the factories to pass the provider's address.

### DIFF
--- a/pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
+++ b/pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
@@ -15,9 +15,11 @@
 pragma solidity ^0.7.0;
 pragma experimental ABIEncoderV2;
 
+import "@balancer-labs/v2-pool-utils/contracts/ProtocolFeeCache.sol";
+
 import "./BaseWeightedPool.sol";
 
-abstract contract InvariantGrowthProtocolFees is BaseWeightedPool {
+abstract contract InvariantGrowthProtocolFees is BaseWeightedPool, ProtocolFeeCache {
     using FixedPoint for uint256;
 
     // This Pool pays protocol fees by measuring the growth of the invariant between joins and exits. Since weights are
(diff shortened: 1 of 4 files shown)


pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
```diff
@@ -26,6 +28,12 @@ abstract contract InvariantGrowthProtocolFees is BaseWeightedPool {
     // This mechanism requires keeping track of the invariant after the last join or exit.
     uint256 private _lastPostJoinExitInvariant;
 
+    constructor(IProtocolFeePercentagesProvider protocolFeeProvider)
```
EndymionJkb, line 31: Because the behavior of WeightedPool and ManagedPool will diverge (ManagedPool needs to have the fixed fee option) - and to conform to the "initialize at the most derived level" convention (e.g., from Stable) - I didn't have a constructor here. InvariantGrowthProtocolFees does derive from ProtocolFeeCache, but I initialize it in WeightedPool

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -67,6 +67,7 @@ contract WeightedPool is BaseWeightedPool, InvariantGrowthProtocolFees {
         uint256[] memory normalizedWeights,
         address[] memory assetManagers,
         uint256 swapFeePercentage,
+        IProtocolFeePercentagesProvider protocolFeeProvider,
```
EndymionJkb, line 70: I guess it doesn't matter, but I had this parameter right under the Vault, like in Stable

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -83,6 +84,7 @@ contract WeightedPool is BaseWeightedPool, InvariantGrowthProtocolFees {
             owner,
             false
         )
+        InvariantGrowthProtocolFees(protocolFeeProvider)
```
EndymionJkb, line 87: I had the `ProtocolFeeCache(protocolFeeProvider, ProtocolFeeCache.DELEGATE_PROTOCOL_SWAP_FEES_SENTINEL)` here; with Weighted it will always delegate

pkg/pool-weighted/contracts/WeightedPool.sol
TomAFrench: Yep, I just noticed that ManagedPool does the same. I'll move this up to match.

EndymionJkb:
There's also more to this; changing WeightedPoolDeployer, etc. Are you handling that?

pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
nventuro: While I agree with @EndymionJkb in this regard, we don't need to worry right now about what the final Managed Pool will look like, and can focus instead on finishing the current iteration of Weighted.

TomAFrench:
Yeah, I'm handling the test failures atm.

EndymionJkb (approved):

PR #1619, opened 2022-08-18
Update benchmarks to use new weighted pool limit

TomAFrench:
There was another usage of 20 as the max number of tokens for weighted pools which was causing benchmarks to blow up.


EndymionJkb:
Pesky benchmarks! Can we define these higher up (in a config.ts file or something) so they only have to change in one place (well, two places counting the contracts)?

PR #1620, opened 2022-08-18
Remove unnecessary paused check in fee calculations

TomAFrench:
In WeightedPool and ManagedPool we bypass some of the fee collection logic if the pool is paused. Due to the introduction of RecoveryMode we don't need to perform this check as all joins and exits are blocked completely while paused so we'll never enter these functions.

I've overridden `_doRecoveryModeExit` in ManagedPool to update the AUM fee collection timestamp to try to recover some of the behaviour we had before however there's a potential manager attack vector that's been opened up, they can just pause the pool to lock funds and extract AUM fees without users being able to exit. This seems fine as they can only do it for the first 90 days of the pool factory existing and so the chances of the EV of attacking outweighing trying to sustain and grow the pool is low.


nventuro:
>they can just pause the pool to lock funds and extract AUM fees without users being able to exit. 

Only governance can pause the pool, so this is actually not possible. We can however revisit ManagedPool behavior and failure modes later, once the mode is complete.

nventuro (approved):

PR #1622, opened 2022-08-18
Pass WeightedPool constructor args in a struct

TomAFrench:
CI will be fixed by #1623 

We're going to run into stack troubles when adding rate providers so I'm splitting out converting the constructor args into a struct in a separate PR for clarity.

### DIFF
--- a/pkg/pool-weighted/contracts/WeightedPool.sol
+++ b/pkg/pool-weighted/contracts/WeightedPool.sol
@@ -59,26 +59,30 @@ contract WeightedPool is BaseWeightedPool, InvariantGrowthProtocolFees {
     uint256 internal immutable _normalizedWeight6;
     uint256 internal immutable _normalizedWeight7;
 
+    struct NewPoolParams {
+        string name;
+        string symbol;
+        IERC20[] tokens;
+        uint256[] normalizedWeights;
+        address[] assetManagers;
+        uint256 swapFeePercentage;
+    }
+
     constructor(
+        NewPoolParams memory params,
         IVault vault,
         IProtocolFeePercentagesProvider protocolFeeProvider,
-        string memory name,
-        string memory symbol,
-        IERC20[] memory tokens,
-        uint256[] memory normalizedWeights,
-        address[] memory assetManagers,
-        uint256 swapFeePercentage,
         uint256 pauseWindowDuration,
         uint256 bufferPeriodDuration,
         address owner
     )
         BaseWeightedPool(
             vault,
-            name,
-            symbol,
-            tokens,
-            assetManagers,
-            swapFeePercentage,
+            params.name,
+            params.symbol,
+            params.tokens,
+            params.assetManagers,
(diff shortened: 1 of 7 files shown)


pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -59,76 +59,80 @@ contract WeightedPool is BaseWeightedPool, InvariantGrowthProtocolFees {
     uint256 internal immutable _normalizedWeight6;
     uint256 internal immutable _normalizedWeight7;
 
+    struct NewPoolParams {
+        string name;
+        string symbol;
+        IERC20[] tokens;
+        uint256[] normalizedWeights;
+        address[] assetManagers;
+        uint256 swapFeePercentage;
```
EndymionJkb, line 68: What do you think about adding the owner to the struct?

In ComposableStable, we had to put everything into one struct because of the stack, but here I think we have a choice. I prefer it like this, where only the "parameters" that can be set by the owner/manager are in the struct. The owner itself is one of those params, though.

EndymionJkb (commented):
Looks good. Do we need to maintain WeightedPoolNoAMFactory.sol? Are we going to have WeightedPools with asset managers, or only ManagedPools?

If the usual case is no asset managers, and we need both, maybe have the main factory with no asset managers, and have a WeightedPoolWithAMFactory?

pkg/pool-weighted/contracts/WeightedPool.sol
TomAFrench: I aimed for consistency with [ManagedPool](https://github.com/balancer-labs/balancer-v2-monorepo/blob/381e8318776c6dbb92c5e2192ecdc13df3ef3d5e/pkg/pool-weighted/contracts/smart/ManagedPool.sol#L166-L178) when creating this struct so I left it out to match that.

Moving it inside gives the benefit of us being able to pass the struct and only the struct to the factory (we have a small bit of inconsistency in the factories atm as ManagedPool takes a struct and owner while WeightedPool takes individual args)

TomAFrench:
> If the usual case is no asset managers, and we need both, maybe have the main factory with no asset managers, and have a WeightedPoolWithAMFactory?

I think if we have something like this deployed already then we should be careful about switching the names (up to a No/With) as it could be easy to mix them up at a glance. I think renaming can happen in a separate PR though.

pkg/pool-weighted/contracts/WeightedPool.sol
TomAFrench: Ah, we leave `owner` outside of the struct as when we have a pool controller then it's not a user chosen value. We'd then have to have two different structs or pass a dummy value for the owner.

TomAFrench: > while WeightedPool takes individual args

and this is due to asset managers being non-optional in one of the factories.

EndymionJkb (approved):

PR #1623, opened 2022-08-18
Fix ManagedPool AUM tests

TomAFrench:
Hoo boy. These tests got broken by me causing fee collection to revert when the pool is paused, which I've fixed but that showed a separate issue which has been sitting around for a while.

It looks like we had managed to duplicate half of the tests at some point (I think around the time at which we added token addition/removals) so I've stripped these out. We need to do another pass on this as it's missing tests for when we're adding a token but this gets us nice green checks again.


EndymionJkb (approved):
Looks good, and glad to remove skips. (And want to fix master before pushing more PRs.)

PR #1624, opened 2022-08-19
Use ProtocolFeeCache in WeightedPool

EndymionJkb:
Since we now have the ProtocolFeeCache providing the protocol fee percentages, we no longer want to use the value passed into joins and exits from the Vault.

I also removed the skipped tests in InvariantGrowthProtocolFees.behavior; since all joins and exits revert, these no longer make sense, at least in that file. They should be tested as part of the regular join/exit suite.

### DIFF
--- a/pkg/pool-weighted/contracts/BaseWeightedPool.sol
+++ b/pkg/pool-weighted/contracts/BaseWeightedPool.sol
@@ -15,7 +15,6 @@
 pragma solidity ^0.7.0;
 pragma experimental ABIEncoderV2;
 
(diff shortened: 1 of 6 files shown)


TomAFrench:
I think that as long as we don't allow the protocol swap fee to be overridden (as in ManagedPool) we could just use the value provided from the Vault. We're performing the SLOAD there anyway so we may as well make use of the value rather than re-reading it again.

TomAFrench (approved):
Agreed to kick the can on `protocolSwapFeePercentage` down the road.

PR #1626, opened 2022-08-19
Post join/exit fees for WeightedPools

EndymionJkb:
Adds calculation of protocol fees for non-proportional joins/exits, closely paralleling the Stable case.

It short-circuits if either the protocol fee is zero, or it is an operation that inherently does not charge protocol fees: initialization, explicit proportional joins/exits, or Managed Pool add/remove. Otherwise, it computes the fee on the taxable amount.

Will make a dependent PR with the tests.


pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
```diff
@@ -64,12 +64,22 @@ abstract contract InvariantGrowthProtocolFees is BaseWeightedPool, ProtocolFeeCa
...
+        uint256[] memory normalizedWeights,
+        uint256 preJoinExitSupply,
+        uint256 postJoinExitSupply
     ) internal virtual override {
         // After all joins and exits we store the post join/exit invariant in order to compute growth due to swap fees
         // in the next one.
+        uint256 protocolSwapFeePercentage;
+        uint256 preJoinExitInvariant;
+
+        if (!isExemptFromProtocolFees) {
```
TomAFrench, line 79:
```suggestion
        // We must calculate the `preJoinExitInvariant` here before we mutate `preBalances` into the post join/exit balances.
        if (!isExemptFromProtocolFees) {
```

pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
```diff
@@ -80,5 +90,64 @@ abstract contract InvariantGrowthProtocolFees is BaseWeightedPool, ProtocolFeeCa
...
+        // then normalize against the final invariant
+        uint256 protocolOwnershipPercentage = invariantDeltaFromFees.divDown(postJoinExitInvariant).mulDown(
+            protocolSwapFeePercentage
+        );
+
+        if (protocolOwnershipPercentage > 0) {
+            uint256 protocolFeeAmount = _calculateAdjustedProtocolFeeAmount(
+                postJoinExitSupply,
+                protocolOwnershipPercentage
+            );
```
TomAFrench, line 148: Shall we abstract this away? There's no difference with the ComposableStablePool code afaict so we could change this into:

```
uint256 protocolFeeAmount = _getJoinExitProtocolSwapFee(
    preJoinExitInvariant,
    postJoinExitInvariant,
    preJoinExitSupply,
    postJoinExitSupply
)
```

EndymionJkb: I'll try; I already moved up part of the calculation into the base protocol fee contract

EndymionJkb:
Superseded by #1636

PR #1629, opened 2022-08-19
Add stable and linear action ids

nventuro:
Now that we have some initial pools, we can compute these.


TomAFrench (approved):
CI is happy so I'm happy.

PR #1630, opened 2022-08-19
Pay protocol fees within WeightedPool rather than InvariantGrowthProtocolFees

TomAFrench:
We're going to have to call into two separate contracts once we add yield fees so we can't just call an inherited version of `_beforeJoinExit` anymore.

To avoid duplicated mints I've changed the `_beforeJoinExit` function in `InvariantGrowthProtocolFees` to return the amount of BPT to mint rather than minting it itself.


nventuro (approved):
This should also make testing easier in general.

There's a bizarre CI failure here, not sure what's up with that.

PR #1631, opened 2022-08-19
Deduplicate invariant growth protocol fee functions

TomAFrench:
I noticed that _calcDueProtocolSwapFeeBptAmountand is just a special case of _getJoinExitProtcolSwapFee (where the start and end supplies are the same.)

I've then pulled this code into the `ProtocolFeeCache` so it can be shared across both contracts. I think we ideally would have it in a separate library maybe (I'm just matching what's been done in #1626 atm).

Tests are a pain and need to get fixed.


TomAFrench:
@nventuro I'm thinking that using the above we can simplify how we charge swap fees in the WeightedPool by removing the call to `_getJoinExitProtocolSwapFee` in the `_beforeJoinExit` hook.

Currently we're processing swap fees twice:
1. From after last join/exit to before the current join/exit
2. From before the current join/exit to after the current join/exit.

The maths _should_ still work if we process swap fees from after the last join/exit to after the current join/exit in a single step.

pkg/pool-weighted/contracts/WeightedMath.sol
```diff
@@ -444,15 +443,12 @@ library WeightedMath {
         //
         // Compute the portion of the invariant increase due to fees
         uint256 supplyGrowthRatio = postJoinExitSupply.divDown(preJoinExitSupply);
-        uint256 feelessInvariant = preJoinExitInvariant.mulDown(supplyGrowthRatio);
 
-        uint256 invariantDeltaFromFees = postJoinExitInvariant - feelessInvariant;
+        uint256 swapFeesPercentage = FixedPoint.ONE - supplyGrowthRatio.divDown(invariantRatio);
 
         // To convert to a percentage of pool ownership, multiply by the rate,
         // then normalize against the final invariant
```
EndymionJkb, line 450: This comment is a bit stale now

TomAFrench: Yep, and the one above as well. Working on it now

pkg/pool-weighted/contracts/WeightedMath.sol
```diff
@@ -407,63 +407,58 @@ library WeightedMath {
...
         // To isolate the amount of increase by fees then, we multiply the original invariant by the supply growth
         // ratio to get the "feeless invariant". The difference between the final invariant and this value is then
         // the amount of the invariant due to fees, which we convert to a percentage by normalizing against the
-        // final (postJoinExit) invariant.
+        // final invariant.
         //
-        // Compute the portion of the invariant increase due to fees
-        uint256 supplyGrowthRatio = postJoinExitSupply.divDown(preJoinExitSupply);
+        // This operation can be simplified to give the below expression for fraction of the Pool which is swap fees.
+        // This allows us to work in terms of ratios of invariants rather than true values.
```
EndymionJkb, line 449: // This operation can be simplified to the expression given below for the fraction of the Pool which is swap fees.
// This allows us to work in terms of ratios of invariants rather than absolute values.

pkg/pool-stable/contracts/ComposableStablePoolProtocolFees.sol
```diff
@@ -235,64 +236,27 @@ abstract contract ComposableStablePoolProtocolFees is
...
 
-        uint256 invariantDeltaFromFees = postJoinExitInvariant - feelessInvariant;
+        uint256 postJoinExitInvariant = StableMath._calculateInvariant(currentAmp, balances);
 
-        // To convert to a percentage of pool ownership, multiply by the rate,
-        // then normalize against the final invariant
-        uint256 protocolOwnershipPercentage = invariantDeltaFromFees.divDown(postJoinExitInvariant).mulDown(
+        // This usage of WeightedMath is intentional.
+        // `_calcDueProtocolSwapFeeBptAmount` is general to several invariants so we can use it here safely.
+        uint256 protocolFeeAmount = WeightedMath._calcDueProtocolSwapFeeBptAmount(
```
nventuro, line 252: Surely this hints at moving it out of WeightedMath?

pkg/pool-weighted/contracts/WeightedMath.sol
```diff
@@ -407,41 +407,58 @@ library WeightedMath {
...
+        // so the growth ratios (invariantGrowthRatio / supplyGrowthRatio) will be equal. In this case, we do not charge
+        // any protocol fees.
 
-        // Assuming the Pool is balanced and token weights have not changed, a growth of the invariant translates into
-        // proportional growth of all token balances. The protocol is due a percentage of that growth: more precisely,
-        // it is due `k = protocol fee * (growth - 1) * balance / growth` for each token.
-        // We compute the amount of BPT to mint for the protocol that would allow it to proportionally exit the Pool and
-        // receive these balances. Note that the total BPT supply will increase when minting, so we need to account for
-        // this in order to compute the percentage of Pool ownership the protocol will have.
+        if (supplyGrowthRatio == invariantGrowthRatio || protocolSwapFeePercentage == 0) return 0;
```
nventuro, line 437:
```suggestion
        if ((supplyGrowthRatio == invariantGrowthRatio) || (protocolSwapFeePercentage == 0)) return 0;
```

plz no operator precedence abuse

pkg/pool-weighted/contracts/WeightedMath.sol
```diff
@@ -407,41 +407,58 @@ library WeightedMath {
...
+        // is due entirely to the balance changes, while the invariant growth also includes swap fees.
+        //
+        // To isolate the amount of increase by fees then, we multiply the original invariant by the supply growth
+        // ratio to get the "feeless invariant". The difference between the final invariant and this value is then
+        // the amount of the invariant due to fees, which we convert to a percentage by normalizing against the
+        // final invariant.
         //
-        // toMint = supply * k / (1 - k)
+        // This operation can be simplified to the expression given below for the fraction of the Pool corresponding
+        // to swap fees. This allows us to work in terms of ratios of invariants rather than absolute values.
```
nventuro, line 449: Could you include a short derivation in the comments? Like 1-2 lines

nventuro: Or at least provide the original and final expressions

pkg/pool-weighted/contracts/WeightedMath.sol
TomAFrench: ser, there's no other way to operate these operators.

pkg/pool-stable/contracts/ComposableStablePoolProtocolFees.sol
TomAFrench: Yeahhhhh, I mentioned this in the OP but I'm prioritising unblocking other PRs atm. I'd to refactor a bunch of usages of this function in tests.

pkg/pool-utils/contracts/InvariantGrowthProtocolSwapFees.sol
```diff
@@ -0,0 +1,88 @@
...
+        // is made up of accumulated swap fees.
+        //
+        // swapFeesPercentage = 1 - supplyGrowthRatio * previousInvariant / currentInvariant
+        //
+        // We then define `invariantGrowthRatio` in a similar fashion to `supplyGrowthRatio` to give the result:
+        //
+        // swapFeesPercentage = 1 - supplyGrowthRatio / invariantGrowthRatio
+        //
+        // Using this form allows us only consider the ratios of the two invariants rather than absolute values,
+        // a useful property as this is sometimes easier than calculating the full invariant twice.
```
EndymionJkb, line 74:
```suggestion
        // Using this form allows us to consider only the ratios of the two invariants, rather than their absolute
        // values: a useful property, as this is sometimes easier than calculating the full invariant twice.
```

TomAFrench: I don't english good.

EndymionJkb:
I also agree we should rename InvariantGrowthProtocolFees to avoid confusion with the library, to something like WeightedProtocolFees. (It would also be parallel to ComposableStableFeeProtocolFees then.)

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -933,7 +934,11 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
             currentBalanceTokenOut.sub(amountOut)
         );
 
-        _payProtocolAndManagementFees(normalizedWeights, preSwapBalances, postSwapBalances);
+        uint256 invariantGrowthRatio = FixedPoint.divDown(
+            WeightedMath._calculateInvariant(normalizedWeights, postSwapBalances),
+            WeightedMath._calculateInvariant(normalizedWeights, preSwapBalances)
+        );
```
nventuro, line 940: You could simply compute the weighted product of the balance ratios here, saving two calls to `pow`.

pkg/pool-weighted/contracts/smart/ManagedPool.sol
```diff
@@ -959,7 +964,11 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
             currentBalanceTokenOut.sub(swapRequest.amount)
         );
 
-        _payProtocolAndManagementFees(normalizedWeights, preSwapBalances, postSwapBalances);
+        uint256 invariantGrowthRatio = FixedPoint.divDown(
+            WeightedMath._calculateInvariant(normalizedWeights, postSwapBalances),
+            WeightedMath._calculateInvariant(normalizedWeights, preSwapBalances)
+        );
+        _payProtocolAndManagementFees(invariantGrowthRatio);
```
nventuro, line 971: Same here

pkg/pool-weighted/contracts/InvariantGrowthProtocolFees.sol
```diff
@@ -53,10 +54,10 @@ abstract contract InvariantGrowthProtocolFees is BaseWeightedPool, ProtocolFeeCa
         uint256 preJoinExitInvariant = WeightedMath._calculateInvariant(normalizedWeights, preBalances);
 
         return
-            WeightedMath._calcDueProtocolSwapFeeBptAmount(
+            InvariantGrowthProtocolSwapFees.calcDueProtocolFees(
+                preJoinExitInvariant.divDown(_lastPostJoinExitInvariant),
+                preJoinExitSupply,
```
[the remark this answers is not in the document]
EndymionJkb: I would also comment it here, since repeating preJoinExitSupply looks weird at first glance.

nventuro (approved):

TomAFrench:
I'm gonna merge this as the CI failures are intermittent and this PR is blocking others waiting to go through.

PR #1632, opened 2022-08-19
Tests for WeightedPool join/exit protocol fees

EndymionJkb:
Tests for afterJoinExit protocol swap fees. Should be close, but apparently I'm doing something wrong checking the post invariant, since it's intermittently failing.


nventuro:
Should this be closed as well?

TomAFrench (approved):

PR #1634, opened 2022-08-22
Reuse _MIN_TOKENS

EndymionJkb:
It's a constant, so no danger making _MIN_TOKENS internal. This avoids redeclaring it in derived classes, and ensures consistency with BasePool.


nventuro:
These two `_MIN_TOKENS` represent different concerns. The composable pool one should instead be renamed to e.g. `_MIN_NON_BPT_TOKENS`.

EndymionJkb:
> _MIN_NON_BPT_TOKENS

It is very closely related: the _MIN_NON_BPT_TOKENS constraint arises from the _MIN_TOKENS constraint of BasePool, and has to be equal, so to me it could go either way. Are we going to define this again in Managed when we add BPT?

I'd originally renamed it _MIN_STABLE_TOKENS, as it was before, but this is clearer.

PR #1635, opened 2022-08-22
Liq. mining - gauges: removing single gauge per recipient restriction

jubeira:
This PR allows creating multiple gauges per recipient (or pool).
It also reorganizes gauge factories' implementation, and deprecates `ISingleRecipientGauge` interface.

Applies to stakeless gauge factories (Polygon, Arbitrum, Optimism, Single recipient) and mainnet's LiquidityGaugeFactory.


pkg/liquidity-mining/contracts/gauges/ethereum/LiquidityGaugeFactory.sol
```diff
@@ -61,22 +35,14 @@ contract LiquidityGaugeFactory is IBaseRootGaugeFactory {
      * Care must be taken to ensure that gauges deployed from this factory are
      * suitable before they are added to the GaugeController.
      *
-     * This factory disallows deploying multiple gauges for a single pool.
+     * It is possible to deploy multiple gauges for a single pool.
```
jubeira, line 38: Note for the reviewer: there's also a check in the adder here: https://github.com/balancer-labs/balancer-v2-monorepo/blob/adf98cc4b017113b016b3b0040d6732046a6e6e4/pkg/liquidity-mining/contracts/admin/GaugeAdder.sol#L127

I understand we won't be deploying a new adder right away. Maybe file a ticket for later?

If we need to deploy a new adder:
- Should we remove that check as well? 
- Do we need to store gauges for a pool, or should we remove the pool --> gauge mapping as we did with stakeless gauges and recipients?

nventuro (commented):
Looks good!

It'd be great to have some basic BaseGaugeFactory tests as well.

pkg/liquidity-mining/contracts/gauges/ethereum/LiquidityGaugeFactory.sol
nventuro: These parts of the adder will be rewored - it will also be aware of L2 recipients, allow for factory removal, potentially redeployment if a gauge is killed, etc. It's not just that change, and at any rate it is not urgent.

jubeira:
Missing test in backlog: #1641

PR #1636, opened 2022-08-23
Support protocol fees on joins and exits (v2)

EndymionJkb:
Replaces #1626 - easier to tack this onto #1631 than trying to merge afterward.

The `_updatePostJoinExit` call sequences is not lovely, but it seemed even worse to special case the _afterJoinExit hook for initialization.


pkg/pool-weighted/contracts/test/MockWeightedPoolProtocolFees.sol
```diff
@@ -0,0 +1,147 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
+
+import "../WeightedPoolProtocolFees.sol";
+
+contract MockWeightedPoolProtocolFees is WeightedPoolProtocolFees {
```
TomAFrench, line 20: We don't actually use this in any tests it seems.

EndymionJkb:
Right - it's preparation. Tests are in another PR (which needs to be
updated)

pkg/pool-weighted/contracts/test/MockWeightedPoolProtocolFees.sol
EndymionJkb: I could remove it and put it in the separate test PR

pkg/pool-weighted/contracts/BaseWeightedPool.sol
```diff
@@ -156,11 +156,18 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool {
...
         uint256[] memory balanceDeltas,
-        uint256[] memory normalizedWeights
+        uint256[] memory normalizedWeights,
+        uint256 preJoinExitSupply,
+        uint256 postJoinExitSupply
     ) internal virtual {
         // solhint-disable-previous-line no-empty-blocks
     }
 
+    // Derived classes may call this to update state after a join or exit.
```
nventuro, line 166:
```suggestion
    // Derived contracts may call this to update state after a join or exit.
```

pkg/pool-weighted/contracts/BaseWeightedPool.sol
```diff
@@ -187,7 +194,8 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool {
         // consistent in Pools with similar compositions but different number of tokens.
         uint256 bptAmountOut = Math.mul(invariantAfterJoin, amountsIn.length);
 
-        _afterJoinExit(true, new uint256[](amountsIn.length), amountsIn, normalizedWeights);
+        // Initialization is still a join, so we need to do post-join work.
+        _updatePostJoinExit(invariantAfterJoin);
```
nventuro, line 198: Maybe mention why we're not calling `_afterJoinExit`

pkg/pool-weighted/contracts/BaseWeightedPool.sol
```diff
@@ -209,14 +217,24 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool {
         uint256[] memory normalizedWeights = _getNormalizedWeights();
 
         _beforeJoinExit(balances, normalizedWeights);
+
+        uint256 preJoinExitSupply = totalSupply();
```
nventuro, line 221: You don't actually need to read it so early, as we haven't yet minted anything.

Also, we might save gas by passing this into `_doJoin`.

pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
```diff
@@ -64,23 +64,43 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
             );
     }
 
-    function _afterJoinExit(
+    function _getJoinExitProtocolFees(
         bool isJoin,
```
nventuro, line 68: Why don't need this any more, since we have both post and pre supply

TomAFrench: @EndymionJkb I think Nico means here that you could replace the usages of `isJoin` with `postJoinExitSupply > preJoinExitSupply`

EndymionJkb: Ah - that makes sense (now addressed with #1652 ). It looked like the *function* wasn't needed anymore somehow

nventuro (approved):

PR #1639, opened 2022-08-23
Divide before multiplying in calcDueProtocolFees

TomAFrench:
We we're dealing with numbers which are less than 1 here, we should divide first and then multiply to keep them as bit as possible.


nventuro (commented):
I don't mind, but these are fixed point multiplications, aren't they? So it shouldn't really matter.

(we could actually use non-fp math and save the scaling factor divs)

nventuro (approved):

PR #1644, opened 2022-08-23
Polygon Root Gauge Factory V2 deployment

jubeira:


nventuro (approved):
I verified the build info matches what's built on commit 94d5d32aa3e1d7780b58a0dca878919346f1814f

PR #1645, opened 2022-08-23
Deploy Mainnet GaugeFactoryV2

nventuro:
These are the so-called microcap gauges for L1 pools.

There's a bunch of expected collisions since the contracts are the same (and hence the selectors match).


TomAFrench (changes requested):
The bytecode onchain for the gauge isn't matching up with what we expect.

nventuro:
@TomAFrench I addressed this in https://github.com/balancer-labs/balancer-v2-monorepo/pull/1645/commits/966f2477962e24d09e1fa8109a5548fac0d77356 - the build info had a leading '0x' sequence, which caused for the string comparison to fail. It was created manually from Vyper output and contained this mistake.

If you could recompile and check again that the build info matches the result of compiling at https://github.com/balancer-labs/balancer-v2-monorepo/pull/1645/commits/966f2477962e24d09e1fa8109a5548fac0d77356 that'd be great.

TomAFrench (approved):
I've checked this again and the bytecode matches what I get when compiling the contracts from source. I've updated the bytecode stored in the `bytecode` dir to not have the leading `0x` either.

PR #1648, opened 2022-08-23
Optimism Root Gauge Factory V2 deployment

jubeira:


nventuro (approved):
I checked the build info at a62913b9d3ec30720ea9577c3661de0a467310da is correct

PR #1652, opened 2022-08-24
Remove `isJoin` argument from _afterJoinExit

TomAFrench:
This PR addresses this review comment: https://github.com/balancer-labs/balancer-v2-monorepo/pull/1636#discussion_r952936468

Made as a separate PR to avoid conflicts on that branch.

### DIFF
--- a/pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
+++ b/pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
@@ -65,14 +65,13 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
     }
 
     function _getJoinExitProtocolFees(
-        bool isJoin,
         uint256[] memory preBalances,
         uint256[] memory balanceDeltas,
         uint256[] memory normalizedWeights,
         uint256 preJoinExitSupply,
         uint256 postJoinExitSupply
     ) internal view returns (uint256, uint256) {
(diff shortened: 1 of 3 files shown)


pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
```diff
@@ -65,14 +65,13 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
...
     function _getJoinExitProtocolFees(
-        bool isJoin,
         uint256[] memory preBalances,
         uint256[] memory balanceDeltas,
         uint256[] memory normalizedWeights,
         uint256 preJoinExitSupply,
         uint256 postJoinExitSupply
     ) internal view returns (uint256, uint256) {
-        uint256 preJoinExitInvariant = WeightedMath._calculateInvariant(normalizedWeights, preBalances);
+        bool isJoin = postJoinExitSupply >= preJoinExitSupply;
```
EndymionJkb, line 74: Since this is only used once, maybe just put `postJoinExitSupply >= preJoinExitSupply` down where it's used? (Maybe with a comment, so it's still clear we're checking for join vs. exit.)

EndymionJkb (approved):
Thanks - should have thought of that :)

pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
TomAFrench: We'll be making the comparison once for every token in the pool in that case.

PR #1654, opened 2022-08-25
Restore preJoin invariant calculation

EndymionJkb:
Maybe this was a merge conflict?

### DIFF
--- a/pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
+++ b/pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
@@ -71,6 +71,7 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
         uint256 preJoinExitSupply,
         uint256 postJoinExitSupply
     ) internal view returns (uint256, uint256) {
+        uint256 preJoinExitInvariant = WeightedMath._calculateInvariant(normalizedWeights, preBalances);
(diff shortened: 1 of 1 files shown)


pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
```diff
@@ -71,6 +71,7 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
         uint256 preJoinExitSupply,
         uint256 postJoinExitSupply
     ) internal view returns (uint256, uint256) {
+        uint256 preJoinExitInvariant = WeightedMath._calculateInvariant(normalizedWeights, preBalances);
```
TomAFrench, line 74:
```suggestion
        // We calculate `preJoinExitInvariant` now before we mutate `preBalances` into the post joinExit balances.
        uint256 preJoinExitInvariant = WeightedMath._calculateInvariant(normalizedWeights, preBalances);
```

TomAFrench (commented):
Ah, this is from me wanting to move `preJoinExitInvariant` closer to where it's used in #1652. We should have a comment explaining why we calculate it early.

EndymionJkb:
> Ah, this is from me wanting to move `preJoinExitInvariant` closer to where it's used in #1652. We should have a comment explaining why we calculate it early.

I believe there was at one time

PR #1655, opened 2022-08-25
Return `preJoinExitSupply` from `_beforeJoinExit`

TomAFrench:
We often use the `_beforeJoinExit` to pay any due protocol fees so we'll always know the total supply and the amount of fees being minted. The pool also needs to know the supply after these fees have been paid in order to properly process the join so we  read the total supply again to pass it into the join/exit handlers.

Instead we can just have `_beforeJoinExit` return the supply to be used for the join/exit as we do in ComposableStablePool

### DIFF
--- a/pkg/pool-weighted/contracts/BaseWeightedPool.sol
+++ b/pkg/pool-weighted/contracts/BaseWeightedPool.sol
@@ -136,7 +136,11 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool {
      * @dev Called before any join or exit operation. Empty by default, but derived contracts may choose to add custom
      * behavior at these steps. This often has to do with protocol fee processing.
      */
-    function _beforeJoinExit(uint256[] memory preBalances, uint256[] memory normalizedWeights) internal virtual {
+    function _beforeJoinExit(uint256[] memory preBalances, uint256[] memory normalizedWeights)
+        internal
+        virtual
+        returns (uint256 preJoinExitSupply)
+    {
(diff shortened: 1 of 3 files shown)


pkg/pool-weighted/contracts/BaseWeightedPool.sol
```diff
@@ -136,8 +136,11 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool {
      * @dev Called before any join or exit operation. Empty by default, but derived contracts may choose to add custom
      * behavior at these steps. This often has to do with protocol fee processing.
```
nventuro, line 137: This is no longer accurate

pkg/pool-weighted/contracts/BaseWeightedPool.sol
```diff
@@ -136,8 +136,11 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool {
...
      */
-    function _beforeJoinExit(uint256[] memory preBalances, uint256[] memory normalizedWeights) internal virtual {
-        // solhint-disable-previous-line no-empty-blocks
+    function _beforeJoinExit(
+        uint256[] memory, /* preBalances */
+        uint256[] memory /* normalizedWeights */
+    ) internal virtual returns (uint256 preJoinExitSupply) {
```
nventuro, line 142:
```suggestion
    ) internal virtual returns (uint256) {
```

nventuro (approved):

PR #1662, opened 2022-08-26
Make _getJoinExitProtocolFees update _lastPostJoinExitInvariant automatically

TomAFrench:
I've had to make `_lastPostJoinExitInvariant` internal so that the mock contract can expose it but I don't think there's any way around that.

### DIFF
--- a/pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
+++ b/pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
@@ -177,6 +177,11 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
         return rateProduct;
     }
 
+    /**
+     * @notice Returns the amount of BPT to be minted to pay protocol fees on yield accrued by the Pool.
+     * @dev Note that this isn't a view function. This function automatically updates `_athRateProduct`  to ensure that
+     * proper accounting is performed to prevent charging duplicate protocol fees.
+     */
     function _getYieldProtocolFee(uint256[] memory normalizedWeights, uint256 preJoinExitSupply)
         internal
         returns (uint256)
@@ -232,6 +237,11 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
             );
     }
 
+    /**
+     * @notice Returns the amount of BPT to be minted to pay protocol fees on swap fees accrued during a join/exit.
+     * @dev Note that this isn't a view function. This function automatically updates `_lastPostJoinExitInvariant` to
+     * ensure that proper accounting is performed to prevent charging duplicate protocol fees.
+     */
     function _getJoinExitProtocolFees(
         uint256[] memory preBalances,
         uint256[] memory balanceDeltas,
--- a/pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
+++ b/pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
@@ -45,7 +45,7 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
     // immutable, the invariant only changes due to accumulated swap fees, which saves gas by freeing the Pool
     // from performing any computation or accounting associated with protocol fees during swaps.
     // This mechanism requires keeping track of the invariant after the last join or exit.
-    uint256 internal _lastPostJoinExitInvariant;
+    uint256 private _lastPostJoinExitInvariant;
 
     constructor(uint256 numTokens, IRateProvider[] memory rateProviders) {
         _require(numTokens <= 8, Errors.MAX_TOKENS);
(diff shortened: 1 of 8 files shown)


pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
```diff
@@ -45,7 +45,7 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
     // immutable, the invariant only changes due to accumulated swap fees, which saves gas by freeing the Pool
     // from performing any computation or accounting associated with protocol fees during swaps.
     // This mechanism requires keeping track of the invariant after the last join or exit.
-    uint256 private _lastPostJoinExitInvariant;
+    uint256 internal _lastPostJoinExitInvariant;
```
EndymionJkb, line 48: I'd really like to keep this private. There is a public getter `getLastInvariant()`, so what you can do is leave this private and call that public getter from the Mock.

Could do the same thing for the yield: make it private, add a public getter, and call that in the Mock.

It's a little dangerous / non-future-proofed to make mutable important things non-private, when the intent is to encapsulate their behavior in this protocol fees contract.

pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
```diff
@@ -238,7 +238,7 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
         uint256[] memory normalizedWeights,
         uint256 preJoinExitSupply,
         uint256 postJoinExitSupply
-    ) internal view returns (uint256, uint256) {
+    ) internal returns (uint256) {
```
EndymionJkb, line 241: Would loudly document the side effect here; it's not a pure getter. (Same for yield.)

pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
TomAFrench: Ah, I missed that getter, I swtich to use that. One downside to this however is that we're forced into spending the bytecode in the production contract for a getter which is only relevant for testing.

TomAFrench: I guess we have the bytecode to spare for now so no harm.

EndymionJkb: Is it only relevant for testing? If so, the getter could be internal (getting a little back).

It seems like the ATH Rate might be useful for UIs, etc. That getter could be public

pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
```diff
@@ -177,6 +177,11 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
         return rateProduct;
     }
 
+    /**
+     * @notice Returns the amount of BPT to be minted to pay protocol fees on yield accrued by the Pool.
```
EndymionJkb, line 181: We don't usually use notice for internal functions. Maybe just @dev for the whole thing? (Same as below)

TomAFrench: I quite like to have both anyway as the presence of a `@dev` tag means that there's a gotcha to the function to be aware of. I've removed this though.

pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
```diff
@@ -105,6 +105,10 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
             );
     }
 
+    function getATHRateProduct() external view returns (uint256) {
```
EndymionJkb, line 108: Well, now `getLastInvariant` should also be external (so we can call it directly from the test, vs exposing it in a Mock).

Can we move put them together near the top (showing the external interface first), and add @notice (since they're external)

TomAFrench: We weren't exposing it through the mock but I've changed `getLastInvariant` to external from public

pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
```diff
@@ -71,12 +71,20 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
     }
 
     /**
-     * @dev Returns the value of the invariant after the last join or exit operation.
+     * @notice Returns the value of the invariant after the last join or exit operation.
      */
-    function getLastInvariant() public view returns (uint256) {
+    function getLastInvariant() external view returns (uint256) {
```
EndymionJkb, line 76: Should we rename this to `getLastPostJoinExitInvariant` for consistency?

EndymionJkb (approved):
You fixed it before I finished typing :) Looks good now

TomAFrench:
Oops, just realised I changed a fork test as well there. I'll revert that in a separate PR

PR #1663, opened 2022-08-26
Merge mocks for weighted pool protocol fees

TomAFrench:


TomAFrench:
The vault's test failure is due to it not being able to download the vyper compiler. I haven't changed anything that would impact the vault so we can ignore this.

EndymionJkb:
> The vault's test failure is due to it not being able to download the vyper compiler. I haven't changed anything that would impact the vault so we can ignore this.

When there's a BS failure like this you can go into Actions and rerun just the failed jobs

EndymionJkb (commented):
Looks good, but I would adjust 1662 first, get that merged, then update this one.

EndymionJkb (approved):

Issue #1507, opened 2022-07-20
Add read-only chained references


nventuro:
We've already thought of a number of scenarios in which they'd prove useful.

nventuro:
We could have a second prefix that marks read-only refs, and e.g. have something like the following:

```
function _getChainedReferenceValue(uint256 ref) internal override returns (uint256 value) {
    (bytes32 slot, bool clear) = _getTempStorageSlot(ref);

    // Since we do manual calculation of storage slots, it is easier (and cheaper) to rely on internal assembly to
    // access it.

    // solhint-disable-next-line no-inline-assembly
    assembly {
        value := sload(slot)
    }

    if (clear) {
        sstore(slot, 0)   
    }
}
```

jubeira:
Closed via #1664.

PR #1664, opened 2022-08-26
[Batch Relayer V4] Introducing temporary and read-only references.

jubeira:
See #1507.

This PR modifies chained references so that they can be kept after reading them, instead of always clearing them.

Summary of the changes:
- Chained references are now identified by the prefix `0xba1` instead of `0xba10`.
- The previous prefix `0xba10` now identifies temporary references. So by default, references used up to this point will be cleared, ensuring backwards compatibility.
- The prefix `0xbaY` where `Y > 0` will identify references not to be deleted after a read (i.e. read-only).

### DIFF
--- a/pkg/standalone-utils/contracts/relayer/BaseRelayerLibrary.sol
+++ b/pkg/standalone-utils/contracts/relayer/BaseRelayerLibrary.sol
@@ -118,6 +118,19 @@ contract BaseRelayerLibrary is IBaseRelayerLibrary {
      * @dev Returns true if `amount` is not actually an amount, but rather a chained reference.
      */
     function _isChainedReference(uint256 amount) internal pure override returns (bool) {
+        // First 3 nibbles are enough to determine if it's a chained reference.
+        return
+            (amount & 0xfff0000000000000000000000000000000000000000000000000000000000000) ==
+            0xba10000000000000000000000000000000000000000000000000000000000000;
+    }
+
+    /**
+     * @dev Returns true if `ref` is temporary reference, i.e. to be deleted after reading it.
+     */
+    function _isTemporaryChainedReference(uint256 amount) internal pure returns (bool) {
+        // First 3 nibbles determine if it's a chained reference.
+        // If the 4th nibble is 0 it is temporary; otherwise it is considered read-only.
+        // In practice, we shall use '0xba11' for read-only references.
         return
             (amount & 0xffff000000000000000000000000000000000000000000000000000000000000) ==
             0xba10000000000000000000000000000000000000000000000000000000000000;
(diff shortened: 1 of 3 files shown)


pkg/standalone-utils/contracts/relayer/BaseRelayerLibrary.sol
```diff
@@ -149,7 +164,13 @@ contract BaseRelayerLibrary is IBaseRelayerLibrary {
         // solhint-disable-next-line no-inline-assembly
         assembly {
             value := sload(slot)
-            sstore(slot, 0)
+        }
+
+        if (_isTemporaryChainedReference(ref)) {
```
jubeira, line 169: Some thoughts about naming this change.

We could invert the checks and search for 'read-only' prefixes instead of 'temporary' ones. But then we would need to either negate the result, or compare to `false` in this `if` clause. IMO it's more readable like this.

pkg/standalone-utils/contracts/relayer/BaseRelayerLibrary.sol
```diff
@@ -138,25 +151,33 @@ contract BaseRelayerLibrary is IBaseRelayerLibrary {
...
-     * @dev Returns the amount referenced by chained reference `ref`. Reading an amount clears it, so they can each
-     * only be read once.
+     * @dev Returns the amount referenced by chained reference `ref`.
+     * If the reference is temporary, it will be cleared after reading it, so they can each only be read once.
+     * If the reference is not temporary (i.e. read-only), it will not be cleared after reading it
+     * (see `_isTemporaryChainedReference` function).
      */
     function _getChainedReferenceValue(uint256 ref) internal override returns (uint256 value) {
-        bytes32 slot = _getTempStorageSlot(ref);
+        bytes32 slot = _getStorageSlot(ref);
```
jubeira, line 160: Note: in #1507 the suggestion was to return a flag in `_getTempStorageSlot`, but the method is also used in `_setChainedReferenceValue`, and we don't need that flag there.

pkg/standalone-utils/contracts/relayer/BaseRelayerLibrary.sol
```diff
@@ -118,6 +118,19 @@ contract BaseRelayerLibrary is IBaseRelayerLibrary {
...
+            0xba10000000000000000000000000000000000000000000000000000000000000;
+    }
+
+    /**
+     * @dev Returns true if `ref` is temporary reference, i.e. to be deleted after reading it.
+     */
+    function _isTemporaryChainedReference(uint256 amount) internal pure returns (bool) {
+        // First 3 nibbles determine if it's a chained reference.
+        // If the 4th nibble is 0 it is temporary; otherwise it is considered read-only.
+        // In practice, we shall use '0xba11' for read-only references.
```
nventuro, line 133: Note that the PR description says that e.g. `0xba2` would be read-only. I prefer the approach as shown in the code, and would update the description.

nventuro (approved):

PR #1667, opened 2022-08-29
Return only manager's BPT from external AUM collection functions

TomAFrench:
In the natspec for the external functions for collecting AUM fees from ManagedPool we say that we return the amount of BPT which is minted to the manager. We actually return the total amount of BPT minted for both the manager and the protocol.

This PR separates out these two values so we can return the correct values here.

### DIFF
--- a/pkg/pool-weighted/contracts/smart/ManagedPool.sol
+++ b/pkg/pool-weighted/contracts/smart/ManagedPool.sol
@@ -822,7 +822,7 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard {
         // This is only necessary if the pool has been initialized (which is indicated by a nonzero total supply).
         uint256 supplyBeforeFeeCollection = totalSupply();
         if (supplyBeforeFeeCollection > 0) {
-            amount = _collectAumManagementFees(supplyBeforeFeeCollection);
+            (, amount) = _collectAumManagementFees(supplyBeforeFeeCollection);
         }
 
         _setManagementAumFeePercentage(managementAumFeePercentage);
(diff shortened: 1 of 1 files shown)


EndymionJkb:
I wonder if we should even have this return value, though. 1) We don't actually use it anywhere; 2) `setters` don't generally return values (or side effects other than emitting events), and 3) the signature in IControlledManagedPool doesn't have a return value.

Also, should _payProtocolFees emit an event then? It doesn't now (other than the transfer, which is how we check it in the tests).

TomAFrench:
Yeah, there's an argument to be made there. I remember we added this return value for a couple reasons:
1. We have the value so why not? It avoids the `exitPool` issue of realising that having a return value here would be useful down the line when it's too late.
2. We avoid the contract existence check when calling these functions which makes them cheaper in solc 0.8.10+

That said there's not much chance of this return value being particularly useful for callers:
1. We can't always predict the value which a caller will be interested it. Do they want to know how much BPT the manager got or just how much fees were collected in general?
2. The fact that these fees are automatically collected on joins/exits and by arbitrary addresses poking the ManagedPool also means that a contract can't perform proper accounting using the return values from these functions as it will be missing most of the data.

---
Having written all this though, I can see a benefit where having a return value simplifies a manager seeing how much fees they can collect. They can just perform a static call to this function and read the return amount rather than having to do "full simulation" and extract the `ManagementAumFeeCollected` event.

TomAFrench:
> Also, should _payProtocolFees emit an event then? It doesn't now (other than the transfer, which is how we check it in the tests).

I think we should avoid this:
1. It would add extra gas costs to every swap in the ManagedPool.
2. Protocol fees can be unambiguously determined by looking for `Transfer(address(0), address(protocolFeeCollector), amount)` events whereas managers can join the pool themselves so we can't distinguish between a join or fees without an event.

(Granted someone could perform a join with the recipient as the protocol fee collector but it's unlikely to happen with any significant amount and it's then treated the same as any other protocol fees.)

EndymionJkb (approved):

PR #1668, opened 2022-08-29
Consolidate mints of protocol fees on joins/exits

TomAFrench:
builds on #1667 as we need to distinguish between manager and protocol BPT.

Currently we calculate protocol fees before and after the join and immediately mint the necessary amount of BPT, resulting in two extra warm SSTOREs (balance + total supply).

I've changed the `_beforeJoinExit` and `_afterJoinExit` hooks to return the amount of protocol fees to be collected and we now mint those BPT at the same time.

### DIFF
--- a/pkg/pool-weighted/contracts/BaseWeightedPool.sol
+++ b/pkg/pool-weighted/contracts/BaseWeightedPool.sol
@@ -133,8 +133,8 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool {
     }
 
     /**
-     * @dev Called before any join or exit operation. Returns the Pool's total supply by default, but derived contracts
-     * may choose to add custom behavior at these steps. This often has to do with protocol fee processing.
+     * @dev Called before any join or exit operation. Derived contracts may add custom behavior here to calculate any
+     * fees to be paid to the protocol (or e.g. the Pool's owner) before the join or exit.
      * @return preJoinExitTotalSupply - the total supply once all protocol fees prior to the join/exit have been paid.
      * @return preJoinExitProtocolFees - the amount of protocol fees which are to be collected before the join/exit.
      */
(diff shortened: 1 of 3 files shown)


pkg/pool-weighted/contracts/BaseWeightedPool.sol
```diff
@@ -139,26 +139,26 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool {
     function _beforeJoinExit(
         uint256[] memory, /* preBalances */
         uint256[] memory /* normalizedWeights */
-    ) internal virtual returns (uint256) {
-        return totalSupply();
+    ) internal virtual returns (uint256, uint256) {
```
EndymionJkb, line 142: The comment says "Returns the Pool's total supply by default" - should also mention the second return value. Maybe something like:
```
 * @dev Called before any join or exit operation. Returns the Pool's total supply and zero due protocol fees
 * by default, but derived contracts may add custom behavior here to calculate a BPT protocol fee amount.
 * This often has to do with optimizing the collection of fees from multiple sources.
```

TomAFrench: I've added some `@return` tags to the natspec. Would you be happy if we removed the references to the default return values in here and just had something like:

```
    /**
     * @dev Called before any join or exit operation. Derived contracts may add custom behavior here to calculate any
     * fees to be paid to the protocol (or e.g. the Pool's owner) before the join or exit.
     * @return preJoinExitTotalSupply - the total supply once all fees prior to the join/exit have been paid.
     * @return preJoinExitProtocolFees - the amount of protocol fees which are to be collected before the join/exit.
     */
```

EndymionJkb: Yes; I noted those were still there and a bit stale. That looks good.

TomAFrench:
With the required changes to ManagedPool (which would result in extra bytecode) I'm becoming a bit less of a fan of this approach tbh. Ideally we could implement this on just WeightedPool without requiring the same on ManagedPool (without lying and returning 0 protocol fees to mint)

PR #1669, opened 2022-08-29
Enforce controlled pool interface on pools

TomAFrench:
Addressing https://github.com/balancer-labs/balancer-v2-monorepo/pull/1667#issuecomment-1230353575.

Currently we aren't enforcing that Pools implement the interface that the `ManagedPoolController` uses to interact with the pool. which could lead to disastrous consequences in future if the Pool interface changes and we don't update the Controller to match. There's been a little bit of this already as the interface expects a `withdrawCollectedManagementFees` function on the pool (a leftover from the `InvestmentPool` days.)

I've removed how `ManagedPoolController` implements the `IControlledManagedPool` interface for the time being as it doesn't 100% line up with how we're handling management fees currently.

### DIFF
--- a/pkg/interfaces/contracts/pool-utils/IControlledManagedPool.sol
+++ b/pkg/interfaces/contracts/pool-utils/IControlledManagedPool.sol
@@ -31,9 +31,9 @@ interface IControlledManagedPool is IControlledPool {
 
     function setMustAllowlistLPs(bool mustAllowlistLPs) external;
 
-    function withdrawCollectedManagementFees(address recipient) external;
+    function collectAumManagementFees() external returns (uint256);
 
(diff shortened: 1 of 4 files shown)


EndymionJkb (approved):
Good catch on withdrawCollectedManagementFees! I'm trying to remember why these didn't implement the interfaces from the beginning... it seems there was some reason, but I don't remember it.

These controllers may be unlikely to survive as is - though eventually we'll have "standard" controllers. (Maybe the V1 "rights" model is obsolete, as we're moving toward controllers with more internal logic, and only high level direction from managers.)

In any case, while we have them, they should be correct and consistent.

TomAFrench:
> it seems there was some reason, but I don't remember it.

I think it was probably due to these being around the time of the circular dependency issue which meant we had to split off the interfaces package.

PR #1670, opened 2022-08-29
Move the "bptForPoolPercentage" logic into a library.

TomAFrench:
This PR started as refactoring the logic in `_calculateAdjustedProtocolFeeAmount ` so that's it's in the `InvariantGrowthProtocolSwapFees` library in a reusable way.

I wasn't a huge fan of this as it results in us using the `InvariantGrowthProtocolSwapFees` library for non-swap fees so I moved this into a generic `ProtocolFees` library, at which point it made sense to make a directory in `pool-utils` specific to protocol fees.

### DIFF
--- a/pkg/pool-utils/contracts/protocol-fees/ProtocolFees.sol
+++ b/pkg/pool-utils/contracts/protocol-fees/ProtocolFees.sol
@@ -23,13 +23,19 @@ library ProtocolFees {
     /**
      * @dev Calculates the amount of BPT necessary to give ownership of a given percentage of the Pool.
      * Note that this function reverts if `poolPercentage` >= 100%, it's expected that the caller will enforce this.
+     * @param totalSupply - The total supply of the pool prior to minting BPT.
+     * @param poolOwnershipPercentage - The desired ownership percentage of the pool to have as a result of minting BPT.
      * @return bptAmount - The amount of BPT to mint such that it is `poolPercentage` of the resultant total supply.
      */
-    function bptForPoolPercentage(uint256 totalSupply, uint256 poolPercentage) internal pure returns (uint256) {
+    function bptForPoolOwnershipPercentage(uint256 totalSupply, uint256 poolOwnershipPercentage)
+        internal
+        pure
+        returns (uint256)
+    {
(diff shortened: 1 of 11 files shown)


pkg/pool-utils/contracts/protocol-fees/ProtocolFees.sol
```diff
@@ -0,0 +1,35 @@
...
+
+library ProtocolFees {
+    using FixedPoint for uint256;
+
+    /**
+     * @dev Calculates the amount of BPT necessary to give ownership of a given percentage of the Pool.
+     * Note that this function reverts if `poolPercentage` >= 100%, it's expected that the caller will enforce this.
+     * @return bptAmount - The amount of BPT to mint such that it is `poolPercentage` of the resultant total supply.
+     */
+    function bptForPoolPercentage(uint256 totalSupply, uint256 poolPercentage) internal pure returns (uint256) {
```
EndymionJkb, line 28: Should this be called something like bptForProtocolOwnershipPercentage? I'm afraid it's confusable with the fee percentage itself, which it *might* correspond to (e.g., for managed AUM fees), but not always, when we're really adjusting a pre-mint to a post-mint percentage of ownership.

Maybe, to be totally clear:

function bptForProtocolOwnershipPercentage(uint256 totalSupply, uint256 preMintProtocolOwnershipPercentage)

EndymionJkb:
Should there be a test file for this, even though it's simple and only one function? (e.g., testing for that revert on 100%).

pkg/pool-utils/contracts/protocol-fees/ProtocolFees.sol
TomAFrench: I'd like to avoid repeating "protocol" as it's included in the library name (also as we'll be using this for manager fees so it's not entirely true). I've changed it to:

`function bptForPoolOwnershipPercentage(uint256 totalSupply, uint256 poolOwnershipPercentage)`

>  we're really adjusting a pre-mint to a post-mint percentage of ownership.

Not sure what you mean by this. The percentage ownership should be the same pre and post mint (just beforehand the protocol's ownership is through debt rather than BPT).

EndymionJkb (approved):

PR #1671, opened 2022-08-30
Composable stable pool virtual supply edge case tests

danielmkm:
A series of tests that showcase arb opportunities that are created by not including unminted protocol fee BPTs in the virtual supply of the ComposableStablePool. The protocol and swap fees used are extreme to highlight the underlying issue.

All test cases currently fail.

In order to allow for calling `StablePool.init` for a pool containing a nested BPT (ie: bb-a-USD-TUSD), a `skipMint` flag was added since its not possible to simply mint the nested BPT.


TomAFrench:
These test cases have been incorporated into #1674 so I'm going to close this PR. Could you take a look at that PR as we had to modify your test cases a little bit.

PR #1672, opened 2022-08-30
[Batch Relayer V4] [2] Public 'peek' for chained references

### LINKED ISSUE #1509: Add a public way to read chained references
Relayer calls are currently a bit obscure in that it is not possible to `eth_call` them and get a sense of the result. While `multicall` does return all of the intermediate return values, most functions (all except swaps?) don't actually return anything. Since we already use the reference mechanism to store values, we could expose a getter for these so that then can be read on an `eth_call` (and then removed in the actual transaction).

This would likely be `peek` instead of `get` (see #1507). So we could e.g. do `[unstake, exit, swap, join({ outputRef: 42 }), peekRef(42), stake(input: 42)]` and get the `join` return value in `results[5]`, and then removing the call to `peekRef` from the `multicall` for the actual call.

jubeira:
Closes #1509.

This PR adds a public `peek` method in `BaseRelayerLibrary` that returns the value of a chained reference without modifying it.

### DIFF
--- a/pkg/standalone-utils/contracts/relayer/BaseRelayerLibrary.sol
+++ b/pkg/standalone-utils/contracts/relayer/BaseRelayerLibrary.sol
@@ -81,6 +81,15 @@ contract BaseRelayerLibrary is IBaseRelayerLibrary {
         token.approve(address(getVault()), amount);
     }
 
+    /**
+     * @notice Returns the amount referenced by chained reference `ref`.
+     * @dev It does not alter the reference (even if it's marked as temporary), and it does not expose the reference
+     * slot (see `_peekChainedReferenceValue`).
+     */
+    function peekChainedReferenceValue(uint256 ref) public view override returns (uint256 value) {
+        (, value) = _peekChainedReferenceValue(ref);
+    }
+
     function _pullToken(
         address sender,
         IERC20 token,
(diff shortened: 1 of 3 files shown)


nventuro:
Not sure why we have merge conflicts - can you address those?

jubeira:
The conflict was because we had chained PRs, this one has multiple commits and we squash-merged the first one in the chain.

Rebased with no changes; now it can be reviewed.

pkg/standalone-utils/contracts/relayer/BaseRelayerLibrary.sol
```diff
@@ -81,6 +81,15 @@ contract BaseRelayerLibrary is IBaseRelayerLibrary {
         token.approve(address(getVault()), amount);
     }
 
+    /**
+     * @notice Returns the amount referenced by chained reference `ref`.
+     * @dev It does not alter the reference (even if it's marked as temporary), and it does not expose the reference
+     * slot (see `_peekChainedReferenceValue`).
```
nventuro, line 87:
```suggestion
     * @dev It does not clear the reference (even if it's marked as temporary).
```

Saying 'it doesn't do X' seems a bit out of scope - you wouldn't expect for this function to return the slot.

pkg/standalone-utils/contracts/relayer/BaseRelayerLibrary.sol
```diff
@@ -157,14 +166,8 @@ contract BaseRelayerLibrary is IBaseRelayerLibrary {
...
-        bytes32 slot = _getStorageSlot(ref);
-
-        // Since we do manual calculation of storage slots, it is easier (and cheaper) to rely on internal assembly to
-        // access it.
-        // solhint-disable-next-line no-inline-assembly
-        assembly {
-            value := sload(slot)
-        }
+        bytes32 slot;
+        (slot, value) = _peekChainedReferenceValue(ref);
```
nventuro, line 170: I'd delete the named return value, declare both slot and value here, and explicitly return value at the end

pkg/standalone-utils/test/BaseRelayerLibrary.test.ts
```diff
@@ -110,6 +110,42 @@ describe('BaseRelayerLibrary', function () {
...
+          expect(await relayerLibrary.peekChainedReferenceValue(reference.add(1))).to.be.eq(0);
+        });
+
+        it('peeks same slot multiple times', async () => {
+          await relayerLibrary.setChainedReferenceValue(reference, 19);
+          expect(await relayerLibrary.peekChainedReferenceValue(reference)).to.be.eq(19);
+          expect(await relayerLibrary.peekChainedReferenceValue(reference)).to.be.eq(19);
+          expect(await relayerLibrary.peekChainedReferenceValue(reference)).to.be.eq(19);
+
+          await expectChainedReferenceContents(reference, 19);
```
nventuro, line 140:
```suggestion
```

nventuro (approved):

jubeira:
Thanks for the review! Comments addressed; will merge after CI gives me the green light.

PR #1674, opened 2022-08-30
Make `getRate` take payable protocol fees into account.

TomAFrench:
This is based of just intuiting how we should make this adjustment so feel free to push back if you disagree on whether it's the right way to go about this.

---

In `getRate` we now calculate protocol fees so that we can calculate what fraction of the pool is owned by the protocol. The true value of the BPT is then `1 - protocolPoolOwnershipPercentage` times what it would otherwise be.

We make use of the same optimisation as we use in `_beforeJoinExit` to avoid recalculation of the invariant in the constant amplification situation.

This PR will need some testing before merging.


nventuro:
(we should merge into `composable-stable-deployment`, but I'm commenting here now regardless)

pkg/pool-stable/contracts/ComposableStablePool.sol
```diff
@@ -934,10 +934,26 @@ contract ComposableStablePool is
...
-
-        return StableMath._getRate(balances, currentAmp, virtualSupply);
+        if (currentAmp == lastJoinExitAmp) {
+            // If the amplitude hasn't changed then we can reuse the already calculated invariant.
+            return oldAmpInvariant.divDown(virtualSupply).mulDown(protocolOwnershipPercentage.complement());
+        } else {
+            return
+                StableMath._getRate(balances, currentAmp, virtualSupply).mulDown(
+                    protocolOwnershipPercentage.complement()
+                );
```
nventuro, line 955: Instead of using `complement` and reasoning about it this way, I'd simply call

```
uint256 protocolFeeAmount = _calculateAdjustedProtocolFeeAmount(
    virtualSupply,
    expectedProtocolOwnershipPercentage
);
```

and add that to `virtualSupply`, to get the _real_ virtual supply. Those BPT should be minted - the only reason why it hasn't happened yet is due to gas optimizations favoring regular swaps. They will in fact be minted before the next interaction that involves BPT (and hence update the virtual supply).

nventuro: So this could become

```
uint256 currentInvariant = (currentAmp == lastJoinExitAmp) ? 
  oldAmplnvariant : 
  StableMath.calculateInvariant(currentAmp, balances);

return currentInvariant.div(virtualSupply.add(protocolFeeAmount));
```

(`StableMath.getRate` is just that division)

nventuro: Note that this clearly shows that updating the protocol fee rate cache manipulates the rate - we need to decide if that's an issue.

pkg/pool-stable/test/ComposableStablePool.test.ts
```diff
@@ -1581,24 +1874,41 @@ describe('ComposableStablePool', () => {
...
+            // Perform a very small proportional join. This ensures that the rate should not increase from swap fees
+            // due to this join so this can't mask issues with the rate.
+            const poolBalances = await pool.getBalances();
+            const amountsIn = poolBalances.map((balance, i) => (i == bptIndex ? bn(0) : balance.div(10000)));
+            await pool.joinGivenIn({ from: lp, amountsIn });
+
+            const rateAfterJoin = await pool.getRate();
+
+            const rateDelta = rateAfterJoin.sub(rateBeforeJoin);
+            expect(rateDelta.abs()).to.be.lte(2);
```
nventuro, line 1911: Why is there a difference here? Rounding in the invariant?

pkg/pool-stable/contracts/ComposableStablePool.sol
nventuro: Actually, I'd consider changing `getVirtualSupply` as well to include this

pkg/pool-stable/test/ComposableStablePool.test.ts
```diff
@@ -1537,6 +1538,298 @@ describe('ComposableStablePool', () => {
           expect(virtualSupply).to.be.equalWithError(senderBptBalance, 0.0001);
         });
       });
+
+      context('nested pools (beethovenxfi: PR 1671)', () => {
```
nventuro, line 1542: I'd consider placing this in a separate file, along with other complex scenarios we might produce, to try to keep things tidy. The sheer size of this file makes it hard to navigate.

EndymionJkb: @TomAFrench do you want me to move this out (since I put it in?) I actually started it as a separate file, but then didn't want to duplicate all the initial setup. We could put the others in the new file too (the rate-lowering one would still fail).

pkg/pool-stable/test/ComposableStablePool.test.ts
TomAFrench: Yeah, I get small errors in the last digit here.

pkg/pool-stable/test/ComposableStablePool.test.ts
```diff
@@ -1581,24 +1874,41 @@ describe('ComposableStablePool', () => {
...
 
-            const expectedRate = invariant.mul(FP_SCALING_FACTOR).div(virtualSupply);
+            const rateAssumingNoProtocolFees = invariant.mul(FP_SCALING_FACTOR).div(virtualSupply);
 
             const rate = await pool.getRate();
 
-            expect(rate).to.be.equalWithError(expectedRate, 0.0001);
+            const rateDifference = rateAssumingNoProtocolFees.sub(rate);
+            // 10000 is chosen as a non-negligible amount to show that the difference is not just from rounding errors.
+            expect(rateDifference).to.be.gt(10000);
```
nventuro, line 1896: It'd be great if this value could be made more meaningful

nventuro:
Superceded by https://github.com/balancer-labs/balancer-v2-monorepo/pull/1683

PR #1675, opened 2022-08-30
Test case showcasing a higher protocol fee collected when rate provider ticks down

danielmkm:
Added this to its own file because the setup was quite heavy. Happy to move it somewhere else if the case is even relevant.

We create two independent pool deployments with the exact same config, accruing protocol fees to both. We then down tick the rate provider on one pool before performing a trivial balanced join on both pools.

We can then see that the pool with the rater provider ticked down mints more BPT to the fee collector.


EndymionJkb:
This has also been incorporated into #1674 (separate file for integration level tests).

PR #1677, opened 2022-08-31
POC for adding tiers to the Swap and Yield fees

danielmkm:
Very rough POC of swap and yield tiers. Add a `feeTier` to `ProtocolFeeCache` that when set exposes the appropriate fee tier value for both `swapFee` and `yieldFee`. This encapsulates all changes in the `ProtocolFeeCache` so that external usage can all remain unchanged.


nventuro:
This has been superceded by https://github.com/balancer-labs/balancer-v2-monorepo/issues/1814.

PR #1678, opened 2022-08-31
Address manipulable getRate on Weighted

EndymionJkb:
First cut at a straightforward adjustment for pending protocol fees on `getRate`; could possibly be refactored to be more efficient / shared across pools. Needs tests.

### DIFF
--- a/pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
+++ b/pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
@@ -136,7 +136,7 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
     /**
      * @dev Returns the percentage of the Pool's supply which corresponds to protocol fees on yield accrued by the Pool.
      * @param normalizedWeights - The Pool's normalized token weights.
-     * @return yieldProtocolFees - The amount of BPT to be minted as protocol fees on yield.
+     * @return yieldProtocolFeesPercentage - The percentage of the Pool which corresponds to protocol fees on yield.
      * @return athRateProduct - The new all-time-high rate product if it has increased, otherwise zero.
      */
     function _getYieldProtocolFeesPoolPercentage(uint256[] memory normalizedWeights)
--- a/pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
+++ b/pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
@@ -16,8 +16,8 @@ pragma solidity ^0.7.0;
 pragma experimental ABIEncoderV2;
 
 import "@balancer-labs/v2-interfaces/contracts/pool-utils/IRateProvider.sol";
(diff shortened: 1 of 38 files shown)


pkg/pool-weighted/test/WeightedPool.test.ts
```diff
@@ -122,4 +132,122 @@ describe('WeightedPool', function () {
...
+
+          await vaultContract.connect(lp).swap(singleSwap, funds, 0, MAX_UINT256);
+        });
+
+        it("doesn't include the value of uncollected protocol fees in the rate", async () => {
+          const rate = await pool.getRate();
+
+          const rateDifference = rate.sub(originalRate);
+          // 10000 is chosen as a non-negligible amount to show that the difference is not just from rounding errors.
+          expect(rateDifference).to.be.gt(10000);
```
TomAFrench, line 225: https://github.com/balancer-labs/balancer-v2-monorepo/blob/ade39e53a74c76a8eabd0104e022b526d738882e/pkg/pool-stable/test/ComposableStablePool.test.ts#L1627-L1629

pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
```diff
@@ -136,7 +136,7 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
     /**
      * @dev Returns the percentage of the Pool's supply which corresponds to protocol fees on yield accrued by the Pool.
      * @param normalizedWeights - The Pool's normalized token weights.
-     * @return yieldProtocolFeesPercentage - The percentage of the Pool which corresponds to protocol fees on yield.
+     * @return yieldProtocolFees - The amount of BPT to be minted as protocol fees on yield.
```
TomAFrench, line 139:
```suggestion
     * @return yieldProtocolFeesPercentage - The percentage of the Pool which corresponds to protocol fees on yield.
```

pkg/pool-weighted/test/WeightedPool.test.ts
```diff
@@ -170,4 +170,108 @@ describe('WeightedPool', function () {
...
+          await vaultContract.connect(lp).swap(singleSwap, funds, 0, MAX_UINT256);
+        });
+
+        it("doesn't include the value of uncollected protocol fees in the rate", async () => {
+          const rate = await pool.getRate();
+
+          // The rate considering fees should be lower. Check that we have a difference of at least 0.01% to discard
+          // rounding error.
+          expect(originalRate).to.be.lt(rate.mul(9999).div(10000));
+        });
```
TomAFrench, line 264: It isn't clear how the description follows from the test here. We're checking that swap fees increase the rate but not that it only counts the non-protocol swap fees.

TomAFrench: I think that for this test we should be calculating the rate based on the balances and ignoring any protocol fees and show that the rate is significantly below this value.

TomAFrench (approved):

PR #1680, opened 2022-09-01
Switch yield protocol fees to use a view function

TomAFrench:
This PR aims to help #1678 by making `_getYieldProtocolFee` a view function. We can then make use of it in the `getRate` function without rewriting the same logic again.

### DIFF
--- a/pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
+++ b/pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
@@ -81,7 +81,7 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
      * @notice Returns the all time high value for the weighted product of the Pool's tokens' rates.
      * @dev Yield protocol fees are only charged when this value is exceeded.
      */
-    function getATHRateProduct() external view returns (uint256) {
+    function getATHRateProduct() public view returns (uint256) {
         return _athRateProduct;
     }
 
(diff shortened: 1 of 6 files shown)


pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
```diff
@@ -189,27 +189,23 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
      * @dev Returns the amount of BPT to be minted to pay protocol fees on yield accrued by the Pool.
      * Note that this isn't a view function. This function automatically updates `_athRateProduct`  to ensure that
      * proper accounting is performed to prevent charging duplicate protocol fees.
```
EndymionJkb, line 191:
```suggestion
```
It is now :)

pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
```diff
@@ -189,27 +189,23 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
...
         uint256 rateProduct = _getRateProduct(normalizedWeights);
 
-        // Initialise `_athRateProduct`. This will occur on the first join/exit after Pool initialisation.
-        // Not initialising this here properly will cause all joins/exits to revert.
-        if (athRateProduct == 0) {
-            _athRateProduct = rateProduct;
-            return 0;
-        }
-
         // Only charge yield fees if we've exceeded the all time high of Pool value generated through yield.
```
EndymionJkb, line 205: Should this big comment block be down in `_getRateProduct`?

EndymionJkb (commented):
Much cleaner to initialize the rate in `_onInitialize` anyway, rather than doing it lazily as before

pkg/pool-weighted/test/WeightedPoolProtocolFees.behavior.ts
```diff
@@ -98,6 +103,111 @@ export function itPaysProtocolFeesFromInvariantGrowth(): void {
...
+            tokens,
+            weights: WEIGHTS.slice(0, numTokens),
+            swapFeePercentage: POOL_SWAP_FEE_PERCENTAGE,
+          });
+        });
+
+        it('is not set on initialization', async () => {
+          await yieldFeeExemptPool.init({ initialBalances });
+
+          // expect(await yieldFeeExemptPool.instance.getATHRateProduct()).to.be.eq(0);
```
nventuro, line 122: ?

pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
TomAFrench: I don't think so. It's more of an explanation of how the yield fees mechanism works and so *why* we use those values. I think putting it in `_getRateProduct` would be hiding motivation inside implementation.

pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
```diff
@@ -229,20 +225,22 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
...
                 preJoinExitSupply,
                 getProtocolFeePercentageCache(ProtocolFeeType.YIELD)
-            );
+            ),
+            rateProduct
+        );
+    }
+
+    function _updateATHRateProduct(uint256 rateProduct) internal {
+        _athRateProduct = rateProduct;
```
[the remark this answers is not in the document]
TomAFrench: Nah, it's not really information that anyone will track so it would be unnecessary gas costs. Similar to the lastJoinExitInvariant.

pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
EndymionJkb: Makes sense - maybe add a note that the logic described is implemented in `_getRateProduct`?

EndymionJkb: That's good; for a second it looked wrong because the comment says invariant but it's about rate products. This clarifies that they're equivalent.

pkg/pool-weighted/test/WeightedPoolProtocolFees.behavior.ts
```diff
@@ -98,6 +103,111 @@ export function itPaysProtocolFeesFromInvariantGrowth(): void {
...
+                data: WeightedPoolEncoder.joinAllTokensInForExactBPTOut((await pool.totalSupply()).div(2)),
+                from: lp,
+              });
+
+              expectEqualWithError(await pool.instance.getATHRateProduct(), expectedRateProduct);
+            });
+
+            it('is updated by exits', async () => {
+              // We only test with a proportional exit, since all exits are treated equally and proportional exits remain
+              // enabled while paused
```
nventuro, line 164: Proportional exits are no longer enabled while paused - and why would that be relevant?

nventuro (approved):

PR #1682, opened 2022-09-01
Perform cleanup operations after disabling recovery mode

TomAFrench:
BasePool now has a `_onDisableRecoveryMode()` hook which allows pools to take any actions they need to do to reset their state after recovery mode is disabled. Draft while adding tests.

### DIFF
--- a/pkg/pool-utils/contracts/BasePool.sol
+++ b/pkg/pool-utils/contracts/BasePool.sol
@@ -232,7 +232,7 @@ abstract contract BasePool is
 
     /**
      * @dev Performs any necessary actions on the disabling of Recovery Mode.
-     * This is usually to reset any fee collection mechanisms to ensure that they operate correctly going forwards.
+     * This is usually to reset any fee collection mechanisms to ensure that they operate correctly going forward.
      */
     function _onDisableRecoveryMode() internal virtual {}
 
--- a/pkg/pool-utils/contracts/BasePool.sol
+++ b/pkg/pool-utils/contracts/BasePool.sol
@@ -227,6 +227,8 @@ abstract contract BasePool is
 
         emit RecoveryModeStateChanged(enabled);
 
+        // Some pools need to update their state when leaving recovery mode to ensure proper functioning of the Pool.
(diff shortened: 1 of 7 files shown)


pkg/pool-utils/contracts/BasePool.sol
```diff
@@ -226,8 +226,16 @@ abstract contract BasePool is
         _miscData = _miscData.insertBool(enabled, _RECOVERY_MODE_BIT_OFFSET);
 
         emit RecoveryModeStateChanged(enabled);
+
+        if (!enabled) _onDisableRecoveryMode();
     }
 
+    /**
+     * @dev Performs any necessary actions on the disabling of Recovery Mode.
+     * This is usually to reset any fee collection mechanisms to ensure that they operate correctly going forwards.
```
EndymionJkb, line 235:
```suggestion
     * This is usually to reset any fee collection mechanisms to ensure that they operate correctly going forward.
```
Unless this is a British thing?

pkg/pool-utils/contracts/BasePool.sol
```diff
@@ -226,8 +226,16 @@ abstract contract BasePool is
         _miscData = _miscData.insertBool(enabled, _RECOVERY_MODE_BIT_OFFSET);
 
         emit RecoveryModeStateChanged(enabled);
+
+        if (!enabled) _onDisableRecoveryMode();
```
EndymionJkb, line 230: Add a comment that we're calling a hook? We already have a warning up in RecoveryMode about not putting any "dangerous" code that might fail here, especially on the enable path. Maybe worth repeating, since we are actually doing something at this level, and more derived contracts might start here?

EndymionJkb (commented):
I really like that we no longer need to override doRecoveryModeExit in Managed now.

nventuro:
Looking good!

Longer term we'll likely want to merge recovery mode, the protocol fee cache and protocol fee computation into a single thing, as they have a number of interactions:

 - recovery mode overrides protocol fees
 - recovery mode deactivation resets due fees
 - protocol fee percentage cache update triggers protocol fee payout

PR #1683, opened 2022-09-01
Make Stable `getRate` consider unpaid protocol fees

nventuro:
This also adds some (but not all, since I didn't quite understand them) tests from #1671

Supercedes https://github.com/balancer-labs/balancer-v2-monorepo/pull/1674

### DIFF
--- a/pkg/pool-stable/contracts/ComposableStablePool.sol
+++ b/pkg/pool-stable/contracts/ComposableStablePool.sol
@@ -926,7 +926,7 @@ contract ComposableStablePool is
 
     /**
      * @dev This function returns the appreciation of BPT relative to the underlying tokens, as an 18 decimal fixed
-     * point number. It is simply the ratio of the invariant over the BPT supply.
+     * point number. It is simply the ratio of the invariant to the BPT supply.
      *
      * The total supply is initialized to equal the invariant, so this value starts at one. During Pool operation the
      * invariant always grows and shrinks either proportionally to the total supply (in scenarios with no price impact,
--- a/pkg/pool-stable/contracts/ComposableStablePool.sol
+++ b/pkg/pool-stable/contracts/ComposableStablePool.sol
@@ -925,20 +925,58 @@ contract ComposableStablePool is
     // BPT rate
 
     /**
-     * @dev This function returns the appreciation of one BPT relative to the
-     * underlying tokens. This starts at 1 when the pool is created and grows over time.
-     * Because of preminted BPT, it uses `getVirtualSupply` instead of `totalSupply`.
+     * @dev This function returns the appreciation of BPT relative to the underlying tokens, as an 18 decimal fixed
+     * point number. It is simply the ratio of the invariant over the BPT supply.
+     *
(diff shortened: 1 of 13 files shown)


nventuro:
There's something off in some test regarding the usage of `getScaledBalances`. I'm looking into that.

pkg/pool-stable/contracts/ComposableStablePool.sol
```diff
@@ -925,19 +925,58 @@ contract ComposableStablePool is
     // BPT rate
 
     /**
-     * @dev This function returns the appreciation of one BPT relative to the
-     * underlying tokens. This starts at 1 when the pool is created and grows over time.
-     * Because of preminted BPT, it uses `getVirtualSupply` instead of `totalSupply`.
+     * @dev This function returns the appreciation of BPT relative to the underlying tokens, as an 18 decimal fixed
+     * point number. It is simply the ratio of the invariant over the BPT supply.
```
EndymionJkb, line 929:
```suggestion
     * point number. It is simply the ratio of the invariant to the BPT supply.
```

pkg/pool-stable/test/ComposableStablePool.test.ts
```diff
@@ -1575,31 +1597,107 @@ describe('ComposableStablePool', () => {
...
 
-            await tokens.mint({ to: lp, amount });
-            await tokens.approve({ from: lp, to: pool.vault });
+          function itReportsRateCorrectly() {
+            it('takes into account unminted protocol fees', async () => {
+              const invariant = await pool.estimateInvariant();
+
+              // The virtual supply does not include the unminted protocol fees. We need to adjust it by computing those.
+              // Since all balances are relatively close and the pool is balanced, we can estimate their prices as being
+              // equal to then price the due protocool fees.
```
EndymionJkb, line 1613:
```suggestion
              // Since all balances are relatively close and the pool is balanced, we can estimate their prices as being
              // equal to then price the due protocool fees.
```
              // Since all balances are relatively close and the pool is balanced, we can simply add the fee amount to
              // the current balances to obtain the final sum.

pkg/pool-stable/test/ComposableStablePool.test.ts
```diff
@@ -1575,31 +1597,107 @@ describe('ComposableStablePool', () => {
...
+
+              // Note that we exit with proportional *unscaled* balances - otherwise we'd need to take their different
+              // scaling factors into account.
+              const { balances: unscaledBalances } = await pool.getTokens();
+              const amountsOut = unscaledBalances.map((balance, i) => (i == bptIndex ? bn(0) : balance.div(100)));
+              await pool.exitGivenOut({ from: lp, amountsOut });
+
+              const rateAfterExit = await pool.getRate();
+
+              // There's some minute diference due to rounding error
```
TomAFrench, line 1666:
```suggestion
              // There's some minute difference due to rounding error
```

nventuro:
I ended up undoing the `getScaledBalances` changes since those cause issues in other places. We need a more thorough refactor of the tests for that.

PR #1684, opened 2022-09-02
Pay protocol fees on cache update

nventuro:
I believe this is what needs to be done to safely update the protocol fee cahce. In its current form the code goes over the bytecode size, but there's some simple refactors that can be introduced (`getRate` and this new hook share a lot of their code). I intend to do that separetly, once all new changes have been introduced.

### DIFF
--- a/pkg/pool-stable/contracts/ComposableStablePool.sol
+++ b/pkg/pool-stable/contracts/ComposableStablePool.sol
@@ -962,7 +962,7 @@ contract ComposableStablePool is
         // supply to make the calculation with the correct amount.
         uint256 actualTotalSupply = virtualSupply.add(protocolFeeAmount);
 
-        // All that's missing now is the invariant. We have the balances required to calculate it already, but are still
+        // All that's missing now is the invariant. We have the balances required to calculate it already, but still
         // need the current amplification factor.
         (uint256 currentAmp, ) = _getAmplificationParameter();
 
--- a/pkg/pool-stable/contracts/ComposableStablePool.sol
+++ b/pkg/pool-stable/contracts/ComposableStablePool.sol
@@ -962,8 +962,8 @@ contract ComposableStablePool is
         // supply to make the calculation with the correct amount.
         uint256 actualTotalSupply = virtualSupply.add(protocolFeeAmount);
 
-        // All that's missing is now the invariant. We have the balances already, but are missing the current
-        // amplification factor.
+        // All that's missing now is the invariant. We have the balances required to calculate it already, but are still
+        // need the current amplification factor.
         (uint256 currentAmp, ) = _getAmplificationParameter();
 
         // It turns out that the process for due protocol fee calculation involves computing the current invariant,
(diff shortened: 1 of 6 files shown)


pkg/pool-stable/contracts/ComposableStablePool.sol
```diff
@@ -962,8 +962,8 @@ contract ComposableStablePool is
         // supply to make the calculation with the correct amount.
         uint256 actualTotalSupply = virtualSupply.add(protocolFeeAmount);
 
-        // All that's missing is now the invariant. We have the balances already, but are missing the current
-        // amplification factor.
+        // All that's missing now is the invariant. We have the balances required to calculate it already, but are still
```
TomAFrench, line 965:
```suggestion
        // All that's missing now is the invariant. We have the balances required to calculate it already, but still
```

pkg/pool-utils/contracts/ProtocolFeeCache.sol
```diff
@@ -126,6 +128,8 @@ abstract contract ProtocolFeeCache is RecoveryMode {
         _updateProtocolFeeCache(_protocolFeeProvider, ProtocolFeeType.AUM);
     }
 
+    function _beforeProtocolFeeCacheUpdate() internal virtual {}
```
TomAFrench, line 131: Lets add some natspec here explaining that we need to pay any protocol fees before updating the cache. This should be pretty similar to some of the comments you have elsewhere in the PR.

TomAFrench (dismissed):

nventuro:
I added some more tests, and added a check for pausability. Some of the tests really don't belong there (and arguably the code doesn't either), but this is not the time to refactor the entire pool or test suite.

pkg/pool-utils/contracts/ProtocolFeeCache.sol
```diff
@@ -126,6 +128,13 @@ abstract contract ProtocolFeeCache is RecoveryMode {
         _updateProtocolFeeCache(_protocolFeeProvider, ProtocolFeeType.AUM);
     }
 
+    /**
+     * @dev Override in derived contracts to perform some action before the cache is updated. This is typically relevant
+     * in Pools that have protocol debt that is dependent in the protocol fee percentages, as it needs to be paid out
+     * before said percentages change.
```
EndymionJkb, line 134:
```suggestion
     * @dev Override in derived contracts to perform some action before the cache is updated. This is typically relevant
     * to Pools that incur protocol debt between operations. To avoid altering the amount due retroactively,
     * this debt needs to be paid before the fee percentages change.
```
Protocol debt would always be dependent on the fee percentages.

pkg/pool-stable/contracts/ComposableStablePool.sol
```diff
@@ -979,6 +979,40 @@ contract ComposableStablePool is
         return currentInvariant.divDown(actualTotalSupply);
     }
 
+    function _beforeProtocolFeeCacheUpdate() internal override {
+        // The `getRate()` function depends on the actual supply, which in turn depends on the cached protocol fee
+        // percentages. Changing these would therefore result in the rate changing, which is not acceptable as this is a
+        // sensitive value.
+        // Because of this, we pay any due protocol fees *before* updating the cached rates, making it so that the new
+        // percentages only affect future operation of the Pool, and not past fees. As a result, `getRate()` is
+        // unaffected by the cached protocol fee percentages changing.
```
TomAFrench, line 988:
```suggestion
        // Because of this, we pay any due protocol fees *before* updating the cache, making it so that the new
        // percentages only affect future operation of the Pool, and not past fees. As a result, `getRate()` is
        // unaffected by the cached protocol fee percentages changing.
```

"Cached rates" reads as the token rates for yield fees to me. Let's just say "cache" as it's clear which cache we're talking about from the function name.

TomAFrench (approved):

PR #1685, opened 2022-09-02
Rework WeightedPoolProtocolFees to return percentages rather than amounts

TomAFrench:
Pulls out all the non rate related code from #1678


EndymionJkb (approved):
Looks familiar :) And cleaner now.

PR #1686, opened 2022-09-02
Gauge Actions test

### LINKED ISSUE #1148: Add GaugeAction tests
This module was added in https://github.com/balancer-labs/balancer-v2-monorepo/pull/1146 with no tests, we should write them before deployment.

jubeira:
This PR adds unit tests for `GaugeActions`; closes #1148 .

Some notes:

###  `gaugeWithdraw` and `gaugeDeposit`
We will only care about token movements between sender, recipient and a staking liquidity gauge; `veBAL` calculations are out of scope. Therefore, we can use actual liquidity gauges and mock the `VeDelegation` when setting up the test. 

The tests call the methods under several conditions, which increases the test complexity a bit:
- Multiple amounts withdraw / deposit (within and outside allowed range).
- Immediate values and chained references.
- Using sender as recipient. and using other as recipient.

Each of those conditions open a new context and go down one level in the call stack, but hopefully it's easy enough to follow.

### `gaugeMint` and `gaugeClaimRewards`

These are more straightforward, since `GaugeActions` mostly forwards the calls. Again, rewards' calculations are out of scope; we just verify transfers between contracts.

### DIFF
--- a/pkg/standalone-utils/test/GaugeActions.test.ts
+++ b/pkg/standalone-utils/test/GaugeActions.test.ts
@@ -29,7 +29,7 @@ describe('GaugeActions', function () {
   let adaptor: Contract, gaugeController: Contract, balMinter: Contract;
   let BAL: Contract, veBAL: Contract, rewardToken: Contract, lpToken: Contract;
 
-  let gaugeFactory: Contract;
+  let liquidityGaugeFactory: Contract, childChainGaugeFactory: Contract;
   let gauge: Contract;
 
   const totalLpTokens = fp(100);
@@ -66,6 +66,8 @@ describe('GaugeActions', function () {
     gaugeController = await deploy('v2-liquidity-mining/MockGaugeController', {
       args: [veBAL.address, adaptor.address],
     });
+    // Type weight is ignored in the mock controller.
+    await gaugeController.add_type('Ethereum', 0);
 
     const balTokenAdmin = await deploy('v2-liquidity-mining/MockBalancerTokenAdmin', {
       args: [vault.address, BAL.address],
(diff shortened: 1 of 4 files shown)


pkg/standalone-utils/test/GaugeActions.test.ts
```diff
@@ -0,0 +1,643 @@
...
+            amount: depositAmountOrRef,
+          }),
+        ]);
+        receipt = await tx.wait();
+      });
+
+      // Short-circuit when no tokens are deposited.
+      expectedAmount.gt(0) &&
+        it('pulls BPT tokens from sender if necessary', async () => {
+          // This transfer is skipped if token sender is relayer, but addresses are undefined outside this scope.
```
jubeira, line 246: Note for the reviewer: I think using the same function that tests everything that happens in each `deposit` or `withdraw` makes sense, even if some steps don't always happen.

IMO short circuiting the test is a good idea and is less confusing than skipping it (e.g. when the amount to deposit is 0).

I would have preferred to do the same when the transfer doesn't happen because `tokenSender == relayer`, but as the comment says it's not possible. I'm open to other approaches as well here.

pkg/standalone-utils/test/GaugeActions.test.ts
```diff
@@ -0,0 +1,659 @@
...
+    const relayerActionIds = await Promise.all(
+      ['setRelayerApproval', 'manageUserBalance'].map((action) => actionId(vault.instance, action))
+    );
+    await vault.grantPermissionsGlobally(relayerActionIds, relayer);
+
+    // Approve relayer by BPT holder
+    await vault.setRelayerApproval(userSender, relayer, true);
+  });
+
+  // We won't be needing the adder; we just need a staking liquidity gauge where tokens can be deposited / withdrawn.
```
nventuro, line 94:
```suggestion
```

nventuro: Took me a while to understand what 'adder' meant, and by the time I realized it was the GaugeAdder I was like 'uh, yea, sure'. I don't think we need to mention this at all :shrug:

pkg/standalone-utils/test/GaugeActions.test.ts
```diff
@@ -0,0 +1,659 @@
...
+    });
+
+    const gaugeImplementation = await deploy('v2-liquidity-mining/LiquidityGaugeV5', {
+      args: [balMinter.address, veBalDelegationProxy.address, adaptor.address],
+    });
+    gaugeFactory = await deploy('v2-liquidity-mining/LiquidityGaugeFactory', { args: [gaugeImplementation.address] });
+
+    gauge = await deployedAt(
+      'v2-liquidity-mining/LiquidityGaugeV5',
+      await deployGauge(gaugeFactory, lpToken.address) // No weight cap.
```
nventuro, line 109: You don't need to go through all this - you can just deploy the gauge directly, skipping the factory entirely.

pkg/standalone-utils/test/GaugeActions.test.ts
```diff
@@ -0,0 +1,659 @@
...
+      args: [vault.address, veBAL.address, veBalDelegation.address],
+    });
+
+    const gaugeImplementation = await deploy('v2-liquidity-mining/LiquidityGaugeV5', {
+      args: [balMinter.address, veBalDelegationProxy.address, adaptor.address],
+    });
+    gaugeFactory = await deploy('v2-liquidity-mining/LiquidityGaugeFactory', { args: [gaugeImplementation.address] });
+
+    gauge = await deployedAt(
+      'v2-liquidity-mining/LiquidityGaugeV5',
```
nventuro, line 108: This only tests the V5 gauge, but we should also test against the `RewardsOnlyGauge` (the ones we deploy on L2s).

pkg/standalone-utils/test/GaugeActions.test.ts
```diff
@@ -0,0 +1,659 @@
...
+        expect(
+          relayerLibrary.connect(userSender).gaugeDeposit(gauge.address, userSender.address, gauge.address, fp(1))
+        ).to.be.revertedWith('Incorrect sender');
+      });
+    });
+
+    context('when caller != sender and sender != relayer', () => {
+      it('reverts', async () => {
+        expect(
+          relayerLibrary.connect(userSender).gaugeDeposit(gauge.address, other.address, gauge.address, fp(1))
```
nventuro, line 131: This call should be made via the relayer, not the library

nventuro, line 131:
```suggestion
          relayerLibrary.connect(other).gaugeDeposit(gauge.address, userSender.address, gauge.address, fp(1))
```

This is more natural - it's the call you'd make, but made by an unauthorized account

pkg/standalone-utils/test/GaugeActions.test.ts
```diff
@@ -0,0 +1,659 @@
...
+            .multicall([
+              encodeDeposit({ gauge: gauge, sender: userSender, recipient: userSender, amount: totalLpTokens.add(1) }),
+            ])
+        ).to.be.reverted;
+      });
+    });
+
+    context('when sender has enough BPT', () => {
+      context('sender = senderUser, recipient = senderUser', () => {
+        sharedBeforeEach('', async () => {
```
nventuro, line 150:
```suggestion
        sharedBeforeEach(async () => {
```

This helper is smart enough to deal with no description being provided :)

pkg/standalone-utils/test/GaugeActions.test.ts
```diff
@@ -0,0 +1,659 @@
...
+            .connect(userSender)
+            .multicall([
+              encodeDeposit({ gauge: gauge, sender: userSender, recipient: userSender, amount: totalLpTokens.add(1) }),
+            ])
+        ).to.be.reverted;
+      });
+    });
+
+    context('when sender has enough BPT', () => {
+      context('sender = senderUser, recipient = senderUser', () => {
```
nventuro, line 149: This says 'senderUser', but the account is called 'userSender' :zany_face: 

tbh I'd rename the description to 'sender = user, recipient = user'. 'sender = senderUser' is a bit hard to parse

pkg/standalone-utils/test/GaugeActions.test.ts
```diff
@@ -0,0 +1,659 @@
...
+        // Start: sender has all the BPT, no gauge tokens minted, recipient is clean unless it is sender.
+        expect(await lpToken.balanceOf(gauge.address)).to.be.eq(0);
+        expect(await lpToken.balanceOf(tokenSenderAddress)).to.be.eq(totalLpTokens);
+        if (tokenSenderAddress != tokenRecipientAddress) {
+          expect(await lpToken.balanceOf(tokenRecipientAddress)).to.be.eq(0);
+        }
+
+        expect(await gauge.balanceOf(gauge.address)).to.be.eq(0);
+        expect(await gauge.balanceOf(tokenSenderAddress)).to.be.eq(0);
+        expect(await gauge.balanceOf(tokenRecipientAddress)).to.be.eq(0);
```
nventuro, line 238: This is fine, but it makes the other tests a bit fragile. I think it'd be better to measure and store the balances _before_ the deposit, and to have the tests work with deltas (i.e. measure balance, assert on `curr - old`)

pkg/standalone-utils/test/GaugeActions.test.ts
```diff
@@ -0,0 +1,659 @@
...
+            recipient: tokenRecipientAddress,
+            amount: depositAmountOrRef,
+          }),
+        ]);
+        receipt = await tx.wait();
+      });
+
+      // Short-circuit when no tokens are deposited.
+      expectedAmount.gt(0) &&
+        it('pulls BPT tokens from sender if necessary', async () => {
```
nventuro, line 253: Do a standard `if` :joy:

pkg/standalone-utils/test/GaugeActions.test.ts
```diff
@@ -0,0 +1,659 @@
...
+        });
+
+        itDepositsWithEnoughTokens();
+      });
+    });
+
+    function itDepositsWithEnoughTokens() {
+      context('when depositing some of the tokens', () => {
+        itDepositsWithRefsAndAmounts(totalLpTokens.div(3));
+      });
```
nventuro, line 191: If you're doing this, might as all also add a test for `total.add(1)` :)

pkg/standalone-utils/test/GaugeActions.test.ts
nventuro: Agreed - this is small enough that it doesn't really matter that we have an empty `it`. In other circumstances we could refactor the tests to address this, but we don't need to go crazy for this.

pkg/standalone-utils/test/GaugeActions.test.ts
```diff
@@ -0,0 +1,659 @@
...
+          }
+        });
+
+      it("approves gauge to use relayer's BPT funds", async () => {
+        expectEvent.inIndirectReceipt(receipt, lpToken.interface, 'Approval', {
+          owner: relayer.address,
+          spender: gauge.address,
+          value: expectedAmount,
+        });
+      });
```
nventuro, line 270: This test is not _wrong_, but the fact that this happens is a bit of an implementation detail on the relayer. This approval doesn't _need_ to happen (maybe the relayer sets e.g. infite approval at some other point or whatever). We can keep this now, but note that what is important is that the deposit succeeds - how we get there doesn't matter as much

pkg/standalone-utils/test/GaugeActions.test.ts
```diff
@@ -0,0 +1,659 @@
...
+        expectTransferEvent(
+          receipt,
+          { from: ZERO_ADDRESS, to: tokenRecipientAddress, value: expectedAmount },
+          gauge.address
+        );
+      });
+    }
+  });
+
+  describe('gaugeWithdraw', () => {
```
nventuro, line 308: Most (all?) of the comments on deposit also apply here

pkg/standalone-utils/test/GaugeActions.test.ts
```diff
@@ -0,0 +1,659 @@
...
+  });
+
+  describe('gaugeMint', () => {
+    sharedBeforeEach('grant mint approval to sender via relayer', async () => {
+      const { v, r, s, deadline } = await BalancerMinterAuthorization.signSetMinterApproval(
+        balMinter,
+        relayer.address,
+        true,
+        userSender
+      );
```
nventuro, line 519: It'd be even better if you had this in the same multicall as the call to mint :)

nventuro: (which is the expected usage)

nventuro (commented):
Looks great! Fantastic first go at testing this complex component :)

pkg/standalone-utils/test/GaugeActions.test.ts
jubeira: There is a small difference: gauges use proxies when created via factories, and the `create` method calls initialize with the right arguments. Otherwise we'd need to call it manually.

Even if it adds a few more lines, I'd vouch for leaving it like this for completion. WDYT?

pkg/standalone-utils/test/GaugeActions.test.ts
jubeira: The difference with the `if` is that the test doesn't appear at all when running it, and I thought that was more accurate than passing an empty test :grimacing: 

I'll do the change anyways :)

pkg/standalone-utils/test/GaugeActions.test.ts
jubeira: Could be, but that's already covered in a context above (`when sender does not have enough BPT`)

pkg/standalone-utils/test/GaugeActions.test.ts
jubeira: Added in https://github.com/balancer-labs/balancer-v2-monorepo/pull/1686/commits/05dfc6f02363311f7f3ab7fe260cd46f6c223f1b and https://github.com/balancer-labs/balancer-v2-monorepo/pull/1686/commits/86eb6da1c8ec14bc379041d221889b1bc3c6b96a.

tl;dr; both gauges now share `deposit` and `withdraw`. Rewards only has a slightly different `claim` test, and no `mint` of course.

pkg/standalone-utils/test/GaugeActions.test.ts
jubeira: Done; also added a mini fix for `almostEqual` in big number comparisons to support negative numbers with some error.

pkg/standalone-utils/test/GaugeActions.test.ts
jubeira: comments applied to deposit as well.

pkg/standalone-utils/test/GaugeActions.test.ts
jubeira: done :)

pkg/standalone-utils/test/GaugeActions.test.ts
nventuro: Ah no, I didn't mean skip the test body, but rather skip the `it` call:
```
if (expectedAmount.gt(0)) {
 it('pulls tokens....', ....);
}
```

pkg/standalone-utils/test/GaugeActions.test.ts
```diff
@@ -0,0 +1,769 @@
...
+              recipient: tokenRecipientAddress,
+              amount: depositAmountOrRef,
+            }),
+          ]);
+          receipt = await tx.wait();
+        });
+
+        it('pulls BPT tokens from sender if necessary', async () => {
+          // This transfer is skipped if token sender is relayer or if 0 tokens are expected to be transferred.
+          if (tokenSenderAddress != relayer.address && expectedAmount.gt(0)) {
```
nventuro, line 424: This is fine though - no need to refactor any further

nventuro (approved):

PR #1687, opened 2022-09-02
Forfeit protocol fees on recovery mode disable

nventuro:


pkg/pool-stable/contracts/ComposableStablePool.sol
```diff
@@ -1013,6 +1013,24 @@ contract ComposableStablePool is
         _updatePostJoinExit(currentAmp, currentInvariant);
     }
 
+    function _onDisableRecoveryMode() internal override {
+        // Enabling recovery mode forcefully sets all protocol fee percentages to zero, increasing the return value of
+        // `getRate()` and effectively forfeiting due protocol fees.
```
EndymionJkb, line 1018:
```suggestion
        // Enabling recovery mode short-circuits protocol fee computations so that all percentages are
        // effectively zero, increasing the return value of `getRate()` and effectively forfeiting due protocol fees.
```
The original wording sounds like there is code running that overwrites rates.

pkg/pool-stable/test/ComposableStablePool.test.ts
```diff
@@ -1683,6 +1683,43 @@ describe('ComposableStablePool', () => {
...
+              expect(newRate).to.be.gt(initialRate);
+
+              // We can compute the new rate by computing the ratio of invariant and total supply, not considering any
+              // due protocol fees (because there should be none).
+              const scaledBalances = arrayFpMul(await pool.getBalances(), await pool.getScalingFactors()).filter(
+                (_, i) => i != bptIndex
+              );
+              const invariant = calculateInvariant(
+                scaledBalances,
+                (await pool.getAmplificationParameter()).value.div(1000)
```
EndymionJkb, line 1703:
```suggestion
                (await pool.getAmplificationParameter()).value.div(AMP_PRECISION)
```

pkg/pool-utils/contracts/RecoveryMode.sol
```diff
@@ -60,7 +60,7 @@ abstract contract RecoveryMode is IRecoveryMode, BasePoolAuthorization {
      * @dev Does not otherwise affect pool operations (beyond deferring payment of protocol fees), though some pools may
      * perform certain operations in a "safer" manner that is less likely to fail, in an attempt to keep the pool
      * running, even in a pathological state. Unlike the Pause operation, which is only available during a short window
-     * after factory deployment, Recovery Mode can always be enableed.
+     * after factory deployment, Recovery Mode can always be enabled.
      */
     function enableRecoveryMode() external override authenticate {
         _setRecoveryMode(true);
```
danielmkm, line 66: Neither `enableRecoveryMode` or `disableRecoveryMode` check the current recovery mode state prior to calling `_setRecoveryMode`.  It doesn't impact the `BasePool` implementation, but logically I'd expect there to be some sort of guard to prevent setting recovery mode to a pool already in recovery mode, and vice-a-versa.

pkg/pool-stable/contracts/ComposableStablePool.sol
danielmkm: The protocol fee computations still run, they just run with 0 values, "short-circuits protocol fee computations" sounds more like it prevents the computations altogether.

What about something along the lines of below, we also try to clarify why getRate increases:

// Enabling recovery mode short-circuits protocol fee percentage retrieval, returning a zero percentage for all fee types. This effectively forfeits any due protocol fees (unminted BPTs), reducing the actualTotalSupply and increasing the value returned by `getRate()`.

pkg/pool-utils/contracts/RecoveryMode.sol
TomAFrench: I don't think there's a good way that we can fully enforce this without considering the Authorizer and permissions as well.

If we had this guard then you could still cause problems by "flicking" the pool in and out of recovery mode whenever you want to trigger the hook (assuming you had permissions for both functions). We'd then need to bake in cooldowns (or similar mechanism) which would be impossible to perfectly set from the outset. This sort of thing probably makes more sense to push out to the periphery and rely on a helper contract or delays on disabling recovery mode.

TomAFrench (approved):
Looks good!

PR #1688, opened 2022-09-02
Add comment on decreasing StablePool token rates

nventuro:


pkg/pool-stable/contracts/ComposableStablePoolProtocolFees.sol
```diff
@@ -155,6 +155,14 @@ abstract contract ComposableStablePoolProtocolFees is
         // Each invariant should be larger than its precedessor. In case any rounding error results in them being
         // smaller, we adjust the subtraction to equal 0.
 
+        // Note: in the unexpected scenario where the rates of the tokens shrink over time instead of grow (i.e. if the
```
EndymionJkb, line 158:
```suggestion
        // Note: in the unexpected scenario where the rates of the tokens shrink over time instead of growing (i.e. if the
```

pkg/pool-stable/contracts/ComposableStablePoolProtocolFees.sol
```diff
@@ -155,6 +155,14 @@ abstract contract ComposableStablePoolProtocolFees is
...
+        // yield is negative), the non-exempt growth invariant might actually be *smaller* than the swap fee growth
+        // invariant. Depending on the order in which swaps, joins/exits and rate changes happen, as well as their
+        // magnitude, it is possible for the Pool to either pay more or less protocol fees than it should.
+        // Due to the complexity that handling all of these cases would introduce, this behavior is considered out of
+        // scope, and is expected to be handled in a case-by-case basis if the token rates were to ever decrease (which
```
EndymionJkb, line 163:
```suggestion
        // scope, and is expected to be handled on a case-by-case basis if the token rates were to ever decrease (which
```

pkg/pool-stable/contracts/ComposableStablePoolProtocolFees.sol
```diff
@@ -155,6 +155,14 @@ abstract contract ComposableStablePoolProtocolFees is
         // Each invariant should be larger than its precedessor. In case any rounding error results in them being
         // smaller, we adjust the subtraction to equal 0.
 
+        // Note: in the unexpected scenario where the rates of the tokens shrink over time instead of grow (i.e. if the
+        // yield is negative), the non-exempt growth invariant might actually be *smaller* than the swap fee growth
+        // invariant. Depending on the order in which swaps, joins/exits and rate changes happen, as well as their
+        // magnitude, it is possible for the Pool to either pay more or less protocol fees than it should.
```
EndymionJkb, line 161:
```suggestion
        // relative magnitudes, it is possible for the Pool to either pay more or less protocol fees than it should.
```

EndymionJkb (approved):
Would surely need linting if you take these changes

pkg/pool-stable/contracts/ComposableStablePoolProtocolFees.sol
```diff
@@ -155,6 +155,14 @@ abstract contract ComposableStablePoolProtocolFees is
         // Each invariant should be larger than its precedessor. In case any rounding error results in them being
         // smaller, we adjust the subtraction to equal 0.
 
+        // Note: in the unexpected scenario where the rates of the tokens shrink over time instead of grow (i.e. if the
+        // yield is negative), the non-exempt growth invariant might actually be *smaller* than the swap fee growth
```
danielmkm, line 159: You only address non-exempt growth invariants here. For decreases of exempt rate providers, it's possible for the totalGrowthInvariant to be smaller than the totalNonExemptGrowthInvariant.

pkg/pool-stable/contracts/ComposableStablePoolProtocolFees.sol
danielmkm: I think grow here sounds better than growing, but just an opinion.

pkg/pool-stable/contracts/ComposableStablePoolProtocolFees.sol
danielmkm: ...,the non-exempt growth invariant might actually be *smaller* than the swap fee growth invariant and the total growth invariant might actually be *smaller* than the non-exempt growth invariant.

danielmkm:
One thing not addressed in the comment is what happens when yield goes negative, but is then later recovered. It would essentially be possible for the protocol to collect a fee on the same yield growth twice in this scenario.

PR #1691, opened 2022-09-05
Add FP helpers

EndymionJkb:
Add fpMul and fpDiv helpers to simplify test code. Abstract away the FP_SCALING_FACTOR in many calculations.


TomAFrench (approved):
Yeah, this is something that we definitely need to avoid tests being extremely verbose.

Could we rename them to be camelcase?

TomAFrench:
:camel:

PR #1697, opened 2022-09-05
Test: peekChainedReferenceValue multicall

### LINKED ISSUE #1693: Add test for peekReferenceValue using multicall
#1672 added `peekChainedReferenceValue` method and tests, but the tests use `relayerLibrary` directly.

We should add tests that use `peek` in a `multicall` static call, to double check that it will work properly for the purpose it was designed for.

jubeira:
Closes #1693.

`peek` already has tests under several conditions (e.g. after a `get`, peeking multiple times, etc). A simple smoke test within a `multicall` should be enough.

### DIFF
--- a/pkg/standalone-utils/test/BaseRelayerLibrary.test.ts
+++ b/pkg/standalone-utils/test/BaseRelayerLibrary.test.ts
@@ -13,8 +13,6 @@ import { ANY_ADDRESS, MAX_UINT256 } from '@balancer-labs/v2-helpers/src/constant
 import Vault from '@balancer-labs/v2-helpers/src/models/vault/Vault';
 import { BigNumberish, bn, fp } from '@balancer-labs/v2-helpers/src/numbers';
 import { toChainedReference } from './helpers/chainedReferences';
-import { random, range } from 'lodash';
-import Decimal from 'decimal.js';
 
 describe('BaseRelayerLibrary', function () {
   let vault: Contract;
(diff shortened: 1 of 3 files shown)


pkg/standalone-utils/test/BaseRelayerLibrary.test.ts
```diff
@@ -294,6 +294,19 @@ describe('BaseRelayerLibrary', function () {
...
+    describe('peekChainedReferenceValue', () => {
+      it('peeks chained reference', async () => {
+        const reference = toChainedReference(174);
+        const value = fp(340);
+
+        const result = await relayer.callStatic.multicall([
+          relayerLibrary.interface.encodeFunctionData('setChainedReferenceValue', [reference, value]),
+          relayerLibrary.interface.encodeFunctionData('peekChainedReferenceValue', [reference]),
+        ]);
+        expect(result).to.be.deep.eq([bn(0), value]);
```
nventuro, line 307: Heh, this made me realize we have a small bug in `setChainedRef`: that function should not return anything.

nventuro (approved):

pkg/standalone-utils/test/BaseRelayerLibrary.test.ts
jubeira: Changed in both mock relayer library implementations, and adjusted test.

PR #1698, opened 2022-09-05
Adding unauthorized call tests to VaultActions

### LINKED ISSUE #992: Add relayer VaultAction tests for relayer sender
As mentioned [in this review comment](https://github.com/balancer-labs/balancer-v2-monorepo/pull/982#discussion_r742150652), now that we allow for the relayer to be the sender of swaps, joins, etc., we should have simple tests that check that this works as expected (i.e. relayer sender is accepted and works, invalid senders are rejetected).

jubeira:
Closes #992.

Diff will be quite large, but this PR only adds a test for each of the methods  in `VaultActions` using an unauthorized caller.
Existing tests were not modified; they are just nested under a new `context`.


nventuro (commented):
Looks good! You're only addressing half of #992 though - we should also have tests where the sender is the relayer (which is allowed).

Also, protip: you can hide whitespace changes in the review, which in scenarios like this one results in a much cleaner diff (since most of the changes are due to indentation)

image

jubeira:
New context added for all tests in https://github.com/balancer-labs/balancer-v2-monorepo/pull/1698/commits/153a033ec25b9288ce7df40e20a2d69ba9d48c59, using relayer as sender.

Quick comment to read the changes:
- There's now a sender account and a recipient account, which allows changing the endpoints of the transactions before executing the tests.
- All the `encode` methods can now set sender / recipient (or use the default set by the context where they are used with a closure).
- Recipient is (almost) always the user, since the methods under test don't follow different codepaths when changing the recipient. The exception is chained operations; we need the recipient to be the sender so as to chain the first operation with the second one.
- When sender is the relayer, it has to be funded with tokens before running the tests.

That being said, tests are almost the same as before. There are slight changes when verifying balance changes when sender != recipient as you'd expect, but other than that it's _almost_ the same.

nventuro (approved):

nventuro:
Looks great! I shared already my only feedback of renaming `senderUser` to `sender`, as it can be hard to figure out otherwise what the user is actually doing (and if they're the sender).

PR #1704, opened 2022-09-06
Add getActualSupply

nventuro:
Not having this would make the SDK very complicated. I created a helper function to reduce bytecode size.

### DIFF
--- a/pkg/pool-stable/contracts/ComposableStablePool.sol
+++ b/pkg/pool-stable/contracts/ComposableStablePool.sol
@@ -925,6 +925,54 @@ contract ComposableStablePool is
     // BPT rate
 
     /**
+     * Many functions require accessing multiple internal values that might at first seem unrelated, but are actually
+     * quite intertwined, and computed at the same time for optimal performance (since calculating some of them also
+     * yields intermediate results useful for other queries). This helper function returns many of these values,
+     * greatly reducing bytecode size.
+     *
+     * The return values are:
+     *  - the current upscaled token balances (not including BPT)
+     *  - the virtual supply
+     *  - the BPT amount of unpaid protocol fees
+     *  - the amplification factor at the last join or exit operation
+     *  - the invariant of the current balances, calculated using the amplification factor at the last join or exit
+     *    operation.
+     */
+    function _getSupplyAndFeesData()
+        private
+        view
+        returns (
+            uint256[] memory balances,
+            uint256 virtualSupply,
+            uint256 protocolFeeAmount,
+            uint256 lastJoinExitAmp,
+            uint256 currentInvariantWithLastJoinExitAmp
+        )
+    {
+        // First we query the Vault for current registered balances (which includes preminted BPT), to then calculate
+        // the current scaled balances and virtual supply.
+        (, uint256[] memory registeredBalances, ) = getVault().getPoolTokens(getPoolId());
+        _upscaleArray(registeredBalances, _scalingFactors());
+        (virtualSupply, balances) = _dropBptItemFromBalances(registeredBalances);
+
+        // Now we need to calculate any BPT due in the form of protocol fees. This requires data from the last join or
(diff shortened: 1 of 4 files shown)


pkg/pool-stable/test/ComposableStablePool.test.ts
```diff
@@ -1551,7 +1551,7 @@ describe('ComposableStablePool', () => {
       });
     });
 
-    describe('getRate and protocol fees', () => {
+    describe.only('getRate and protocol fees', () => {
```
TomAFrench, line 1554:
```suggestion
    describe('getRate and protocol fees', () => {
```

pkg/pool-stable/contracts/ComposableStablePool.sol
```diff
@@ -924,6 +924,54 @@ contract ComposableStablePool is
...
+     * yields intermediate results useful for other queries). This helper function returns many of these values,
+     * greatly reducing bytecode size.
+     *
+     * The return values are:
+     *  - the current upscaled token balances (not including BPT)
+     *  - the virtual supply
+     *  - the BPT amount of unpaid protocol fees
+     *  - the amplification factor at the last join or exit operation
+     *  - the invariant of the current balances, calculated using the amplification factor at the last join or exit
+     *    operation.
```
TomAFrench, line 939:
```suggestion
     * The return values are:
     *  @return balances - The current upscaled token balances (not including BPT)
     *  @return virtualSupply - The Pool's virtual supply
     *  @return protocolFeeAmount - The amount of unpaid protocol fees in BPT
     *  @return lastJoinExitAmp - The Pool's amplification factor at the last join or exit operation
     *  @return currentInvariantWithLastJoinExitAmp - The invariant of the current balances, calculated using the amplification factor at the last join or exit
     *    operation.
```

Will likely break linting.

pkg/pool-stable/contracts/ComposableStablePool.sol
```diff
@@ -979,6 +1015,23 @@ contract ComposableStablePool is
...
 
+    /**
+     * @dev Returns the effective BPT supply.
+     *
+     * In other pools, this would be the same as `totalSupply`, but there are two key differences here:
+     *  - this pool pre-mints BPT and holds it in the Vault as a token, and as such we need to subtract the Vault's
+     *    balance to get the total "circulating supply". This is called the 'virtualSupply'.
+     *  - the Pool owes debt to the Protocol in the form of unminted BPT, which will be minted immediately before the
+     *    next join or exit. We need to take these into account since, even if they don't yet exist, they will
+     *    effectively be included in any Pool operation that involves BPT.
```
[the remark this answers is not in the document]
nventuro: But is not true in old Pools (though they are wrong in a different way)

pkg/pool-stable/contracts/ComposableStablePoolStorage.sol
```diff
@@ -376,6 +376,9 @@ abstract contract ComposableStablePoolStorage is BasePool {
     /**
      * @dev Returns the number of tokens in circulation.
      *
+     * WARNING: in the vast majority of cases this is not a useful value, since it does not include the debt the Pool
```
EndymionJkb, line 379: Should we consider not even exposing the "raw" getVirtualSupply? As with the actual supply, they could calculate it externally if they really needed it, and it's not part of the BasePool interface. Would also save a little bytecode.

pkg/pool-stable/contracts/test/MockComposableStablePool.sol
```diff
@@ -69,4 +69,17 @@ contract MockComposableStablePool is ComposableStablePool, MockFailureModes {
     {
         return _beforeJoinExit(registeredBalances);
     }
+
+    function getVirtualSupply() external view returns (uint256) {
```
EndymionJkb, line 73: Should also be renamed here. And in StablePool.ts

PR #1706, opened 2022-09-06
[Deployments] Batch relayer V4 preparation.

jubeira:
- Binaries, scripts and fork test (same as in V3).
- Deprecating V3.
- Updating README.


pkg/deployments/tasks/20220906-batch-relayer-v4/readme.md
```diff
@@ -0,0 +1,10 @@
+# 2022-09-06 - Batch Relayer V4
+
+Deployment of the fourth `BalancerRelayer` using `BatchRelayerLibrary`, for combining multiple operations (swaps, joins, etc.) in a single transaction.
+
+This new version adds support for Liquidity Gauge interactions, including staking, unstaking, minting, etc.
```
nventuro, line 5: This is what v3 does :stuck_out_tongue: 

I'd replace this with something like 'this new version provides quality of life improvements, like read only chained references and ways to read them'.

nventuro (commented):
Marking as do-not-merge since we don't yet have the final build info.

Can you do a goerli deployment of the current version however, so that the SDK team can test it?

jubeira:
Done: https://github.com/balancer-labs/balancer-v2-monorepo/pull/1706/commits/34e5a7e1ed6370f5ed822238c10bcde36464140f

jubeira:
@johngrantuk the new deployment for the batch relayer in Goerli is ready; you can see more info here https://github.com/balancer-labs/balancer-v2-monorepo/pull/1706/commits/2cda6a76cb35da45c85ea48e79e1a437d3c19490.

This includes the changes in #1763 and #1765.
/cc @nventuro

TomAFrench:
When building [2d310c3](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1706/commits/2d310c3805c427b5edb6a92e52f73ea7a21acb1f) I get the same build-info as committed

TomAFrench (approved):

nventuro:
We're looking into an issue that seems to cause for the build info file to not be identical on every run - the file id is different, and some fields are in different order. But the actual contents of the source code, bytecode, etc., are all the same.

PR #1707, opened 2022-09-06
Prepare Composable Stable Pool deployment

nventuro:


TomAFrench:
I get the same build-info with hash b30de0ab284a701dd9b9a09a921a987e when building commit 838411b

PR #1710, opened 2022-09-06
FF semantics in tests

EndymionJkb:
When reviewing tests, I noticed we use fp(1) a lot in different contexts. Sometimes it's just arbitrary/convenient amount (e.g.,  bptAmountOut), but very often it's a meaningful "FixedPoint.ONE" value, as with a rate or ratio, or it means 100% (as when subtracting a percentage).

Expressing these semantics with constants makes the tests a little bit more readable. For completion, I also defined a ZERO value.


pkg/pool-linear/test/LinearPool.test.ts
```diff
@@ -625,8 +625,8 @@ describe('LinearPool', function () {
     });
 
     const itAdaptsTheScalingFactorsCorrectly = () => {
-      const expectedBptScalingFactor = fp(1);
-      const expectedMainTokenScalingFactor = fp(1);
+      const expectedBptScalingFactor = FP_SCALING_FACTOR;
+      const expectedMainTokenScalingFactor = FP_SCALING_FACTOR;
```
TomAFrench, line 629: I'm not a fan of this as it seems to be overloading what `SCALING_FACTOR` means in `FP_SCALING_FACTOR`.

`FP_SCALING_FACTOR` = "Scaling factor to convert a number into a fixed point number. (This is always 1e18)
`expectedBptScalingFactor` = "Scaling factor to convert a token to 18 decimals represented as fixed point number` (This is often 1e18 but sometimes not).

I'd change this to be `FP_ONE` as then it's clear that the scaling factor is 1 but we're showing it in a FP way.

TomAFrench (commented):
Yeah, we really needed a muldiv so this is great.

I think that we should not export `FP_SCALING_FACTOR` from `numbers.ts` anymore tbh, we were doing the wrong thing by exporting it in the first place.

I'm torn on `FP_100_PCT`, I think of the percentages as from [0, 1] so `FP_ONE` works just fine for me but it could help readability for people who think of the percentages as [0,100]

TomAFrench (approved):

PR #1713, opened 2022-09-07
Weighted pool changes

TomAFrench:
Results from WeightedPool walkthrough.

### DIFF
--- a/pkg/pool-utils/contracts/protocol-fees/ProtocolFeeCache.sol
+++ b/pkg/pool-utils/contracts/protocol-fees/ProtocolFeeCache.sol
@@ -51,7 +51,7 @@ abstract contract ProtocolFeeCache is RecoveryMode {
 
     FeeTypeCache private _cache;
 
-    event ProtocolFeePercentageCacheUpdated(uint256 indexed feeType, uint256 protocolSwapFeePercentage);
+    event ProtocolFeePercentageCacheUpdated(uint256 indexed feeType, uint256 protocolFeePercentage);
 
     // Swap fees can be set to a fixed value at construction, or delegated to the ProtocolFeePercentagesProvider if
     // passing the special sentinel value.
--- a/pkg/pool-weighted/contracts/WeightedPool.sol
+++ b/pkg/pool-weighted/contracts/WeightedPool.sol
@@ -348,8 +348,8 @@ contract WeightedPool is BaseWeightedPool, WeightedPoolProtocolFees {
         _updatePostJoinExit(getInvariant());
 
         // If the Pool has any protocol yield fees accrued then we update the athRateProduct to zero these out.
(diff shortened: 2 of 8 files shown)


pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -343,26 +343,20 @@ contract WeightedPool is BaseWeightedPool, WeightedPoolProtocolFees {
...
-    }
-
     function _onDisableRecoveryMode() internal override {
         // Update the postJoinExitInvariant to the value of the currentInvariant, zeroing out any protocol swap fees.
         _updatePostJoinExit(getInvariant());
 
-        // Update the athRateProduct to the value of the current rateProduct, zeroing out any protocol yield fees.
+        // If the Pool has any protocol yield fees accrued then we update the athRateProduct to zero these out.
+        // If the current rate product is less than the athRateProduct then we do not perform this update, this prevents
+        // the Pool from paying protocol fees on the same yield twice.
```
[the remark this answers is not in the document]
EndymionJkb: Could also expand a little on the last line. Not checking could result in lowering the athRateProduct, which is how it can result in paying fees twice on the difference between the current rate and previous ATH.

pkg/pool-weighted/contracts/WeightedPoolProtocolFees.sol
```diff
@@ -70,6 +63,19 @@ abstract contract WeightedPoolProtocolFees is BaseWeightedPool, ProtocolFeeCache
         _rateProvider7 = numTokens > 7 ? rateProviders[7] : IRateProvider(0);
     }
 
+    function _getYieldFeeExemption(IRateProvider[] memory rateProviders) internal pure returns (bool) {
+        // If we know that no rate providers are set then we can skip yield fees logic.
+        // If so then set `_exemptFromYieldFees` to true, otherwise leave it false.
```
EndymionJkb, line 68:
```suggestion
        // If any tokens have rate providers, then set `_exemptFromYieldFees` to false, otherwise leave it true.
```

EndymionJkb (commented):
Did we also want to add the "> 0" check inside _payProtocolFees, so that we can remove the external checks for non-zero fee amounts?

pkg/pool-utils/contracts/protocol-fees/ProtocolFeeCache.sol
```diff
@@ -51,7 +51,7 @@ abstract contract ProtocolFeeCache is RecoveryMode {
 
     FeeTypeCache private _cache;
 
-    event ProtocolFeePercentageCacheUpdated(uint256 indexed feeType, uint256 protocolSwapFeePercentage);
+    event ProtocolFeePercentageCacheUpdated(uint256 indexed feeType, uint256 protocolFeePercentage);
```
EndymionJkb, line 54: Need to update the tests here too

PR #1716, opened 2022-09-08
Flaky Stable test

EndymionJkb:
This addresses an intermittent failure in the Stable tests, caused by a (very slightly) changing amp value.


nventuro:
Let's try this version instead and see what results we get - we have a test right next to this one that tests deep equality in the amplification, so this doesn't look like a race condition.

PR #1717, opened 2022-09-08
Reaper linear pool and supporting utilities

danielmkm:
Implementation of the ReaperLinearPool, already deployed to the Fantom Opera network on Beethoven X, modified to support the new Rebalancer model. The ReaperLinearPool supports the latest iteration of Reaper farm's single asset vaults. Common instances of this vault include rfUSDC, rfDAi, rfUSDT, etc. The implementation leverages the vault exposed price per full share (pricePerFullShare) when inferring the wrappedTokenRate. The bulk of the contract code was written by @tess3rac7 and @degenicus from the reaper team.

This PR contains the following:

ReaperLinearPool
ReaperLinearPoolFactory
ReaperLinearPoolRebalancer
ReaperWrapping (BatchRelayer extension)


pkg/pool-linear/contracts/reaper/ReaperLinearPoolRebalancer.sol
```diff
@@ -0,0 +1,55 @@
...
+        // Withdrawing into underlying (i.e. DAI, USDC, etc. instead of rfDAI or rfUSDC). Approvals are not necessary
+        // here as the wrapped token is simply burnt.
+        IReaperTokenVault(address(_wrappedToken)).withdraw(amount);
+    }
+
+    function _getRequiredTokensToWrap(uint256 wrappedAmount) internal view override returns (uint256) {
+        IReaperTokenVault tokenVault = IReaperTokenVault(address(_wrappedToken));
+        
+        // wrappedAmount * pricePerFullShare / 10^decimals
+        return wrappedAmount.mul(tokenVault.getPricePerFullShare()).divUp(10**tokenVault.decimals());
```
TomAFrench, line 53: We can store `10**tokenVault.decimals()` in an immutable variable to save gas here.

danielmkm: thanks for the feedback! I moved the divisor into an immutable variable. I wasn't sure on the preferred way to access the `wrappedToken` in the constructor. If there is a preferred way, can adjust it.

danielmkm: Also adjusted the yearn linear pool to the same approach.

pkg/interfaces/contracts/pool-linear/IReaperTokenVault.sol
```diff
@@ -0,0 +1,48 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+
+import "../solidity-utils/openzeppelin/IERC20.sol";
+
+interface IReaperTokenVault is IERC20 {
```
nventuro, line 19: Can we add a link to the reference where we got this from?

pkg/pool-linear/contracts/reaper/ReaperLinearPool.sol
```diff
@@ -0,0 +1,95 @@
...
+     * ie: If ppfs === 1e18, 1 DAI === 1 rfDAI
+     * ie: If ppfs === 1e18, 1 WBTC === 0.000_000_000_100_000_000 rfWBTC
+     * -----------
+     * Internally, the LinearPool scales all balances and rates up to 18 decimal places, meaning that 1 USDC is 
+     * represented as 1e18 by the LinearPool. Since the rfUSDC is already 18 decimals,
+     * but in a different representation,we need to account for that in our wrappedTokenRate.
+     */
+    function _getWrappedTokenRate() internal view override returns (uint256) {
+        ERC20 underlying = ERC20(_tokenVault.token());
+        uint256 underlyingDecimals = underlying.decimals();
```
nventuro, line 81: These two values don't change - these could be made `immutable`, or even better make `10**(underlyingDecimals - 18)` an immutable state var.

pkg/pool-linear/contracts/reaper/ReaperLinearPool.sol
```diff
@@ -0,0 +1,95 @@
...
+
+        uint256 sharePrice = _tokenVault.getPricePerFullShare();
```
nventuro, line 83: I'm going to assume this meets all the LinearPool rate requirements: critically, this value must not be manipulable, and anyone should be able to get the exact same rate by interacting with Reaper.

pkg/pool-linear/contracts/reaper/ReaperLinearPoolRebalancer.sol
```diff
@@ -0,0 +1,57 @@
...
+
+    function _unwrapTokens(uint256 amount) internal override {
+        // Withdrawing into underlying (i.e. DAI, USDC, etc. instead of rfDAI or rfUSDC). Approvals are not necessary
+        // here as the wrapped token is simply burnt.
+        IReaperTokenVault(address(_wrappedToken)).withdraw(amount);
+    }
+
+    function _getRequiredTokensToWrap(uint256 wrappedAmount) internal view override returns (uint256) {
+        // wrappedAmount * pricePerFullShare / 10^decimals
+        return wrappedAmount.mul(IReaperTokenVault(address(_wrappedToken)).getPricePerFullShare()).divUp(_divisor);
```
nventuro, line 55: It'd be good to have a small comment explaining that rounding up here is important: see the one in the AaveRebalancer:

```
// We add one to ensure the returned value will always be enough to get `wrappedAmount` when unwrapping. This might result in some
// dust being left in the Rebalancer.
```

pkg/pool-linear/test/ReaperLinearPool.test.ts
```diff
@@ -0,0 +1,203 @@
...
+describe('ReaperLinearPool', function () {
+  let poolFactory: Contract;
+  let lp: SignerWithAddress, owner: SignerWithAddress;
+  let vault: Vault;
+  let funds: FundManagement;
+
+  const POOL_SWAP_FEE_PERCENTAGE = fp(0.01);
+
+  before('setup', async () => {
+    [lp, owner] = await ethers.getSigners();
```
nventuro, line 24:
```suggestion
    [, lp, owner] = await ethers.getSigners();
```

pkg/standalone-utils/contracts/relayer/ReaperWrapping.sol
```diff
@@ -0,0 +1,108 @@
...
+
+        // The unwrap caller is the implicit sender of tokens, so if the goal is for the tokens
+        // to be sourced from outside the relayer, we must first pull them here.
+        if (sender != address(this)) {
+            require(sender == msg.sender, "Incorrect sender");
+            _pullToken(sender, vaultToken, amount);
+        }
+
+        IERC20 underlyingToken = IERC20(vaultToken.token());
+        //determine the amount of underlying token that is present on the batch relayer prior to leaving.
```
nventuro, line 51:
```suggestion
        // Determine the amount of underlying token that is present on the batch relayer prior to unwrapping.
```

Please adjust the other comments to also have the same style

pkg/standalone-utils/contracts/relayer/ReaperWrapping.sol
```diff
@@ -0,0 +1,108 @@
...
+        //This should always be 0, but we want to be certain
+        uint256 tokenAmountBefore = underlyingToken.balanceOf(address(this));
+
+        //burn the rf shares and receive the underlying token
+        vaultToken.withdraw(amount);
+
+        //determine the amount of underlying returned for the shares burned
+        uint256 withdrawnAmount = underlyingToken.balanceOf(address(this)) - tokenAmountBefore;
```
nventuro, line 59: Use checked arithmetic here via `SafeMath` to avoid unnecessary exposure to reentrancy issues

pkg/standalone-utils/contracts/relayer/ReaperWrapping.sol
```diff
@@ -0,0 +1,108 @@
...
+
+        //send the shares to the recipient
+        underlyingToken.transfer(recipient, withdrawnAmount);
```
nventuro, line 62: Is there a reason to not just send the entire relayer balance? This would be cheaper as: a) we'd skip the `balanceOf` calls and b) we'd set the relayer's balance to 0, triggering a gas refund

pkg/standalone-utils/contracts/relayer/ReaperWrapping.sol
nventuro: See comment below - might be better to just not have `tokenAmountBefore` at all

pkg/standalone-utils/contracts/relayer/ReaperWrapping.sol
```diff
@@ -0,0 +1,108 @@
...
+        uint256 sharesBefore = vaultToken.balanceOf(address(this));
+
+        //deposit the tokens into the vault
+        vaultToken.deposit(amount);
+
+        //determine the amount of shares gained from depositing
+        uint256 sharesGained = vaultToken.balanceOf(address(this)) - sharesBefore;
+
+        //send the shares to the recipient
+        vaultToken.transfer(recipient, sharesGained);
```
nventuro, line 102: Same comment re. sending the entire balance

pkg/interfaces/contracts/pool-linear/IReaperTokenVault.sol
danielmkm: Added a comment with a github link and a small summary

pkg/pool-linear/contracts/reaper/ReaperLinearPool.sol
danielmkm: Attached is an excerpt from a conversation with a yearn dev. The same applies for reaper vaults. It's possible to manipulate the ppfs upwards, but the cost of this attack is not financially viable for the linear pool. All reaper vaults we will deploy linear pools for will be non upgradable single strategy vaults where assets are deposited into AAVE (no folding).

pkg/pool-linear/contracts/reaper/ReaperLinearPoolRebalancer.sol
danielmkm: Added a comment

pkg/standalone-utils/contracts/relayer/ReaperWrapping.sol
danielmkm: Agreed, we've updated the implementation to send the full relayer balance.

pkg/pool-linear/contracts/reaper/ReaperLinearPool.sol
nventuro: I'd argue that'd be a true price change, and not manipulation (since they cannot undo the change). That sort of behavior is fine.

The other important requirement is that the rate accurately reflects the underlying exchange rate. The LinearPool offers trades at that exact rate with no swap fees of any kind: if the rate is off even by a small percentage, then the Pool can be trivially drained by swapping with it back and forth.

danielmkm: Price per full share accurately represents the current and real time exchange rate of vault token to underlying asset 👍🏽

pkg/pool-linear/contracts/reaper/ReaperLinearPool.sol
```diff
@@ -54,6 +56,18 @@ contract ReaperLinearPool is LinearPool {
...
+        // Reaper vault tokens are always 18 decimal places, but are stored in the precision of the mainToken
+        // ie: If ppfs === 1e18, 1 USDC === 0.000_000_000_001_000_000 rfUSDC
+        // ie: If ppfs === 1e18, 1 DAI === 1 rfDAI
+        // ie: If ppfs === 1e18, 1 WBTC === 0.000_000_000_100_000_000 rfWBTC
+        // -----------
+        // Internally, the LinearPool scales all balances and rates up to 18 decimal places, meaning that 1 USDC is 
+        // represented as 1e18 by the LinearPool. Since the rfUSDC is already 18 decimals,
+        // but in a different representation, we need to account for that in our wrappedTokenRate.
+        // Since we only accept tokens with <= 18 decimals, we know the smallest this can be is 10^0 === 1
+        _rateScaleFactor = 10**(18 - ERC20(tokenVault.token()).decimals());
```
nventuro, line 69: I'd use `SafeMath.sub` to make sure we revert on any tokens with more than 18 decimals

nventuro:
Looks great! Could you address the linter issues? Once those are sorted out I think we'll be ready to merge this.

nventuro:
Closed accidentally, sorry!

danielmkm:
Sorry, so used to having prettier run on save, didn't realize it wasn't on. All of the lint errors should be resolved.

PR #1718, opened 2022-09-08
Return `preJoinExitInvariant` from `_beforeJoinExit` in WeightedPool

TomAFrench:
We now return `preJoinExitInvariant` from `_beforeJoinExit` to avoid recalculating it in `_afterJoinExit`.

The issues with ManagedPool were worse than I expected as when adding/removing tokens then we can't calculate the invariant to be returned. For reasons explained in the code I've just set it to return a zero value.


nventuro (approved):
Everything makes a lot of sense. Thanks!

PR #1721, opened 2022-09-08
Do weighted v2 deployment

nventuro:


TomAFrench (approved):
All deployment CI has passed so we're good to go

PR #1730, opened 2022-09-12
Add library for managing information on ManagedPool tokens

TomAFrench:
This PR adds `ManagedPoolTokenLib` which handles all data operations to do with the `_tokenState` mapping, i.e. it encapsulates token scaling factors and weights. As the data layout is the same as current, I've modified ManagedPool to make use of this new library.

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPool.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPool.sol
@@ -33,6 +33,8 @@ import "../lib/WeightCompression.sol";
 
 import "../BaseWeightedPool.sol";
 
+import "./ManagedPoolTokenLib.sol";
+
 /**
  * @dev Weighted Pool with mutable tokens and weights, designed to be used in conjunction with a pool controller
  * contract (as the owner, containing any specific business logic). Since the pool itself permits "dangerous"
@@ -209,6 +211,13 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
         _setManagementSwapFeePercentage(params.managementSwapFeePercentage);
 
         _setManagementAumFeePercentage(params.managementAumFeePercentage);
+
+        // Write the scaling factors for each token into their token state.
+        // We do this before setting the weights in `_startGradualWeightChange` so we start from a empty token state.
+        for (uint256 i = 0; i < totalTokens; i++) {
+            IERC20 token = params.tokens[i];
+            _tokenState[token] = ManagedPoolTokenLib.setTokenScalingFactor(bytes32(0), token);
+        }
 
         // Initialize the denormalized weight sum to ONE. This value can only be changed by adding or removing tokens.
         _denormWeightSum = FixedPoint.ONE;
@@ -683,9 +692,15 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
         // pass an empty array for this parameter.
         getVault().registerTokens(getPoolId(), tokensToAdd, new address[](1));
 
-        // `_encodeTokenState` performs an external call to `token` (to get its decimals). Nevertheless, this is
-        // reentrancy safe. View functions are called in a STATICCALL context, and will revert if they modify state.
-        _tokenState[token] = _encodeTokenState(token, normalizedWeight, normalizedWeight, newDenormWeightSum);
+        // `ManagedPoolTokenLib.setTokenScalingFactor` performs an external call to `token` (to get its decimals).
+        // Nevertheless, this is reentrancy safe. View functions are called in a STATICCALL context, and will revert
+        // if they modify state.
+        _tokenState[token] = ManagedPoolTokenLib.setTokenWeight(
+            ManagedPoolTokenLib.setTokenScalingFactor(bytes32(0), token),
+            normalizedWeight,
+            normalizedWeight,
+            newDenormWeightSum
+        );
         _totalTokensCache += 1;
     }
 
@@ -869,7 +884,7 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
     }
 
     function _scalingFactor(IERC20 token) internal view virtual override returns (uint256) {
-        return _readScalingFactor(_getTokenData(token));
+        return ManagedPoolTokenLib.getTokenScalingFactor(_getTokenData(token));
     }
 
     function _scalingFactors() internal view virtual override returns (uint256[] memory scalingFactors) {
@@ -879,24 +894,17 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
         scalingFactors = new uint256[](numTokens);
 
         for (uint256 i = 0; i < numTokens; i++) {
-            scalingFactors[i] = _readScalingFactor(_tokenState[tokens[i]]);
+            scalingFactors[i] = ManagedPoolTokenLib.getTokenScalingFactor(_tokenState[tokens[i]]);
         }
     }
 
     function _getNormalizedWeight(IERC20 token) internal view override returns (uint256) {
-        bytes32 tokenData = _getTokenData(token);
-        uint256 startWeight = tokenData.decodeUint(_START_DENORM_WEIGHT_OFFSET, 64).decompress(64, _MAX_DENORM_WEIGHT);
-        uint256 endWeight = tokenData.decodeUint(_END_DENORM_WEIGHT_OFFSET, 64).decompress(64, _MAX_DENORM_WEIGHT);
-
         bytes32 poolState = _getMiscData();
         uint256 startTime = poolState.decodeUint(_WEIGHT_START_TIME_OFFSET, 32);
         uint256 endTime = poolState.decodeUint(_WEIGHT_END_TIME_OFFSET, 32);
 
-        return
-            _normalizeWeight(
-                GradualValueChange.getInterpolatedValue(startWeight, endWeight, startTime, endTime),
-                _denormWeightSum
-            );
+        uint256 weightChangeProgress = GradualValueChange.calculateValueChangeProgress(startTime, endTime);
+        return ManagedPoolTokenLib.getTokenWeight(_getTokenData(token), weightChangeProgress, _denormWeightSum);
     }
 
     // This could be simplified by simply iteratively calling _getNormalizedWeight(), but this routine is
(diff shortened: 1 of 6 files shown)


TomAFrench:
Looking into test case failure.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -573,10 +564,13 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
         // is achieved efficiently by simply updating the sum of the denormalized weights.
         _denormWeightSum = weightSumAfterAdd;
 
-        // Finally, we store the new token's weight.
-        // `_encodeTokenState` performs an external call to `token` (to get its decimals), but this is
-        // reentrancy safe since that is a static call.
-        _tokenState[token] = _encodeTokenState(token, normalizedWeight, normalizedWeight, weightSumAfterAdd);
+        // Finally, we store the new token's weight and scaling factor.
+        _tokenState[token] = ManagedPoolTokenLib.setTokenWeight(
+            ManagedPoolTokenLib.setTokenScalingFactor(bytes32(0), token),
```
nventuro, line 569: Sneakily initializing the state to include the scaling factor is rather odd - I had to spend about 3 minutes going back and forth trying to understand what this line was doing.

An issue here is that there's no guarantee the full state is scaling factor + weight. We should perhaps have a function that returns a fully initialied state, so we can be sure we're not forgetting anything.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -708,7 +695,7 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
         // all other token weights accordingly.
         // Clean up data structures and update the token count
         delete _tokenState[token];
-        _denormWeightSum -= _denormalizeWeight(tokenNormalizedWeight, _denormWeightSum);
+        _denormWeightSum -= tokenNormalizedWeight.mulUp(_denormWeightSum);
```
nventuro, line 698: DId you feel these didn't add much?

pkg/pool-weighted/contracts/test/MockManagedPoolTokenLib.sol
```diff
@@ -0,0 +1,85 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
+
+import "@balancer-labs/v2-interfaces/contracts/solidity-utils/openzeppelin/IERC20.sol";
+
+import "../managed/ManagedPoolTokenLib.sol";
+
+contract MockManagedPoolTokenLib {
```
nventuro, line 22: If only there was a way to test libraries directly from Solidity

pkg/pool-weighted/test/WeightedPool.test.ts
```diff
@@ -62,7 +62,7 @@ describe('WeightedPool', function () {
 
         // Perform a recovery mode exit. This will reduce the invariant but this isn't tracked due to recovery mode.
         const preExitInvariant = await pool.getLastPostJoinExitInvariant();
-        await pool.recoveryModeExit({ from: lp, bptIn: fp(100) });
+        await pool.recoveryModeExit({ from: lp, bptIn: fp(10) });
```
nventuro, line 65: Why this change?

pkg/pool-weighted/test/ManagedPoolTokenLib.test.ts
```diff
@@ -0,0 +1,169 @@
...
+    const clearedResult = mask.and(result);
+    const clearedWord = mask.and(word);
+    expect(clearedResult).to.equal(clearedWord);
+  }
+
+  describe('token scaling factor', () => {
+    const DECIMAL_DIFF_OFFSET = 128;
+    const DECIMAL_DIFF_WIDTH = 5;
+
+    async function assertTokenScalingFactor(
```
nventuro, line 34: I'd have this function also receive an `expected` argument, and also check against that.

pkg/pool-weighted/contracts/managed/ManagedPoolTokenLib.sol
```diff
@@ -0,0 +1,167 @@
...
+        returns (uint256, uint256)
+    {
+        return (
+            _decodeWeight(tokenState.decodeUint(_START_DENORM_WEIGHT_OFFSET, _DENORM_WEIGHT_WIDTH), denormWeightSum),
+            _decodeWeight(tokenState.decodeUint(_END_DENORM_WEIGHT_OFFSET, _DENORM_WEIGHT_WIDTH), denormWeightSum)
+        );
+    }
+
+    function getMinimumTokenEndWeight(
+        mapping(IERC20 => bytes32) storage tokenStates,
```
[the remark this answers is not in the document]
TomAFrench: Yeah, I toyed with the idea of building the array of `tokenStates` inside `ManagedPool` but extra memory cost + and avoids `ManagedPool` from having to think as much.

pkg/pool-weighted/contracts/test/MockManagedPoolTokenLib.sol
TomAFrench: I haven't found rly anything that would allow that.

pkg/pool-weighted/contracts/managed/ManagedPoolTokenLib.sol
```diff
@@ -0,0 +1,167 @@
...
+    // Getters
+
+    /**
+     * @notice Returns the token's scaling factor.
+     */
+    function getTokenScalingFactor(bytes32 tokenState) internal pure returns (uint256) {
+        uint256 decimalsDifference = tokenState.decodeUint(_DECIMAL_DIFF_OFFSET, _DECIMAL_DIFF_WIDTH);
+
+        // This is equivalent to `10**(18+decimalsDifference)` but this form optimizes for 18 decimal tokens.
+        return FixedPoint.ONE * 10**decimalsDifference;
```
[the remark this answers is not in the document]
TomAFrench: It's an EXP opcode which has a cost of `10 + 50 * exponent_byte_size`. I was thinking that we could do the multiplication for free if we shove `FixedPoint.ONE` up in the exponent but then we'd always have a non-zero exponent and pay 50 rather than 10 gas for most tokens.

pkg/pool-weighted/test/WeightedPool.test.ts
TomAFrench: The tests were failing due to `lp` not having enough BPT

pkg/pool-weighted/contracts/managed/ManagedPool.sol
TomAFrench: Through offloading most of the usages into `ManagedPoolTokenLib` this ended up being the only place where `_denormalizeWeight` was used. I think at this point inlining it makes sense.

pkg/pool-weighted/contracts/managed/ManagedPoolTokenLib.sol
```diff
@@ -0,0 +1,167 @@
...
+     * @notice Updates a token's start and end weights.
+     * @dev To be called when initiating a gradual weight change to update the two values to interpolate between.
+     */
+    function setTokenWeight(
+        bytes32 tokenState,
+        uint256 normalizedStartWeight,
+        uint256 normalizedEndWeight,
+        uint256 denormWeightSum
+    ) internal pure returns (bytes32) {
+        return
```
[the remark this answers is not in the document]
TomAFrench: I think that times in which we write new state are going to be fairly localised and only change a single field. This means that we can merge the modification and setting of state.

```
bytes32 state = _tokenState[token];
/* calculations of start and end weights */
_tokenState[token] =  ManagedPoolTokenLib.setTokenWeight(state, newStartWeight, newEndWeight, denormWeightSum); 
```

The chance of missing a write only occurs if we need to perform multiple updates to the token state while reading values from a half-updated state which I don't think would ever happen.

pkg/pool-weighted/contracts/managed/ManagedPoolTokenLib.sol
```diff
@@ -0,0 +1,167 @@
...
+    // Denormalized weights are stored using the WeightCompression library as a percentage of the maximum absolute
+    // denormalized weight: independent of the current _denormWeightSum, which avoids having to recompute the denorm
+    // weights as the sum changes.
+    uint256 private constant _MAX_DENORM_WEIGHT = 1e22; // FP 10,000
+
+    // Store scaling factor and start/end denormalized weights for each token
+    // Mapping should be more efficient than trying to compress it further
+    // [ 123 bits |  5 bits  |  64 bits   |   64 bits    |
+    // [ unused   | decimals | end denorm | start denorm |
+    // |MSB                                           LSB|
```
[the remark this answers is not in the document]
TomAFrench: This library does require some knowledge of the `denormWeightSum` as it takes it as an argument to normalize weights. Agreed on needing to update comments.

pkg/pool-weighted/contracts/managed/ManagedPoolTokenLib.sol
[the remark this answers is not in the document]
TomAFrench: If I had to chose between the two approaches though I would definitely go b).

pkg/pool-weighted/test/ManagedPoolTokenLib.test.ts
TomAFrench: As in an expected value for `result`? I'm inclined to work just off of observable values instead as this is all the Pool cares about (the pool should only use `_tokenState` to feed it into this lib).

TomAFrench: This would likely get pretty yucky as we'd have to be able to predict the rounding errors involved in compressing the weights as we can't do fuzzy matching on the entire state without decoding it first (at which point we'd just be back to the current testing regime but we're testing everything in every test.)

TomAFrench:
Note that I've expanded `_MAX_DENORM_WEIGHT` from `1e22` to `2**64`. I'm not sure why we were not using the full range beforehand.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
TomAFrench: [95bcb3b](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1730/commits/95bcb3bac511e86250b6a9862572c5ce0b6661ae)

pkg/pool-weighted/contracts/managed/ManagedPoolTokenLib.sol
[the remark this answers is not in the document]
TomAFrench: Shall we wait until we have `ManagedBasePool` and start flattening this to decide how to move down this path?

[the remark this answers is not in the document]
nventuro: We decided to go with alternative c) leave everything as is, since the arguments presented above are good.

pkg/pool-weighted/contracts/test/MockManagedPoolTokenLib.sol
nventuro: *standing ovation*

pkg/pool-weighted/test/WeightedPool.test.ts
nventuro: But why did the tests start failing? Did we change anything else?

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -210,6 +200,13 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
 
         _setManagementAumFeePercentage(params.managementAumFeePercentage);
 
+        // Write the scaling factors for each token into their token state.
+        // We do this before setting the weights in `_startGradualWeightChange` so we start from a empty token state.
+        for (uint256 i = 0; i < totalTokens; i++) {
+            IERC20 token = params.tokens[i];
+            _tokenState[token] = ManagedPoolTokenLib.setTokenScalingFactor(bytes32(0), token);
```
nventuro, line 207: Bit odd that we don't use the initialization function here

nventuro (approved):

pkg/pool-weighted/test/WeightedPool.test.ts
TomAFrench: I've reverted the change and it seems to work again. I can't reproduce locally.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
TomAFrench: Will address separately.

PR #1731, opened 2022-09-12
Add specialized invariant variant to calculate managed swap fees.

TomAFrench:
To use `WeightedMath._calculateInvariant` we have to create dynamic arrays which in turn means that we have to use the `ArrayHelpers.arrayFill` function which is a bit of a kludge.

This PR adds a two token variant of `WeightedMath._calculateInvariant` which we can call with fixed size arrays. This should save gas on swaps with a bytecode cost of <10 bytes.


nventuro (commented):
I like how we now measure features in terms of bytecode footprint.

Have you considered simply passing 4 stack values instead of two memory arrays? I'm curious about the memory usage since we're not using `for` loops etc.

PR #1734, opened 2022-09-12
Add ManagedPoolStorageLib

TomAFrench:
This is the equivalent library for the Pool state (`_miscData`) as #1730 is for the token state.

I've moved "end swap fee" out of the top 62 bits and down to sit next to "start swap fee", making this library incompatible with the current form of ManagedPool so I haven't integrated it. Even then, Basepool prevents us from setting these top bits ourselves so we'd need to remove this protection first anyway (which I think should happen in a separate PR).

It also doesn't have any concept of Recovery mode as it wasn't mentioned in the layout described in `ManagedPool` so this will need to be added at some point.


nventuro:
>It also doesn't have any concept of Recovery mode as it wasn't mentioned in the layout described in ManagedPool

Note that this is simply outdated docs: the recovery mode bit is bundled by BasePool in misc data.

pkg/pool-weighted/contracts/managed/ManagedPoolStorageLib.sol
```diff
@@ -0,0 +1,156 @@
...
+    uint256 private constant _SWAP_FEE_START_TIME_OFFSET = 64;
+    uint256 private constant _SWAP_FEE_END_TIME_OFFSET = 96;
+    uint256 private constant _SWAP_FEE_START_PERCENTAGE_OFFSET = 128;
+    uint256 private constant _SWAP_FEE_END_PERCENTAGE_OFFSET = 192;
+    uint256 private constant _SWAP_ENABLED_OFFSET = 254;
+    uint256 private constant _MUST_ALLOWLIST_LPS_OFFSET = 255;
+
+    uint256 private constant _TIMESTAMP_WIDTH = 32;
+    uint256 private constant _SWAP_FEE_START_PERCENTAGE_WIDTH = 64;
+    uint256 private constant _SWAP_FEE_END_PERCENTAGE_WIDTH = 62;
```
nventuro, line 43: I'm fond of autocomputing these as:

```
weight_start_offset = 0
wieght_end_offset = weight_start_offset + timestamp_width
swap_fee_offset = weight_end_offset + timestamp_width
...
```

But not everyone likes that style :(

pkg/pool-weighted/contracts/managed/ManagedPoolStorageLib.sol
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    function setSwapFeeData(
+        bytes32 miscData,
+        uint256 startTime,
+        uint256 endTime,
+        uint256 startSwapFeePercentage,
+        uint256 endSwapFeePercentage
+    ) internal pure returns (bytes32) {
+        // Add scope to prevent "Stack too deep"
```
nventuro, line 117: Do we really need this? That's a bit surprising - I imagine it has to do with the temporary values that result from the chained `.insert` calls.

It'd be interesting to change this to alternative forms and see if we get a noticeable impact on e.g bytecode size, say
```
data = data.insert(...)
data = data.insert(...)
data = data.insert(...)
return data
```

nventuro (commented):
I like this pattern. Most of my comments re. overall strategy would be duplicates of the ones in #1730, so let's just continue the conversation there.

pkg/pool-weighted/contracts/managed/ManagedPoolStorageLib.sol
TomAFrench: The bytecode cause by the two approaches seems to be identical.

TomAFrench:
> Note that this is simply outdated docs: the recovery mode bit is bundled by BasePool in misc data.

Yep, I haven't tackled this as it is has the same issue as the swap fee.

nventuro:
Do you want to merge this as-is and then afterwords use this in Managed, or to do all at once here? It seems like the former would be easier.

TomAFrench:
It should definitely be a two step process. I just wanted to tackle the integration before we commit to a final form of this library so we don't run into issues.

Let's merge this now and then we can start looking at #1748 and how to smooth out it's inclusion.

nventuro (approved):

PR #1736, opened 2022-09-12
Set `_lastAumFeeCollectionTimestamp` on initialisation

TomAFrench:
Rather than faffing about with a zero check on `lastCollection` on every join/exit, let's just set it at initialisation.


pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -1268,6 +1268,24 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
         return (protocolBptAmount, managerBPTAmount);
     }
 
+    /**
+     * @dev Pays any due protocol and manager fees before updating the cached protocol fee percentages.
+     */
+    function _beforeProtocolFeeCacheUpdate() internal override {
```
nventuro, line 1275: We didn't have this? :grimacing:

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -1268,6 +1268,24 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
...
+        // We pay any due protocol or manager fees *before* updating the cache. This ensures that the new
+        // percentages only affect future operation of the Pool, and not past fees.
+
+        // Given that this operation is state-changing and relatively complex, we only allow it as long as the Pool is
+        // not paused.
+        _ensureNotPaused();
+
+        uint256 supplyBeforeFeeCollection = totalSupply();
+        if (supplyBeforeFeeCollection > 0) {
```
nventuro, line 1284: I'd get rid of this `if` and have this blow up tbh

TomAFrench: `_collectAumManagementFees` doesn't blow up if `totalSupply == 0` but does a zero value mint for the manager.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
TomAFrench: :shushing_face:

pkg/pool-weighted/contracts/managed/ManagedPool.sol
TomAFrench: I could make it blow up instead but it would be a conscious choice rather than just allowing it to fail.

nventuro: Don't we skip zero value mints? If the `if` does nothing I'd just remove it. We don't need to have it blow up - ideally we don't do anything special here (assuming it is fine to not do anything).

We should still have a test for this scenario though to catch regressions.

TomAFrench: > Don't we skip zero value mints? If the if does nothing I'd just remove it.

Not currently. This will be fixed with #1742 though.

nventuro: I was referring to https://github.com/balancer-labs/balancer-v2-monorepo/pull/1722

TomAFrench: We don't have a zero value mint check for the manager though.

nventuro: Ah, sadness. Can we remove the `if` and adjust the test as needed if the behavior later changes? :upside_down_face:

TomAFrench: Yep, I'm happy to remove the `if` once #1742 is in.

PR #1740, opened 2022-09-13
Use Hardhat's linking

TomAFrench:
Addresses below comment from `pvt/helpers/src/contract.ts`

> // From https://github.com/nomiclabs/hardhat/issues/611#issuecomment-638891597, temporary workaround until
> // https://github.com/nomiclabs/hardhat/issues/1716 is addressed.


nventuro (approved):
Fantatsic, thanks!

PR #1742, opened 2022-09-13
Move AUM fee calculation into a library

TomAFrench:
Builds on #1736 

This PR moves the AUM fee calculation into a simple library so that `_collectAUMFeeCalculation` only shows logic relevant for the pool (i.e. updating state and distributing fees).

### DIFF
--- a/pkg/pool-utils/contracts/protocol-fees/ProtocolAUMFees.sol
+++ b/pkg/pool-utils/contracts/protocol-fees/ProtocolAUMFees.sol
@@ -40,11 +40,12 @@ library ProtocolAUMFees {
 
         uint256 annualBptAmount = ProtocolFees.bptForPoolOwnershipPercentage(totalSupply, annualAumFeePercentage);
 
-        // We want to collect fees so that after a year the Pool will have paid `annualAumFeePercentage` of it's AUM as
+        // We want to collect fees so that after a year the Pool will have paid `annualAumFeePercentage` of its AUM as
         // fees. In normal operation however, we will collect fees regularly over the course of the year so we
         // multiply `annualBptAmount` by the fraction of the year which has elapsed since we last collected fees.
         uint256 elapsedTime = currentTime - lastCollection;
 
+        // Like with all other fees, we round down, favoring LPs.
         return Math.divDown(Math.mul(annualBptAmount, elapsedTime), 365 days);
     }
 }
--- a/pkg/pool-utils/contracts/protocol-fees/ProtocolAUMFees.sol
+++ b/pkg/pool-utils/contracts/protocol-fees/ProtocolAUMFees.sol
@@ -55,6 +55,9 @@ library ProtocolAUMFees {
         // fees_to_collect = expected_yearly_fees * time_since_last_collection / 1 year
         //                 = 52.63e18 * 7 / 365
         //                 ~= 1.009 BPT
+        //
+        // Note that if we were to mint expected_yearly_fees BPT then the recipient would own 52.63e18 out of
+        // 1052.63e18 BPT. This agrees with the recipient being expected to own 5% of the Pool *after* fees are paid.
 
         // Like with all other fees, we round down, favoring LPs.
         return Math.divDown(Math.mul(annualBptAmount, elapsedTime), 365 days);
--- a/pkg/pool-utils/contracts/protocol-fees/ProtocolAUMFees.sol
+++ b/pkg/pool-utils/contracts/protocol-fees/ProtocolAUMFees.sol
@@ -45,6 +45,17 @@ library ProtocolAUMFees {
         // multiply `annualBptAmount` by the fraction of the year which has elapsed since we last collected fees.
         uint256 elapsedTime = currentTime - lastCollection;
 
+        // As an example for this calculate, consider a pool with a total supply of 1000e18 BPT, AUM fees are charged
+        // at 5% yearly and it's been 7 days since the last collection of AUM fees. The expected fees are then:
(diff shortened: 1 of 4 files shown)


pkg/pool-utils/test/ProtocolAUMFees.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  context('when no time has passed since last collection', () => {
+    const lastCollectionTime = random(1000, 10000);
+    const currentTime = lastCollectionTime - 1;
+
+    it('returns zero', async () => {
+      const totalSupply = fp(random(1, 100));
+      const aumFeePercentage = fp(random(0, 0.9));
+
+      expect(await lib.getAumFeesBptAmount(totalSupply, currentTime, lastCollectionTime, aumFeePercentage)).to.be.eq(
```
nventuro, line 36: - Mom can we have fuzzing?
- We have fuzzing at home.

Fuzzing at home:

pkg/pool-utils/test/ProtocolAUMFees.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+      context('when total supply is nonzero', () => {
+        const totalSupply = fp(random(1, 100));
+
+        it('returns the expected amount', async () => {
+          const expectedBptAmount = expectedAUMFees(totalSupply, aumFeePercentage, currentTime - lastCollectionTime);
+          expect(
+            await lib.getAumFeesBptAmount(totalSupply, currentTime, lastCollectionTime, aumFeePercentage)
+          ).to.be.eq(expectedBptAmount);
+        });
```
nventuro, line 79: I think it'd be a good practice to have simple numerical examples in the source code to make sure the reader understands what we're trying to do, and to then have a very dumb test that uses those constants to check the result is correct. This would be low effort, would improve readability, and would decrease the burden of the helper functions (in this case `expectedAUMFees`) being correct.

An an example, we could here describe a scenario where total supply is 1000, 7 days have passed, and the fee is 5%. 
7 days is ~1.92% of the year, and so we need to mint xyz. 

Writing this actually made me thing that we're perhaps getting APR and APY mixed up - are you sure we should compute the anual percentage and then divide instead of computing the e.g. daily fee % and then applying that? Regardless of which one is the right one, this seems like good indication of the need to have these examples I suggest.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -1230,33 +1232,26 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
...
-        // exit the Pool and receive this fraction of the Pool's assets.
-        uint256 annualizedFee = ProtocolFees.bptForPoolOwnershipPercentage(totalSupply, managementAumFeePercentage);
-
-        // This value is annualized: in normal operation we will collect fees regularly over the course of the year.
-        // We then multiply this value by the fraction of the year which has elapsed since we last collected fees.
-        uint256 elapsedTime = currentTime - lastCollection;
-        uint256 bptAmount = Math.divDown(Math.mul(annualizedFee, elapsedTime), 365 days);
+        // As we update `_lastAumFeeCollectionTimestamp` when updating `_managementAumFeePercentage`, we only need to
+        // update `_lastAumFeeCollectionTimestamp` when non-zero AUM fees are paid. This avoids an SSTORE on zero-length
+        // collections.
```
nventuro, line 1251: This would be priced as a warm sload, but yeah.

pkg/pool-utils/contracts/protocol-fees/ProtocolAUMFees.sol
```diff
@@ -0,0 +1,50 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
+
+import "@balancer-labs/v2-solidity-utils/contracts/math/Math.sol";
+
+import "./ProtocolFees.sol";
+
+library ProtocolAUMFees {
```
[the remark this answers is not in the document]
TomAFrench: Agreed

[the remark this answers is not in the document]
TomAFrench: #1747

pkg/pool-utils/contracts/protocol-fees/ProtocolAUMFees.sol
```diff
@@ -0,0 +1,62 @@
...
+        // fees. In normal operation however, we will collect fees regularly over the course of the year so we
+        // multiply `annualBptAmount` by the fraction of the year which has elapsed since we last collected fees.
+        uint256 elapsedTime = currentTime - lastCollection;
+
+        // As an example for this calculate, consider a pool with a total supply of 1000e18 BPT, AUM fees are charged
+        // at 5% yearly and it's been 7 days since the last collection of AUM fees. The expected fees are then:
+        //
+        // expected_yearly_fees = totalSupply * annualAumFeePercentage / (1 - annualAumFeePercentage)
+        //                      = 1000e18 * 0.05 / 0.95
+        //                      ~= 52.63e18 BPT
```
nventuro, line 53:
```suggestion
        //                      ~= 52.63e18 BPT
        //
        // (if we minted those BPT, the total supply would increase to ~1052.63e18, and whomever held these tokens would
        // then own 5% of the total BPT)
```

nventuro (approved):

Issue #945, opened 2021-10-15
[relayer] Add 'relayer balance' sentinel value


nventuro:
We've identified scenarios where there are valid 'given out' actions while some leftover (e.g. unstake + exit given tokens out, resulting in dust BPT). We could provide a escape hatch mechanism in the form of a sentinel amount value that stands for 'use all relayer balance', and sort these out that way.

nventuro:
This idea is still good and worth pursuing, but we're not including the feature in relayer v4.

EndymionJkb:
A good idea, but it was deferred from relayer v4 and no further relayer versions will ship on frozen V2. Closing.

PR #1743, opened 2022-09-13
[WIP] Use relayer balance for exits in vault actions (BPT in).

jubeira:
See #945.

This initial draft replaces BPT amount in for the relayer balance when a sentinel value (`0xa11`) is found in the amount.

Some notes:
- Only for exits, when BPT in is given.
 - Does it make sense to implement this for tokens given out? On paper we only care about the amount of tokens given out in those cases, not how much BPT we use. In other words we'd be ignoring the operation kind, which sounds counter-intuitive.
- It doesn't make much sense to use all relayer balance when sender is not relayer, so we should probably revert if that's not the case when the sentinel value is found.
- Do we want to extend the sentinel value functionality to other operations as well? It'll probably mean an extra `else if` when we look for chained references as in this draft.

Input is welcome!


pkg/standalone-utils/contracts/relayer/VaultActions.sol
```diff
@@ -236,52 +236,77 @@ abstract contract VaultActions is IBaseRelayerLibrary {
...
+            bptAmountIn = IERC20(VaultHelpers.toPoolAddress(poolId)).balanceOf(address(this));
+            return abi.encode(WeightedPoolUserData.ExitKind.EXACT_BPT_IN_FOR_TOKENS_OUT, bptAmountIn);
         } else {
             // Save gas by only re-encoding the data if we actually performed a replacement
             return userData;
         }
     }
+
+    function _useRelayerBalance(uint256 bptAmountIn) internal pure returns (bool) {
+        // Look for 0xa11 sentinel value to use all of relayer's balance.
```
nventuro, line 307: Lol does this stand for 'all'? :joy:

jubeira: if `0xba1` stands for BAL, why not? :upside_down_face:

PR #1745, opened 2022-09-13
Vendor BasePool for ManagedPool

TomAFrench:
This copies `BasePool`, `BaseMinimalSwapInfoPool` and `BaseWeightedPool` into the `managed/vendor` directory so `ManagedPool` has it's own copy for us to tear up.


nventuro (approved):
Beginning of the end

PR #1746, opened 2022-09-13
Begin flattening ManagedPool

TomAFrench:
Here's some low-hanging fruit changes we can make.

### DIFF
--- a/pkg/pool-weighted/contracts/managed/vendor/BasePool.sol
+++ b/pkg/pool-weighted/contracts/managed/vendor/BasePool.sol
@@ -632,23 +632,6 @@ abstract contract BasePool is
     // Scaling
 
     /**
-     * @dev Returns a scaling factor that, when multiplied to a token amount for `token`, normalizes its balance as if
-     * it had 18 decimals.
-     */
-    function _computeScalingFactor(IERC20 token) internal view returns (uint256) {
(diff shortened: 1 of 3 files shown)


pkg/pool-weighted/contracts/managed/vendor/BasePool.sol
```diff
@@ -631,23 +631,6 @@ abstract contract BasePool is
...
-            return FixedPoint.ONE;
-        }
-
-        // Tokens that don't implement the `decimals` method are not supported.
-        uint256 tokenDecimals = ERC20(address(token)).decimals();
-
-        // Tokens with more than 18 decimals are not supported.
-        uint256 decimalsDifference = Math.sub(18, tokenDecimals);
-        return FixedPoint.ONE * 10**decimalsDifference;
-    }
```
nventuro, line 649: Is this because we already bundle this on the token storage initialization?

nventuro (approved):

pkg/pool-weighted/contracts/managed/vendor/BasePool.sol
TomAFrench: Yes, this is now the responsibility of `ManagedPoolTokenLib` so there's no need for it to live in the pool.

PR #1748, opened 2022-09-13
ManagedPoolStorageLib integration

TomAFrench:
Draft as I want to be sure of test coverage before we merge this.

This PR does away with `_miscData` and replaces it with `_poolState` which is managed by `ManagedPoolStorageLib`. Any usage of `_miscData` has been replaced with the appropriate getter/setter from `ManagedPoolStorageLib`

`BasePool._setSwapFeePercentage` has been inlined into `ManagedPool._setSwapFeePercentage` but I'll replicate this change on master as well.

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPool.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPool.sol
@@ -297,7 +297,12 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
         return ManagedPoolStorageLib.getSwapFeeFields(_getPoolState());
     }
 
-    function _setSwapFeePercentage(uint256 swapFeePercentage) internal virtual override {
+    /**
+     * @notice Set the swap fee percentage.
+     * @dev This is a permissioned function, and disabled if the pool is paused. The swap fee must be within the
+     * bounds set by MIN_SWAP_FEE_PERCENTAGE/MAX_SWAP_FEE_PERCENTAGE. Emits the SwapFeePercentageChanged event.
+     */
+    function setSwapFeePercentage(uint256 swapFeePercentage) external override authenticate whenNotPaused {
         // Do not allow setting if there is an ongoing fee change
         uint256 currentTime = block.timestamp;
         bytes32 poolState = _getPoolState();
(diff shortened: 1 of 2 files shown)


TomAFrench:
This needs the fixes made as a result of [this comment](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1749#discussion_r969892359) applied

pkg/pool-weighted/contracts/managed/vendor/BasePool.sol
```diff
@@ -285,16 +253,15 @@ abstract contract BasePool is
...
     /**
-     * @dev Inserts data into the least-significant 192 bits of the misc data storage slot.
-     * Note that the remaining 64 bits are used for the swap fee percentage and cannot be overloaded.
+     * @dev Inserts data into the pool state storage slot.
      */
-    function _setMiscData(bytes32 newData) internal {
-        _miscData = _miscData.insertBits192(newData, 0);
+    function _setPoolState(bytes32 newPoolState) internal {
+        _poolState = newPoolState;
     }
```
nventuro, line 265: Is this just to have more explicit state access syntax?

nventuro: Imagine if we had `this.poolState = foo` :smiling_face_with_tear:

nventuro: You could try a simple bytecode size measurement. Reading the slot shouldn't be much more than `push1 slot sload`, whereas the function call requires pushing the destination, jump, jumpdest, etc.

pkg/pool-weighted/contracts/managed/vendor/BasePool.sol
```diff
@@ -180,27 +168,7 @@ abstract contract BasePool is
...
-        _require(swapFeePercentage >= _getMinSwapFeePercentage(), Errors.MIN_SWAP_FEE_PERCENTAGE);
-        _require(swapFeePercentage <= _getMaxSwapFeePercentage(), Errors.MAX_SWAP_FEE_PERCENTAGE);
-
-        _miscData = _miscData.insertUint(
-            swapFeePercentage,
-            _SWAP_FEE_PERCENTAGE_OFFSET,
-            _SWAP_FEE_PERCENTAGE_BIT_LENGTH
-        );
-
-        emit SwapFeePercentageChanged(swapFeePercentage);
```
nventuro, line 202: This seems out of scope - isn't this a duplicate of what https://github.com/balancer-labs/balancer-v2-monorepo/pull/1749does?

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -231,14 +212,14 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
      * @notice Returns whether swaps are enabled.
      */
     function getSwapEnabled() public view returns (bool) {
```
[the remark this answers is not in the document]
TomAFrench: Agreed. We do make use of this internally atm however and inlining ManagedPoolStorageLib calls at these places would add a ton of noise to the PR for little benefit. I'm aiming for this PR to be "The old ManagedPool you know... but with ManagedPoolStorageLib!" and we can do funky stuff later.

pkg/pool-weighted/contracts/managed/vendor/BasePool.sol
nventuro: Should this be rebased now that PR is merged? Note that one changed somewhat.

nventuro (approved):
Merge once the swap fee duplication of #1749 is addressed

PR #1749, opened 2022-09-13
Remove protections on upper bits of `_miscData` and inline `_setSwapFeePercentage`

TomAFrench:
This PR lays some of the groundwork for #1748 as removing the limitations on setting the upper bits is required for that PR.

~~There's an easter egg related to `endSwapFeePercentage` that I noticed while making this PR.~~ Misread stuff so this isn't an issue.

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPool.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPool.sol
@@ -327,7 +327,7 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
      * @dev This is a permissioned function, and disabled if the pool is paused. The swap fee must be within the
      * bounds set by MIN_SWAP_FEE_PERCENTAGE/MAX_SWAP_FEE_PERCENTAGE. Emits the SwapFeePercentageChanged event.
      */
-    function setSwapFeePercentage(uint256 swapFeePercentage) external authenticate whenNotPaused {
+    function setSwapFeePercentage(uint256 swapFeePercentage) external override authenticate whenNotPaused {
         // Do not allow setting if there is an ongoing fee change
         uint256 currentTime = block.timestamp;
         bytes32 poolState = _getMiscData();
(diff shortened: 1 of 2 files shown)


pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -335,9 +335,17 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
             );
         }
 
+        _require(swapFeePercentage >= _getMinSwapFeePercentage(), Errors.MIN_SWAP_FEE_PERCENTAGE);
+        _require(swapFeePercentage <= _getMaxSwapFeePercentage(), Errors.MAX_SWAP_FEE_PERCENTAGE);
```
nventuro, line 339: I'd move these checks to the top of the function

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -1081,7 +1089,7 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
         uint256 endSwapFeePercentage
     ) internal virtual {
         if (startSwapFeePercentage != getSwapFeePercentage()) {
-            super._setSwapFeePercentage(startSwapFeePercentage);
```
nventuro, line 1084: Whoah, this changes behavior, doesn't it? With this change, we'd no longer allow a gradual change if the start value is not the current one _and_ there's an ongoing change.

It's also interesting that with this change we potentially write the end time twice, since both here and in the call we call `_setSwapFeeData()`. At any rate, the whole structure is confusing. It'd be great to eventually get rid of `setSwapFeePercentage()` entirely.

TomAFrench: > Whoah, this changes behavior, doesn't it? With this change, we'd no longer allow a gradual change if the start value is not the current one and there's an ongoing change.

Ah this is true, I'll address this later.

TomAFrench: I've hoisted up `setSwapFeePercentage` to ManagedPool to hold the behaviour blocking changing the swap fee when there's a gradual update ongoing.

Agreed on the structure being confusing. Once `ManagedPoolStorageLib` is in then we can refactor this to only do a single write.

nventuro (approved):

PR #1752, opened 2022-09-13
Adding benchmark for relayer multicall

### LINKED ISSUE #936: [relayer] Measure gas cost of returning call results
The current implementation of the relayer returns the return values of all delegatecalls it makes. This can be useful, though I don't think we care about it in most cases. It'd be good to perform a simple measurement of how much extra gas this adds.

```
function multicall(bytes[] calldata data) external payable override nonReentrant returns (bytes[] memory results) {
    results = new bytes[](data.length);
    for (uint256 i = 0; i < data.length; i++) {
        results[i] = _implementation.functionDelegateCall(data[i]);
    }
    ...
```

jubeira:
Closes #936.

Gas usage depends on the amount of bytes returned by the delegate call. Some measurements below:

### Measurements with return values (status quo)

- Gas for input size 0: 30.3k
- Gas for input size 256: 36.3k
- Average gas per call (inputs from length 0 to length 256): 33.5k

### Measurements without return values

For this test, relayer's `multicall` was compiled without considering return values:

```
    function multicall(bytes[] calldata data) external payable override nonReentrant {
        for (uint256 i = 0; i < data.length; i++) {
            _library.functionDelegateCall(data[i]);
        }

        _refundETH();
    }
```

The results were:
- Gas for input size 0: 29.5k
- Gas for input size 256: 34.9k
- Average gas per call (inputs from length 0 to length 256): 32.3k

### Conclusion

Removing the return values saves at least ~800 gas for a call; the amount grows larger with the size of the returned data. Probably not negligible, but not huge either IMO.

### DIFF
--- a/pkg/standalone-utils/contracts/test/MockBaseRelayerLibrary.sol
+++ b/pkg/standalone-utils/contracts/test/MockBaseRelayerLibrary.sol
@@ -15,6 +15,7 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
 
 //import "@balancer-labs/v2-interfaces/contracts/standalone-utils/IBaseRelayerLibrary.sol";
 
@@ -35,5 +36,9 @@ contract MockBaseRelayerLibrary is BaseRelayerLibrary {
 
     function getChainedReferenceValue(uint256 ref) public {
         emit ChainedReferenceValueRead(_getChainedReferenceValue(ref));
+    }
+
+    function bytesTunnel(bytes memory input) public pure returns (bytes memory) {
+        return input;
     }
 }
--- a/pvt/benchmarks/relayer.ts
+++ b/pvt/benchmarks/relayer.ts
@@ -1,0 +1,46 @@
+import { BigNumber, Contract, ethers } from 'ethers';
+
+import { setupEnvironment } from './misc';
+import { bn, printGas } from '@balancer-labs/v2-helpers/src/numbers';
+import { BytesLike } from 'ethers/lib/utils';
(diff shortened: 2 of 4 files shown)


jubeira:
We could also try measuring the cost of chaining multiple calls with different data sizes as well if that adds any value.

pkg/standalone-utils/contracts/test/MockBaseRelayerLibrary.sol
```diff
@@ -36,4 +37,8 @@ contract MockBaseRelayerLibrary is BaseRelayerLibrary {
     function getChainedReferenceValue(uint256 ref) public {
         emit ChainedReferenceValueRead(_getChainedReferenceValue(ref));
     }
+
+    function bytesTunnel(bytes memory input) public pure returns (bytes memory) {
```
nventuro, line 41: :joy:

jubeira: :grimacing:

pvt/benchmarks/relayer.ts
```diff
@@ -0,0 +1,46 @@
...
+  relayerLibrary = await deploy('v2-standalone-utils/MockBaseRelayerLibrary', { args: [vault.address] });
+  relayer = await deploy('v2-standalone-utils/BalancerRelayer', { args: [vault.address, relayerLibrary.address] });
+  let totalGasUsed = bn(0);
+
+  console.log('== Measuring multicall gas usage ==\n');
+  for (let i = 0; i <= maxInputLength; i++) {
+    totalGasUsed = totalGasUsed.add(await testMulticall(i));
+  }
+
+  console.log(`\n# Total gas used: ${printGas(totalGasUsed)}`);
```
nventuro, line 25:
```suggestion
```

nventuro (approved):

PR #1755, opened 2022-09-14
Add `ManagedPoolSwapFees` contract

TomAFrench:
builds on #1748 

This new contract is designed to act as a module responsible for handling any write operations for swap fee data, it also exposes any external getters for swap fee data.

This module has no concept of access control so it's the responsibility of ManagedPool to expose any permissioned functions e.g. `setSwapFeePercentage` and `startGradualSwapFeeChange` necessary to call into the functions in this contract.

There's some slight yuckiness due to the inheritance structure where we need to manually resolve some functions but that should go away as ManagedPool is flattened more.


pkg/pool-weighted/contracts/managed/ManagedPoolSwapFees.sol
```diff
@@ -0,0 +1,127 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
+
+import "@balancer-labs/v2-interfaces/contracts/vault/IBasePool.sol";
+
+import "./ManagedPoolStorageLib.sol";
+
+abstract contract ManagedPoolSwapFees is IBasePool {
```
nventuro, line 22: It's sort of annoying how this makes the contract abstract and greatly increases complexity in the mock.

What if we didn't declare any of the external functions, and left that up for the ManagedPool? We'd then have no need for this module to be able to access state directly (because it is only ever called internally by someone who provides it)

pkg/pool-weighted/test/ManagedPoolSwapFees.test.ts
```diff
@@ -0,0 +1,175 @@
+import { expect } from 'chai';
+import { BigNumber, Contract } from 'ethers';
+import { WEEK, DAY, MINUTE, advanceToTimestamp, currentTimestamp } from '@balancer-labs/v2-helpers/src/time';
+import { fp } from '@balancer-labs/v2-helpers/src/numbers';
+import * as expectEvent from '@balancer-labs/v2-helpers/src/test/expectEvent';
+import { deploy } from '@balancer-labs/v2-helpers/src/contract';
+
+describe('ManagedPoolSwapFees', function () {
```
nventuro, line 8: We're missing a test for the fact that start time is resolved.

nventuro: We're also not testing the gradual behavior. Not sure what the correct strategy should be here - surely we don't want to restest the entire gradual change behavior, but shouldn't we at least have one test where we start a change and query the fee after some time?

pkg/pool-weighted/test/ManagedPool.test.ts
```diff
@@ -733,79 +729,32 @@ describe('ManagedPool', function () {
...
 
-          expect(updateParams.startTime).to.equalWithError(startTime, 0.001);
-          expect(updateParams.endTime).to.equalWithError(endTime, 0.001);
-          expect(updateParams.startSwapFeePercentage).to.equal(START_SWAP_FEE);
-          expect(updateParams.endSwapFeePercentage).to.equal(END_SWAP_FEE);
+        expectEvent.inReceipt(await receipt.wait(), 'GradualSwapFeeUpdateScheduled', {
+          startTime: startTime,
+          endTime: endTime,
+          startSwapFeePercentage: START_SWAP_FEE,
+          endSwapFeePercentage: END_SWAP_FEE,
```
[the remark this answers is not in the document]
TomAFrench: This would be the responsibility of whatever contract implements the external setter.

pkg/pool-weighted/test/ManagedPoolSwapFees.test.ts
TomAFrench: > surely we don't want to retest the entire gradual change behaviour, but shouldn't we at least have one test where we start a change and query the fee after some time?

I'd argue that verifying the return values from `getGradualSwapFeeUpdateParams` is sufficient. The tests for `ManagedPoolStorageLib` show that we'll interpolate between these values if the start and end swap fees are different.

TomAFrench: > We're missing a test for the fact that start time is resolved.

True, I'll add this.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -433,17 +390,9 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
         uint256 startSwapFeePercentage,
         uint256 endSwapFeePercentage
     ) external authenticate whenNotPaused nonReentrant {
-        _validateSwapFeePercentage(startSwapFeePercentage);
-        _validateSwapFeePercentage(endSwapFeePercentage);
-
         startTime = GradualValueChange.resolveStartTime(startTime, endTime);
```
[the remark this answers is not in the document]
TomAFrench: I kinda like the separation of

1. ManagedPoolStorageLib - handles data read and write.
2. ManagedPoolSwapFees - handles validation and events.
3. ManagedPool - handles access control.

pkg/pool-weighted/test/ManagedPool.test.ts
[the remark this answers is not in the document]
TomAFrench: Ah, just noticed you commented on `ManagedPool` rather than `ManagedPoolSwapFees`

nventuro:
>[9bbe2cc](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1755/commits/9bbe2cce310a9c9f19bf98762f2ce810b2a5f2c8) adds 13(!) bytes taking us to 28190 so we might want to step back the inlining when we're closer to the end.

Not sure if joking? Surely you're not concerned about 13 bytes. Also, that commit is great in that it uncovers a lot of hidden state dependencies, which result in warm storage reads. I imagine as the flattening continues and the hooks are unwound, we'll be able to reuse the read state in more and more places.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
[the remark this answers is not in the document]
nventuro: Right, that makes sense to me. But surely `resolveStartTime` is not access control?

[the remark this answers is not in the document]
TomAFrench: Ugh, github windows offsets strike again. I thought you were wanting         `_validateSwapFeePercentage(endSwapFeePercentage)` in `ManagedPoolStorageLib`. I've moved `resolveStartTime` to ManagedPoolSwapFees.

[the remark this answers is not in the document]
nventuro: Sorry :sweat_smile: I'll be sure to be more clear in the future

TomAFrench:
> Not sure if joking?

Lol, 95%. It depends on how desperate the bytecode cramming gets at the end, but not concerned right now.

pkg/pool-weighted/test/ManagedPool.test.ts
```diff
@@ -706,107 +704,91 @@ describe('ManagedPool', function () {
...
-        tokens: poolTokens,
-        weights: poolWeights,
-        owner: owner.address,
-        swapFeePercentage: POOL_SWAP_FEE_PERCENTAGE,
-        poolType: WeightedPoolType.MANAGED_POOL,
-        swapEnabledOnStart: true,
-      };
-      pool = await WeightedPool.create(params);
-    });
+  describe.only('update swap fee gradually', () => {
```
nventuro, line 707: in b4 linter

TomAFrench: I don't see that commit. Please don't leave fake reviews on my PRs.

pkg/pool-weighted/contracts/managed/ManagedPoolSwapFees.sol
```diff
@@ -0,0 +1,100 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
+
+import "./ManagedPoolStorageLib.sol";
+
+abstract contract ManagedPoolSwapFees {
```
nventuro, line 20: Libraries can have events, so this does not need to be a contract anymore.

TomAFrench: Unfortunately this results in ManagedPool's ABI not containing these events anymore. We can leave it as, revert, or add dummy events just for the ABI.

TomAFrench:
Quick check on how much we save by making this library external based on [bbb1080](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1755/commits/bbb108074d7f286ff0514e8010fbbccb88469df6):

Internal library: 28114
External library: 28031 (-83)

pkg/pool-weighted/contracts/managed/ManagedPoolSwapFees.sol
nventuro: I want to cry

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -256,16 +253,8 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
...
-        (
-            uint256 startTime,
-            uint256 endTime,
-            uint256 startSwapFeePercentage,
-            uint256 endSwapFeePercentage
-        ) = ManagedPoolStorageLib.getSwapFeeFields(_getPoolState());
-
-        return
-            GradualValueChange.getInterpolatedValue(startSwapFeePercentage, endSwapFeePercentage, startTime, endTime);
+    function getSwapFeePercentage() public view override returns (uint256) {
```
nventuro, line 256:
```suggestion
    function getSwapFeePercentage() external view override returns (uint256) {
```

pkg/pool-weighted/contracts/managed/vendor/BasePool.sol
```diff
@@ -241,14 +214,14 @@ abstract contract BasePool is
...
-    function _getPoolState() internal view returns (bytes32) {
+    function _getPoolState() internal view virtual returns (bytes32) {
         return _poolState;
     }
 
     /**
      * @dev Inserts data into the pool state storage slot.
      */
-    function _setPoolState(bytes32 newPoolState) internal {
+    function _setPoolState(bytes32 newPoolState) internal virtual {
```
nventuro, line 224:
```suggestion
    function _setPoolState(bytes32 newPoolState) internal {
```

pkg/pool-weighted/contracts/managed/vendor/BasePool.sol
```diff
@@ -241,14 +214,14 @@ abstract contract BasePool is
         _setPaused(false);
     }
 
-    function _getPoolState() internal view returns (bytes32) {
+    function _getPoolState() internal view virtual returns (bytes32) {
```
nventuro, line 217:
```suggestion
    function _getPoolState() internal view returns (bytes32) {
```

pkg/pool-weighted/contracts/test/MockManagedPoolSwapFeesLib.sol
```diff
@@ -0,0 +1,66 @@
...
+import "../managed/ManagedPoolSwapFeesLib.sol";
+
+contract MockManagedPoolSwapFeesLib {
+    bytes32 private _poolState;
+
+    // Helpers to decode Pool state
+
+    function getSwapFeePercentage() external view returns (uint256) {
+        return ManagedPoolStorageLib.getSwapFeePercentage(_poolState);
+    }
```
nventuro, line 27: Wouldn't it be better to pass and receive the state, instead of having it be on storage? Not a huge deal either way

TomAFrench:
Let's get #1774 in first

pvt/helpers/src/test/expectEvent.ts
```diff
@@ -54,7 +54,8 @@ export function inIndirectReceipt(
     .map((log) => {
       try {
         return emitter.parseLog(log);
-      } catch {
+      } catch (e) {
+        console.log(e);
```
nventuro, line 58: Leftover?

nventuro (approved):

pkg/pool-weighted/contracts/test/MockManagedPoolSwapFeesLib.sol
TomAFrench: This is a hack so that we don't need to be smart about decoding it in the tests. This will go away with #1776

PR #1757, opened 2022-09-14
Flatten `_isOnlyOwnerAction`

TomAFrench:
*squash*

Chucking all these tests into `ManagedPool.test.ts` is going to be pretty unwieldy. Let's use break out on different subunits into the `test/managed` dir


pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -1079,29 +1079,22 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
...
+            (actionId == getActionId(ManagedPool.setSwapFeePercentage.selector)) ||
             (actionId == getActionId(ManagedPool.addAllowedAddress.selector)) ||
             (actionId == getActionId(ManagedPool.removeAllowedAddress.selector)) ||
             (actionId == getActionId(ManagedPool.setMustAllowlistLPs.selector)) ||
             (actionId == getActionId(ManagedPool.addToken.selector)) ||
             (actionId == getActionId(ManagedPool.removeToken.selector)) ||
             (actionId == getActionId(ManagedPool.setManagementSwapFeePercentage.selector)) ||
             (actionId == getActionId(ManagedPool.setManagementAumFeePercentage.selector)) ||
+            (actionId == getActionId(BasePool.setAssetManagerPoolConfig.selector)) ||
             super._isOwnerOnlyAction(actionId);
```
nventuro, line 1098: We don't actually need to call super any more now, do we?

nventuro (approved):

nventuro:
The composable tests blew up, but that cannot be because of these changes. Did we just find a new source of flakyness?

pkg/pool-weighted/contracts/managed/ManagedPool.sol
TomAFrench: It was to remind us of why we're doing this.

TomAFrench:
> Did we just find a new source of flakyness?

I hope not.

PR #1758, opened 2022-09-14
Remove `BaseMinimalSwapInfoPool`

TomAFrench:
*splosh*

And then there were two.


nventuro (approved):
Away it goes! Albeit to the wrong place :joy:

TomAFrench:
> Albeit to the wrong place joy

It'll need to get `_poolState` at some point so needs to know something of ManagedPool so I'm trying to bring it closer to that. We can be smart later.

PR #1763, opened 2022-09-14
[Batch Relayer V4] Using same storage slot for read-only and chained refs

### LINKED ISSUE #1762: Remove prefix when calculating chained reference slot
We want to read chained references more than once, mixing temporary and read-only ones. E.g.:

- swap(output: 15)
- approve(read only ref: 15)
- join(temp ref: 15)

This is currently not possible as the prefix is used to calculate the slot, so a read-only ref and a temp ref will point to different slots.

jubeira:
Closes #1762.


nventuro (approved):
@johngrantuk this is the fix I mentioned earlier

johngrantuk:
> @johngrantuk this is the fix I mentioned earlier

Thanks. Please let me know when you get a chance to redeploy on Goerli.

PR #1765, opened 2022-09-14
Removing return values from VaultActions

### LINKED ISSUE #1753: Remove return values from relayer library operations
As stated in https://github.com/balancer-labs/balancer-v2-monorepo/issues/936#issue-1025753042, in most cases we are not using the return values.

We do want to preserve the mechanism in `multicall` to be able to peek chained references (see #1509), but we should remove the return values from the functions that have still them for consistency.

jubeira:
Closes #1753.

### DIFF
--- a/pkg/standalone-utils/contracts/relayer/VaultActions.sol
+++ b/pkg/standalone-utils/contracts/relayer/VaultActions.sol
@@ -48,7 +48,7 @@ abstract contract VaultActions is IBaseRelayerLibrary {
         uint256 deadline,
         uint256 value,
         uint256 outputReference
-    ) external payable returns (uint256) {
+    ) external payable {
         require(funds.sender == msg.sender || funds.sender == address(this), "Incorrect sender");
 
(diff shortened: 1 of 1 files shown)


pkg/standalone-utils/contracts/relayer/VaultActions.sol
```diff
@@ -95,8 +93,6 @@ abstract contract VaultActions is IBaseRelayerLibrary {
             // parameter.
             _setChainedReferenceValue(outputReferences[i].key, Math.abs(results[outputReferences[i].index]));
         }
-
-        return results;
     }
```
jubeira, line 96: We could also check whether the output reference is required before allocating memory for the results instead of always allocating it.

nventuro: Hm, might be a bit overkill. Also, solc likely checks the return data and copies it to memory, even if we don't use it. You can write a simple test to try out this hypothesis.

nventuro (approved):

PR #1768, opened 2022-09-14
Automatically renormalize on gradual weight changes

TomAFrench:
This avoids the issue we identified a while back where a Pool owner can manipulate `_denormWeightSum` by manipulating the weights in combination with adding or removing tokens

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPool.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPool.sol
@@ -187,22 +187,23 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
             _tokenState[token] = ManagedPoolTokenLib.setTokenScalingFactor(bytes32(0), token);
         }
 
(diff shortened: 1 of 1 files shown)


pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -1057,7 +1062,7 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
                 _tokenState[token],
                 startWeights[i],
                 endWeight,
-                denormWeightSum
+                FixedPoint.ONE
```
nventuro, line 1065: Hm, interesting. This does seem correct. It'd be good to review the existing tests to get a sense of what the current coverage of all of this looks like.

TomAFrench:
One thing I like about this is that we don't need to pass in a `denormWeightSum` if we push this out to an abstract contract/library. We just need to make sure to add a `_denormWeightSum = FixedPoint.ONE` after calls to this function.

nventuro (approved):

PR #1769, opened 2022-09-14
Cache token weight in _onSwapGivenIn/Out

TomAFrench:
This saves 6 warm reads per swap.

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPool.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPool.sol
@@ -822,12 +822,15 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
     ) internal virtual override returns (uint256) {
         _require(getSwapEnabled(), Errors.SWAPS_DISABLED);
 
+        uint256 tokenInWeight = _getNormalizedWeight(swapRequest.tokenIn);
+        uint256 tokenOutWeight = _getNormalizedWeight(swapRequest.tokenOut);
+
         // balances (and swapRequest.amount) are already upscaled by BaseWeightedPool.onSwap
         uint256 amountOut = WeightedMath._calcOutGivenIn(
             currentBalanceTokenIn,
-            _getNormalizedWeight(swapRequest.tokenIn),
+            tokenInWeight,
             currentBalanceTokenOut,
-            _getNormalizedWeight(swapRequest.tokenOut),
+            tokenOutWeight,
             swapRequest.amount
         );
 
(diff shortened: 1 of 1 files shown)


TomAFrench:
6856979 takes us from 28544B back down to 28102B

pkg/pool-weighted/contracts/managed/vendor/BaseWeightedPool.sol
```diff
@@ -75,31 +75,35 @@ abstract contract BaseWeightedPool is IMinimalSwapInfoPool, BasePool {
     /**
      * @dev Returns the normalized weight of `token`. Weights are fixed point numbers that sum to FixedPoint.ONE.
      */
-    function _getNormalizedWeight(IERC20 token) internal view virtual returns (uint256);
+    function _getNormalizedWeight(IERC20 token, uint256 weightChangeProgress) internal view virtual returns (uint256);
+
+    /**
+     * @notice Returns all normalized weights, in the same order as the Pool's tokens.
+     */
+    function getNormalizedWeights() public view returns (uint256[] memory) {
```
nventuro, line 83:
```suggestion
    function getNormalizedWeights() external view returns (uint256[] memory) {
```

So we can be sure we never call this version ourselves

pkg/pool-weighted/contracts/managed/vendor/BaseWeightedPool.sol
```diff
@@ -75,31 +75,35 @@ abstract contract BaseWeightedPool is IMinimalSwapInfoPool, BasePool {
...
     /**
      * @dev Returns all normalized weights, in the same order as the Pool's tokens.
      */
-    function _getNormalizedWeights() internal view virtual returns (uint256[] memory);
+    function _getNormalizedWeights(IERC20[] memory tokens) internal view virtual returns (uint256[] memory);
 
     /**
      * @dev Returns the current value of the invariant.
      */
     function getInvariant() public view returns (uint256) {
```
nventuro, line 96:
```suggestion
    function getInvariant() external view returns (uint256) {
```

Same reasoning

pkg/pool-weighted/contracts/managed/vendor/BaseWeightedPool.sol
```diff
@@ -196,7 +200,7 @@ abstract contract BaseWeightedPool is IMinimalSwapInfoPool, BasePool {
         uint256[] memory scalingFactors,
         bytes memory userData
     ) internal virtual override returns (uint256, uint256[] memory) {
-        uint256[] memory normalizedWeights = _getNormalizedWeights();
+        uint256[] memory normalizedWeights = getNormalizedWeights();
```
nventuro, line 203: Ah, we _do_ call it ourselves. Bummer

nventuro (approved):
The dependency explicitooooor

pkg/pool-weighted/contracts/managed/vendor/BaseWeightedPool.sol
TomAFrench: We do actually call this ourselves in cases where we don't have the token array onhand. Prior to [6856979](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1769/commits/6856979b4a2d85d22cb84ed698e6578c03a3846e) we did have this external but it ballooned bytecode size.

PR #1772, opened 2022-09-15
Avoid double zero duration check in `calculateValueChangeProgress`

TomAFrench:
Noticed this while mucking around with fuzz tests on the foundry branch. Drops gas cost by 23.


TomAFrench:
Bundled in some unnecessary safemath as we know that delta is bounded so the result can't over/underflow

nventuro (approved):
Nice

PR #1773, opened 2022-09-15
Fix deployments package

nventuro:
We can't import hardhat at the top level for this module, because we use it in places were hardhat hasn't loaded yet.


TomAFrench:
Can't we use a non-hardhat import of ethers for this? We just want a random address-y string at the end of the day.

TomAFrench (approved):

PR #1774, opened 2022-09-15
Pull _poolState up into ManagedPool

TomAFrench:
It makes no sense for `_poolState` to live in `BasePool` as it's so strongly intertwined with ManagedPool. This results in us needing fudges like `_getPoolState` and `_setPoolState`.

This PR pulls `_poolState` up into `ManagedPool` so we can get rid of `_getPoolState` and `_setPoolState`. BasePool then just defines virtual getters for the fields it needs.

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPool.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPool.sol
@@ -248,7 +248,7 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
      * @dev Computes the current swap fee percentage, which can change every block if a gradual swap fee
      * update is in progress.
      */
-    function getSwapFeePercentage() public view virtual override returns (uint256) {
+    function getSwapFeePercentage() public view override returns (uint256) {
         return ManagedPoolStorageLib.getSwapFeePercentage(_poolState);
     }
 
@@ -306,7 +306,7 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
         _setSwapFeePercentage(swapFeePercentage);
     }
 
-    function _setSwapFeePercentage(uint256 swapFeePercentage) internal virtual override {
+    function _setSwapFeePercentage(uint256 swapFeePercentage) internal override {
         _validateSwapFeePercentage(swapFeePercentage);
 
         _poolState = ManagedPoolStorageLib.setSwapFeeData(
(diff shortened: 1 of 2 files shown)


pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -1223,6 +1217,26 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
 
     // Recovery Mode
 
+    /**
+     * @notice Returns whether the pool is in Recovery Mode.
+     */
+    function inRecoveryMode() public view override returns (bool) {
```
nventuro, line 1223:
```suggestion
    function inRecoveryMode() external view override returns (bool) {
```

Same for the rest

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -1146,7 +1140,7 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
             (actionId == getActionId(BasePool.setAssetManagerPoolConfig.selector));
     }
 
-    function _getMaxSwapFeePercentage() internal pure virtual override returns (uint256) {
+    function _getMaxSwapFeePercentage() internal pure override returns (uint256) {
         return _MAX_SWAP_FEE_PERCENTAGE;
     }
```
nventuro, line 1145: I think we can begin getting rid of these and make them just be constants

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -237,6 +243,15 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
         return !getMustAllowlistLPs() || _allowedAddresses[member];
     }
 
+    /**
+     * @notice Returns the current value of the swap fee percentage.
+     * @dev Computes the current swap fee percentage, which can change every block if a gradual swap fee
+     * update is in progress.
+     */
+    function getSwapFeePercentage() public view override returns (uint256) {
```
nventuro, line 251:
```suggestion
    function getSwapFeePercentage() external view override returns (uint256) {
```

nventuro (approved):

pkg/pool-weighted/contracts/managed/ManagedPool.sol
TomAFrench: This is necessary for now due to `_addSwapFeeAmount` and `_subtractSwapFeeAmount`

pkg/pool-weighted/contracts/managed/ManagedPool.sol
TomAFrench: This breaks `_ensureInRecoveryMode()`

PR #1776, opened 2022-09-15
Replace ManagedPoolStorageLib tests with foundry fuzzing

TomAFrench:
This PR moves the tests for ManagedPoolStorageLib over from hardhat to foundry.

We'll likely want to move the `Hevm` interface somewhere centrally along with any other solidity test helpers. We also probably want to move all of our mock contracts out of `contracts/test` into `contract/mock`.


pkg/pool-weighted/contracts/test/ManagedPoolStorageLib.t.sol
```diff
@@ -0,0 +1,198 @@
...
+    uint256 private constant _SWAP_ENABLED_OFFSET = _SWAP_FEE_END_PCT_OFFSET + _SWAP_FEE_PCT_WIDTH;
+    uint256 private constant _MUST_ALLOWLIST_LPS_OFFSET = _SWAP_ENABLED_OFFSET + 1;
+    uint256 private constant _RECOVERY_MODE_OFFSET = _MUST_ALLOWLIST_LPS_OFFSET + 1;
+
+    uint256 private constant _TIMESTAMP_WIDTH = 32;
+    uint256 private constant _SWAP_FEE_PCT_WIDTH = 62;
+
+    uint256 private constant _MAX_SWAP_FEE = 1 << _SWAP_FEE_PCT_WIDTH;
+
+    function clearWordAtPosition(
```
EndymionJkb, line 40: `clearWordAtPosition` and `assertOtherStateUnchanged` would be generally useful. Maybe put them up with WordCodec (solidity-utils/helpers/FoundryHelpers)?

pkg/pool-weighted/contracts/test/Hevm.sol
```diff
@@ -0,0 +1,110 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
+
+// solhint-disable
+
+interface Hevm {
```
EndymionJkb, line 7: Should this just go up in /interfaces?

pkg/pool-weighted/contracts/test/ManagedPoolStorageLib.t.sol
```diff
@@ -0,0 +1,198 @@
...
+        assertEq(recoveredEndTime, endTime);
+    }
+
+    function testWeightChangeProgress(
+        bytes32 poolState,
+        uint32 startTime,
+        uint32 endTime,
+        uint32 now
+    ) public {
+        hevm.warp(now);
```
nventuro: Ah this is also a good idea

pkg/pool-weighted/contracts/test/ManagedPoolStorageLib.t.sol
```diff
@@ -0,0 +1,198 @@
...
+
+    function testSwapFeeData(
+        bytes32 poolState,
+        uint32 startTime,
+        uint32 endTime,
+        uint64 startSwapFeePercentage,
+        uint64 endSwapFeePercentage
+    ) public {
+        hevm.assume(startSwapFeePercentage < _MAX_SWAP_FEE);
+        hevm.assume(endSwapFeePercentage < _MAX_SWAP_FEE);
```
nventuro: Don't we also have a minimum fee?

nventuro: Also this should be <=

pkg/pool-weighted/contracts/test/Hevm.sol
TomAFrench: Looks like I was being dumb and not importing the right test helper so we don't need this at all.

pkg/pool-weighted/contracts/test/ManagedPoolStorageLib.t.sol
```diff
@@ -0,0 +1,198 @@
...
+        bytes32 clearedOldState = clearWordAtPosition(oldPoolState, offset, bitLength);
+        bytes32 clearedNewState = clearWordAtPosition(newPoolState, offset, bitLength);
+        assertEq(clearedOldState, clearedNewState);
+    }
+
+    function testSetRecoveryMode(bytes32 poolState, bool enabled) public {
+        bytes32 newPoolState = ManagedPoolStorageLib.setRecoveryModeEnabled(poolState, enabled);
+        assertOtherStateUnchanged(poolState, newPoolState, _RECOVERY_MODE_OFFSET, 1);
+
+        assertTrue(ManagedPoolStorageLib.getRecoveryModeEnabled(newPoolState) == enabled);
```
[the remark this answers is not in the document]
TomAFrench: No `function assertEq(bool a, bool b)` on DSTest...

pkg/pool-weighted/contracts/test/ManagedPoolStorageLib.t.sol
```diff
@@ -0,0 +1,198 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+
+import { DSTest } from "ds-test/test.sol";
+import "./Hevm.sol";
+
+import "../managed/ManagedPoolStorageLib.sol";
+
+contract ManagedPoolStorageLibTest is DSTest {
```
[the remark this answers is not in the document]
EndymionJkb: Are we also creating test/mock directories? (Might leave the top level /test empty though.)

pkg/pool-weighted/contracts/test/ManagedPoolStorageLib.t.sol
TomAFrench: Not for the purposes of `ManagedPoolStorageLib`. `_MAX_SWAP_FEE` here is over 100% as we're just defining the range of inputs which can be written into the slot.

pkg/pool-weighted/contracts/test/ManagedPoolStorageLib.t.sol
TomAFrench: Yeah, we'll need them for `ManagedPoolTokenLib`, etc. I'll move these once we add the second test contract which uses them.

nventuro (approved):

PR #1778, opened 2022-09-15
Inline `getSwapEnabled()` inside swap hooks

TomAFrench:
In these cases we need `_poolState` anyway so we can save a read by inlining.


pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -801,12 +801,17 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
         uint256 currentBalanceTokenIn,
         uint256 currentBalanceTokenOut
     ) internal override returns (uint256) {
-        _require(getSwapEnabled(), Errors.SWAPS_DISABLED);
+        uint256 tokenInWeight;
+        uint256 tokenOutWeight;
+        {
```
nventuro, line 806: It this because of stack too deep? Let's add a comment if so

nventuro (approved):

PR #1779, opened 2022-09-15
Update `_startGradualWeightChange` to return updated pool state

TomAFrench:
`_startGradualSwapFeeChange` returns the updated pool state whereas `_startGradualWeightChange` writes it to storage directly. If we are to create another library (`ManagedPoolWeightsLib.sol`) then we need to make this change.


pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -190,23 +190,29 @@ contract ManagedPool is BaseWeightedPool, ProtocolFeeCache, ReentrancyGuard, ICo
         // Initialize the denormalized weight sum to ONE. This value can only be changed by adding or removing tokens.
         _denormWeightSum = FixedPoint.ONE;
 
+        bytes32 poolState;
```
nventuro, line 193: It feels like this is due some more ceremony: we're initializing all `poolState` fields, and should not forget any. We could at least list them here

nventuro (approved):

PR #1782, opened 2022-09-15
Add some random fuzz tests

nventuro:
I also wanted to have some fun fuzzing


pkg/solidity-utils/test/foundry/InputHelpers.t.sol
```diff
@@ -0,0 +1,39 @@
...
+        } else {
+            bool unsorted = false;
+            for (uint256 i = 1; i < addresses.length; ++i) {
+                if (addresses[i - 1] >= addresses[i]) {
+                    unsorted = true;
+                }
+            }
+
+            assertTrue(unsorted);
+        }
```
TomAFrench, line 37:
```suggestion
        bool unsorted = false;
        for (uint256 i = 1; i < addresses.length; ++i) {
            if (addresses[i - 1] >= addresses[i]) {
                unsorted = true;
                break;
            }
        }

        if (unsorted) vm.expectRevert("BAL#101");
        InputHelpers.ensureArrayIsSorted(addresses);
```

pkg/solidity-utils/test/foundry/WordCodec.t.sol
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.7.0;
+
+import "forge-std/Test.sol";
+
+import "../../contracts/helpers/WordCodec.sol";
+
+contract WordCodecTest is Test {
+    function testEncodeUint255Bits(uint256 input) external {
+        vm.assume(input < (2 << (255 - 1)));
```
TomAFrench, line 10:
```suggestion
        vm.assume(input < (1 << 255) - 1));
```
`input == 0` here.

pkg/solidity-utils/test/foundry/WordCodec.t.sol
```diff
@@ -0,0 +1,67 @@
...
+    )
+        private
+        returns (
+            uint256,
+            uint8,
+            uint256
+        )
+    {
+        vm.assume(bits > 0);
+        vm.assume(input < (2 << (255 - 1)));
```
TomAFrench, line 56: Same here.

pkg/solidity-utils/test/foundry/InputHelpers.t.sol
nventuro: Ah nice. Also it seems the docs for `vm.expectRevert` are misleanding - we don't actually need a call.

PR #1785, opened 2022-09-16
Replace maths functions with branches with assembly

TomAFrench:
These functions are far enough away from the business logic that I'm comfortable with making them less readable like this.

Most of these optimisations are to do with removing branches from the code to avoid JUMPs. As we're breaking type safety in assembly we can treat boolean values as a uint256 and then do maths with it directly.

One of the functions (`FixedPoint.mulUp`) goes _up_ in cost based on the snapshot. This is due to the new version being worse when `a == 0`, which I'd say is worth it for being cheaper the rest of the time.


EndymionJkb:
This is great savings, but I think we should retain the original code in comments above it. That way we're not sacrificing readability either. (And people can independently verify the correctness of the assembly.)

pkg/solidity-utils/test/foundry/FixedPoint.t.sol
```diff
@@ -0,0 +1,58 @@
...
+            FixedPoint.divDown(a, b);
+        } else if (a != 0 && (a * FixedPoint.ONE) / FixedPoint.ONE != a) {
+            vm.expectRevert("BAL#005"); // DIV_INTERNAL
+            FixedPoint.divDown(a, b);
+        } else {
+            uint256 referenceDivDown = a == 0 ? 0 : (a * FixedPoint.ONE) / b;
+            uint256 divDown = FixedPoint.divDown(a, b);
+
+            assertEq(divDown, referenceDivDown);
+        }
```
nventuro, line 41:
```suggestion
        if (b == 0) {
            vm.expectRevert("BAL#004"); // ZERO_DIVISION
        } else if (a != 0 && (a * FixedPoint.ONE) / FixedPoint.ONE != a) {
            vm.expectRevert("BAL#005"); // DIV_INTERNAL
        }

        uint256 divDown = FixedPoint.divDown(a, b);
        uint256 referenceDivDown = a == 0 ? 0 : (a * FixedPoint.ONE) / b;

        assertEq(divDown, referenceDivDown);
```

pkg/solidity-utils/test/foundry/Math.t.sol
```diff
@@ -11,19 +11,30 @@ contract MathTest is Test {
...
     function testMax(uint256 a, uint256 b) external {
-        uint256 referenceMax = (a < b) ? b : a;
         uint256 max = Math.max(a, b);
+        assertGe(max, a);
+        assertGe(max, b);
+        assertTrue((max == a) || (max == b));
 
+        uint256 referenceMax = (a < b) ? b : a;
         assertEq(max, referenceMax);
     }
```
nventuro, line 29: I would've split this into two tests - one comparing against the ref, and one testing props

pkg/solidity-utils/test/foundry/FixedPoint.t.sol
TomAFrench: I don't think I agree with this. IMO, `vm.expectRevert` should be kept as close as possible to the call we expect to blow up.

nventuro: Fair enough

nventuro (approved):
:scared:

PR #1786, opened 2022-09-16
Misc changes to ManagedPool/BasePool

TomAFrench:
I'd recommend looking through commit by commit.


TomAFrench:
Looking at test failures now

nventuro (approved):

TomAFrench:
I've ended up just removing `setAssetManagerPoolConfig` from all BasePool stuff as it's causing interface issues and we don't use it elsewhere anyway.

PR #1790, opened 2022-09-16
Make base pool better

TomAFrench:


pkg/pool-weighted/contracts/managed/vendor/BasePool.sol
```diff
@@ -272,30 +239,22 @@ abstract contract BasePool is IBasePool, BasePoolAuthorization, BalancerPoolToke
...
             return (amountsIn, new uint256[](balances.length));
         } else {
-            _upscaleArray(balances, scalingFactors);
             (uint256 bptAmountOut, uint256[] memory amountsIn) = _onJoinPool(
                 poolId,
                 sender,
                 recipient,
                 balances,
                 lastChangeBlock,
                 inRecoveryMode() ? 0 : protocolSwapFeePercentage, // Protocol fees are disabled while in recovery mode
```
nventuro, line 250: We can change the internal handler to not receive data we know we don't care about, such as the fee percentage, recipient, or last change block

nventuro (approved):
Lovely

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -1045,13 +999,6 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
         return _totalTokensCache;
     }
 
-    function _getTokenData(IERC20 token) private view returns (bytes32 tokenData) {
```
EndymionJkb, line 1048: `getTokenData` was just there for the error checking. If you call getScalingFactor with an invalid token,  it would return 1 (kind of ok). But if you call getTokenWeight with a zero bytes32, it would return zero. It would probably revert downstream anyway, but it might be less obvious why.

TomAFrench: This is fine, we never call into this function with a user provided token so this is fine.

PR #1792, opened 2022-09-19
Circuit Breaker 

EndymionJkb:
Starting with the storage, this sets/retrieves circuit breaker data and performs all associated error checks and computations (e.g., bounds with changing weights).

I have a lot of simulations to go with this (see #1399 ), and have some concerns, but this is the current state of the design.

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPoolCircuitBreakerLib.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPoolCircuitBreakerLib.sol
@@ -115,7 +115,7 @@ library ManagedPoolCircuitBreakerLib {
         returns (uint256, uint256)
     {
         // Retrieve the reference bptPrice and weight factors, stored at the time the circuit breaker was set.
-        uint256 bptPrice = circuitBreakerState.decodeUint(_REFERENCE_BPT_PRICE_OFFSET, _BPT_PRICE_WIDTH);
+        uint256 referenceBptPrice = circuitBreakerState.decodeUint(_REFERENCE_BPT_PRICE_OFFSET, _BPT_PRICE_WIDTH);
         uint256 referenceWeightFactor = circuitBreakerState
             .decodeUint(_WEIGHT_FACTOR_OFFSET, _WEIGHT_FACTOR_WIDTH)
             .decompress(_WEIGHT_FACTOR_WIDTH);
@@ -124,11 +124,15 @@ library ManagedPoolCircuitBreakerLib {
             // If the weight factor hasn't changed since the circuit breaker was set, we can use the precomputed
             // boundary expressions.
             return (
-                bptPrice.mulDown(
-                    circuitBreakerState.decodeUint(_LOWER_BOUND_RATIO_CACHE_OFFSET, _BOUND_RATIO_CACHE_WIDTH)
+                referenceBptPrice.mulDown(
+                    circuitBreakerState
+                        .decodeUint(_LOWER_BOUND_RATIO_CACHE_OFFSET, _BOUND_RATIO_CACHE_WIDTH)
+                        .decompress(_BOUND_RATIO_CACHE_WIDTH, _MAX_BOUND_PERCENTAGE)
                 ),
-                bptPrice.mulUp(
-                    circuitBreakerState.decodeUint(_UPPER_BOUND_RATIO_CACHE_OFFSET, _BOUND_RATIO_CACHE_WIDTH)
+                referenceBptPrice.mulUp(
+                    circuitBreakerState
+                        .decodeUint(_UPPER_BOUND_RATIO_CACHE_OFFSET, _BOUND_RATIO_CACHE_WIDTH)
+                        .decompress(_BOUND_RATIO_CACHE_WIDTH, _MAX_BOUND_PERCENTAGE)
                 )
             );
         } else {
@@ -149,7 +153,7 @@ library ManagedPoolCircuitBreakerLib {
                 currentWeightFactor
             );
 
-            return (bptPrice.mulDown(lowerBoundRatioCache), bptPrice.mulUp(upperBoundRatioCache));
+            return (referenceBptPrice.mulDown(lowerBoundRatioCache), referenceBptPrice.mulUp(upperBoundRatioCache));
         }
     }
 
--- a/pkg/pool-weighted/contracts/managed/ManagedPoolCircuitBreakerLib.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPoolCircuitBreakerLib.sol
@@ -1,0 +1,249 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
+
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
+
+import "../lib/ValueCompression.sol";
+
+/**
+ * @title Managed Pool Circuit Breaker Library
+ * @notice Library for storing and manipulating state related to circuit breakers in ManagedPool.
+ * @dev BPT price = supply / balance / weight. The lower and upper bounds are percentages.
+ * For instance, a lower bound of 0.8 means the circuit breaker will trip if the token price
+ * drops more than 20% (relative to other tokens). An upper bound ratio of 2.0 means it will trip
+ * if the price more than doubles.
+ */
+library ManagedPoolCircuitBreakerLib {
+    using ValueCompression for uint256;
+    using FixedPoint for uint256;
+    using WordCodec for bytes32;
+
+    struct CircuitBreakerParams {
+        uint256 referenceBptPrice;
+        uint256 currentWeightFactor;
+        uint256 lowerBoundPercentage;
+        uint256 upperBoundPercentage;
+    }
+
+    // Store circuit breaker information per token
+    // [  24 bits |  24 bits |   16 bits   |   16 bits   |    64 bits    |  112 bits |
+    // [ UB ratio | LB ratio | upper bound | lower bound | weight factor | ref price |
+    // |MSB                                                                       LSB|
+    uint256 private constant _REFERENCE_BPT_PRICE_OFFSET = 0;
+    uint256 private constant _WEIGHT_FACTOR_OFFSET = _REFERENCE_BPT_PRICE_OFFSET + _BPT_PRICE_WIDTH;
+    uint256 private constant _LOWER_BOUND_PCT_OFFSET = _WEIGHT_FACTOR_OFFSET + _WEIGHT_FACTOR_WIDTH;
+    uint256 private constant _UPPER_BOUND_PCT_OFFSET = _LOWER_BOUND_PCT_OFFSET + _BOUND_PERCENTAGE_WIDTH;
+    uint256 private constant _LOWER_BOUND_RATIO_CACHE_OFFSET = _UPPER_BOUND_PCT_OFFSET + _BOUND_PERCENTAGE_WIDTH;
+    uint256 private constant _UPPER_BOUND_RATIO_CACHE_OFFSET = _LOWER_BOUND_RATIO_CACHE_OFFSET +
+        _BOUND_RATIO_CACHE_WIDTH;
+
+    uint256 private constant _BOUND_PERCENTAGE_WIDTH = 16;
+    uint256 private constant _BOUND_RATIO_CACHE_WIDTH = 24;
+    uint256 private constant _WEIGHT_FACTOR_WIDTH = 64;
+    uint256 private constant _BPT_PRICE_WIDTH = 112;
+
+    // We are compressing the ratios to 64-bit resolution, but the upper bound would normally be > 1,
+    // so we cannot use the default compression. For consistency, use the same maximum uncompressed value,
+    // even though we expect the lower bounds to be < 1.
+    uint256 private constant _MAX_BOUND_PERCENTAGE = 10e18; // FP 10
+
+    /**
+     * @notice Returns the reference BPT price, reference weight factor, and upper and lower bounds for a given token.
+     * @dev If a bound value is zero, it means there is no circuit breaker in that direction for the given token.
+     * @param circuitBreakerState - The bytes32 state of the token of interest.
+     */
+    function getCircuitBreakerFields(bytes32 circuitBreakerState)
+        internal
+        pure
+        returns (
+            uint256,
+            uint256,
+            uint256,
+            uint256
+        )
+    {
+        return (
+            circuitBreakerState.decodeUint(_REFERENCE_BPT_PRICE_OFFSET, _BPT_PRICE_WIDTH),
+            circuitBreakerState.decodeUint(_WEIGHT_FACTOR_OFFSET, _WEIGHT_FACTOR_WIDTH).decompress(
+                _WEIGHT_FACTOR_WIDTH
+            ),
+            circuitBreakerState.decodeUint(_LOWER_BOUND_PCT_OFFSET, _BOUND_PERCENTAGE_WIDTH).decompress(
+                _BOUND_PERCENTAGE_WIDTH,
+                _MAX_BOUND_PERCENTAGE
+            ),
+            circuitBreakerState.decodeUint(_UPPER_BOUND_PCT_OFFSET, _BOUND_PERCENTAGE_WIDTH).decompress(
+                _BOUND_PERCENTAGE_WIDTH,
+                _MAX_BOUND_PERCENTAGE
+            )
+        );
+    }
+
+    /**
+     * @notice Returns the dynamic upper and lower bounds for a given token, at the current weight.
+     * @dev If a bound value is zero, it means there is no circuit breaker in that direction for the given token.
+     * The current boundary is given as: referenceBptPrice * (raw boundary percentage)**(currentWeightFactor).
+     *
+     * The weight factor calculation attempts to isolate changes in the balance due to arbers responding to external
+     * prices, and internal price changes from moving weights. There is a non-linear relationship between "spot" price
+     * changes and BPT price changes, so this calculation transforms one into the other.
+     *
+     * The raw thresholds are simple percentages: 0.8 means "tolerate a 20% drop in external price." To check the
+     * circuit breaker at runtime, we need to transform that into a corresponding BPT price ratio (relative to the
+     * BPT price at the time the breaker is set), such that the final boundary to be checked is simply the original
+     * reference BPT price multiplied by this conversion ratio.
+     * @param circuitBreakerState - The bytes32 state of the token of interest.
+     * @param currentWeightFactor - The combined weight of all other tokens: (_denormWeightSum - weight of this token)
+     * @return - lower and upper BPT price bounds, which can be directly compared against the current BPT price.
+     */
+    function getCurrentCircuitBreakerBounds(bytes32 circuitBreakerState, uint256 currentWeightFactor)
+        internal
+        pure
+        returns (uint256, uint256)
+    {
+        // Retrieve the reference bptPrice and weight factors, stored at the time the circuit breaker was set.
+        uint256 bptPrice = circuitBreakerState.decodeUint(_REFERENCE_BPT_PRICE_OFFSET, _BPT_PRICE_WIDTH);
+        uint256 referenceWeightFactor = circuitBreakerState
+            .decodeUint(_WEIGHT_FACTOR_OFFSET, _WEIGHT_FACTOR_WIDTH)
+            .decompress(_WEIGHT_FACTOR_WIDTH);
+
+        if (referenceWeightFactor == currentWeightFactor) {
+            // If the weight factor hasn't changed since the circuit breaker was set, we can use the precomputed
+            // boundary expressions.
+            return (
+                bptPrice.mulDown(
+                    circuitBreakerState.decodeUint(_LOWER_BOUND_RATIO_CACHE_OFFSET, _BOUND_RATIO_CACHE_WIDTH)
+                ),
+                bptPrice.mulUp(
+                    circuitBreakerState.decodeUint(_UPPER_BOUND_RATIO_CACHE_OFFSET, _BOUND_RATIO_CACHE_WIDTH)
+                )
+            );
+        } else {
+            // Something has changed - either the weight of the token, or the total weight (e.g., another token was
+            // added or removed), so we must retrieve the raw bounds and do the full calculation.
+            uint256 lowerBoundPercentage = circuitBreakerState
+                .decodeUint(_LOWER_BOUND_PCT_OFFSET, _BOUND_PERCENTAGE_WIDTH)
+                .decompress(_BOUND_PERCENTAGE_WIDTH, _MAX_BOUND_PERCENTAGE);
+
+            uint256 upperBoundPercentage = circuitBreakerState
+                .decodeUint(_UPPER_BOUND_PCT_OFFSET, _BOUND_PERCENTAGE_WIDTH)
+                .decompress(_BOUND_PERCENTAGE_WIDTH, _MAX_BOUND_PERCENTAGE);
+
+            // Use these ratios to onvert raw percentage bounds to BPT price bounds.
+            (uint256 lowerBoundRatioCache, uint256 upperBoundRatioCache) = _getBoundaryConversionRatios(
+                lowerBoundPercentage,
+                upperBoundPercentage,
+                currentWeightFactor
+            );
+
+            return (bptPrice.mulDown(lowerBoundRatioCache), bptPrice.mulUp(upperBoundRatioCache));
+        }
+    }
+
+    /**
+     * @notice Sets the reference BPT price, and upper and lower bounds for a token.
+     * @dev If a bound is zero, it means there is no circuit breaker in that direction for the given token.
+     * @param circuitBreakerState - The bytes32 state of the token of interest.
+     * @param params - CircuitBreakerParams (introduced in part to address stack issues) has the following components:
+     * - referenceBptPrice: The BptPrice of the token at the time the circuit breaker is set. The BPT Price
+     *   of a token is generally given by: supply * weight / balance.
+     * - currentWeightFactor: This is _denormWeightSum - currentWeight of the token.
+     * - lowerBoundPercentage: The value of the lower bound. Any operation that would cause the effective
+     *   BPT Price to fall below lowerBoundRatio * referenceBptPrice should revert.
+     * - upperBoundPercentage: The value of the upper bound. If non-zero, any operation that would cause the
+     *   effective BPT Price to rise above upperBoundRatio * referenceBptPrice should revert.
+     */
+    function setCircuitBreakerFields(bytes32 circuitBreakerState, CircuitBreakerParams memory params)
+        internal
+        pure
+        returns (bytes32)
+    {
+        // It's theoretically not required for the lower bound to be < 1, but it wouldn't make much sense otherwise:
+        // the circuit breaker would immediately trip. Note that this explicitly allows setting both to 0, disabling
+        // the circuit breaker for the token.
+        _require(params.lowerBoundPercentage <= FixedPoint.ONE, Errors.INVALID_CIRCUIT_BREAKER_BOUNDS);
+        _require(params.upperBoundPercentage <= _MAX_BOUND_PERCENTAGE, Errors.INVALID_CIRCUIT_BREAKER_BOUNDS);
+        if (params.upperBoundPercentage > 0) {
+            _require(params.upperBoundPercentage >= params.lowerBoundPercentage, Errors.INVALID_CIRCUIT_BREAKER_BOUNDS);
+        }
+
+        // Set the basic parameters, and chain to `_setCircuitBreakerState` for the rest.
+        circuitBreakerState = circuitBreakerState
+            .insertUint(params.referenceBptPrice, _REFERENCE_BPT_PRICE_OFFSET, _BPT_PRICE_WIDTH)
+            .insertUint(
+            params.currentWeightFactor.compress(_WEIGHT_FACTOR_WIDTH),
+            _WEIGHT_FACTOR_OFFSET,
+            _WEIGHT_FACTOR_WIDTH
+        );
+
+        return _setCircuitBreakerState(circuitBreakerState, params);
+    }
+
+    // This function is only needed to address stack-too-deep issues
+    function _setCircuitBreakerState(bytes32 circuitBreakerState, CircuitBreakerParams memory params)
+        private
+        pure
+        returns (bytes32)
+    {
+        // Add the raw percentage boundaries
+        circuitBreakerState = circuitBreakerState
+            .insertUint(
+            params.lowerBoundPercentage.compress(_BOUND_PERCENTAGE_WIDTH, _MAX_BOUND_PERCENTAGE),
+            _LOWER_BOUND_PCT_OFFSET,
+            _BOUND_PERCENTAGE_WIDTH
+        )
+            .insertUint(
+            params.upperBoundPercentage.compress(_BOUND_PERCENTAGE_WIDTH, _MAX_BOUND_PERCENTAGE),
+            _UPPER_BOUND_PCT_OFFSET,
+            _BOUND_PERCENTAGE_WIDTH
+        );
+
+        // Precompute and store the conversion ratios; if the weights aren't changing, we can use these directly.
+        (uint256 lowerBoundRatioCache, uint256 upperBoundRatioCache) = _getBoundaryConversionRatios(
+            params.lowerBoundPercentage,
+            params.upperBoundPercentage,
+            params.currentWeightFactor
+        );
+
+        return
+            circuitBreakerState
+                .insertUint(
+                lowerBoundRatioCache.compress(_BOUND_RATIO_CACHE_WIDTH, _MAX_BOUND_PERCENTAGE),
+                _LOWER_BOUND_RATIO_CACHE_OFFSET,
+                _BOUND_RATIO_CACHE_WIDTH
+            )
+                .insertUint(
+                upperBoundRatioCache.compress(_BOUND_RATIO_CACHE_WIDTH, _MAX_BOUND_PERCENTAGE),
+                _UPPER_BOUND_RATIO_CACHE_OFFSET,
+                _BOUND_RATIO_CACHE_WIDTH
+            );
+    }
+
+    // Convert percentage bounds to BPT price bounds
+    function _getBoundaryConversionRatios(
+        uint256 lowerBoundPercentage,
+        uint256 upperBoundPercentage,
+        uint256 currentWeightFactor
+    ) private pure returns (uint256 lowerBoundRatioCache, uint256 upperBoundRatioCache) {
+        if (lowerBoundPercentage > 0) {
+            lowerBoundRatioCache = LogExpMath.pow(lowerBoundPercentage, currentWeightFactor);
+        }
+
+        if (upperBoundPercentage > 0) {
+            upperBoundRatioCache = LogExpMath.pow(upperBoundPercentage, currentWeightFactor);
+        }
+    }
+}
--- a/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
@@ -92,6 +92,10 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
     // See `ManagedPoolTokenLib.sol` for data layout.
     mapping(IERC20 => bytes32) private _tokenState;
 
+    // Store the circuit breaker configuration for each token.
+    // See `ManagedPoolCircuitBreakerLib.sol` for data layout.
+    mapping(IERC20 => bytes32) private _circuitBreakerState;
+
     // If mustAllowlistLPs is enabled, this is the list of addresses allowed to join the pool
     mapping(address => bool) private _allowedAddresses;
 
@@ -133,6 +137,7 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
     event AllowlistAddressRemoved(address indexed member);
     event TokenAdded(IERC20 indexed token, uint256 normalizedWeight);
     event TokenRemoved(IERC20 indexed token, uint256 normalizedWeight, uint256 tokenAmountOut);
+    event CircuitBreakerSet(IERC20 indexed token, uint256 lowerBound, uint256 upperBound);
 
     struct NewPoolParams {
         string name;
@@ -1050,6 +1055,10 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
 
         // A valid token can't be zero (must have non-zero weights)
         _require(tokenData != 0, Errors.INVALID_TOKEN);
+    }
+
+    function _getCircuitBreakerData(IERC20 token) internal view returns (bytes32) {
+        return _circuitBreakerState[token];
     }
 
     /**
--- a/pkg/pool-weighted/test/ManagedPoolCircuitBreakerLib.test.ts
+++ b/pkg/pool-weighted/test/ManagedPoolCircuitBreakerLib.test.ts
@@ -1,0 +1,154 @@
+import { expect } from 'chai';
+import { Contract, BigNumber } from 'ethers';
+import { deploy } from '@balancer-labs/v2-helpers/src/contract';
+import { fp, fpMul, FP_ZERO, fromFp } from '@balancer-labs/v2-helpers/src/numbers';
+import { random } from 'lodash';
+import { ZERO_BYTES32 } from '@balancer-labs/v2-helpers/src/constants';
+import { sharedBeforeEach } from '@balancer-labs/v2-common/sharedBeforeEach';
+import { CircuitBreakerParams } from '@balancer-labs/v2-helpers/src/models/pools/weighted/types';
+
+describe('ManagedPoolCircuitBreakerLib', () => {
+  const BPT_PRICE = fp(0.4212);
+  const WEIGHT_FACTOR = fp(0.2); // 1 - 0.8
+  const MAX_BOUND = fp(10);
+  const MAX_RELATIVE_ERROR = 0.01;
+  const LOWER_BOUND = fp(0.8);
+  const UPPER_BOUND = fp(2);
+
+  let lib: Contract;
+
+  before('deploy lib', async () => {
+    lib = await deploy('MockManagedPoolCircuitBreakerLib');
+  });
+
+  async function assertCircuitBreakerState(
+    getter: (word: string) => Promise<BigNumber[]>,
+    setter: (tokenState: string, circuitBreakerParams: CircuitBreakerParams) => Promise<string>,
+    lowerBoundPct: BigNumber,
+    upperBoundPct: BigNumber
+  ) {
+    const circuitBreakerParams: CircuitBreakerParams = {
+      referenceBptPrice: BPT_PRICE,
+      currentWeightFactor: WEIGHT_FACTOR,
+      lowerBoundPercentage: lowerBoundPct,
+      upperBoundPercentage: upperBoundPct,
+    };
+
+    const data = await setter(ZERO_BYTES32, circuitBreakerParams);
+
+    // Bounds set correctly.
+    const [referenceBptPrice, currentWeightFactor, lowerBoundPercentage, upperBoundPercentage] = await getter(data);
+
+    expect(referenceBptPrice).to.equal(BPT_PRICE);
+    expect(currentWeightFactor).to.equal(WEIGHT_FACTOR);
+    // These are high precision random numbers, and there is compression, so it's not exact
+    expect(lowerBoundPercentage).to.almostEqual(lowerBoundPct, MAX_RELATIVE_ERROR);
+    expect(upperBoundPercentage).to.almostEqual(upperBoundPct, MAX_RELATIVE_ERROR);
+  }
+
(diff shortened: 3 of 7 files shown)


pkg/pool-weighted/contracts/managed/ManagedPoolCircuitBreakerLib.sol
```diff
@@ -0,0 +1,253 @@
...
+        _BOUND_RATIO_CACHE_WIDTH;
+
+    uint256 private constant _BOUND_PERCENTAGE_WIDTH = 16;
+    uint256 private constant _BOUND_RATIO_CACHE_WIDTH = 24;
+    uint256 private constant _WEIGHT_FACTOR_WIDTH = 64;
+    uint256 private constant _BPT_PRICE_WIDTH = 112;
+
+    // We are compressing the ratios to 64-bit resolution, but the upper bound would normally be > 1,
+    // so we cannot use the default compression. For consistency, use the same maximum uncompressed value,
+    // even though we expect the lower bounds to be < 1.
```
TomAFrench, line 61:
```suggestion
    // We compress the ratios into 16 bits from a range of [0, 10e18], chosen to allow the upper bound to exceed 1.
    // For consistency, use the same maximum uncompressed value, even though we the lower bound is less than 1.
```

The first line and a half kinda threw me.

pkg/pool-weighted/contracts/managed/ManagedPoolCircuitBreakerLib.sol
```diff
@@ -0,0 +1,253 @@
...
+
+            return (referenceBptPrice.mulDown(lowerBoundRatioCache), referenceBptPrice.mulUp(upperBoundRatioCache));
+        }
+    }
+
+    /**
+     * @notice Sets the reference BPT price, and upper and lower bounds for a token.
+     * @dev If a bound is zero, it means there is no circuit breaker in that direction for the given token.
+     * @param circuitBreakerState - The bytes32 state of the token of interest.
+     * @param params - CircuitBreakerParams (introduced in part to address stack issues) has the following components:
```
TomAFrench, line 164:
```suggestion
     * @param params - CircuitBreakerParams has the following components:
```

TomAFrench (commented):
Considering how this is more complex than the other similar libraries (compounding errors from `pow` and compression) I think it would be useful to do some fuzz testing to scope out the usable ranges of the upper and lower bounds.

pkg/pool-weighted/test/ManagedPoolCircuitBreakerLib.test.ts
```diff
@@ -0,0 +1,154 @@
+import { expect } from 'chai';
+import { Contract, BigNumber } from 'ethers';
+import { deploy } from '@balancer-labs/v2-helpers/src/contract';
+import { fp, fpMul, FP_ZERO, fromFp } from '@balancer-labs/v2-helpers/src/numbers';
+import { random } from 'lodash';
+import { ZERO_BYTES32 } from '@balancer-labs/v2-helpers/src/constants';
+import { sharedBeforeEach } from '@balancer-labs/v2-common/sharedBeforeEach';
+import { CircuitBreakerParams } from '@balancer-labs/v2-helpers/src/models/pools/weighted/types';
+
+describe('ManagedPoolCircuitBreakerLib', () => {
```
nventuro, line 10: It looks like we could do a lot of fuzz testing on this library, since we're effectively testing using arbitrary values here.

pkg/pool-weighted/contracts/managed/ManagedPoolCircuitBreakerLib.sol
```diff
@@ -0,0 +1,248 @@
...
+     */
+    function getCircuitBreakerFields(bytes32 circuitBreakerState)
+        internal
+        pure
+        returns (
+            uint256,
+            uint256,
+            uint256,
+            uint256
+        )
```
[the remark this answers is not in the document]
EndymionJkb: I'll return the struct, that will address both issues. I'm using a struct both to help with the stack, and in case we need to change the fields later.

pkg/pool-weighted/test/ManagedPoolCircuitBreakerLib.test.ts
EndymionJkb: Working on it

pkg/pool-weighted/contracts/managed/ManagedPoolCircuitBreakerLib.sol
```diff
@@ -0,0 +1,248 @@
...
+
+    // Convert percentage bounds to BPT price bounds
+    function _getBoundaryConversionRatios(
+        uint256 lowerBoundPercentage,
+        uint256 upperBoundPercentage,
+        uint256 currentWeightFactor
+    ) private pure returns (uint256 lowerBoundRatioCache, uint256 upperBoundRatioCache) {
+        lowerBoundRatioCache = LogExpMath.pow(lowerBoundPercentage, currentWeightFactor);
+        upperBoundRatioCache = LogExpMath.pow(upperBoundPercentage, currentWeightFactor);
+    }
```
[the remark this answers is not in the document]
EndymionJkb: It's a function because it's called in two places (when setting and getting). To make it more lenient (i.e., fail less often), I can use powDown for the lower bound and powUp for the upper bound.

pkg/pool-weighted/contracts/managed/ManagedPoolCircuitBreakerLib.sol
```diff
@@ -0,0 +1,248 @@
...
+    // [  24 bits |  24 bits |   16 bits   |   16 bits   |    64 bits    |  112 bits |
+    // [ UB ratio | LB ratio | upper bound | lower bound | weight factor | ref price |
+    // |MSB                                                                       LSB|
+    uint256 private constant _REFERENCE_BPT_PRICE_OFFSET = 0;
+    uint256 private constant _WEIGHT_FACTOR_OFFSET = _REFERENCE_BPT_PRICE_OFFSET + _BPT_PRICE_WIDTH;
+    uint256 private constant _LOWER_BOUND_PCT_OFFSET = _WEIGHT_FACTOR_OFFSET + _WEIGHT_FACTOR_WIDTH;
+    uint256 private constant _UPPER_BOUND_PCT_OFFSET = _LOWER_BOUND_PCT_OFFSET + _BOUND_PERCENTAGE_WIDTH;
+    uint256 private constant _LOWER_BOUND_RATIO_CACHE_OFFSET = _UPPER_BOUND_PCT_OFFSET + _BOUND_PERCENTAGE_WIDTH;
+    uint256 private constant _UPPER_BOUND_RATIO_CACHE_OFFSET = _LOWER_BOUND_RATIO_CACHE_OFFSET +
+        _BOUND_RATIO_CACHE_WIDTH;
```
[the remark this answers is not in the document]
EndymionJkb: Well, it's a cache in the sense that it is stored in the _circuitBreakerState when it is originally set along with the weight factor. If nothing has changed (no ongoing weight change, or tokens added/removed), we can just use these values directly, instead of performing expensive exponentiation.

I suppose I could call everything stored for this purpose the "reference" value, like the bptPrice is, to avoid overloading the "cache" concept.

pkg/pool-weighted/contracts/managed/ManagedPoolCircuitBreakerLib.sol
```diff
@@ -0,0 +1,248 @@
...
+            params.currentWeightFactor.compress(_WEIGHT_FACTOR_WIDTH),
+            _WEIGHT_FACTOR_OFFSET,
+            _WEIGHT_FACTOR_WIDTH
+        );
+
+        return _setCircuitBreakerState(circuitBreakerState, params);
+    }
+
+    // This function is only needed to address stack-too-deep issues
+    function _setCircuitBreakerState(bytes32 circuitBreakerState, CircuitBreakerParams memory params)
```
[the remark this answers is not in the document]
EndymionJkb: Ok; will try it that way.

pkg/pool-weighted/contracts/managed/ManagedPoolCircuitBreakerLib.sol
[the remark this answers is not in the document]
nventuro: If using structs keep https://github.com/ethereum/solidity/issues/13475 in mind

pkg/pool-weighted/contracts/managed/ManagedPoolCircuitBreakerLib.sol
```diff
@@ -0,0 +1,248 @@
...
+        _BOUND_RATIO_CACHE_WIDTH;
+
+    uint256 private constant _BOUND_PERCENTAGE_WIDTH = 16;
+    uint256 private constant _BOUND_RATIO_CACHE_WIDTH = 24;
+    uint256 private constant _WEIGHT_FACTOR_WIDTH = 64;
+    uint256 private constant _BPT_PRICE_WIDTH = 112;
+
+    // We compress the ratios into 16 bits from a range of [0, 10e18], chosen to allow the upper bound to exceed 1.
+    // For consistency, use the same maximum uncompressed value, even though we the lower bound is less than 1.
+    uint256 private constant _MAX_BOUND_PERCENTAGE = 10e18; // FP 10
```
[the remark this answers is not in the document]
EndymionJkb: Should also change the FP 10,000 in ManagedPoolTokenLib then.

pkg/pool-weighted/contracts/managed/ManagedPoolCircuitBreakerLib.sol
```diff
@@ -0,0 +1,248 @@
...
+            circuitBreakerState.decodeUint(_UPPER_BOUND_PCT_OFFSET, _BOUND_PERCENTAGE_WIDTH).decompress(
+                _BOUND_PERCENTAGE_WIDTH,
+                _MAX_BOUND_PERCENTAGE
+            )
+        );
+    }
+
+    /**
+     * @notice Returns the dynamic upper and lower bounds for a given token, at the current weight.
+     * @dev If a bound value is zero, it means there is no circuit breaker in that direction for the given token.
```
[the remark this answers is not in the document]
EndymionJkb: No; whether you retrieve the upper and lower bounds as percentages or prices, 0 = no breaker has been set. I will clarify as part of the general documentation update.

EndymionJkb:
Inspired by #1816 , I have a version of this with a 66-bit weight complement that likewise removes compression, and contains a detailed range analysis (to remove compression from the other fields as well).

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -0,0 +1,285 @@
...
+     * If the value of the weight complement has not changed, we can use the reference conversion ratios stored
+     * when the breaker was set. Otherwise, we need to calculate them.
+     *
+     * As described in the general comments above, the weight complement calculation attempts to isolate changes
+     * in the balance due to arbitrageurs responding to external prices, from internal price changes caused by an
+     * ongoing weight update, or changes to the pool composition. There is a non-linear relationship between "spot"
+     * price changes and BPT price changes. This calculation transforms one into the other.
+     *
+     * @param circuitBreakerState - The bytes32 state of the token of interest.
+     * @param currentWeightComplement - The complement of this token's weight: (_denormWeightSum - token weight)
```
nventuro, line 144: Same comment re. what the complement is. I don't see how it could be sum - weight, as that is an arbitrary value.

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -0,0 +1,285 @@
...
+                circuitBreakerState.decodeUint(_UPPER_BOUND_PCT_OFFSET, _BOUND_PERCENTAGE_WIDTH).decompress(
+                    _BOUND_PERCENTAGE_WIDTH,
+                    _MAX_BOUND_PERCENTAGE
+                ),
+                currentWeightComplement
+            );
+
+            // Use these ratios to convert raw percentage bounds to BPT price bounds.
+            // To maximize the valid trading range, round the lower bound down, and the upper bound up.
+            return (referenceBptPrice.mulDown(lowerBoundRatio), referenceBptPrice.mulUp(upperBoundRatio));
```
nventuro, line 194: The ratios are not explained on the other return value. I recommend having a common `return` statement outside the `if/else` construct, so that the onl difference is how the ratios are computed.

Also, why do we want to maximize the trading range? That means the breaker could not trip when it should. I'd argue we should be err on the side of being restrictive.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -103,6 +107,7 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
     event AllowlistAddressRemoved(address indexed member);
     event TokenAdded(IERC20 indexed token, uint256 normalizedWeight);
     event TokenRemoved(IERC20 indexed token);
+    event CircuitBreakerSet(IERC20 indexed token, uint256 lowerBound, uint256 upperBound);
```
nventuro, line 110: See [this review comment](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1792#discussion_r974576492)

pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol
```diff
@@ -0,0 +1,159 @@
...
+        upperBound = bound(upperBound, FixedPoint.ONE, _MAX_BOUND_PERCENTAGE);
+
+        CircuitBreakerLib.CircuitBreakerParams memory params = CircuitBreakerLib.CircuitBreakerParams({
+            referenceBptPrice: bptPrice,
+            referenceWeightComplement: weightComplement,
+            lowerBoundPercentage: lowerBound,
+            upperBoundPercentage: upperBound
+        });
+
+        CircuitBreakerLib.CircuitBreakerParams memory result = _roundTrip(params);
```
nventuro, line 73: Inline this call - it being a function simply makes the code harder to follow and understand what is actually being tested.

Also, the name of the test should be updated to reflect what is being tested.

pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol
```diff
@@ -0,0 +1,159 @@
...
+
+    function testCircuitBreakerBounds(
+        uint64[_NUM_WEIGHT_TRIALS] memory newWeightComplements,
+        uint112 refBptPrice,
+        uint256 initialWeightComplement,
+        uint256 lowerBound,
+        uint256 upperBound
+    ) public {
+        // With refBptPrice ~ 0, rounding errors make it fail
+        vm.assume(refBptPrice > _MIN_BPT_PRICE);
```
nventuro, line 89: Prefer `bound`

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -0,0 +1,285 @@
...
+    // |MSB                                                                                     LSB|
+    uint256 private constant _REFERENCE_BPT_PRICE_OFFSET = 0;
+    uint256 private constant _REFERENCE_WEIGHT_COMPLEMENT_OFFSET = _REFERENCE_BPT_PRICE_OFFSET +
+        _REFERENCE_BPT_PRICE_WIDTH;
+    uint256 private constant _REFERENCE_LOWER_BOUND_RATIO_OFFSET = _UPPER_BOUND_PCT_OFFSET + _BOUND_PERCENTAGE_WIDTH;
+    uint256 private constant _REFERENCE_UPPER_BOUND_RATIO_OFFSET = _REFERENCE_LOWER_BOUND_RATIO_OFFSET +
+        _BOUND_RATIO_WIDTH;
+    uint256 private constant _LOWER_BOUND_PCT_OFFSET = _REFERENCE_WEIGHT_COMPLEMENT_OFFSET +
+        _REFERENCE_WEIGHT_COMPLEMENT_WIDTH;
+    uint256 private constant _UPPER_BOUND_PCT_OFFSET = _LOWER_BOUND_PCT_OFFSET + _BOUND_PERCENTAGE_WIDTH;
```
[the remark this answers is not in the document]
EndymionJkb: Maybe sometimes numbers are better? :) I made them long, consistent, and descriptive. I was trying to group the reference things together, vs dynamic things, but I can put them in order. I can also make the names in the bit layout comment closer, though they're slightly abbreviated.

I also think "Ref" is sufficient to abbreviate reference, which makes everything fit on one line

pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol
```diff
@@ -0,0 +1,159 @@
...
+            lowerBoundPercentage: lowerBound,
+            upperBoundPercentage: upperBound
+        });
+
+        CircuitBreakerLib.CircuitBreakerParams memory result = _roundTrip(params);
+
+        assertEq(result.referenceBptPrice, bptPrice);
+        assertApproxEqRel(result.referenceWeightComplement, weightComplement, _MAX_RELATIVE_ERROR);
+        assertApproxEqRel(result.lowerBoundPercentage, lowerBound, _MAX_BOUND_ERROR);
+        assertApproxEqRel(result.upperBoundPercentage, upperBound, _MAX_BOUND_ERROR);
```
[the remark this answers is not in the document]
EndymionJkb: They're compressed to 16 bits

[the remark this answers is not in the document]
EndymionJkb: Well, those are computed values, not passed in. That computation is tested in the next test. However, it would be an improvement to add an external calculation (not just calling the same function the contract does, though it does do the same thing), and test that here as well. Will do.

pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol
EndymionJkb: Yes; missed that one.

pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol
```diff
@@ -0,0 +1,159 @@
...
+    uint256 private constant _MAX_RELATIVE_ERROR = 1e16;
+    uint256 private constant _MAX_BOUND_ERROR = 2e16;
+ 
+    uint256 private constant _NUM_WEIGHT_TRIALS = 10;
+
+    MockCircuitBreakerLib private _mock;
+
+    function setUp() external {
+        _mock = new MockCircuitBreakerLib();
+    }
```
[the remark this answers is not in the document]
EndymionJkb: It was going to be for future gas reporting, but if there's a setting where you can see the report without it (and you apparently don't want to do reporting anyway), happy to remove it.

pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol
```diff
@@ -0,0 +1,159 @@
...
+
+import "../../contracts/test/MockCircuitBreakerLib.sol";
+import "../../contracts/lib/CircuitBreakerLib.sol";
+
+contract CircuitBreakerLibTest is Test {
+    using FixedPoint for uint256;
+
+    uint256 private constant _MAX_BOUND_PERCENTAGE = 2e18; // 2.0 (max uncompressed is 10)
+    uint256 private constant _MINIMUM_LOWER_BOUND = 1e17;  // 0.1
+    uint256 private constant _MAX_WEIGHT_COMPLEMENT = 3.3e18;
```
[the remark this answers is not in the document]
EndymionJkb: Partly reasoning about how people would set real world bounds (input from integrations), and partly empirical limits from fuzzing (in the case of the weight).

I have a branch that reduces compression, and further refines these values, with some boundary testing and further reasoning. Could backport that to here.

pkg/pool-weighted/contracts/managed/ManagedPoolCircuitBreakerLib.sol
```diff
@@ -0,0 +1,248 @@
...
+     * The current boundary is given as: referenceBptPrice * (raw boundary percentage)**(currentWeightFactor).
+     *
+     * The weight factor calculation attempts to isolate changes in the balance due to arbers responding to external
+     * prices, and internal price changes from moving weights. There is a non-linear relationship between "spot" price
+     * changes and BPT price changes, so this calculation transforms one into the other.
+     *
+     * The raw thresholds are simple percentages: 0.8 means "tolerate a 20% drop in external price." To check the
+     * circuit breaker at runtime, we need to transform that into a corresponding BPT price ratio (relative to the
+     * BPT price at the time the breaker is set), such that the final boundary to be checked is simply the original
+     * reference BPT price multiplied by this conversion ratio.
```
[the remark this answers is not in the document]
EndymionJkb: I reworded this (especially after we got rid of denormWeight). (It is too long, and in the end probably most of it should be in external documentation.)

pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol
```diff
@@ -0,0 +1,159 @@
...
+        return _mock.getCircuitBreakerFields(newPoolState);
+    }
+
+    function testCircuitBreakerPrice(
+        uint128 bptPrice,
+        uint256 weightComplement,
+        uint256 lowerBound,
+        uint256 upperBound
+    ) public {
+        vm.assume(bptPrice < (1 << 112));
```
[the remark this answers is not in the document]
EndymionJkb: It uses bound now, between min/max.

pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol
```diff
@@ -0,0 +1,144 @@
...
+
+    uint256 private constant _MINIMUM_BOUND_PERCENTAGE = 1e17;  // 0.1
+    // The weight complement (1 - w) is bounded by the min/max token weights
+    uint256 private constant _MINIMUM_TOKEN_WEIGHT = 1e16; // 0.01 (1%)
+    uint256 private constant _MAXIMUM_TOKEN_WEIGHT = 99e16; // 0.99 (99%)
+    uint256 private constant _MAX_BOUND_PERCENTAGE = 2e18; // 2.0
+    uint256 private constant _MIN_BPT_PRICE = 1e6;
+
+    uint256 private constant _MAX_RELATIVE_ERROR = 1e16;
+    uint256 private constant _MAX_BPT_PRICE = type(uint112).max;
```
nventuro, line 33: This is a bit extreme - it corresponds to a price of 10^15

pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol
```diff
@@ -0,0 +1,144 @@
...
+    function testReferenceParamsAndBoundRatios(
+        uint256 bptPrice,
+        uint256 weightComplement,
+        uint256 lowerBound,
+        uint256 upperBound
+    ) public {
+        bptPrice = bound(bptPrice, _MIN_BPT_PRICE, _MAX_BPT_PRICE);
+        weightComplement = bound(weightComplement, _MINIMUM_TOKEN_WEIGHT, _MAXIMUM_TOKEN_WEIGHT);
+        lowerBound = bound(lowerBound, _MINIMUM_BOUND_PERCENTAGE, FixedPoint.ONE);
+        upperBound = bound(upperBound, FixedPoint.ONE, _MAX_BOUND_PERCENTAGE);
```
nventuro, line 44:
```suggestion
        lowerBound = bound(lowerBound, _MINIMUM_BOUND_PERCENTAGE, FixedPoint.ONE);
        upperBound = bound(upperBound, lowerBound, _MAX_BOUND_PERCENTAGE);
```

pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol
```diff
@@ -0,0 +1,144 @@
...
+        CircuitBreakerLib.CircuitBreakerParams memory params = CircuitBreakerLib.CircuitBreakerParams({
+            bptPrice: bptPrice,
+            weightComplement: weightComplement,
+            lowerBound: lowerBound,
+            upperBound: upperBound
+        });
+
+        // The setter overwrites all state, so the previous state doesn't matter
+        // If we find we need to set fields individually (e.g., only the bounds),
+        // we could add tests that the previous state was not altered.
```
nventuro, line 55:
```suggestion
```

I just found this confusing - there's no previous state, as can be seen in the fact that the function doesn't take one in.

pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol
```diff
@@ -0,0 +1,144 @@
...
+
+        assertApproxEqRel(initialLowerBptPriceBoundary, expectedLowerBoundBptPrice, _MAX_RELATIVE_ERROR);
+        assertApproxEqRel(initialUpperBptPriceBoundary, expectedUpperBoundBptPrice, _MAX_RELATIVE_ERROR);
+
+        // Test that calling it with the original weightComplement retrieves exact values from the ratio cache
+        (uint256 cachedLowerBptPriceBoundary, uint256 cachedUpperBptPriceBoundary) =
+            CircuitBreakerLib.getCurrentCircuitBreakerBounds(initialPoolState, weightComplement);
+
+        assertEq(cachedLowerBptPriceBoundary, initialLowerBptPriceBoundary);
+        assertEq(cachedUpperBptPriceBoundary, initialUpperBptPriceBoundary);
```
nventuro, line 79: This should be a standalone test, not included in the other one. The first part tests the encoding, while this tests the computation of the bounds when the weight complement has not changed.

nventuro, line 79:
```suggestion
```

This test is wrong - you're just calling the same `pure` function twice with the same arguments, by definition it'll return the same result.

pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    function testDynamicBoundRatios(
+        uint256 initialBptPrice,
+        uint256 initialWeightComplement,
+        uint256 newWeightComplement,
+        uint256 lowerBound,
+        uint256 upperBound
+    ) public {
+        // With refBptPrice ~ 0, rounding errors make it fail
```
nventuro, line 89:
```suggestion
```

pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol
```diff
@@ -0,0 +1,144 @@
...
+        initialBptPrice = bound(initialBptPrice, _MIN_BPT_PRICE, _MAX_BPT_PRICE);
+        lowerBound = bound(lowerBound, _MINIMUM_BOUND_PERCENTAGE, FixedPoint.ONE);
+        upperBound = bound(upperBound, FixedPoint.ONE, _MAX_BOUND_PERCENTAGE);
```
nventuro, line 92:
```suggestion
        upperBound = bound(upperBound, lowerBound, _MAX_BOUND_PERCENTAGE);
```

pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol
```diff
@@ -0,0 +1,144 @@
...
+            lowerBound: lowerBound,
+            upperBound: upperBound
+        });
+
+        // Set the initial state of the breaker
+        bytes32 initialPoolState = CircuitBreakerLib.setCircuitBreakerFields(params);
+        (uint256 lowerBptPriceBoundary, uint256 upperBptPriceBoundary) =
+            CircuitBreakerLib.getCurrentCircuitBreakerBounds(initialPoolState, newWeightComplement);
+
+        _validateWithNewComplement(
```
nventuro, line 108: I'd remove this function and just inline it - we gain nothing with this encapsulation

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -0,0 +1,299 @@
...
+    // Convert percentage bounds to BPT price bounds
+    function getBoundaryConversionRatios(
+        uint256 lowerBound,
+        uint256 upperBound,
+        uint256 currentWeightComplement
+    ) internal pure returns (uint256 lowerBoundRatio, uint256 upperBoundRatio) {
+        // Rounding down for the lower bound, and up for the upper bound will maximize the
+        // "operating range" - the BPT price range that will not trigger the circuit breaker -
+        // of the pool for traders.
+        lowerBoundRatio = lowerBound.powDown(currentWeightComplement);
```
nventuro, line 296: Related to https://github.com/balancer-labs/balancer-v2-monorepo/issues/1844:

If I recall correctly, for `pow(x,y)` to be valid, then `log(x) * y > -41`. The worst case for `y` is 0.99, which seems to mean that `x` could indeed go extremely low. Do we need to though? Why can the lower bound be so small but the upper bound is just 10x?

nventuro (commented):
Did one more pass over some things that I found slightly confusing, but this seems to be in a very good state.

I'm however concerned that I've been looking at this for so long that it's hard to re-review it - it'd be great if we could get a fresh set of eyes. @TomAFrench could you take a look?

One thing in particular that seems a bit concerning is test coverage: we only have tests for actual bounds (i.e. there's no tests for null bounds), and the fuzz tests are simply a reimplementation of the logic inside the contract, so they're not testing much other than lack of reverts.

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -0,0 +1,298 @@
...
+        pure
+        returns (uint256, uint256)
+    {
+        uint256 bptPrice = circuitBreakerState.decodeUint(_BPT_PRICE_OFFSET, _BPT_PRICE_WIDTH);
+        uint256 weightComplement = circuitBreakerState.decodeUint(_WEIGHT_COMPLEMENT_OFFSET, _WEIGHT_COMPLEMENT_WIDTH);
+
+        uint256 lowerBoundRatio;
+        uint256 upperBoundRatio;
+
+        if (weightComplement == currentWeightComplement) {
```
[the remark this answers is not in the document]
EndymionJkb: I was moving away from using "Reference" in the names. In the data structure they are just "bptPrice" and "weightComplement". I can use initialWeightComplement here: the value it had when the breaker was set. (It's not really "computed", just passed in.)

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -0,0 +1,298 @@
...
+            (lowerBoundRatio, upperBoundRatio) = getBoundaryConversionRatios(
+                circuitBreakerState.decodeUint(_LOWER_BOUND_OFFSET, _BOUND_WIDTH).decompress(
+                    _BOUND_WIDTH,
+                    _MAX_BOUND_PERCENTAGE
+                ),
+                circuitBreakerState.decodeUint(_UPPER_BOUND_OFFSET, _BOUND_WIDTH).decompress(
+                    _BOUND_WIDTH,
+                    _MAX_BOUND_PERCENTAGE
+                ),
+                currentWeightComplement
```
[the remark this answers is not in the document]
EndymionJkb: They are stored - and retrieved if initialWeightComplement == weightComplement. If the weight changed, they have to be recomputed.

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -0,0 +1,299 @@
...
+        uint256 lowerBound,
+        uint256 upperBound,
+        uint256 currentWeightComplement
+    ) internal pure returns (uint256 lowerBoundRatio, uint256 upperBoundRatio) {
+        // Rounding down for the lower bound, and up for the upper bound will maximize the
+        // "operating range" - the BPT price range that will not trigger the circuit breaker -
+        // of the pool for traders.
+        lowerBoundRatio = lowerBound.powDown(currentWeightComplement);
+        upperBoundRatio = upperBound.powUp(currentWeightComplement);
+    }
```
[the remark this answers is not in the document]
EndymionJkb: Not sure what happened here - I reversed that in response to an earlier comment. Maybe it snuck back in with a merge conflict somehow? Anyway, it should round to be conservative (the opposite of how it is here). Will restore and make sure it's consistent everywhere.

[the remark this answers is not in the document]
EndymionJkb: You're right about codeblindness - didn't notice this

pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol
EndymionJkb: What do you think it should be? 96 bits? That's about 10^10

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -0,0 +1,299 @@
...
+    //
+    // We then compute and store the current BPT price, and the lower and upper bound conversion ratios, used to
+    // convert the percentage bounds into BPT prices that can be directly compared to the "runtime" BPT prices.
+    //
+    // Since the price bounds shift along with the token weight, in general these bound ratios would need to be
+    // computed every time. However, if the weight of the token and composition of the pool have not changed since
+    // the circuit breaker was set, these stored values can still be used, avoiding a heavy computation.
+    //
+    // [      24 bits      |      24 bits      |  112 bits |   64 bits    |   16 bits   |   16 bits   |
+    // [ upper bound ratio | lower bound ratio | BPT price | weight comp. | upper bound | lower bound |
```
TomAFrench, line 89:
```suggestion
    // [      24 bits      |      24 bits      |  112 bits |   60 bits    |   18 bits   |   18 bits   |
    // [ upper bound ratio | lower bound ratio | BPT price | weight comp. | upper bound | lower bound |
```

As mentioned, in the comment the weight complement fits in 60 bits so we can get some extra precision on the bounds for free like this.

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
EndymionJkb: This all needs to be revisited given the switch to normalized weights. I updated the code, but the test range can now be expanded.

pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol
EndymionJkb: It was needed for stack issues at one time; now able to inline it.

pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol
EndymionJkb: Refactored - still not a good test, but less egregiously so

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -0,0 +1,299 @@
...
+ * BPT prices - circuit breakers are set using simple percentages. Intuitively, a lower bound of 0.8 means the
+ * token can lose 20% of its value before triggering the circuit breaker, and an upper bound of 3.0 means it
+ * can triple before being halted. These percentages are then transformed into BPT prices for comparison to the
+ * "reference" state of the pool when the circuit breaker was set.
+ *
+ * Prices can change in two ways: arbitrage traders responding to external price movement can change the balances,
+ * or an ongoing gradual weight update (or change in pool composition) can change the weights. In order to isolate
+ * the balance changes due to price movement, the bounds are dynamic, adjusted for the current weight.
+ */
+library CircuitBreakerLib {
```
TomAFrench, line 64: Seems like this should expose a `hasCircuitBreakerTripped(bytes32 circuitBreakerState, uint256 currentWeightComplement, uint256 tokenBalance)` function. This would avoid us having to calculate the BPT price inside `ManagedPool`.

TomAFrench (commented):
Had a quick look over the library itself. I'll likely look at the tests, etc. tomorrow.

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
TomAFrench: Ah, just saw you mentioned this in slack.

EndymionJkb: What about the BPT price? It's now 112 bits (max Vault balance). If that could be, say, 96 bits (~10^10) we'd get a bunch more. I also think any extra precision might be better assigned to the bound ratios - those are the most sensitive. Presumably the bounds will be set to "round" numbers like 0.8 or 0.9 anyway.

I have another version that removes compression from bounds entirely.

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
EndymionJkb: Look at the next PR (#1834). You have to calculate the bptPrice anyway when setting it, since it's one of the inputs. I have a getBptPrice method (public, since it might be externally, e.g., for an aggregator predicting the behavior of the pool).

I suppose it could be pushed down to the library, but it needs so much "pool stuff" the library doesn't know about to calculate it: supply, scaling factor for the balance, etc., I thought it made more sense to be in the pool.

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
TomAFrench: huh, I'm a bit confused about naming here in that case.

Working off of the name of this constant
https://github.com/balancer-labs/balancer-v2-monorepo/blob/ce65e40000d8f8b2cea132eb8483e0d7b00e5aaf/pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol#L105
along with these lines
https://github.com/balancer-labs/balancer-v2-monorepo/blob/ce65e40000d8f8b2cea132eb8483e0d7b00e5aaf/pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol#L120-L127

I was assuming that we're only compressing the bounds, and so that's where we need extra bits.

It looks like we're compressing the bound ratios as well.
https://github.com/balancer-labs/balancer-v2-monorepo/blob/ce65e40000d8f8b2cea132eb8483e0d7b00e5aaf/pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol#L277-L287

In this case I think we should add a `_MAX_BOUND_RATIO` constant (even if it's just `uint256 private constant _MAX_BOUND_RATIO = _MAX_BOUND;`) as it isn't immediately obvious that these values should be compressed to the same range.

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
TomAFrench: I think we should definitely have a version of this function in the library for a couple reasons.
1. It doesn't really feel complete to have a circuit breaker library which doesn't have a "am I tripped" function.
2. It would be a good example of usage even if we were to have a parallel implementation somewhere else. This would be useful for testing.
3. It fits in the ethos of `ManagedPool` being glue code and we'll have all the relevant inputs calculated already so we can just feed them into this function quite naturally.

TomAFrench:
I checked and we can actually remove `CircuitBreakerParams` as we don't get stack issues if we just pass in the the 4 fields separately. This could help readability and simplify the integration code inside `ManagedPoolSettings`.

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -0,0 +1,326 @@
...
+            circuitBreakerState,
+            normalizedWeight.complement()
+        );
+        uint256 currentBptPrice = totalSupply.mulUp(normalizedWeight).divDown(upscaledBalance);
+
+        return (
+            lowerBoundBptPrice != 0 && currentBptPrice < lowerBoundBptPrice,
+            upperBoundBptPrice != 0 && currentBptPrice > upperBoundBptPrice
+        );
+    }
```
TomAFrench, line 255: I was originally thinking that we should just return a single boolean here for whether either of the breakers are flipped. Is the reasoning here that we only check the lower breaker if we're selling this token to the pool and vice versa?

TomAFrench: I guess this has the benefit that traders can push the pool back towards the desired range without having to do it in a single trade.

EndymionJkb:
> I checked and we can actually remove `CircuitBreakerParams` as we don't get stack issues if we just pass in the the 4 fields separately. This could help readability and simplify the integration code inside `ManagedPoolSettings`.

Gone!

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -0,0 +1,325 @@
...
+
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
+
+import "../lib/ValueCompression.sol";
+
+/**
+ * @title Circuit Breaker Library
+ * @notice Library for storing and manipulating state related to circuit breakers.
+ * @dev The intent of circuit breakers is to halt trading of a given token if its value changes drastically -
+ * in either direction - with respect to other tokens in the pool. For instance, a stable coin might de-peg
```
TomAFrench, line 25:
```suggestion
 * in either direction - with respect to other tokens in the pool. For instance, a stablecoin might de-peg
```

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -0,0 +1,325 @@
...
+ * and go to zero. With no safeguards, arbitrageurs could drain the pool by selling large amounts of the
+ * token to the pool at inflated internal prices.
+ *
+ * The circuit breaker mechanism establishes a "safe trading range" for each token, expressed in terms of
+ * the BPT price. Both lower and upper bounds can be set, and if a trade would result in moving the BPT price
+ * of any token involved in the operation outside that range, the breaker is "tripped", and the operation
+ * should revert. Each token is independent, since some might have very "tight" valid trading ranges, such as
+ * stable coins, and others are more volatile.
```
TomAFrench, line 33:
```suggestion
 * stablecoins, and others are more volatile.
```

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -0,0 +1,325 @@
...
+        return (
+            lowerBoundBptPrice != 0 && currentBptPrice < lowerBoundBptPrice,
+            upperBoundBptPrice != 0 && currentBptPrice > upperBoundBptPrice
+        );
+    }
+
+    /**
+     * @notice Sets the reference BPT price, weight complement, and upper and lower bounds for a token.
+     * @dev If a bound is zero, it means there is no circuit breaker in that direction for the given token.
+     * @param bptPrice: The BptPrice of the token at the time the circuit breaker is set. The BPT Price
```
TomAFrench, line 257:
```suggestion
     * @param bptPrice: The BPT price of the token at the time the circuit breaker is set. The BPT price
```

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -0,0 +1,325 @@
...
+    uint256 private constant _BPT_PRICE_OFFSET = _WEIGHT_COMPLEMENT_OFFSET + _WEIGHT_COMPLEMENT_WIDTH;
+    uint256 private constant _LOWER_BOUND_RATIO_OFFSET = _BPT_PRICE_OFFSET + _BPT_PRICE_WIDTH;
+    uint256 private constant _UPPER_BOUND_RATIO_OFFSET = _LOWER_BOUND_RATIO_OFFSET + _BOUND_RATIO_WIDTH;
+
+    uint256 private constant _WEIGHT_COMPLEMENT_WIDTH = 64;
+    uint256 private constant _BPT_PRICE_WIDTH = 96;
+    uint256 private constant _BOUND_WIDTH = 16;
+    uint256 private constant _BOUND_RATIO_WIDTH = 32;
+
+    // We allow the  bounds to range over two orders of magnitude: 0.1 - 10. The maximum upper bound is set to 10.0
```
TomAFrench, line 95:
```suggestion
    // We allow the bounds to range over two orders of magnitude: 0.1 - 10. The maximum upper bound is set to 10.0
```

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -0,0 +1,325 @@
...
+    // there is no need for compression.
+    //
+    // We then compute and store the current BPT price, and the lower and upper bound conversion ratios, used to
+    // convert the percentage bounds into BPT prices that can be directly compared to the "runtime" BPT prices.
+    //
+    // Since the price bounds shift along with the token weight, in general these bound ratios would need to be
+    // computed every time. However, if the weight of the token and composition of the pool have not changed since
+    // the circuit breaker was set, these stored values can still be used, avoiding a heavy computation.
+    //
+    // [      32 bits      |      32 bits      |   96 bits |   64 bits    |   16 bits   |   16 bits   |
```
TomAFrench, line 80:
```suggestion
    // [      32 bits      |      32 bits      |  96 bits  |   64 bits    |   16 bits   |   16 bits   |
```

TomAFrench (approved):
Happy to merge this now, just a couple of suggestions about comments.

One potential improvement I can see is the addition of a function along the lines of

```
function updateBoundRatios(bytes32 circuitBreakerState, uint256 weightComplement) internal returns (bytes32) {

        (uint256 lowerBoundRatio, uint256 upperBoundRatio) = getBoundaryConversionRatios(
            circuitBreakerState.decodeUint(_LOWER_BOUND_OFFSET, _BOUND_WIDTH) << _BOUND_SHIFT_BITS,
            circuitBreakerState.decodeUint(_UPPER_BOUND_OFFSET, _BOUND_WIDTH) << _BOUND_SHIFT_BITS,
            weightComplement
        );
        return
            circuitBreakerState
                .insertUint(
                weightComplement,
                _WEIGHT_COMPLEMENT_OFFSET,
                _WEIGHT_COMPLEMENT_WIDTH
            )
                .insertUint(
                lowerBoundRatio.compress(_BOUND_RATIO_WIDTH, _MAX_BOUND_PERCENTAGE),
                _LOWER_BOUND_RATIO_OFFSET,
                _BOUND_RATIO_WIDTH
            )
                .insertUint(
                upperBoundRatio.compress(_BOUND_RATIO_WIDTH, _MAX_BOUND_PERCENTAGE),
                _UPPER_BOUND_RATIO_OFFSET,
                _BOUND_RATIO_WIDTH
            );
}
```

This seems like it would allow us to trustlessly recompute the cached values for bound ratios using the end weights when we initiate a weight change. As a result, we'd get cheaper swaps once the weight change completed without having any input from the manager.

**I think if we want to pursue this is can be in a separate PR however.**

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
EndymionJkb: Tom the ASCII art enforceooooor

PR #1793, opened 2022-09-19
Expose getter for expected compression error on ValueCompression.sol

TomAFrench:
I'm intending this to be used in cases where we're testing libraries which store and read compressed values such as `ManagedPoolTokenLib` or `ManagedPoolCircuitBreakerLib.sol`.

### DIFF
--- a/pkg/pool-weighted/test/foundry/ValueCompression.t.sol
+++ b/pkg/pool-weighted/test/foundry/ValueCompression.t.sol
@@ -1,0 +1,86 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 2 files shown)


pkg/pool-weighted/test/foundry/ValueCompression.t.sol
```diff
@@ -0,0 +1,86 @@
...
+            msb += 2;
+        }
+        if (x >= 2**1) {
+            // No need to shift x any more.
+            msb += 1;
+        }
+    }
+
+    function testCompression(
+        uint64 value,
```
nventuro, line 64: Can't we test over the entire uint256 range?

TomAFrench: I was originally testing the `FixedPoint.ONE` variant and did this to reduce the fraction of inputs that would be rejected. Forgot to relax this when moving to general case.

nventuro (approved):

PR #1794, opened 2022-09-19
Remove `_getTotalTokens` internal getter

TomAFrench:
We actually do length checking of the `balances` array when upscaling so this check is unnecessary.

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
@@ -153,8 +153,6 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
 
         InputHelpers.ensureInputLengthMatch(totalTokens, params.normalizedWeights.length, params.assetManagers.length);
 
-        _totalTokensCache = totalTokens;
-
         // Validate and set initial fees
         _setManagementSwapFeePercentage(params.managementSwapFeePercentage);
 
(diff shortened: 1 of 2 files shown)


pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -792,7 +790,6 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
 
         // Finally, we store the new token's weight and scaling factor.
         _tokenState[token] = ManagedPoolTokenLib.initializeTokenState(token, normalizedWeight, weightSumAfterAdd);
-        _totalTokensCache += 1;
```
nventuro, line 793: We should also delete the variable itself

pkg/pool-weighted/contracts/managed/vendor/BasePool.sol
```diff
@@ -332,8 +329,6 @@ abstract contract BasePool is
         uint256 protocolSwapFeePercentage,
         bytes memory userData
     ) external override returns (uint256 bptOut, uint256[] memory amountsIn) {
-        InputHelpers.ensureInputLengthMatch(balances.length, _getTotalTokens());
```
nventuro, line 329: This is safe to do because we also don't check the length in eg. `onJoinPool`. `_onJoin` and `_onExit` **must** implement this check.

nventuro (approved):

pkg/pool-weighted/contracts/managed/vendor/BasePool.sol
TomAFrench: However we have must stronger trust in the caller in those cases. If the Vault has fallen out of sync with the Pool then the Pool has acted improperly. Users can provide arbitrary data however.

PR #1796, opened 2022-09-19
Remove internal versions of setter functions and inline them into constructor

TomAFrench:
We might want to leave this until the end as it requires us to keep the setters and the constructor in sync. This saves ~50~ 13 bytes however.


TomAFrench:
Closing this as ManagedPool has been deployed.

PR #1800, opened 2022-09-20
Inline swap fee functions in ManagedPool

TomAFrench:
I've consolidated all of the pool state reads into `_onMinimalSwap()` and replaced `_addSwapFeeAmount` and `_subtractSwapFeeAmount` with the underlying FixedPoint maths so that we only perform a single read of `_poolState`.

This has the benefit of reducing bytecode from 27.479kB to 27.407kB.

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPool.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPool.sol
@@ -78,14 +78,16 @@ contract ManagedPool is ManagedPoolSettings {
         balanceTokenIn = _upscale(balanceTokenIn, scalingFactorTokenIn);
         balanceTokenOut = _upscale(balanceTokenOut, scalingFactorTokenOut);
 
+        uint256 swapFeeComplement = getSwapFeePercentage().complement();
+
         if (request.kind == IVault.SwapKind.GIVEN_IN) {
             // Fees are subtracted before scaling, to reduce the complexity of the rounding direction analysis.
-            request.amount = _subtractSwapFeeAmount(request.amount);
+            request.amount = request.amount.mulDown(swapFeeComplement);
 
             // All token amounts are upscaled.
             request.amount = _upscale(request.amount, scalingFactorTokenIn);
 
-            uint256 amountOut = _onSwapGivenIn(request, balanceTokenIn, balanceTokenOut);
+            uint256 amountOut = _onSwapGivenIn(request, balanceTokenIn, balanceTokenOut, swapFeeComplement);
 
             // amountOut tokens are exiting the Pool, so we round down.
             return _downscaleDown(amountOut, scalingFactorTokenOut);
@@ -93,13 +95,13 @@ contract ManagedPool is ManagedPoolSettings {
             // All token amounts are upscaled.
             request.amount = _upscale(request.amount, scalingFactorTokenOut);
 
-            uint256 amountIn = _onSwapGivenOut(request, balanceTokenIn, balanceTokenOut);
+            uint256 amountIn = _onSwapGivenOut(request, balanceTokenIn, balanceTokenOut, swapFeeComplement);
 
             // amountIn tokens are entering the Pool, so we round up.
             amountIn = _downscaleUp(amountIn, scalingFactorTokenIn);
 
             // Fees are added after scaling happens, to reduce the complexity of the rounding direction analysis.
-            return _addSwapFeeAmount(amountIn);
+            return amountIn.divUp(swapFeeComplement);
         }
     }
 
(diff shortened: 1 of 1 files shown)


TomAFrench:
I've added named return values to `_onSwapGivenIn` and `_onSwapGivenOut` as otherwise we have too many stack variables.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -80,26 +89,42 @@ contract ManagedPool is ManagedPoolSettings {
 
         if (request.kind == IVault.SwapKind.GIVEN_IN) {
             // Fees are subtracted before scaling, to reduce the complexity of the rounding direction analysis.
-            request.amount = _subtractSwapFeeAmount(request.amount);
+            // This returns amount - fee amount, so we round down (favoring a higher fee amount).
+            request.amount = request.amount.mulDown(swapFeeComplement);
```
nventuro, line 93:
```suggestion
            // We round the amount in down (favoring a higher fee amount).
            request.amount = request.amount.mulDown(swapFeeComplement);
```

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -163,13 +177,11 @@ contract ManagedPool is ManagedPoolSettings {
         uint256 invariantGrowthRatio = WeightedMath._calculateTwoTokenInvariant(
             tokenInWeight,
             tokenOutWeight,
-            FixedPoint.ONE.add(_addSwapFeeAmount(swapRequest.amount).divDown(currentBalanceTokenIn)),
+            FixedPoint.ONE.add(swapRequest.amount.divUp(swapFeeComplement).divDown(currentBalanceTokenIn)),
             FixedPoint.ONE.sub(amountOut.divDown(currentBalanceTokenOut))
         );
 
         _payProtocolAndManagementFees(invariantGrowthRatio);
```
nventuro, line 184: I find this to be a bit too hidden - I'm particularly not a fan of the fact that we apply the fee outside and deapply it here.

I'd consider refactoring this so that we return `(amount in, amount out)`, and have this function handle all fees. For the purposes of calling weighted math, it removes the fees, but the full amount in does include it. Then we'd pay protocol fees on the callsite (for both given in and given out). This will also be useful once we add circuit breakers, which will similarly require access to the full amount incl. fees.

nventuro: That'd mean upscaling before applying the fees, but I think this is totally fine. Upscaling is a well-understood operation, and it'll be clearer to always upscale all balances at the same time, instead of the weird thing we have now were we first take fees and then upscale.

It'll also make this more correct: we're currently taking unscaled fees, but deapplying them scaled for external fee collection purposes, leading to some rounding error.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -120,128 +141,63 @@ contract ManagedPool is ManagedPoolSettings {
      *
      * Returns the amount of tokens that will be taken from the Pool in return.
      *
-     * All amounts inside `swapRequest`, `balanceTokenIn`, and `balanceTokenOut` are upscaled. The swap fee has already
-     * been deducted from `swapRequest.amount`.
+     * All amounts inside `request`, `balanceTokenIn`, and `balanceTokenOut` are upscaled.
```
EndymionJkb, line 144:
```suggestion
     * All amounts inside `request`, `currentBalanceTokenIn`, and `currentBalanceTokenOut` are upscaled.
```

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -120,128 +141,63 @@ contract ManagedPool is ManagedPoolSettings {
...
-        return amountOut;
     }
 
     /*
      * @dev Called when a swap with the Pool occurs, where the amount of tokens exiting the Pool is known.
      *
      * Returns the amount of tokens that will be granted to the Pool in return.
      *
-     * All amounts inside `swapRequest`, `balanceTokenIn`, and `balanceTokenOut` are upscaled.
+     * All amounts inside `request`, `balanceTokenIn`, and `balanceTokenOut` are upscaled.
```
EndymionJkb, line 176:
```suggestion
     * All amounts inside `request`, `currentBalanceTokenIn`, and `currentBalanceTokenOut` are upscaled.
```

EndymionJkb:
We can remove `WeightedMath._calculateTwoTokenInvariant` now too, as it's no longer called. (Unless we'll need it for something later?)

TomAFrench:
Let's remove it for now. We can always fish it out of the git history if necessary.

nventuro (approved):

PR #1802, opened 2022-09-20
Pass asset manager in addToken

nventuro:
This is preparation for the changes to `removeToken`: we'll need to have a way to withdraw all balance in order to be able to remove.

I'm not using the MockAssetManager yet, but I will in follow up PRs.


pkg/pool-weighted/contracts/test/MockWithdrawDepositAssetManager.sol
```diff
@@ -0,0 +1,82 @@
...
+
+    function withdrawFromPool(
+        bytes32 poolId,
+        IERC20 token,
+        uint256 amount
+    ) external {
+        // Tokens can be withdrawn from the Vault with a 'withdraw' operation, but that will create 'managed' balance
+        // and leave the 'total' balance unchanged. We therefore have to perform two operations: one to withdraw, and
+        // another to clear the 'managed' balance (as the tokens withdrawn are about to be wrapped or unwrapped, and
+        // therefore lost to the Pool in their current format).
```
TomAFrench, line 37:
```suggestion
        // Tokens can be withdrawn from the Vault with a 'withdraw' operation, but that will create 'managed' balance
        // and leave the 'total' balance unchanged. We therefore have to perform two operations: one to withdraw, and
        // another to clear the 'managed' balance.
```

pkg/pool-weighted/contracts/test/MockWithdrawDepositAssetManager.sol
```diff
@@ -0,0 +1,82 @@
...
+
+    function depositToPool(
+        bytes32 poolId,
+        IERC20 token,
+        uint256 amount
+    ) external {
+        // Tokens can be deposited to the Vault with a 'deposit' operation, but that requires a prior 'managed'
+        // balance to exist. We therefore have to perform two operations: one to set the 'managed' balance (representing
+        // the new tokens that resulted from wrapping or unwrapping and which we are managing for the Pool), and
+        // another to deposit.
```
TomAFrench, line 63:
```suggestion
        // Tokens can be deposited to the Vault with a 'deposit' operation, but that requires a prior 'managed'
        // balance to exist. We therefore have to perform two operations: one to set the 'managed' balance (representing
        // the new tokens which we are managing for the Pool), and another to deposit.
```

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -761,36 +761,41 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
+        // since one of its tokens has a balance of zero (making the invariant also zero). After that, we can freely
+        // modify the Pool state (e.g. alter weights).
+        PoolRegistrationLib.registerToken(getVault(), getPoolId(), token, assetManager);
+
         // We need to check that both the new weight is valid, and that it won't make any of the existing weights
         // invalid.
         uint256 weightSumAfterAdd = _validateNewWeight(normalizedWeight);
 
+        // We don't need to check that the new token is not already in the Pool, as the Vault will simply revert if we
+        // try to register it again.
```
TomAFrench, line 790: This should be up above where we register the token.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -761,36 +761,41 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
      * Emits the TokenAdded event.
      *
      * @param token - The ERC20 token to be added to the Pool.
+     * @param assetManager - The asset manager for the token.
```
TomAFrench, line 764: We should add some natspec to here explaining that they need an asset manager which is able to add  the neccessary tokens (and reduce the token balance to zero again if they want to remove the token in future). nbd if it's in a separate PR later.

TomAFrench (approved):

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -802,6 +794,14 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
             _mintPoolTokens(recipient, mintAmount);
         }
 
+        // Once we've updated the internal state, we register the token in the Vault. This makes the Pool enter an
+        // invalid state, since one of its tokens has a balance of zero (making the invariant also zero). The Asset
+        // Manager must be used to deposit some initial balancae and restore regular operation.
```
EndymionJkb, line 799:
```
        // Manager must be used to deposit some initial balance and restore regular operation.
```

PR #1803, opened 2022-09-20
Foundry gas reporting

EndymionJkb:
This adds yarn tasks for measuring and reporting gas in Foundry tests. It only works when there is a deployed contract; hence changes to use existing Mock contracts instead of calling the library directly. (ManagedPoolStorageLib tests could be covered if there were a Mock for it.)

`yarn test-gas` runs the gas report on the Foundry test functions.
`yarn test-snapshot-gas` creates a .gas-snapshot file with current results
`yarn test-snapshot-gas --diff` runs tests and compares them to the stored .snapshot, letting you see any changes (e.g., if you're testing to see whether a given change saves gas).

You can control what the gas report covers with the `gas_reports` option in foundry.toml:
`gas_reports = ["*"]`
tests everything, which seems to be the default, so I don't think we need to change anything in the .toml files unless we have a reason to exclude something.


TomAFrench:
Should we update CI to check that these snapshots are kept up to date?

nventuro (commented):
>It only works when there is a deployed contract; hence changes to use existing Mock contracts instead of calling the library directly. (ManagedPoolStorageLib tests could be covered if there were a Mock for it.)

Is this really the case? I'm getting snapshots for tests even if they have no mocks. Also, the measurement seems to be for the test run, not the underlying function. I guess it is true however that if the test remains the same, changes in the function will translate into changes in the test.

I'm overall hesitant about the approach however. I'd rather we measured directly things that interest us (such as cost of swap/join/exit) and ignored everything else.

EndymionJkb:
> > It only works when there is a deployed contract; hence changes to use existing Mock contracts instead of calling the library directly. (ManagedPoolStorageLib tests could be covered if there were a Mock for it.)
> 
> Is this really the case? I'm getting snapshots for tests even if they have no mocks. 

Can our settings be different somehow? I get no output without a mock.

EndymionJkb:
Closed for lack of interest

PR #1805, opened 2022-09-20
(Managed) Base pool test

jubeira:
Summary of differences with respect to weighted base pool:
- Swap fee tests removed.
- Recovery mode enable / disable tests removed.
 - Joins and exits in recovery mode are preserved.
- Small tweaks to join / exit tests to consider relevant parameters instead of protocol fee.
- Using sender / recipient abstractions.
- `beforeSwapJoinExit` tests removed.
- misc data tests removed.

### DIFF
--- a/pkg/pool-weighted/test/managed/BasePool.test.ts
+++ b/pkg/pool-weighted/test/managed/BasePool.test.ts
@@ -23,7 +23,6 @@ describe('BasePool', function () {
   let tokens: TokenList;
 
   const MIN_SWAP_FEE_PERCENTAGE = fp(0.000001);
-  const MAX_SWAP_FEE_PERCENTAGE = fp(0.1);
   const DELEGATE_OWNER = '0xBA1BA1ba1BA1bA1bA1Ba1BA1ba1BA1bA1ba1ba1B';
 
   const PAUSE_WINDOW_DURATION = MONTH * 3;
@@ -57,6 +56,7 @@ describe('BasePool', function () {
       pauseWindowDuration,
       bufferPeriodDuration,
       owner,
+      from,
     } = params;
     if (!poolTokens) poolTokens = tokens;
     if (!assetManagers) assetManagers = Array(poolTokens.length).fill(ZERO_ADDRESS);
@@ -64,9 +64,10 @@ describe('BasePool', function () {
     if (!pauseWindowDuration) pauseWindowDuration = MONTH;
     if (!bufferPeriodDuration) bufferPeriodDuration = 0;
     if (!owner) owner = ZERO_ADDRESS;
+    if (!from) from = deployer;
 
-    return deploy('MockBasePool', {
-      from: params.from,
+    return deploy('v2-pool-weighted/MockBasePool', {
+      from,
       args: [
         vault.address,
         PoolSpecialization.GeneralPool,
@@ -74,7 +75,6 @@ describe('BasePool', function () {
         'BPT',
         Array.isArray(poolTokens) ? poolTokens : poolTokens.addresses,
         assetManagers,
-        swapFeePercentage,
         pauseWindowDuration,
         bufferPeriodDuration,
         TypesConverter.toAddress(owner),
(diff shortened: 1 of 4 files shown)


pkg/pool-weighted/test/managed/BasePool.test.ts
```diff
@@ -0,0 +1,830 @@
...
+        let singleSwap: SingleSwap;
+
+        sharedBeforeEach(() => {
+          singleSwap = {
+            poolId: minimalPoolId,
+            kind: SwapKind.GivenIn,
+            assetIn: tokens.get(0).instance.address,
+            assetOut: tokens.get(1).instance.address,
+            amount: 1, // Needs to be > 0
+            userData: '0xdeadbeef',
```
jubeira, line 720: We don't need anything in particular in this field, but the test fails if this is set to `'0x'`.

This is because when we check the event's fields in `expectEvent.inIndirectReceipt` [here](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1805/files#diff-e5c00b397fb93defb4fa12d0ec28996b558552b4b2c72efb274347b17201c6d0R745), it ends up doing a `deep.eq` here: https://github.com/balancer-labs/balancer-v2-monorepo/blob/ad3ece6be2ee42e00cd64c52f4063676bdc65928/pvt/helpers/src/test/expectEvent.ts#L110

which in turn tries to convert every element of the object under comparison to a big number, and when it reaches `'0x'` it fails.

In short, this seems to be an issue with how TS interprets the contents and how `deep.eq` is implemented. I'm not 100% sure what a reasonable fix would be.

nventuro: Yes, our helper is not powerful enough to do this type of complex object comparison. Our options are a) make the helper better, but this will be hard and not very useful, or b) not pass `request` to the helper, which will make it not compare it, and instead use `inIndirectReceipt`'s return value (the event) to do a custom comparison yourself in your test.

pkg/pool-weighted/test/managed/BasePool.test.ts
```diff
@@ -0,0 +1,830 @@
...
+          const funds: FundManagement = {
+            sender: poolOwner.address,
+            recipient: poolOwner.address,
+            fromInternalBalance: false,
+            toInternalBalance: false,
+          };
+
+          // min amount: 0, deadline: max.
+          await expect(vault.connect(poolOwner).swap(singleSwap, funds, 0, MAX_UINT256)).to.be.revertedWith('PAUSED');
+        });
```
nventuro, line 289: There's two swap handlers, one for the GeneralPool specialization, and one for the other two. This is only testing the General handler (because the Pool is a General Pool).

pkg/pool-weighted/test/managed/BasePool.test.ts
```diff
@@ -0,0 +1,830 @@
...
+
+        it('joins revert', async () => {
+          const OTHER_JOIN_KIND = 1;
+
+          const request: JoinPoolRequest = {
+            assets: tokens.addresses,
+            maxAmountsIn: Array(tokens.length).fill(0),
+            userData: defaultAbiCoder.encode(['uint256'], [OTHER_JOIN_KIND]),
+            fromInternalBalance: false,
```
nventuro, line 298: You're also not testing that initialization fails while paused

pkg/pool-weighted/test/managed/BasePool.test.ts
```diff
@@ -0,0 +1,830 @@
...
+
+      it('can unpause', async () => {
+        await pool.connect(sender).unpause();
+
+        const { paused } = await pool.getPausedState();
+        expect(paused).to.be.false;
+      });
+
+      it('cannot unpause after the pause window', async () => {
+        await advanceTime(PAUSE_WINDOW_DURATION + DAY);
```
nventuro, line 330:
```suggestion
        await advanceTime(PAUSE_WINDOW_DURATION + 1);
```

pkg/pool-weighted/test/managed/BasePool.test.ts
```diff
@@ -0,0 +1,830 @@
...
+              .grantPermissions([pauseAction, unpauseAction], sender.address, [ANY_ADDRESS, ANY_ADDRESS]);
+          });
+
+          itCanPause();
+        });
+      });
+    });
+  });
+
+  describe('recovery mode', () => {
```
nventuro, line 418: Shouldn't this also test the recovery mode exit?

pkg/pool-weighted/test/managed/BasePool.test.ts
jubeira: Thanks; I just wanted to raise this up in case I was missing anything.

For the purpose of this test, this value is just as good as any other (maybe even better than an empty `'0x'`). So I would leave it like this since it also simplifies the test.

pkg/pool-weighted/test/managed/BasePool.test.ts
```diff
@@ -0,0 +1,830 @@
...
+import { defaultAbiCoder } from 'ethers/lib/utils';
+import { sharedBeforeEach } from '@balancer-labs/v2-common/sharedBeforeEach';
+
+describe('BasePool', function () {
+  let admin: SignerWithAddress, poolOwner: SignerWithAddress, deployer: SignerWithAddress, other: SignerWithAddress;
+  let authorizer: Contract, vault: Contract;
+  let tokens: TokenList;
+
+  const MIN_SWAP_FEE_PERCENTAGE = fp(0.000001);
+  const DELEGATE_OWNER = '0xBA1BA1ba1BA1bA1bA1Ba1BA1ba1BA1bA1ba1ba1B';
```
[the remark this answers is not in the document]
jubeira: Done.

pkg/pool-weighted/test/managed/BasePool.test.ts
```diff
@@ -0,0 +1,830 @@
...
+      pauseWindowDuration,
+      bufferPeriodDuration,
+      owner,
+      from,
+    } = params;
+    if (!specialization) specialization = PoolSpecialization.GeneralPool;
+    if (!poolTokens) poolTokens = tokens;
+    if (!assetManagers) assetManagers = Array(poolTokens.length).fill(ZERO_ADDRESS);
+    if (!swapFeePercentage) swapFeePercentage = MIN_SWAP_FEE_PERCENTAGE;
+    if (!pauseWindowDuration) pauseWindowDuration = MONTH;
```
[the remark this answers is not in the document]
jubeira: Done.

pkg/pool-weighted/test/managed/BasePool.test.ts
```diff
@@ -0,0 +1,830 @@
...
+    let pool: Contract;
+
+    sharedBeforeEach('deploy pool', async () => {
+      pool = await deployBasePool();
+    });
+
+    it('returns minimum BPT', async () => {
+      expect(await pool.getMinimumBpt()).to.be.eq(bn(1e6));
+    });
+  });
```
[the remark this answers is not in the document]
jubeira: Done.

pkg/pool-weighted/test/managed/BasePool.test.ts
jubeira: Good catch; added.

pkg/pool-weighted/test/managed/BasePool.test.ts
jubeira: Done.

pkg/pool-weighted/test/managed/BasePool.test.ts
jubeira: The recovery mode exit test is under the `describe` that tests swaps / joins / exits under a context.

We can move it here if you think it's worth it.

pkg/pool-weighted/test/managed/BasePool.test.ts
jubeira: Initialization added in a separate describe, including a `paused` condition.

jubeira:
I've added a few more tests that check the actual return values of the hooks under different conditions (general / minimal swaps, recovery mode exits, initial join, and regular joins / exits).

pkg/pool-weighted/test/managed/BasePool.test.ts
```diff
@@ -0,0 +1,1000 @@
...
+        const receipt = await (
+          await vault.connect(sender).joinPool(poolId, sender.address, recipient.address, request)
+        ).wait();
+
+        expectEvent.inIndirectReceipt(receipt, pool.interface, 'InnerOnInitializePoolCalled', {
+          userData,
+        });
+      });
+
+      it('burns minimum bpt', async () => {
```
nventuro, line 969:
```suggestion
      it('locks the minimum bpt in the zero address', async () => {
```

This BPT is not burnt (i.e. destroyed), it is sent to the zero address so that it can never be claimed for tokens (and burnt), meaning the supply can never drop to zero again.

nventuro (approved):

PR #1806, opened 2022-09-20
Add fuzz tests for RecoveryMode

EndymionJkb:
I was worried about 0-decimal tokens here, etc. Ensure it works for a wide range of inputs and token counts.


pkg/pool-utils/test/foundry/RecoveryMode.t.sol
```diff
@@ -0,0 +1,44 @@
...
+
+      uint256[] memory amounts = new uint256[](balances.length);
+      for (uint256 i = 0; i < balances.length; i++) {
+        amounts[i] = balances[i];
+      }
+
+      uint256[] memory amountsOut = _mock.computeProportionalAmountsOut(
+          amounts,
+          uint256(_TOTAL_SUPPLY).mulUp(FixedPoint.ONE),
+          uint256(bptAmountOut).mulUp(FixedPoint.ONE)
```
TomAFrench, line 35: These `mulUp`s look like noops, can't we just remove them?

pkg/pool-utils/test/foundry/RecoveryMode.t.sol
```diff
@@ -0,0 +1,44 @@
...
+import "@balancer-labs/v2-solidity-utils/contracts/math/FixedPoint.sol";
+import "@balancer-labs/v2-solidity-utils/contracts/math/Math.sol";
+
+import "../../contracts/RecoveryMode.sol";
+import "../../contracts/test/MockRecoveryMode.sol";
+
+contract RecoveryModeTest is Test {
+    using FixedPoint for uint256;
+
+    uint16 private constant _TOTAL_SUPPLY = type(uint16).max;
```
TomAFrench, line 15:
```suggestion
    uint256 private constant _TOTAL_SUPPLY = type(uint16).max;
```

This avoids some casting later.

pkg/pool-utils/test/foundry/RecoveryMode.t.sol
```diff
@@ -0,0 +1,44 @@
...
+        _mock = new MockRecoveryMode(address(0));
+    }
+
+    function testComputeProportionalAmountsOut(uint112[] memory balances, uint16 bptAmountOut) public {
+      vm.assume(balances.length <= _MAX_TOKENS);
+
+      uint256[] memory amounts = new uint256[](balances.length);
+      for (uint256 i = 0; i < balances.length; i++) {
+        amounts[i] = balances[i];
+      }
```
TomAFrench, line 30:
```suggestion
    function testComputeProportionalAmountsOut(uint256[] memory balances, uint16 bptAmountOut) public {
      vm.assume(balances.length <= _MAX_TOKENS);

      for (uint256 i = 0; i < balances.length; i++) {
        balances[i] = bound(balances[i], 1e6, type(uint112).max);
      }
```

I think that we should probably be using bound in these cases, especially when dealing with arrays.

pkg/pool-utils/test/foundry/RecoveryMode.t.sol
```diff
@@ -0,0 +1,44 @@
...
+
+      uint256[] memory amountsOut = _mock.computeProportionalAmountsOut(
+          amounts,
+          uint256(_TOTAL_SUPPLY).mulUp(FixedPoint.ONE),
+          uint256(bptAmountOut).mulUp(FixedPoint.ONE)
+      );
+
+      uint256 ratio = uint256(bptAmountOut).mulDown(FixedPoint.ONE).divDown(_TOTAL_SUPPLY);
```
TomAFrench, line 38:
```suggestion
      uint256 ratio = uint256(bptAmountOut).divDown(_TOTAL_SUPPLY);
```

nventuro (commented):
I'm skeptical about this - all this does is compare the implementation in RecoveryMode.sol to an identical implementation in this file.

pkg/pool-utils/test/foundry/RecoveryMode.t.sol
EndymionJkb: Yes, that's cleaner! Didn't realize I could use bound anywhere (not just at the top); still getting the hang of this.

nventuro (commented):
I still think this test is not useful, but refactored it anyway to show how it could be better written. My resulting test has much wider coverage than the original one and runs faster.

Summary of the changes introduced in ba6ae9c66f5b35af93e0e8432db2aa15bd0b7eea:
  - `totalSupply` is now a random variable (it was unnecessarily constant)
  - `bptAmountIn` is bound to be smaller than totalSupply (instead of using `vm.assume` - this prevents unnecessary rejections)
  - the rest now receives a full array with all balances, and simply ignores some of them. This is much more performant than doing `vm.assume` on the array lenght, as it results in zero rejections.
  - I also added some logging. It doesn't show up when the tests pass, but it prints the values on failure (which makes debugging failing cases much easier).

PR #1808, opened 2022-09-20
Refactor remove token

nventuro:
The refactor is relatively simply, since we're just removing stuff. I also adjusted the tests (best reviewed ignoring whitespace changes).

I ended up disallowing add/remove while uninitialized for simplicity. Am considering also disallowing burning from the zero address, so that the no re-init invariant is kept.

I had to remove a test about remove token collecting fees (which I imagine it did because it used to be an exit?). At any rate, we can review add/remove and fee interations later.

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
@@ -844,8 +844,8 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
 
     /**
      * @notice Removes a token from the Pool's list of tradeable tokens.
-     * @dev Tokens can only be removed if the Pool has more than 2 tokens, as it can never have fewer than 2. Token removal
-     * is also forbidden during a weight change, or if one is scheduled to happen in the future.
+     * @dev Tokens can only be removed if the Pool has more than 2 tokens, as it can never have fewer than 2. Token
+     * removal is also forbidden during a weight change, or if one is scheduled to happen in the future.
      *
      * Emits the TokenRemoved event. This is a permissioned function.
      *
@@ -857,8 +857,8 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
      */
     function removeToken(
         IERC20 token,
-        address sender,
-        uint256 burnAmount
+        uint256 burnAmount,
+        address sender
     ) external authenticate nonReentrant whenNotPaused {
         _require(totalSupply() > 0, Errors.UNINITIALIZED);
 
--- a/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
@@ -844,10 +844,7 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
 
     /**
      * @notice Removes a token from the Pool's list of tradeable tokens.
-     * @dev Removes a token from the Pool's composition, withdraws all funds from the Vault (sending them to
-     * `recipient`), and finally adjusts the weights of all other tokens.
-     *
-     * Tokens can only be removed if the Pool has more than 2 tokens, as it can never have fewer than 2. Token removal
+     * @dev Tokens can only be removed if the Pool has more than 2 tokens, as it can never have fewer than 2. Token removal
      * is also forbidden during a weight change, or if one is scheduled to happen in the future.
      *
      * Emits the TokenRemoved event. This is a permissioned function.
@@ -856,6 +853,7 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
      * in some scenarios to account for the fact that the Pool now has fewer tokens. This is a permissioned function.
      * @param token - The ERC20 token to be removed from the Pool.
      * @param burnAmount - The amount of BPT to be burned after removing `token` from the Pool.
+     * @param sender - The address to burn BPT from.
      */
     function removeToken(
         IERC20 token,
@@ -896,6 +894,9 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
         delete _tokenState[token];
 
         if (burnAmount > 0) {
+            // We disallow burning from the zero address, as that would allow potentially returning the Pool to the
+            // uninitialized state.
+            _require(sender != address(0), Errors.BURN_FROM_ZERO);
             _burnPoolTokens(sender, burnAmount);
         }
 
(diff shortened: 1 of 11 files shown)


pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -853,75 +855,58 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
-        // Exit the pool, returning the full balance of the token to the recipient
-        (IERC20[] memory tokens, uint256[] memory unscaledBalances, ) = getVault().getPoolTokens(getPoolId());
+        // Before this function is called, the caller must have withdrawn all balance for `token` from the Pool. This
+        // means that the Pool is in an invalid state, since among other things the invariant is zero. Because we're not
+        // in a valid state and all value-changing operations will revert, we are free to modify the Pool state (e.g.
+        // alter weights).
+        // We don't need to test the zero balance since the Vault will simply revert on deregistration if this is not
+        // the case.
+
+        // Removing a token will cause for the weights of all other tokens to increase. This is fine, as there is no
```
EndymionJkb, line 878:
```suggestion
        // Removing a token will cause the weights of all other tokens to increase. This is fine, as there is no
```

EndymionJkb: We could mention there is a maximum *total* weight, even if there isn't a maximum per-token weight.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -853,75 +855,58 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
     function removeToken(
         IERC20 token,
-        address recipient,
-        uint256 burnAmount,
-        uint256 minAmountOut
-    ) external authenticate nonReentrant whenNotPaused returns (uint256) {
-        // We require the pool to be initialized (shown by the total supply being nonzero) in order to remove a token,
-        // maintaining the behaviour that no exits can occur before the pool has been initialized.
-        // This prevents the AUM fee calculation being triggered before the pool contains any assets.
+        address sender,
```
TomAFrench, line 862: This allows the manager to burn BPT from unrelated addresses which is why we hardcoded this to `msg.sender`.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -853,75 +855,58 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
+        // the case.
+
+        // Removing a token will cause for the weights of all other tokens to increase. This is fine, as there is no
+        // maximum weight. We also don't need to check that the new token exists in the Pool, as the Vault will simply
+        // revert if we try to deregister a token that is not registered.
+        // We do, however, want to check that the Pool will end up with at least two tokens. This simplifies some
+        // assumptions made elsewhere (e.g. the denormalized weight sum will always be non-zero), and doesn't greatly
+        // restrict the controller.
+
+        (IERC20[] memory tokens, , ) = getVault().getPoolTokens(getPoolId());
```
TomAFrench, line 885: There may be a cheaper way to check the length of this array. scaling factors perhaps?

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
nventuro: They can also mint to any address via `addToken`, which is the same as they can dillute everyone. They can also add bad tokens and drain the pool. I don't think this changes the trust model in any way.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
nventuro: We don't track any length ourselves - the scaling factors are stored in mapping entries. This is because we favor 1 SLOAD access in swaps (where we know the addresses of the tokens involved), at the cost of having to fetch the tokens array in other operations such as joins.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -853,75 +855,58 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
-        uint256 burnAmount,
-        uint256 minAmountOut
-    ) external authenticate nonReentrant whenNotPaused returns (uint256) {
-        // We require the pool to be initialized (shown by the total supply being nonzero) in order to remove a token,
-        // maintaining the behaviour that no exits can occur before the pool has been initialized.
-        // This prevents the AUM fee calculation being triggered before the pool contains any assets.
+        address sender,
+        uint256 burnAmount
+    ) external authenticate nonReentrant whenNotPaused {
         _require(totalSupply() > 0, Errors.UNINITIALIZED);
```
TomAFrench, line 865: We can remove this condition as removing a token doesn't result in an exit anymore.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -775,6 +775,8 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
         uint256 mintAmount,
         address recipient
     ) external authenticate whenNotPaused {
+        _require(totalSupply() > 0, Errors.UNINITIALIZED);
```
TomAFrench, line 778: There's no need for this imo.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
TomAFrench: Yeah, forgot that _scalingFactors begins with this exact query.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
TomAFrench: It's a question about blanket vs targeted actions. Similar to DAI vs USDC, if you hold DAI then Circle could blacklist Maker and make any USDC-backed DAI worthless but they can't blacklist your wallet specifically.

TomAFrench: We agreed on this being an issue for the controller rather than the pool.

nventuro: We decided that this is ultimately a controller concern - how they use burn amount and sender is very much up to them, and part of their trust model.

TomAFrench (approved):
Happy to merge this once the check preventing burning tokens from the zero address is in.

TomAFrench:
There's also some inconsistency in argument ordering between adding/removing tokens but that's not a blocker.

PR #1815, opened 2022-09-21
fix: removeToken argument order changed

EndymionJkb:
Last argument reorder broke a test.


nventuro:
I ninjafixed it on [6316fc9](https://github.com/balancer-labs/balancer-v2-monorepo/commit/6316fc9740f62659ddd5b79ec6ffa7432a35e545) :grimacing:

PR #1816, opened 2022-09-21
Stop compressing token weights

TomAFrench:
After some thinking about we're able to modify the `denormWeightSum` in the situation where we renormalize it on weight changes, we noticed that with a couple of extra bits we can avoid all compression of weights in `ManagedPoolTokenLib.sol`.

Essentially we can start from a 50:50 pool and show that no matter how we add tokens then the `denormWeightSum` is bounded by `50e18`. The worst case denormalized weight is then 98% of this value and so we just need to provide enough bits to contain this.

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPoolTokenLib.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPoolTokenLib.sol
@@ -20,7 +20,6 @@ import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/ERC20.sol";
 
 import "../lib/GradualValueChange.sol";
-import "../lib/ValueCompression.sol";
 
 /**
  * @title Managed Pool Token Library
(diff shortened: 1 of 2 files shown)


pkg/pool-weighted/contracts/managed/ManagedPoolTokenLib.sol
```diff
@@ -39,27 +38,25 @@ import "../lib/ValueCompression.sol";
...
     uint256 private constant _END_DENORM_WEIGHT_OFFSET = _START_DENORM_WEIGHT_OFFSET + _DENORM_WEIGHT_WIDTH;
     uint256 private constant _DECIMAL_DIFF_OFFSET = _END_DENORM_WEIGHT_OFFSET + _DENORM_WEIGHT_WIDTH;
 
-    uint256 private constant _DENORM_WEIGHT_WIDTH = 64;
+    // The maximum allowable value for `denormWeightSum` is `50e18`, produced by the situation where a token is added
+    // to a Pool with equal weights such that all the other tokens are reduced to the minimum weight.
+    // The maximum denormalized weight of a single token is then given by the 3 token case of a 1%:1%:98% weight Pool
+    // where the 98% token has a denormalized weight of `49e18`.
+    // A 66 bit value allows storing values up to ~73e18 and so is suitable for storing all potential token weights.
+    uint256 private constant _DENORM_WEIGHT_WIDTH = 66;
```
nventuro, line 57: I don't think this is accurate. E.g. weights are 50/50. Add two tokens at 49, they become 1/1/49/49. The sum is `0.5+0.5+0.5 *49 + 0.5 * 49 = 50`. I then remove the two tokens at 1%, the weights are 50/50 and the sum is 49. Repeat ad infinitum.

nventuro:
We're likely going to instead store normalized weights directly.

TomAFrench:
Superceded by #1842

PR #1818, opened 2022-09-21
Add shortcircuit for calculating powDown(FixedPoint.ONE, y) in `_computeJoinExactTokensInInvariantRatio`

TomAFrench:
This avoids having to calculate powDown(FixedPoint.ONE, y) on L233 as this is somewhat expensive and always results in a factor of ~1

### DIFF
--- a/pkg/pool-weighted/contracts/WeightedMath.sol
+++ b/pkg/pool-weighted/contracts/WeightedMath.sol
@@ -223,6 +223,9 @@ library WeightedMath {
                 amountInWithoutFee = nonTaxableAmount.add(taxableAmount.sub(swapFee));
             } else {
                 amountInWithoutFee = amountsIn[i];
+                if (amountInWithoutFee == 0) {
+                    continue;
+                }
             }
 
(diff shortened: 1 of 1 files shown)


pkg/pool-weighted/contracts/WeightedMath.sol
```diff
@@ -223,6 +223,9 @@ library WeightedMath {
                 amountInWithoutFee = nonTaxableAmount.add(taxableAmount.sub(swapFee));
             } else {
                 amountInWithoutFee = amountsIn[i];
+                if (amountInWithoutFee == 0) {
```
nventuro, line 226:
```suggestion
                amountInWithoutFee = amountsIn[i];
                
                // If a token's amount in is not being charged a swap fee then it might be zero (e.g. when joining a Pool with
                // only a subset of tokens). In this case, `balanceRatio` will equal `FixedPoint.ONE`, and the `invariantRatio`
                // will not change at all. We therefore skip to the next iteration, avoiding the costly `powDown` call.
                if (amountInWithoutFee == 0) {
```

nventuro (approved):

PR #1821, opened 2022-09-21
Add single token version of `_calcBptOutGivenExactTokensIn`

TomAFrench:
Includes changes in #1820, #1819, #1818

### DIFF
--- a/pkg/pool-weighted/contracts/WeightedMath.sol
+++ b/pkg/pool-weighted/contracts/WeightedMath.sol
@@ -197,6 +197,50 @@ library WeightedMath {
         return bptOut;
     }
 
+    function _calcBptOutGivenExactTokenIn(
+        uint256 balance,
+        uint256 normalizedWeight,
+        uint256 amountIn,
+        uint256 bptTotalSupply,
+        uint256 swapFeePercentage
+    ) internal pure returns (uint256) {
+        // BPT out, so we round down overall.
+
+        uint256 amountInWithoutFee;
+        {
+            uint256 balanceRatioWithFee = balance.add(amountIn).divDown(balance);
+            uint256 invariantRatioWithFees = balanceRatioWithFee.mulDown(normalizedWeight).add(
+                normalizedWeight.complement()
+            );
+
+            if (balanceRatioWithFee > invariantRatioWithFees) {
+                uint256 nonTaxableAmount = invariantRatioWithFees > FixedPoint.ONE
+                    ? balance.mulDown(invariantRatioWithFees - FixedPoint.ONE)
+                    : 0;
+                uint256 taxableAmount = amountIn.sub(nonTaxableAmount);
+                uint256 swapFee = taxableAmount.mulUp(swapFeePercentage);
+
+                amountInWithoutFee = nonTaxableAmount.add(taxableAmount.sub(swapFee));
+            } else {
+                amountInWithoutFee = amountIn;
+                // If a token's amount in is not being charged a swap fee then it might be zero.
+                // In this case, it's clear that the sender should receive no BPT.
+                if (amountInWithoutFee == 0) {
+                    return 0;
+                }
+            }
+        }
+
+        uint256 balanceRatio = balance.add(amountInWithoutFee).divDown(balance);
+
+        uint256 invariantRatio = balanceRatio.powDown(normalizedWeight);
+
+        uint256 bptOut = (invariantRatio > FixedPoint.ONE)
+            ? bptTotalSupply.mulDown(invariantRatio - FixedPoint.ONE)
+            : 0;
+        return bptOut;
+    }
+
     /**
      * @dev Intermediate function to avoid stack-too-deep errors.
      */
@@ -225,7 +269,7 @@ library WeightedMath {
                 amountInWithoutFee = amountsIn[i].sub(swapFee);
             } else {
                 amountInWithoutFee = amountsIn[i];
-                
+
                 // If a token's amount in is not being charged a swap fee then it might be zero (e.g. when joining a
                 // Pool with only a subset of tokens). In this case, `balanceRatio` will equal `FixedPoint.ONE`, and
                 // the `invariantRatio` will not change at all. We therefore skip to the next iteration, avoiding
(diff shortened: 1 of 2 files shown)


nventuro:
Should this be rebased now that all three prerequisites have been merged?

pkg/pool-weighted/contracts/WeightedMath.sol
```diff
@@ -192,7 +192,49 @@ library WeightedMath {
...
+                    : 0;
+                uint256 taxableAmount = amountIn.sub(nonTaxableAmount);
+                uint256 swapFee = taxableAmount.mulUp(swapFeePercentage);
+
+                amountInWithoutFee = nonTaxableAmount.add(taxableAmount.sub(swapFee));
+            } else {
+                amountInWithoutFee = amountIn;
+                if (amountInWithoutFee == 0) {
+                    return 0;
+                }
```
nventuro, line 228: I'd add the same comments we have in `master` here

pkg/pool-weighted/contracts/WeightedMath.sol
```diff
@@ -192,7 +192,49 @@ library WeightedMath {
...
+        uint256 bptTotalSupply,
+        uint256 swapFeePercentage
+    ) internal pure returns (uint256) {
+        // BPT out, so we round down overall.
+
+        uint256 amountInWithoutFee;
+        {
+            uint256 balanceRatioWithFee = balance.add(amountIn).divDown(balance);
+            uint256 invariantRatioWithFees = balanceRatioWithFee.mulDown(normalizedWeight).add(
+                normalizedWeight.complement()
```
nventuro, line 213:
```suggestion
            // This assumes that the sum of all weights equals FixedPoint.ONE, which may not be the case during a
            // gradual weight change due to interpolation rounding error. This will result in a small difference between the
            // output of this function and the equivalent `_calcBptOutGivenExactTokensIn` call.
            uint256 invariantRatioWithFees = balanceRatioWithFee.mulDown(normalizedWeight).add(
                normalizedWeight.complement()
```

nventuro: It'd be good to better explain *what* this difference will be, once we understand it.

pkg/pool-weighted/test/foundry/WeightedMath.t.sol
```diff
@@ -0,0 +1,98 @@
...
+import "@balancer-labs/v2-solidity-utils/contracts/math/FixedPoint.sol";
+
+import "../../contracts/WeightedMath.sol";
+
+contract WeightedMathTest is Test {
+    using FixedPoint for uint256;
+
+    function testJoinSwaps(
+        uint256[20] memory balancesFixed,
+        uint256[20] memory normalizedWeightsFixed,
```
nventuro, line 27: We could go higher here, though I guess it doesn't quite matter in this test.

pkg/pool-weighted/test/foundry/WeightedMath.t.sol
```diff
@@ -0,0 +1,98 @@
...
+        for (uint256 i = 0; i < arrayLength; i++) {
+            normalizedWeights[i] = normalizedWeightsFixed[i].divDown(denormalizedWeightSum);
+            vm.assume(normalizedWeights[i] >= WeightedMath._MIN_WEIGHT);
+            normalizedWeightSum += normalizedWeights[i];
+        }
+
+        // Note: Due to compression errors, this normalization property of weights may not always hold.
+        // This causes the two forms of join to produce slightly different outputs due to
+        // `WeightedMath._calcBptOutGivenExactTokenIn` assuming perfect normalization.
+        if (normalizedWeightSum < FixedPoint.ONE) {
```
nventuro, line 59:
```suggestion
        // We therefore adjust the last weight to produce a scenario in which the two functions should yield the exact
        // same result.
        if (normalizedWeightSum < FixedPoint.ONE) {
```

pkg/pool-weighted/test/foundry/WeightedMath.t.sol
```diff
@@ -0,0 +1,98 @@
...
+            amountsIn[tokenIndex],
+            bptTotalSupply,
+            swapFeePercentage
+        );
+
+        if (joinSwap > 1e10) {
+            assertApproxEqRel(joinSwap, properJoin, 1e10);
+        } else {
+            assertApproxEqAbs(joinSwap, properJoin, 100000);
+        }
```
nventuro, line 96: Do we not get exact equality with the weight adjustment?

pkg/pool-weighted/test/foundry/WeightedMath.t.sol
```diff
@@ -0,0 +1,98 @@
...
+
+        // Note: Due to compression errors, this normalization property of weights may not always hold.
+        // This causes the two forms of join to produce slightly different outputs due to
+        // `WeightedMath._calcBptOutGivenExactTokenIn` assuming perfect normalization.
+        if (normalizedWeightSum < FixedPoint.ONE) {
+            normalizedWeights[arrayLength - 1] += FixedPoint.ONE - normalizedWeightSum;
+        }
+
+        amountIn = bound(amountIn, 0, balances[tokenIndex] / 100);
```
nventuro, line 63: Can this be relaxed?

pkg/pool-weighted/test/foundry/WeightedMath.t.sol
TomAFrench: Pushing too high here results in too many rejected runs due to minimum weight constraints.

pkg/pool-weighted/test/foundry/WeightedMath.t.sol
TomAFrench: I've allowed it to be up to `balances[tokenIndex]`

pkg/pool-weighted/contracts/WeightedMath.sol
TomAFrench: This also occurs in the situation where we have no weight change due to normalization of the stored weights.

nventuro: I was already thinking of the no normalization case :upside_down_face:

nventuro (approved):

PR #1825, opened 2022-09-22
Enforce MaxDenormWeight

EndymionJkb:
If you repeatedly add and remove tokens (e.g., per the example in the original PR), removing the tokens decreases the sum again, so it shouldn't grow without bound. If you add two tokens at 0.49, the sum is over 3.6, but if you then remove them, it's back to 1. (If you could remove the original low weight ones instead, it would end up much higher - but they have non-zero balances, and the remaining ones would have zero balances, so as a practical matter the asset manager would have to intervene to make that possible, so it shouldn't happen.)

There was a _MAX_DENORM_WEIGHT normalization factor when we were compressing - but were not checking that it was never exceeded (so, it could potentially have overflowed).

In practice - especially since weight changes (that will reset the sum) are typically associated with add/removes as part of rebalancing - it's unlikely the denorm weight will reach extremes with normal use, and much more likely to fail with MIN_WEIGHT even then.

So we could still safely avoid compression by setting and enforcing a max weight that fits in 64 bits. (Could also set it to 50 and go back to 66 bits, or 36 for 65 bits, but 18 seems high enough.)

### DIFF
--- a/pkg/balancer-js/src/utils/errors.ts
+++ b/pkg/balancer-js/src/utils/errors.ts
@@ -125,6 +125,7 @@ const balancerErrorCodes: Record<string, string> = {
   '442': 'SAFE_CAST_VALUE_CANT_FIT_UINT64',
   '443': 'UNHANDLED_FEE_TYPE',
   '444': 'BURN_FROM_ZERO',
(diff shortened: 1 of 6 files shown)


EndymionJkb:
Like the parent PR #1816, superseded by #1842

PR #1826, opened 2022-09-22
Remove unused BasePool args

nventuro:
We don't use any of these values (nor will we), so there's no point in passing them around.


jubeira (approved):
Looking good!

PR #1827, opened 2022-09-22
Deleting stale comment in ProtocolFeeCache.

jubeira:


EndymionJkb:
I noticed that too, and in the current design it's now stale.

I don't want to forget the concept of non-delegated fees though (especially given possible future developments like https://forum.balancer.fi/t/bip-77-friendly-balancer-protocol-fees-for-integrators/3732, and other use cases I know where the protocol fees have to be zero or negotiated amounts for certain pools), but that's a future release.

PR #1828, opened 2022-09-22
Enforce allowlist on pool initialization

TomAFrench:
We now enforce that the initial LP must be on the allowlist. I've added tests that the pool is properly initialised (including that it sets the AUM fee timestamp)

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPool.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPool.sol
@@ -217,7 +217,14 @@ contract ManagedPool is ManagedPoolSettings {
 
     // Initialize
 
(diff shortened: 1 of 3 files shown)


EndymionJkb (commented):
There is just a little duplication here with the tests in Settings. Those should be focused on the add/remove functions in isolation: e.g., checking for events, verifying integrity of the list when you toggle it on and off, permissions, etc.

ManagedPool tests should be functional (init/join/exit), checking that people can only join when there's no allowlist, or there is a list and they're on it. (And maybe that it doesn't block exits, etc.)

That's mostly the case, but "when mustAllowlistLPs is toggled", for instance, duplicates the event and add/remove when disabled tests from Settings.

nventuro (approved):

PR #1830, opened 2022-09-22
Add groundwork for composable swaps

TomAFrench:
`_onSwapMinimal` is now the dispatch function for whether we're performing a "joinSwap", "exitSwap" or a "tokenSwap". As BPT doesn't have a scaling factor the responsibility for upscaling tokens appropriately has been pushed into the `_onJoinSwap`, `_onExitSwap` and `_onTokenSwap` functions.

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPool.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPool.sol
@@ -123,8 +123,8 @@ contract ManagedPool is ManagedPoolSettings {
             // TODO: Once all weights are normalized, change to
             // uint256 tokenInWeight = ManagedPoolTokenLib.getTokenWeight(tokenInState, weightChangeProgress)
             // uint256 tokenOutWeight = ManagedPoolTokenLib.getTokenWeight(tokenOutState, weightChangeProgress)
-            tokenInWeight = _getNormalizedWeight(request.tokenIn, weightChangeProgress);
-            tokenOutWeight = _getNormalizedWeight(request.tokenOut, weightChangeProgress);
+            tokenInWeight = _getNormalizedWeight(tokenInState, weightChangeProgress);
+            tokenOutWeight = _getNormalizedWeight(tokenOutState, weightChangeProgress);
 
             scalingFactorTokenIn = ManagedPoolTokenLib.getTokenScalingFactor(tokenInState);
             scalingFactorTokenOut = ManagedPoolTokenLib.getTokenScalingFactor(tokenOutState);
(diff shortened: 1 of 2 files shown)


pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -80,11 +87,44 @@ contract ManagedPool is ManagedPoolSettings {
...
+            uint256 tokenInWeight = _getNormalizedWeight(request.tokenIn, weightChangeProgress);
+            uint256 tokenOutWeight = _getNormalizedWeight(request.tokenOut, weightChangeProgress);
+            return
+                _onTokenSwap(
+                    request,
+                    balanceTokenIn,
+                    balanceTokenOut,
+                    tokenInWeight,
+                    tokenOutWeight,
+                    swapFeePercentage.complement()
```
nventuro, line 106: We'll need to read the scaling factors inside this funcion, which will read the same state `_getNormalizedWeight` has already accesses. It also doesn't seem like we gain much by computing the weights outside of it.

What if this function simply read the pool state, checked for swaps to be enabled, and then dispatched to all three handlers, passing the state and leaving computation of fees, weights, etc. up to them?

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -95,13 +135,15 @@ contract ManagedPool is ManagedPoolSettings {
             // All token amounts are upscaled.
             request.amount = _upscale(request.amount, scalingFactorTokenIn);
 
-            uint256 amountOut = _onSwapGivenIn(
-                request,
+            // We round the amount in down (favoring a higher fee amount).
+            request.amount = request.amount.mulDown(swapFeeComplement);
+
+            uint256 amountOut = WeightedMath._calcOutGivenIn(
```
nventuro, line 141: Big fan of this inlining.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -116,33 +108,41 @@ contract ManagedPool is ManagedPoolSettings {
...
-        // TODO: Once all weights are normalized, change to
-        // uint256 tokenInWeight = ManagedPoolTokenLib.getTokenWeight(tokenInState, weightChangeProgress)
-        // uint256 tokenOutWeight = ManagedPoolTokenLib.getTokenWeight(tokenOutState, weightChangeProgress)
-        uint256 tokenInWeight = _getNormalizedWeight(request.tokenIn, weightChangeProgress);
-        uint256 tokenOutWeight = _getNormalizedWeight(request.tokenOut, weightChangeProgress);
-
-        uint256 scalingFactorTokenIn = ManagedPoolTokenLib.getTokenScalingFactor(tokenInState);
-        uint256 scalingFactorTokenOut = ManagedPoolTokenLib.getTokenScalingFactor(tokenOutState);
+        bytes32 poolState
+    ) internal view returns (uint256 swapOutput) {
```
nventuro, line 112: We've called this 'amount calculated' (as opposed to amount given) in the past - not sure if we want to reproduce that here as it's a bit too general maybe. 'output' might be confusing in that it could be an amount in.

nventuro (approved):

PR #1832, opened 2022-09-23
Add tokens to UpdateWeightsGradually

EndymionJkb:
Change updateWeightsGradually (for ManagedPools) to take an array of tokens (part of the prep for composability).


nventuro (approved):
Lovely, thanks!

We'll need to change the helper once we add composability so that BPT is not included in the array, but that should be easy enough.

PR #1833, opened 2022-09-23
Protocol fee cache ids: fee type remappings

### LINKED ISSUE #1814: Pass fieIds to ProtocolFeeCache
The cache currently uses hardcoded ids - it'd be good if we could pass ids on construction, to be stored as immutable state variables, and use those instead. This would mean that we'd e.g. be able to use the flash loan fee as a swap fee, but we won't try to prevent said errors.

We'll need to add a getter for the fee ids for each fee type.

jubeira:
Closes #1814.


pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -135,7 +136,10 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
             bufferPeriodDuration,
             owner
         )
-        ProtocolFeeCache(protocolFeeProvider)
+        ProtocolFeeCache(
+            protocolFeeProvider,
+            ProviderFeeIDs(ProtocolFeeType.SWAP, ProtocolFeeType.YIELD, params.aumFeeId)
```
nventuro, line 141: There should be a test in ManagedPoolSettings for this.

jubeira: Done in https://github.com/balancer-labs/balancer-v2-monorepo/pull/1833/commits/02e200eeb68232680847849207e0dc9c346ee526

pkg/pool-utils/contracts/protocol-fees/ProtocolFeeCache.sol
```diff
@@ -109,6 +135,6 @@ abstract contract ProtocolFeeCache is RecoveryMode {
             _revert(Errors.UNHANDLED_FEE_TYPE);
         }
 
-        emit ProtocolFeePercentageCacheUpdated(feeType, currentValue);
+        emit ProtocolFeePercentageCacheUpdated(feeType, providerFeeId, currentValue);
```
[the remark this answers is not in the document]
jubeira: Done in https://github.com/balancer-labs/balancer-v2-monorepo/pull/1833/commits/def865334ffddfb8c8dfb45b117a2e4c928cd731

pkg/pool-utils/contracts/protocol-fees/ProtocolFeeCache.sol
```diff
@@ -34,6 +34,10 @@ import "../RecoveryMode.sol";
 abstract contract ProtocolFeeCache is RecoveryMode {
     using SafeCast for uint256;
 
+    uint256 public immutable swapFeeId;
+    uint256 public immutable yieldFeeId;
+    uint256 public immutable aumFeeId;
+
```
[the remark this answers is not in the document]
jubeira: Done in https://github.com/balancer-labs/balancer-v2-monorepo/pull/1833/commits/def865334ffddfb8c8dfb45b117a2e4c928cd731

pkg/pool-utils/contracts/protocol-fees/ProtocolFeeCache.sol
```diff
@@ -96,19 +131,21 @@ abstract contract ProtocolFeeCache is RecoveryMode {
         // solhint-disable-previous-line no-empty-blocks
     }
 
-    function _updateProtocolFeeCache(IProtocolFeePercentagesProvider protocolFeeProvider, uint256 feeType) private {
-        uint256 currentValue = protocolFeeProvider.getFeeTypePercentage(feeType);
+    function _updateSwapProtocolFeeCache(IProtocolFeePercentagesProvider protocolFeeProvider, uint256 feeId) private {
+        uint256 swapFee = protocolFeeProvider.getFeeTypePercentage(feeId);
```
nventuro, line 135: I'd either:
 a) inline this entire function
 b) hardcode `_swapFeeId` here instead of receiving it as an argument

If you want to inline, you can change this slightly by doing a single storage write with

```
_cache = FeeTypeCache({ swapFee: swapFee; yieldFee: yieldFee; aumFee: aumFee; });
```

pkg/pool-utils/test/ProtocolFeeCache.test.ts
```diff
@@ -51,94 +57,129 @@ describe('ProtocolFeeCache', () => {
     );
   });
 
-  sharedBeforeEach('deploy fee cache', async () => {
-    protocolFeeCache = await deploy('MockProtocolFeeCache', {
-      args: [vault.protocolFeesProvider.address],
-      from: admin,
+  context('with valid fee type ids', () => {
+    context('using default mapping', () => {
```
nventuro, line 61:
```suggestion
    context('using default fee ids', () => {
```

pkg/pool-utils/test/ProtocolFeeCache.test.ts
```diff
@@ -51,94 +57,129 @@ describe('ProtocolFeeCache', () => {
...
+    it('reverts when querying unknown protocol fees', async () => {
+      await expect(protocolFeeCache.getProtocolFeePercentageCache(17)).to.be.revertedWith('UNHANDLED_FEE_TYPE');
+    });
 
-        it('returns the same value as in the provider', async () => {
-          expect(await protocolFeeCache.getProtocolFeePercentageCache(feeType)).to.equal(
-            await vault.protocolFeesProvider.getFeeTypePercentage(feeType)
-          );
-        });
+    it('stores fee type id mappings correctly', async () => {
```
nventuro, line 97:
```suggestion
    it('stores fee type ids correctly', async () => {
```

nventuro (approved):

pkg/pool-utils/contracts/protocol-fees/ProtocolFeeCache.sol
jubeira: Unfortunately hardcoding is not possible because the ID is immutable and we are calling this in the constructor.
We can merge the three updates in just one, but we'd still need to pass the IDs as arguments.

nventuro: You can inline the function into both constrcutor and runtime. I'd measure bytecode size with both approaches, and if inlining is much better then go for it.

PR #1834, opened 2022-09-23
Circuit breaker interactions

### LINKED ISSUE #680: Implement Circuit Breakers

EndymionJkb:
This is the beginning of integrating circuit breakers with the pool. It allows getting and setting breakers through ManagedPoolSettings, and emitting the associated event. Setting a circuit breaker is a permissioned action.

Resolves #680. (Well, not entirely.)


pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -938,6 +949,96 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
+                tokenBalance = _upscale(balances[i], ManagedPoolTokenLib.getTokenScalingFactor(_tokenState[token]));
+                break;
+            }
+        }
+
+        if (tokenBalance == 0) {
+            _revert(Errors.INVALID_TOKEN);
+        }
+
+        return totalSupply().mulUp(normalizedWeight).divDown(tokenBalance);
```
TomAFrench, line 1039:
```suggestion
        (uint256 cash, uint256 managed, , ) = vault.getPoolTokenInfo(getPoolId(), token);
        uint256 tokenBalance = _upscale(cash + managed, ManagedPoolTokenLib.getTokenScalingFactor(_tokenState[token]));
        return totalSupply().mulUp(normalizedWeight).divDown(tokenBalance);
```

`getPoolTokenInfo` reverts if the token is not registered (note this means that BPT will pass so we may want to explicitly reject this.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -323,6 +325,15 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
     }
 
+    /**
+     * @dev Returns the normalized weight of a single token.
+     */
+    function _getNormalizedWeight(IERC20 token) internal view returns (uint256) {
+        uint256 weightChangeProgress = ManagedPoolStorageLib.getGradualWeightChangeProgress(_poolState);
+
+        return ManagedPoolTokenLib.getTokenWeight(_tokenState[token], weightChangeProgress);
+    }
```
TomAFrench, line 335: I think if we're only using this getter in one place then we can just inline it.

EndymionJkb: I was trying to get rid of it, but it is used in two places.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -94,6 +99,7 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
     event AllowlistAddressRemoved(address indexed member);
     event TokenAdded(IERC20 indexed token, uint256 normalizedWeight);
     event TokenRemoved(IERC20 indexed token);
+    event CircuitBreakerSet(IERC20 indexed token, uint256 lowerBoundPercentage, uint256 upperBoundPercentage);
```
TomAFrench, line 102:
```suggestion
    event CircuitBreakerSet(IERC20 indexed token, uint256 bptPrice, uint256 lowerBoundPercentage, uint256 upperBoundPercentage);
```
I think we'd need to value to be able to make the necessary calculations offchain.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -932,6 +947,110 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
+
+    /**
+     * @notice Get the price of a single token in terms of BPT, given the weight.
+     * @dev Returns an 18-decimal floating point number. Could be internal, but might be generally useful
+     * externally.
+     */
+    function getBptPrice(IERC20 token, uint256 normalizedWeight) public view returns (uint256) {
+        if (token == IERC20(this)) {
+            // BPT price of the BPT token itself is undefined.
+            _revert(Errors.INVALID_TOKEN);
```
TomAFrench, line 1041:
```suggestion
            return FixedPoint.ONE;
```
I don't think we need to revert here as the price of BPT in terms of BPT is defined.

EndymionJkb: I went back and forth on that. I had it just returning ONE at first, then thought if somebody's doing this it's probably a mistake. But I suppose with highly composed pools it might get called this way.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -921,6 +937,111 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
+        return CircuitBreakerLib.getCircuitBreakerFields(_circuitBreakerState[token]);
+    }
+
+    /**
+     * @notice Get the current BPT price bounds for the given token.
+     * @dev The bounds are set as percentages, but retrieved as BPT prices which can be directly compared to the
+     * current BPT price. Though the percentages are fixed when the breaker is set, the BPT price bounds are dynamic.
+     * If there is an ongoing gradual weight update, they will shift up or down with the current weight.
+     */
+    function getCurrentCircuitBreakerBounds(IERC20 token) public view returns (uint256, uint256) {
```
TomAFrench, line 967:
```suggestion
    function getCurrentCircuitBreakerBounds(IERC20 token) external view returns (uint256, uint256) {
```
Lets have this external until (if) it's needed

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -921,6 +937,111 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
+
+    /**
+     * @notice Get the price of a single token in terms of BPT, given the weight.
+     * @dev Returns an 18-decimal floating point number. Could be internal, but might be generally useful
+     * externally.
+     */
+    function getBptPrice(IERC20 token, uint256 normalizedWeight) public view returns (uint256) {
+        if (token == IERC20(this)) {
+            return FixedPoint.ONE;
+        }
```
TomAFrench, line 1033: I just noticed that this allows us to set a circuit breaker for BPT as it is. We'll likely want to block that.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -921,6 +937,111 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
+            upperBoundPercentage
+        );
+
+        emit CircuitBreakerSet(token, bptPrice, lowerBoundPercentage, upperBoundPercentage);
+    }
+
+    /**
+     * @notice Get the price of a single token in terms of BPT, given the weight.
+     * @dev Returns an 18-decimal floating point number. Could be internal, but might be generally useful
+     * externally.
```
TomAFrench, line 1028:
```suggestion
     * @dev Returns an 18-decimal floating point number.
```

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -921,6 +937,111 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
+     * of 0.8 means the breaker should prevent trades that result in the value of the token dropping 20% or
+     * more relative to the rest of the pool.
+     */
+    function setCircuitBreaker(
+        IERC20 token,
+        uint256 lowerBoundPercentage,
+        uint256 upperBoundPercentage
+    ) external authenticate whenNotPaused {
+        _setCircuitBreaker(token, lowerBoundPercentage, upperBoundPercentage);
+    }
```
TomAFrench, line 1001: I've got some questions about whether this should accept an array of inputs. This would be consistent with `updateWeightsGradually` and avoid the manager having to call in many times.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
EndymionJkb: I am already using it in hasCircuitBreakerTripped

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
EndymionJkb: Will leave this, and specifically check that when setting

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
EndymionJkb: We do have a requirement to be able to set them individually, but there's no requirement here that the token list match the Vault tokens (e.g., you could pass in an array of 1), so that should be fine.

TomAFrench: Yeah, that's what I was picturing. Sorry the comparison to `updateWeightsGradually` was a bit confusing.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -921,6 +937,117 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
+     */
+    function hasCircuitBreakerTripped(IERC20 token) external view returns (bool, bool) {
+        return
+            CircuitBreakerLib.hasCircuitBreakerTripped(
+                _circuitBreakerState[token],
+                getActualSupply(),
+                _getNormalizedWeight(token),
+                _getUpscaledTokenBalance(token)
+            );
+    }
```
nventuro, line 986: This is a bit of a weird function - I imagine it'd almost always return false, since the whole point of circuit breakers is to prevent them from tripping by rejecting bad swaps. 

The only way this could return true is if the weights change to such a value that the current balances are bad. But even then, that is not useful information: it is possible to construct a swap that goes in the opposite direction and is accepted by the breaker.

Given this I'd just remove this function, as it is misleading and not very useful. Am I missing something?

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -921,6 +937,117 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
 
+    // Circuit Breakers
+
+    /**
+     * @notice Return the circuit breaker parameters for the given token.
+     * @dev These are the reference values (BPT price and weight complement) computed when the breaker was set,
+     * along with the percentage bounds. To get the current BPT price boundaries needed to check whether the
+     * circuit breaker should trip, call `getCurrentCircuitBreakerBounds`.
+     */
+    function getCircuitBreakerFields(IERC20 token)
```
nventuro, line 948: I'd merge this with `getCurrentBounds` and even `getBptPrice`, and return a single thing with the entire state. Realistically they'll all be used at the same time, and they all read essentially the same data.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -921,6 +937,117 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
+        uint256 lowerBoundPercentage,
+        uint256 upperBoundPercentage
+    ) private {
+        // Do not allow setting a circuit breaker on the BPT itself
+        _require(address(token) != address(this), Errors.INVALID_TOKEN);
+
+        uint256 normalizedWeight = _getNormalizedWeight(token);
+
+        // Note that `getBptPrice` will revert if the token is invalid.
+        uint256 bptPrice = getBptPrice(token, normalizedWeight);
```
nventuro, line 1019: We can't compute this value ourselves, as this is fully manipulable by third parties via sandwich attacks. This lets such an attacker set the bpt price to whatever they desire, granting them full control of the breaker. 

We've discussed this with @rabmarut already, and the Pool owner will send the reference bpt price for each breaker.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -921,6 +937,117 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
+        );
+
+        emit CircuitBreakerSet(token, bptPrice, lowerBoundPercentage, upperBoundPercentage);
+    }
+
+    /**
+     * @notice Get the price of a single token in terms of BPT, given the weight.
+     * @dev Returns an 18-decimal floating point number.
+     */
+    function getBptPrice(IERC20 token, uint256 normalizedWeight) public view returns (uint256) {
```
nventuro, line 1036: Why receive `weight`? We know that token's weight, there's not much of a point in this being an argument.

nventuro (changes requested):
Left some comments on the API, tldr:
 - we have the receive the bpt price
 - i'm not sure has breker tripped is useful
 - i'd strongly consider merging all or some of the getters

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
TomAFrench: I'd agree with this. The library function is useful for within swaps/joins/exits but probably isn't worth exposing like this.

EndymionJkb: If it's not useful for aggregators, etc., I can remove it. They could do their own calculations based on the settings they observe from the events.

TomAFrench: Yeah, it's more important to have a "what's the biggest trade I can do" function but that can be in a separate contract.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -925,6 +941,99 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
+        bytes32 circuitBreakerState = _circuitBreakerState[token];
+
+        (bptPrice, weightComplement, lowerBound, upperBound) = CircuitBreakerLib.getCircuitBreakerFields(
+            circuitBreakerState
+        );
+
+        (lowerBptPriceBound, upperBptPriceBound) = CircuitBreakerLib.getCurrentCircuitBreakerBounds(
+            circuitBreakerState,
+            _getNormalizedWeight(token).complement()
+        );
```
TomAFrench, line 973: We may want to split this getter in two. Generally when we want to use the breakers we'll know which direction the price is being pushed in, e.g. we only want to check the lower breaker on the ingoing token on a swap.

This means that during weight changes we'll be doing a costly computation of a bound which we don't need on every swap (x2 as this happens for each token).

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -925,6 +941,99 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
+     * @notice Get the price of a single token in terms of BPT, given the weight.
+     * @dev Returns an 18-decimal floating point number.
+     */
+    function getBptPrice(IERC20 token) external view returns (uint256) {
+        if (token == IERC20(this)) {
+            return FixedPoint.ONE;
+        }
+
+        return getActualSupply().mulUp(_getNormalizedWeight(token)).divDown(_getUpscaledTokenBalance(token));
+    }
```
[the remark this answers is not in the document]
nventuro: (though it'll be more expensive to implement externally)

nventuro (approved):

PR #1835, opened 2022-09-23
Add composable pool registration to PoolRegistrationLib

TomAFrench:
We ensure that the BPT is placed in the first position to ensure that it remains at a constant index.

### DIFF
--- a/pkg/pool-utils/contracts/lib/PoolRegistrationLib.sol
+++ b/pkg/pool-utils/contracts/lib/PoolRegistrationLib.sol
@@ -78,7 +78,7 @@ library PoolRegistrationLib {
         IVault.PoolSpecialization specialization,
         IERC20[] memory tokens,
         address[] memory assetManagers
-    ) internal returns (bytes32) {
+    ) private returns (bytes32) {
         bytes32 poolId = vault.registerPool(specialization);
 
         // We don't need to check that tokens and assetManagers have the same length, since the Vault already performs
--- a/pkg/pool-utils/contracts/lib/PoolRegistrationLib.sol
+++ b/pkg/pool-utils/contracts/lib/PoolRegistrationLib.sol
@@ -59,6 +59,10 @@ library PoolRegistrationLib {
 
         IERC20[] memory composableTokens = new IERC20[](tokens.length + 1);
         // We insert the Pool's BPT address into the first position.
+        // This allows us to know the position of the BPT token in the tokens array without explicitly tracking it.
+        // When deregistering a token, the token at the end of the array is moved into the index of the deregistered
+        // token, changing its index. By placing BPT at the beginning of the tokens array we can be sure that its index
+        // will never change unless it is deregistered itself (something which composable pools must prevent anyway).
         composableTokens[0] = IERC20(address(this));
         for (uint256 i = 0; i < tokens.length; i++) {
             composableTokens[i + 1] = tokens[i];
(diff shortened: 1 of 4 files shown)


pkg/pool-utils/test/PoolRegistrationLib.test.ts
```diff
@@ -112,6 +118,85 @@ describe('PoolRegistrationLib', function () {
...
+      const poolId = await registerComposablePool(PoolSpecialization.GeneralPool, assetManagers);
+      const { tokens: poolTokens } = await vault.getPoolTokens(poolId);
+
+      // The library mock is fulfilling the role of the Pool in this test.
+      const composableTokens = [lib.address, ...tokens.addresses];
+
+      expect(poolTokens).to.deep.eq(composableTokens);
+    });
+
+    it('reverts if the tokens are not sorted', async () => {
```
nventuro, line 155: Also add a test that checks we revert if one of the tokens is the Pool (library)

pkg/pool-utils/contracts/lib/PoolRegistrationLib.sol
```diff
@@ -41,6 +41,44 @@ library PoolRegistrationLib {
...
+    ) internal returns (bytes32) {
+        // The Vault only requires the token list to be ordered for the Two Token Pools specialization. However,
+        // to make the developer experience consistent, we are requiring this condition for all the native pools.
+        //
+        // Note that for Pools which can register and deregister tokens after deployment, this property may not hold
+        // as tokens which are added to the Pool after deployment are always added to the end of the array.
+        InputHelpers.ensureArrayIsSorted(tokens);
+
+        IERC20[] memory composableTokens = new IERC20[](tokens.length + 1);
+        // We insert the Pool's BPT address into the first position.
```
nventuro, line 61: We need to explain _why_ we're doing this (as long as it is never removed, BPT will always be at the zero index regardless of how many tokens are added and removed because of how the Vault performs that mutation, which is useful information to have as we can remove it by dropping the first elem and add it back by doing the opposite, without having to search for it).

nventuro: We could potentially fuzz this by deploying a Vault, calling this function with some initial tokens, and then repeatedly calling add or remove with arbitrary tokens. The BPT should always be at the first index.

Constructing such a test might be a bit tricky, but I have some ideas on how to do it if you think this is worth it.

TomAFrench: I've added a (pretty janky) fuzz test for this. Let me know if your ideas on how to do this are more elegant.

pkg/pool-utils/test/foundry/PoolRegistrationLib.t.sol
```diff
@@ -0,0 +1,51 @@
...
+    IVault private _vault;
+
+    function setUp() external {
+        _vault = new Vault(IAuthorizer(0), IWETH(0), 0, 0);
+    }
+
+    function testPositionZeroTokenFixed(uint8[10] memory tokenIds) external {
+        IERC20 bpt = IERC20(address(this));
+
+        IERC20[] memory tokens = new IERC20[](10);
```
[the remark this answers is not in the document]
TomAFrench: This is a noop.

pkg/pool-utils/test/foundry/PoolRegistrationLib.t.sol
```diff
@@ -0,0 +1,51 @@
...
+        bytes32 poolId = PoolRegistrationLib.registerComposablePool(
+            _vault,
+            IVault.PoolSpecialization.GENERAL,
+            tokens,
+            assetManagers
+        );
+
+        for (uint256 i = 0; i < tokenIds.length; i++) {
+            IERC20 token = IERC20(bound(tokenIds[i], 1, 10));
+            (IERC20[] memory poolTokens, , ) = _vault.getPoolTokens(poolId);
```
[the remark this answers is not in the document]
TomAFrench: We then need to sort the 'world' of tokens in order to build the initial state which is a bit of a pain.

nventuro (approved):

PR #1836, opened 2022-09-23
Refactor pool registration so that it happens on outermost layer

TomAFrench:
This allows us to modify how we register the pool (e.g. depending on whether the pool is composable or not) without BasePool having to know anything about this.


nventuro (approved):
And the refactor was thus complete, and they wept, for they saw it was Good.

PR #1837, opened 2022-09-26
Recovery Mode consistency

EndymionJkb:
Ensure the pool is in the opposite state before enabling or disabling Recovery Mode.

Done both ways, for consistency and future-proofing, though enabling it twice shouldn't hurt anything.


nventuro (approved):
Thanks for taking care of this!

PR #1839, opened 2022-09-26
Simplify ValueCompression condition

EndymionJkb:
When writing docs on Foundry tests, found a slight simplfiication.

### DIFF
--- a/pkg/pool-weighted/test/foundry/ValueCompression.t.sol
+++ b/pkg/pool-weighted/test/foundry/ValueCompression.t.sol
@@ -67,7 +67,7 @@ contract ValueCompressionTest is Test {
     ) external {
         vm.assume(maxUncompressedValue > 0);
         vm.assume(value <= maxUncompressedValue);
-        vm.assume(bitLength >= 2 && bitLength <= 255);
+        vm.assume(bitLength > 1);
 
         // Prevent internal overflows
         vm.assume(bitLength < 256 - mostSignificantBit(maxUncompressedValue));


TomAFrench:
This is written to match this condition in `ValueCompression`

https://github.com/balancer-labs/balancer-v2-monorepo/blob/3d4594cc2679deed7e7e8da24ee26894a9d889b7/pkg/pool-weighted/contracts/lib/ValueCompression.sol#L58

I think the benefit of matching the require statements in the real code (so we know that we're testing all potential inputs) is worth the extra verbosity.

nventuro:
@TomAFrench I think @EndymionJkb's point is that `bitLength` is a `uint8` in `WordCodec`, so there's no need to check for `<= 255`. We could also make it be a `uint8` in `ValueCompression`.

EndymionJkb:
> @TomAFrench I think @EndymionJkb's point is that `bitLength` is a `uint8` in `WordCodec`, so there's no need to check for `<= 255`. We could also make it be a `uint8` in `ValueCompression`.

Yes; I was writing a Notion doc about the option of using typed uints for input restriction in Foundry (so that it does it for you, and you don't need either assume or bound), using this as an example.

TomAFrench:
> I think @EndymionJkb's point is that bitLength is a uint8 in WordCodec, so there's no need to check for <= 255. We could also make it be a uint8 in ValueCompression.

I'm not entirely sure what you mean here. `WordCodec` takes a uint256 and enforces validity through require statements (one of which being equivalent to restricting values to a uint8). Are you suggesting we change the libraries or the tests?

My point is that fuzzing tests should optimise for being clear about what the range of inputs are and why. Having the condition as it is makes it clear that we're matching that particular require statement to give meaning on why we're adding that `vm.assume`

EndymionJkb:
> I'm not entirely sure what you mean here. `WordCodec` takes a uint256 and enforces validity through require statements (one of which being equivalent to restricting values to a uint8). Are you suggesting we change the libraries or the tests?
> 
> My point is that fuzzing tests should optimise for being clear about what the range of inputs are and why. Having the condition as it is makes it clear that we're matching that particular require statement to give meaning on why we're adding that `vm.assume`

We've generally always used uint256 for libraries; I'd say don't change anything there. To retain consistency, maybe make it uint256 in the test as well, and `bound(bitLength, 2, 255)` ?

nventuro:
Ah I'm sorry, I misread - thought the test was the code. Agree with Tom then - the extra verbosity in the test does no damage at all, and makes it clear we're imposing the exact same condition.

PR #1843, opened 2022-09-26
Add exitSwap implementation and fuzz test

### LINKED ISSUE #1812: Provide specialized join/exitswap variants for WeightedMath
In ComposableStablePool we simply used the versions that receive arrays and zeroed out most of the inputs, but here we won't have all balances. These functions don't actually care about balances however if the inputs are zero, so we can simply remove those.

It'd be good to have strong e.g. fuzz testing to check for equivalence.

TomAFrench:
closes #1812


nventuro (approved):
I added some comments for clarity, plus a new `assume` that lets us test a wider range of input values without increasing total run time.

PR #1846, opened 2022-09-26
Removing setSwapFeePercentage in managed pool.

### LINKED ISSUE #1750: Streamline protocol fee setting
Managed Pools currently have multiple ways to change their swap fee. We should reduce these to just one (with start and end dates), to simplify the overall model. See [this comment](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1749#discussion_r969892359) for reference.

jubeira:
Gradual swap fee update can achieve the same with the right parameters.

Closes #1750.

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
@@ -228,30 +228,9 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
     }
 
     /**
-     * @notice Set the swap fee percentage.
-     * @dev This is a permissioned function, and disabled if the pool is paused. The swap fee must be within the
-     * bounds set by MIN_SWAP_FEE_PERCENTAGE/MAX_SWAP_FEE_PERCENTAGE. Emits the SwapFeePercentageChanged event.
-     */
-    function setSwapFeePercentage(uint256 swapFeePercentage) external override authenticate whenNotPaused {
-        // Do not allow setting if there is an ongoing fee change
-        uint256 currentTime = block.timestamp;
-        bytes32 poolState = _poolState;
-        (uint256 startTime, uint256 endTime, , ) = ManagedPoolStorageLib.getSwapFeeFields(poolState);
-
-        if (currentTime < endTime) {
-            _revert(
-                currentTime < startTime ? Errors.SET_SWAP_FEE_PENDING_FEE_CHANGE : Errors.SET_SWAP_FEE_DURING_FEE_CHANGE
-            );
-        }
-
-        _poolState = ManagedPoolSwapFeesLib.setSwapFeePercentage(poolState, swapFeePercentage);
-    }
-
-    /**
      * @notice Schedule a gradual swap fee update.
      * @dev The swap fee will change from the given starting value (which may or may not be the current
-     * value) to the given ending fee percentage, over startTime to endTime. Calling this with a starting
-     * value avoids requiring an explicit external `setSwapFeePercentage` call.
+     * value) to the given ending fee percentage, over startTime to endTime.
      *
      * Note that calling this with a starting swap fee different from the current value will immediately change the
      * current swap fee to `startSwapFeePercentage` (including emitting the SwapFeePercentageChanged event),
@@ -269,7 +248,7 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
         uint256 endTime,
         uint256 startSwapFeePercentage,
         uint256 endSwapFeePercentage
-    ) external authenticate whenNotPaused nonReentrant {
+    ) external override authenticate whenNotPaused nonReentrant {
         _poolState = ManagedPoolSwapFeesLib.startGradualSwapFeeChange(
             _poolState,
             startTime,
@@ -913,7 +892,6 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
             (actionId == getActionId(ManagedPoolSettings.updateWeightsGradually.selector)) ||
             (actionId == getActionId(ManagedPoolSettings.updateSwapFeeGradually.selector)) ||
             (actionId == getActionId(ManagedPoolSettings.setSwapEnabled.selector)) ||
-            (actionId == getActionId(ManagedPoolSettings.setSwapFeePercentage.selector)) ||
             (actionId == getActionId(ManagedPoolSettings.addAllowedAddress.selector)) ||
             (actionId == getActionId(ManagedPoolSettings.removeAllowedAddress.selector)) ||
             (actionId == getActionId(ManagedPoolSettings.setMustAllowlistLPs.selector)) ||
--- a/pkg/pool-weighted/test/ManagedPool.test.ts
+++ b/pkg/pool-weighted/test/ManagedPool.test.ts
@@ -2,7 +2,13 @@ import { ethers } from 'hardhat';
 import { expect } from 'chai';
 import { BigNumber, ContractReceipt } from 'ethers';
 
-import { DAY, advanceTime, receiptTimestamp } from '@balancer-labs/v2-helpers/src/time';
+import {
+  DAY,
+  advanceTime,
+  receiptTimestamp,
+  currentTimestamp,
+  setNextBlockTimestamp,
+} from '@balancer-labs/v2-helpers/src/time';
 import { BigNumberish, bn, fp, pct } from '@balancer-labs/v2-helpers/src/numbers';
 import * as expectEvent from '@balancer-labs/v2-helpers/src/test/expectEvent';
 import TokenList from '@balancer-labs/v2-helpers/src/models/tokens/TokenList';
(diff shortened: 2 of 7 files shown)


TomAFrench:
We should remove `ManagedPoolSwapFeesLib.setSwapFeePercentage` and inline the rest of `ManagedPoolSwapFeesLib` into `ManagedPoolSettings` imo..

pkg/pool-weighted/contracts/managed/ManagedPoolSwapFeesLib.sol
```diff
@@ -1,104 +0,0 @@
...
-        uint256 startTime,
-        uint256 endTime,
-        uint256 startSwapFeePercentage,
-        uint256 endSwapFeePercentage
-    ) internal returns (bytes32) {
-        validateSwapFeePercentage(startSwapFeePercentage);
-        validateSwapFeePercentage(endSwapFeePercentage);
-
-        if (startSwapFeePercentage != ManagedPoolStorageLib.getSwapFeePercentage(poolState)) {
-            poolState = setSwapFeePercentage(poolState, startSwapFeePercentage);
```
jubeira, line 88: Note for the reviewer: with these changes, we are not emitting `SwapFeePercentageChanged` event when the start swap fee is different from the current one. Let's double check if this is alright before merging.

jubeira:
Some comments as a quick guide to consider while reviewing:
- Swap fees lib was deleted, and its functionality incorporated to `ManagedPoolSettings`, with the exception of the `setSwapFeePercentage` function.
- Swap fee lib tests were incorporated to managed pool settings test with minimal changes that (mostly) don't affect behavior.
- The only difference in behavior should be https://github.com/balancer-labs/balancer-v2-monorepo/pull/1846#discussion_r982321031.
- About the changes in the helper: I was a bit hesitant to add a mock type to the weighted pool infrastructure. I thought it'd be useful to be able to deploy a mock managed pool, given that the setup is rather complex and I wanted to be able to expose internal methods for testing -which may be useful in the future as well-. For this PR, however, the only method that needed to be exposed for testing was `validateSwapFeePercentage`; the other internal methods from the swap fee lib already had external versions in `ManagedPoolSettings`

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -60,6 +59,13 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
 
     uint256 private constant _MAX_MANAGEMENT_AUM_FEE_PERCENTAGE = 1e17; // 10%
 
+    uint256 private constant _MIN_SWAP_FEE_PERCENTAGE = 1e12; // 0.0001%
+
+    // The swap fee cannot be 100%: calculations that divide by (1-fee) would revert with division by zero.
+    // Swap fees close to 100% can still cause reverts when performing join/exit swaps, if the calculated fee
+    // amounts exceed the pool's token balances in the Vault. 80% is a very high, but relatively safe maximum value.
+    uint256 private constant _MAX_SWAP_FEE_PERCENTAGE = 80e16; // 80%
+
```
TomAFrench, line 68: Can we put this above `_MAX_MANAGEMENT_AUM_FEE_PERCENTAGE`?

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -82,6 +88,16 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
 
     // Event declarations
 
+    // ManagedPool swap fees can change over time: these periods are expected to be long enough (e.g. days)
+    // that any timestamp manipulation would achieve very little.
+    // solhint-disable not-rely-on-time
```
TomAFrench, line 93:
```suggestion
```
This is already included at the top of the contract.

pkg/pool-weighted/test/ManagedPool.test.ts
```diff
@@ -319,7 +327,17 @@ describe('ManagedPool', function () {
 
     context('with the max swap fee', () => {
       sharedBeforeEach('set swap fee to the max value (< 100%)', async () => {
-        await pool.setSwapFeePercentage(owner, MAX_SWAP_FEE_PERCENTAGE);
+        // In practice, a separate contract would call `updateSwapFeeGradually` using `block.timestamp` both as start
+        // and endTime to make the change immediately.
+        const blockTimestamp = (await currentTimestamp()).add(1);
+        await setNextBlockTimestamp(blockTimestamp);
```
TomAFrench, line 333:
```suggestion
        const blockTimestamp = await currentTimestamp();
        await setNextBlockTimestamp(blockTimestamp.add(1));
```

pkg/pool-weighted/contracts/managed/ManagedPoolSwapFeesLib.sol
TomAFrench: This is fine, If you're tracking the swap fee through events then you can just ignore these events entirely and just listen to the gradual update events.

pkg/pool-weighted/test/ManagedPool.test.ts
jubeira: I've rephrased this one a bit.

If we send `start == end` to `updateSwapFeeGradually` and the timestamp is in the past, the function will resolve start time to the present, and revert because `resolvedStart > end`. For this to work 'instantly', we need to mock the call to `updateSwapFeeGradually` with `block.timestamp`.

pvt/helpers/src/models/pools/weighted/WeightedPoolDeployer.ts
```diff
@@ -122,6 +124,32 @@ export default {
         });
         break;
       }
+      case WeightedPoolType.MOCK_MANAGED_POOL: {
+        assert(mockContractName != undefined, 'Mock contract name required to deploy mock base pool');
```
TomAFrench, line 128:
```suggestion
        if (mockContractName == undefined) {
           throw new Error('Mock contract name required to deploy mock base pool');
        }
```

pvt/helpers/src/models/pools/weighted/WeightedPool.ts
```diff
@@ -527,12 +528,17 @@ export default class WeightedPool extends BasePool {
...
+    return this.poolType == WeightedPoolType.MANAGED_POOL || this.poolType == WeightedPoolType.MOCK_MANAGED_POOL;
+  }
+
   async setSwapEnabled(from: SignerWithAddress, swapEnabled: boolean): Promise<ContractTransaction> {
     const pool = this.instance.connect(from);
     return pool.setSwapEnabled(swapEnabled);
   }
 
   async setSwapFeePercentage(from: SignerWithAddress, swapFeePercentage: BigNumberish): Promise<ContractTransaction> {
+    assert(!this._isManagedPool(), 'Not available in managed pool');
```
TomAFrench, line 541:
```suggestion
         if (this._isManagedPool()) {
           throw new Error('Not available in managed pool');
        }
```

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -80,6 +86,13 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
 
     // Event declarations
 
+    event GradualSwapFeeUpdateScheduled(
+        uint256 startTime,
+        uint256 endTime,
+        uint256 startSwapFeePercentage,
+        uint256 endSwapFeePercentage
+    );
+
```
TomAFrench, line 95:
```suggestion
```

TomAFrench (approved):

pvt/helpers/src/models/pools/weighted/WeightedPoolDeployer.ts
jubeira: Thanks; I was in doubt because I only saw it in benchmarks :)

nventuro (commented):
Didn't submit these in time before the merge :(

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -294,6 +285,45 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
+     * @return poolState - The modified Pool state with the updated swap fee data. It's the responsiblity of the caller
+     * to write this to storage so this value is persisted.
+     */
+    function _startGradualSwapFeeChange(
+        bytes32 poolState,
+        uint256 startTime,
+        uint256 endTime,
+        uint256 startSwapFeePercentage,
+        uint256 endSwapFeePercentage
+    ) internal returns (bytes32) {
```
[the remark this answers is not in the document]
jubeira: if we go that way we should probably do the same with `_startGradualWeightChange` for consistency; it sounds counter-intuitive to have them behave differently.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -56,6 +55,13 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
     // creation gas consumption.
     uint256 private constant _MAX_TOKENS = 38;
 
+    // The swap fee cannot be 100%: calculations that divide by (1-fee) would revert with division by zero.
+    // Swap fees close to 100% can still cause reverts when performing join/exit swaps, if the calculated fee
+    // amounts exceed the pool's token balances in the Vault. 80% is a very high, but relatively safe maximum value.
+    uint256 private constant _MAX_SWAP_FEE_PERCENTAGE = 80e16; // 80%
```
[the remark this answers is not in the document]
rabmarut: Target: 99%
Compromise: 95%
Threshold: 90%

PR #1847, opened 2022-09-26
Add some groundwork for composable tests

TomAFrench:
This PR updates the ManagedPool tests so that they're more friendly to making ManagedPool composable. Essentially we're making the WeightedPool model query the pool's tokens before joins/exits and testing transitions rather than expected final states (as we should anyway) so we can hook in the composable logic more easily.


nventuro (approved):
So we'd modify `getTokens` to drop BPT? Sounds like a plan!

TomAFrench:
Yep, that's the plan.

PR #1850, opened 2022-09-26
Add getters for the ManagedPool's actualSupply

TomAFrench:
This PR adds a basic implementation of `getActualSupply` for ManagedPool and the associated tests to make it easier to make ManagedPool composable in future.

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
@@ -170,6 +170,36 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
         return _tokenState[token];
     }
 
+    // Actual Supply
+
+    /**
+     * @notice Returns the effective BPT supply.
+     *
+     * @dev The Pool owes debt to the Protocol in the form of unminted BPT, which will be minted immediately before the
+     * next join or exit. We need to take these into account since, even if they don't yet exist, they will
+     * effectively be included in any Pool operation that involves BPT.
+     *
+     * In the vast majority of cases, this function should be used instead of `totalSupply()`.
+     */
+    function getActualSupply() external view returns (uint256) {
+        return _getActualSupply(totalSupply());
+    }
+
+    function _getActualSupply(uint256 virtualSupply) internal view returns (uint256) {
+        if (inRecoveryMode()) {
+            // If we're in recovery mode then we bypass any fee logic and perform an early return.
+            return virtualSupply;
+        }
+
+        uint256 aumFeesAmount = ProtocolAUMFees.getAumFeesBptAmount(
+            virtualSupply,
+            block.timestamp,
+            _lastAumFeeCollectionTimestamp,
+            getManagementAumFeePercentage()
+        );
+        return virtualSupply.add(aumFeesAmount);
+    }
+
     // Swap fees
 
     /**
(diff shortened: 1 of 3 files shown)


pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -496,6 +496,27 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
         emit MustAllowlistLPsSet(mustAllowlistLPs);
     }
 
+    // Actual Supply
+
+    function getActualSupply() external view returns (uint256) {
```
nventuro, line 501:
```suggestion
    // Actual Supply
    
    /**
     * @dev Returns the effective BPT supply.
     *
     * In other pools, this would be the same as `totalSupply`, but there are two key differences here:
     *  - this pool pre-mints BPT and holds it in the Vault as a token, and as such we need to subtract the Vault's
     *    balance to get the total "circulating supply". This is called the 'virtualSupply'.
     *  - the Pool owes debt to the Protocol in the form of unminted BPT, which will be minted immediately before the
     *    next join or exit. We need to take these into account since, even if they don't yet exist, they will
     *    effectively be included in any Pool operation that involves BPT.
     *
     * In the vast majority of cases, this function should be used instead of `totalSupply()`.
     */
    function getActualSupply() external view returns (uint256) {
```

I pasted the comment from ComposableStablePool - it should be different here since a) Settings doesn't technically know about composability, and b) we owe both protocol _and_ manager fees. Feel free to edit as you see fit.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -496,6 +496,27 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
+        return _getActualSupply(totalSupply());
+    }
+
+    function _getActualSupply(uint256 virtualSupply) internal view returns (uint256) {
+        if (inRecoveryMode()) {
```
nventuro, line 506: `inRecoveryMode` implicitly access `_poolState`: it should be `external` to avoid accidentally calling it like this, and we should call a version that receives the state.

(or learn to stop worrying and love global variables)

pkg/pool-weighted/test/ManagedPool.test.ts
```diff
@@ -388,6 +388,28 @@ describe('ManagedPool', function () {
...
+        it('syncs the total supply to the actual supply', async () => {
+          // As we're performing a join or exit here we need to account for the change in the BPT total supply due to
+          // the join/exit. We do this by tracking the user's balance.
+          const balanceBefore = await pool.balanceOf(other);
+          const totalSupplyBefore = await pool.totalSupply();
+          const expectedManagementFeeBpt = expectedAUMFees(totalSupplyBefore, managementAumFeePercentage, timeElapsed);
+
+          const expectedActualSupply = totalSupplyBefore.add(expectedManagementFeeBpt);
+          const actualSupplyBefore = await pool.getActualSupply();
+          expect(actualSupplyBefore).to.be.equalWithError(expectedActualSupply, 0.0001);
```
nventuro, line 401: The fees are likely very small. Can we decrease the relative error here, or even test for strict equality?

pkg/pool-weighted/test/ManagedPoolSettings.test.ts
```diff
@@ -660,13 +660,16 @@ describe('ManagedPoolSettings', function () {
   });
 
   describe('recovery mode', () => {
+    const managementAumFeePercentage = fp(0.01);
```
nventuro, line 663: We can make this much larger so that it doesn't get lost in rounding error

pkg/pool-weighted/test/ManagedPool.test.ts
```diff
@@ -388,6 +388,28 @@ describe('ManagedPool', function () {
             bptAmount: actualManagementFeeBpt,
           });
         });
+
+        it('syncs the total supply to the actual supply', async () => {
+          // As we're performing a join or exit here we need to account for the change in the BPT total supply due to
+          // the join/exit. We do this by tracking the user's balance.
+          const balanceBefore = await pool.balanceOf(other);
+          const totalSupplyBefore = await pool.totalSupply();
```
nventuro, line 396: I think these tests are doing too many things. I'd have a) one test that checks getActualSupply returns the correct value, b) one that checks that it gets synced after the operation, and c) one that checks that the operation mints the difference between virtual and actual

nventuro: (test c) is more of a maybe)

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
TomAFrench: This would a be change to `RecoveryMode` so I'm going to leave the visibility alone but change to access `_poolState` directly.

nventuro (approved):

PR #1851, opened 2022-09-27
Pass `virtualSupply` to `_beforeJoinExit`

TomAFrench:
When ManagedPool is composable then we'll need to pass in the virtual supply rather than calling `totalSupply()`. To smooth this transition we can pass in the total supply prior to the switch.

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPool.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPool.sol
@@ -186,15 +186,14 @@ contract ManagedPool is ManagedPoolSettings {
      * @dev Called before any join or exit operation. Returns the Pool's total supply by default, but derived contracts
      * may choose to add custom behavior at these steps. This often has to do with protocol fee processing.
      */
-    function _beforeJoinExit() internal returns (uint256) {
+    function _beforeJoinExit(uint256 virtualSupply) internal returns (uint256) {
         // The AUM fee calculation is based on inflating the Pool's BPT supply by a target rate.
         // We then must collect AUM fees whenever joining or exiting the pool to ensure that LPs only pay AUM fees
         // for the period during which they are an LP within the pool: otherwise an LP could shift their share of the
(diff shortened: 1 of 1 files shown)


pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -186,15 +186,14 @@ contract ManagedPool is ManagedPoolSettings {
...
         // The AUM fee calculation is based on inflating the Pool's BPT supply by a target rate.
         // We then must collect AUM fees whenever joining or exiting the pool to ensure that LPs only pay AUM fees
         // for the period during which they are an LP within the pool: otherwise an LP could shift their share of the
         // AUM fees onto the remaining LPs in the pool by exiting before they were paid.
-        uint256 supplyBeforeFeeCollection = totalSupply();
-        (uint256 protocolAUMFees, uint256 managerAUMFees) = _collectAumManagementFees(supplyBeforeFeeCollection);
+        (uint256 protocolAUMFees, uint256 managerAUMFees) = _collectAumManagementFees(virtualSupply);
 
-        return supplyBeforeFeeCollection.add(protocolAUMFees + managerAUMFees);
+        return virtualSupply.add(protocolAUMFees + managerAUMFees);
```
nventuro, line 196: Unrelated to this PR, but I'm not a fan of this function. I'd vote for inlining.

nventuro (approved):

pkg/pool-weighted/contracts/managed/ManagedPool.sol
nventuro: 'before join exit' conveys pretty much no information, and it adds a (very small) layer of obscurity. I'd rather the thing be more explicit about what it's doing.

TomAFrench: Yeah, seems like it's just a holdover from ComposableStablePool at this point.

PR #1855, opened 2022-09-27
Inline `_beforeJoinExit`

TomAFrench:


pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -244,14 +230,20 @@ contract ManagedPool is ManagedPoolSettings {
         uint256[] memory scalingFactors = _scalingFactors(tokens);
         _upscaleArray(balances, scalingFactors);
 
-        uint256 preJoinExitSupply = _beforeJoinExit(totalSupply());
+        // The AUM fee calculation is based on inflating the Pool's BPT supply by a target rate.
+        // We then must collect AUM fees whenever joining or exiting the pool to ensure that LPs only pay AUM fees
+        // for the period during which they are an LP within the pool: otherwise an LP could shift their share of the
+        // AUM fees onto the remaining LPs in the pool by exiting before they were paid.
+        uint256 preFeeCollectionSupply = totalSupply();
+        (uint256 protocolAUMFees, uint256 managerAUMFees) = _collectAumManagementFees(preFeeCollectionSupply);
```
nventuro, line 238: Can't this just return the sum?

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -244,14 +230,20 @@ contract ManagedPool is ManagedPoolSettings {
...
+        uint256 postFeeCollectionSupply = preFeeCollectionSupply + protocolAUMFees + managerAUMFees;
```
nventuro, line 239: The safemath gods are not pleased by this

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -244,14 +230,20 @@ contract ManagedPool is ManagedPoolSettings {
         uint256[] memory scalingFactors = _scalingFactors(tokens);
         _upscaleArray(balances, scalingFactors);
 
-        uint256 preJoinExitSupply = _beforeJoinExit(totalSupply());
+        // The AUM fee calculation is based on inflating the Pool's BPT supply by a target rate.
+        // We then must collect AUM fees whenever joining or exiting the pool to ensure that LPs only pay AUM fees
+        // for the period during which they are an LP within the pool: otherwise an LP could shift their share of the
+        // AUM fees onto the remaining LPs in the pool by exiting before they were paid.
+        uint256 preFeeCollectionSupply = totalSupply();
```
nventuro, line 237: Careful when merging with #1854 l

nventuro, line 237: We could call this `actualSupply`, no?

nventuro (approved):

pkg/pool-weighted/contracts/managed/ManagedPool.sol
nventuro: (well, not this, but this + the fees)

pkg/pool-weighted/contracts/managed/ManagedPool.sol
TomAFrench: The external function `collectAumManagementFees()` returns just the managed portion (to give the manager a way to calculate how many fees they'd receive. To make this only return a single value we'd have to change the return value from  `collectAumManagementFees()`.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
TomAFrench: The safemath gods are sated by a previous offering from when these fees were minted.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
nventuro: @TomAFrench ?

TomAFrench: This is now calling `_getVirtualSupply` but once we make it composable it will be using a local value so this is very temporary.

PR #1861, opened 2022-09-28
Add updateBoundRatios

EndymionJkb:
Feature suggested by Tom, to optimize circuit breaker bounds checking for weight changes. Assuming the weights will be static for comparatively long periods, and that we will likely not reset the bounds entirely (i.e., including updating the BPT price), every time weights are changed.

This resets the weight complement and recalculates the bound ratios of an existing circuit breaker, without changing the BPT price or bounds.

PR #1862, opened 2022-09-28
Add a function to refresh the cache of the circuit breaker bound ratios

TomAFrench:
As shown in the fuzz test we can call `CircuitBreakerLib.updateBoundRatios(circuitBreakerState, newWeightComplement)` to update the cached bound ratios to target a new weight complement. This reduces the costs of checking the circuit breakers while introducing negligible error.

This should save us about 5k gas per token.


TomAFrench:
Closing in favour of #1861

PR #1864, opened 2022-09-28
Addressing managed pool settings outstanding comments

jubeira:
See https://github.com/balancer-labs/balancer-v2-monorepo/pull/1846#pullrequestreview-1124092238.

https://github.com/balancer-labs/balancer-v2-monorepo/pull/1846#discussion_r982687923 was addressed in a separate commit https://github.com/balancer-labs/balancer-v2-monorepo/commit/40428a9d767adbb259887e7cffdd0ea8ab32ebdc in case we don't want to go that way.

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
@@ -253,8 +253,8 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
      * value) to the given ending fee percentage, over startTime to endTime.
      *
      * Note that calling this with a starting swap fee different from the current value will immediately change the
-     * current swap fee to `startSwapFeePercentage` (including emitting the SwapFeePercentageChanged event),
-     * before commencing the gradual change at `startTime`. Emits the GradualSwapFeeUpdateScheduled event.
+     * current swap fee to `startSwapFeePercentage`, before commencing the gradual change at `startTime`.
+     * Emits the GradualSwapFeeUpdateScheduled event.
      * This is a permissioned function.
      *
      * @param startTime - The timestamp when the swap fee change will begin.
--- a/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
@@ -142,15 +142,8 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
             _tokenState[token] = ManagedPoolTokenLib.setTokenScalingFactor(bytes32(0), token);
         }
 
-        // This bytes32 holds a lot of the core Pool state which is read on most interactions, by keeping it in a single
-        // word we can save gas from unnecessary storage reads. It includes items like:
-        // - Swap fees
-        // - Weight change progress
-        // - Various feature flags
-        bytes32 poolState;
-
-        poolState = _startGradualWeightChange(
-            poolState,
+        // We write the pool state here, as both `_setSwapEnabled` and `_setMustAllowlistLPs` read it from storage.
+        _startGradualWeightChange(
             block.timestamp,
             block.timestamp,
             params.normalizedWeights,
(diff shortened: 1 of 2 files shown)


TomAFrench:
This seems to have a deployment task mixed in as well

jubeira:
Thanks Tom -- the base had a WIP commit from another branch :/

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -142,33 +142,22 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
-        // This bytes32 holds a lot of the core Pool state which is read on most interactions, by keeping it in a single
-        // word we can save gas from unnecessary storage reads. It includes items like:
-        // - Swap fees
-        // - Weight change progress
-        // - Various feature flags
-        bytes32 poolState;
-
-        poolState = _startGradualWeightChange(
-            poolState,
+        // We write the pool state here, as both `_setSwapEnabled` and `_setMustAllowlistLPs` read it from storage.
```
TomAFrench, line 145:
```suggestion
```

I think we can get rid of this. We only had this comment because we were moving from manipulating the state in memory to reading from storage.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -295,37 +278,33 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
...
         uint256 startSwapFeePercentage,
         uint256 endSwapFeePercentage
-    ) internal returns (bytes32) {
+    ) internal {
         _validateSwapFeePercentage(startSwapFeePercentage);
         _validateSwapFeePercentage(endSwapFeePercentage);
 
         startTime = GradualValueChange.resolveStartTime(startTime, endTime);
 
         emit GradualSwapFeeUpdateScheduled(startTime, endTime, startSwapFeePercentage, endSwapFeePercentage);
```
TomAFrench, line 299: I'd like to move this below the assignment to `_poolState` now we aren't returning it.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -462,7 +433,7 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, ReentrancyG
 
         emit GradualWeightUpdateScheduled(startTime, endTime, startWeights, endWeights);
```
TomAFrench, line 434: ditto.

pkg/pool-weighted/test/ManagedPoolSettings.test.ts
```diff
@@ -641,72 +641,56 @@ describe('ManagedPoolSettings', function () {
...
+                blockTimestamp,
+                blockTimestamp.add(DAY),
+                UPDATE_FEE_START,
+                MIN_SWAP_FEE
+              );
+            });
+
+            it('instantly sets the swap fee with the starting value', async () => {
+              expect(await pool.getSwapFeePercentage()).to.be.equal(UPDATE_FEE_START);
             });
```
TomAFrench, line 660: I'm not a fan of how we're not using the defined start and end timestamps and swap fees in here as it goes against why we encapsulate these tests into a function.

I would do something like this.
```
    // Enforce that the Pool's current swap fee is different from the starting swap fee.
    await pool.updateSwapFeeGradually(caller, now, endTime, MAX_SWAP_FEE, MAX_SWAP_FEE);
    expect(await pool.getSwapFeePercentage()).to.not.equal(START_SWAP_FEE);
});

it('instantly sets the swap fee with the starting value', async () => {
    await pool.updateSwapFeeGradually(caller, startTime, endTime, START_SWAP_FEE, END_SWAP_FEE);

    expect(await pool.getSwapFeePercentage()).to.be.equal(START_SWAP_FEE);
});
```

TomAFrench (approved):
Looks good. Some minor rearranging in the contracts and I'd do the tests slightly differently.

pkg/pool-weighted/test/ManagedPoolSettings.test.ts
jubeira: I like it!

jubeira:
Thanks for the input! Comments addressed; I'll merge after CI goes green :)

PR #1865, opened 2022-09-28
Replace references to total supply in tests with virtual supply

TomAFrench:
To make the `ManagedPool/ManagedPoolSettings` tests agnostic to whether the pool is composable or not we should replace any references to the total supply in the tests with virtual supply.

### DIFF
--- a/pkg/pool-weighted/contracts/test/MockManagedPool.sol
+++ b/pkg/pool-weighted/contracts/test/MockManagedPool.sol
@@ -29,6 +29,10 @@ contract MockManagedPool is ManagedPool {
         // solhint-disable-previous-line no-empty-blocks
     }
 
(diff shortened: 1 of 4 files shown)


EndymionJkb (approved):
Looks like it's doing what it says: replacing total with virtual.

I never liked these names, but I guess we're constrained by other deployments. To me, "virtual" (aka phantom) *should* mean the preminted BPT in the Vault. We've referred to "total - preminted" as the circulatingSupply in the docs, which is a better name than "virtual". (And then we have actualSupply, which is circulating/virtual + pending BPT fees.)

TomAFrench:
Yeah, it doesn't sit well with me either but I didn't want to open that can of worms just yet.

PR #1866, opened 2022-09-28
Deployment preparation: managed pool

jubeira:
Prepared on top of #1864; base to be adjusted when the dust is settled.


pkg/deployments/tasks/20221021-managed-pool/index.ts
```diff
@@ -0,0 +1,16 @@
...
+  const input = task.input() as ManagedPoolDeployment;
+
+  const addRemoveTokenLib = await task.deployAndVerify('ManagedPoolAddRemoveTokenLib');
+  const circuitBreakerLib = await task.deployAndVerify('CircuitBreakerLib');
+
+  const args = [input.Vault, input.ProtocolFeePercentagesProvider];
+  await task.deployAndVerify('ManagedPoolFactory', args, from, force, {
+    CircuitBreakerLib: circuitBreakerLib.address,
+    ManagedPoolAddRemoveTokenLib: addRemoveTokenLib.address,
+  });
```
nventuro, line 15: You can also call `getWeightedMath` and verify that

PR #1856, opened 2022-09-27
Make ManagedPool composable

TomAFrench:
Draft PR as a number of the tests are broken due to `totalSupply` being inflated. I'd like to move `ManagedPoolSettings` tests over to use a non-composable mock before we proceed with this PR.

I've got PRs for adding joinSwaps and exitSwaps locally but they needs tests added to them.

PR #1867, opened 2022-09-29
Add joinSwaps to ManagedPool

TomAFrench:
Builds on #1856

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPool.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPool.sol
@@ -120,19 +120,86 @@ contract ManagedPool is ManagedPoolSettings {
         SwapRequest memory request,
         uint256 balanceTokenIn,
         uint256 balanceTokenOut
-    ) internal view override returns (uint256) {
+    ) internal override returns (uint256) {
         bytes32 poolState = _getPoolState();
         _require(ManagedPoolStorageLib.getSwapsEnabled(poolState), Errors.SWAPS_DISABLED);
 
         // solhint-disable no-empty-blocks
         if (request.tokenOut == IERC20(this)) {
-            // Do a joinSwap
+            // Check allowlist for LPs, if applicable
+            _require(isAllowedAddress(request.from), Errors.ADDRESS_NOT_ALLOWLISTED);
+
+            // balanceTokenOut is the amount of BPT held by the Pool in the Vault.
+            // Subtracting this from the total supply gives us the virtual supply.
+            uint256 virtualSupply = totalSupply() - balanceTokenOut;
+
+            // The AUM fee calculation is based on inflating the Pool's BPT supply by a target rate.
+            // We then must collect AUM fees whenever joining or exiting the pool to ensure that LPs only pay AUM fees
+            // for the period during which they are an LP within the pool: otherwise an LP could shift their share of
+            // the AUM fees onto the remaining LPs in the pool by exiting before they were paid.
+            virtualSupply += _collectAumManagementFees(virtualSupply);
+
+            return _onJoinSwap(request, balanceTokenIn, virtualSupply, poolState);
         } else if (request.tokenIn == IERC20(this)) {
             // Do an exitSwap
         } else {
             return _onTokenSwap(request, balanceTokenIn, balanceTokenOut, poolState);
         }
         // solhint-enable no-empty-blocks
+    }
+
+    /*
+     * @dev Called when a swap with the Pool occurs, where the tokens leaving the Pool are BPT.
+     *
+     * This function is responsible for upscaling any amounts received, in particular `balanceTokenIn`
+     * and `request.amount`.
+     *
+     * The return value is expected to be downscaled (appropriately rounded based on the swap type) ready to be passed
+     * to the Vault.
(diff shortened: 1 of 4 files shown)


TomAFrench:
The failures on `exitPool` tests must be something to do with switching from a real to mocked Vault.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -120,13 +120,26 @@ contract ManagedPool is ManagedPoolSettings {
...
+
+            // balanceTokenOut is the amount of BPT held by the Pool in the Vault.
+            // Subtracting this from the total supply gives us the virtual supply.
+            uint256 virtualSupply = totalSupply() - balanceTokenOut;
+
+            // The AUM fee calculation is based on inflating the Pool's BPT supply by a target rate.
+            // We then must collect AUM fees whenever joining or exiting the pool to ensure that LPs only pay AUM fees
+            // for the period during which they are an LP within the pool: otherwise an LP could shift their share of
+            // the AUM fees onto the remaining LPs in the pool by exiting before they were paid.
+            uint256 actualSupply = virtualSupply + _collectAumManagementFees(virtualSupply);
```
nventuro, line 140: This seems rather odd - why do this here and not in `_onJoinSwap`?

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -135,6 +148,60 @@ contract ManagedPool is ManagedPoolSettings {
...
+                tokenInWeight,
+                request.amount,
+                actualSupply,
+                swapFeePercentage
+            );
+
+            // BPT doesn't need scaling so we can return immediately.
+            return amountOut;
+        } else {
+            uint256 amountIn = WeightedMath._calcTokenInGivenExactBptOut(
```
nventuro, line 192:
```suggestion
            // request.amount is a BPT amount, so it doesn't need scaling.
            uint256 amountIn = WeightedMath._calcTokenInGivenExactBptOut(
```

pkg/pool-weighted/contracts/managed/ManagedPool.sol
nventuro: I guess ultimately it doesn't matter too much, we can always rearrange these later on.

pkg/pool-weighted/test/ManagedPool.test.ts
```diff
@@ -95,30 +103,158 @@ describe('ManagedPool', function () {
...
+
+            // BPT is leaving the Vault and so is represented as a negative value.
+            expect(joinSwapResult[0].mul(-1)).to.be.eq(joinResult.bptOut);
+          });
+
+          it('takes the same amount of tokens as the equivalent join', async () => {
+            const joinSwapResult = await queryJoinSwap();
+            const joinResult = await queryEquivalentJoin();
+
+            expect(joinSwapResult[1]).to.be.eq(joinResult.amountsIn[2]);
```
nventuro, line 159: Can we use contants for some of these magical indices? Specially token indices that are shifted due to :musical_score:

nventuro: Or even better, receive them as arguments

pkg/pool-weighted/test/ManagedPool.test.ts
```diff
@@ -95,30 +103,158 @@ describe('ManagedPool', function () {
...
+        });
+      }
+
+      context('given in', () => {
+        itPerformsAJoinSwapCorrectly(
+          () => pool.swapGivenIn({ in: 1, out: BPT_INDEX, amount: fp(0.1), from: other, recipient: other }),
+          () =>
+            pool.vault.queryBatchSwap({
+              kind: SwapKind.GivenIn,
+              assets: [pool.address, poolTokens.first.address],
```
nventuro, line 214: The Pool has more tokens than this. I'm slightly surprised the test passes given this, since we do call the vault to ask for token addresses

nventuro (approved):

pkg/pool-weighted/contracts/managed/ManagedPool.sol
TomAFrench: It feels more right to me to do handling of fees and permissioning here and then have  `_onJoinSwap` be a pure "I'm getting A so you get B" function.

pkg/pool-weighted/test/ManagedPool.test.ts
TomAFrench: This array is only for the assets which are involved in the swap. It's for the vault to use to translate `assetInIndex` and `assetOutIndex` into addresses so we only need to provide the token we're joining with and BPT.

nventuro: Ahh sorry - I thought this was `joinPool`'s `assets`

pkg/pool-weighted/test/ManagedPool.test.ts
TomAFrench: It's a bit complex as we have 3 different types of indexes floating about but hopefully it should be fairly clear now.

PR #1871, opened 2022-09-29
Rename ProtocolAUMFees

### LINKED ISSUE #1747: Rebrand "Protocol Fee" libraries to "External Fees"
What we call the "protocol fee" mechanisms are used to pay BPT to any external party, e.g. managers of `ManagedPools`. This could cause some confusion on why a manager is receiving "protocol fees" which could be avoided with a rename.

EndymionJkb:
Minimal rename of ProtocolAUMFees, to reflect the fact that the fees are going to third parties.

There might still be some gray areas: the directory "protocol-fees" has both this and the protocol fee contracts. (If the DAO is "external," they could all be considered external-fees.)

The ProtocolFees library (which defines `bptForPoolOwnershipPercentage`) is also not protocol-fee-specific, and could be called ExternalFees: but renaming that would change a lot of existing code.

Closes #1747.

### DIFF
--- a/pkg/pool-utils/contracts/protocol-fees/ExternalAUMFees.sol
+++ b/pkg/pool-utils/contracts/protocol-fees/ExternalAUMFees.sol
@@ -19,7 +19,7 @@ import "@balancer-labs/v2-solidity-utils/contracts/math/Math.sol";
 
 import "./ProtocolFees.sol";
 
-library ProtocolAUMFees {
+library ExternalAUMFees {
     /**
      * @notice Calculates the amount of BPT to mint to pay AUM fees accrued since the last collection.
      * @dev This calculation assumes that the Pool's total supply is constant over the fee period.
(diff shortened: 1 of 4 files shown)


pkg/pool-utils/contracts/external-fees/ExternalAUMFees.sol
```diff
@@ -39,7 +39,7 @@ library ExternalAUMFees {
         if (currentTime <= lastCollection || annualAumFeePercentage == 0) return 0;
 
         // Reuse the "BPT amount for pool percentage" calculation used to compute protocol fees.
```
nventuro, line 41:
```suggestion
```

pkg/pool-utils/test/foundry/ProtocolFees.t.sol
```diff
@@ -5,22 +5,22 @@ import "forge-std/Test.sol";
 
 import "@balancer-labs/v2-solidity-utils/contracts/math/FixedPoint.sol";
 
-import "../../contracts/protocol-fees/ProtocolFees.sol";
+import "../../contracts/external-fees/ExternalFees.sol";
 
 contract ProtocolFeesTest is Test {
```
nventuro, line 10: This should be ExternalFeesTest

pkg/pool-utils/contracts/test/MockProtocolFees.sol
```diff
@@ -14,14 +14,14 @@
 
 pragma solidity ^0.7.0;
 
-import "../protocol-fees/ProtocolFees.sol";
+import "../external-fees/ExternalFees.sol";
 
 contract MockProtocolFees {
```
nventuro, line 19: This should be MockExternalFees

Issue #1263, opened 2022-04-25
Implement sensible limits on ManagedPool values, or remove them


nventuro:
We don't quite know what values might be good max values for e.g. AUM fees. Perhaps we could not have limits at all and leave that up to the controller.

TomAFrench:
Rab's thoughts on max swap fees.

> Target: 99%
> Compromise: 95%
> Threshold: 90%

https://github.com/balancer-labs/balancer-v2-monorepo/pull/1846#discussion_r983640596

PR #1872, opened 2022-09-29
Adjust managed pool limits

nventuro:
These are all up to the owner.

Closes https://github.com/balancer-labs/balancer-v2-monorepo/issues/1263


EndymionJkb:
I think we just have to document to the owners that some large transactions might revert under certain pathological conditions.

It shouldn't be a problem under normal high-liquidity circumstances: it only blows up when the amountIn with fees (which can increase very rapidly as the fee approaches 100%) exceeds the pool balance of the token. 95% should be plenty of "room", while still being prohibitively high.

nventuro:
Agh forgot to update the tests, so dumb :facepalm: 

>I think we just have to put in the docs that some large transactions might revert under certain pathological conditions.

Yes, I'm not really _that_ concerned. But we don't _need_ extremely high fees, so keeping them 'reasonable' (as opposed to e.g. 99.9999999%) sounds like a good idea.

PR #1873, opened 2022-09-29
Circuit breaker tests refactor

TomAFrench:
Builds on #1834

I've made a few changes to how we lay out the tests.

- Flattened the test structure so we're not testing the calculation of BPT prices inside a `when the sender is the owner` context, etc.
- Added testing for setting circuit breakers in a delegated ownership context (granted I don't expect any delegated ownership Managed Pools)
- Moved the tests out of the `permissioned actions` block as this is extremely broad and shouldn't exist at all imo.


EndymionJkb (approved):
Looks better, thanks! It's not crazy to test delegated either; I could imagine it happening for some partner pools. Or maybe circuit breaker setting will be delegated to some third party like Gauntlet someday.

PR #1876, opened 2022-09-30
Move add/remove token logic into an external library.

TomAFrench:
This PR aims to move the `addToken` and `removeToken` functions into an external library to save bytecode space.

Draft as I still need to:
- [ ] clean up the functions.
- [ ] update the docs.
- [ ] make the library external.


nventuro:
This drops bytecode by ~600 bytes. We could push this even further if we didn't pass the arrays and had the library perform queries, but this is good enough for now. Let's get back to this PR once the have the final contract size and pick our compression strategy.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -722,34 +711,31 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, IControlled
...
-        // With the token registered, we fetch the new list of Pool tokens (which will include it). This is also a good
-        // opportunity to check we have not added too many tokens.
+        {
+            // This complex operation might mint BPT, altering the supply. For simplicity, we forbid adding tokens
+            // before initialization (i.e. before BPT is first minted). We must also collect AUM fees every time the
+            // BPT supply changes. For consistency, we do this always, even if the amount to mint is zero.
+            uint256 supply = _getVirtualSupply();
+            _require(supply > 0, Errors.UNINITIALIZED);
+            _collectAumManagementFees(supply);
+        }
```
EndymionJkb, line 721: Could move this into a function for both add and remove, save bytecode and not need the scope:

```
   /**
     * @dev Complex Add/Remove operations might mint or burn BPT, altering the supply. For simplicity, we forbid
     * adding or removing tokens before initialization (i.e. before BPT is first minted). We must also collect
     * AUM fees every time the BPT supply changes. For consistency, we do this always, even if the amount to
     * mint is zero.
     * This should be called on all adds/removes.
     */
    function _ensureAumManagementFeesCollected() private {
        uint256 supply = _getVirtualSupply();
        _require(supply > 0, Errors.UNINITIALIZED);
        _collectAumManagementFees(supply);
    }
```

pkg/pool-weighted/contracts/managed/ManagedPoolAddRemoveTokenLib.sol
```diff
@@ -0,0 +1,225 @@
...
+        IVault vault,
+        bytes32 poolId,
+        bytes32 poolState,
+        IERC20[] memory currentTokens,
+        uint256[] memory currentWeights,
+        IERC20 tokenToAdd,
+        address assetManager,
+        uint256 tokenToAddNormalizedWeight
+    )
+        public
```
nventuro, line 78:
```suggestion
        external
```

pkg/pool-weighted/contracts/managed/ManagedPoolAddRemoveTokenLib.sol
```diff
@@ -0,0 +1,225 @@
...
+     */
+    function removeToken(
+        IVault vault,
+        bytes32 poolId,
+        bytes32 poolState,
+        IERC20[] memory currentTokens,
+        uint256[] memory currentWeights,
+        IERC20 tokenToRemove,
+        uint256 tokenToRemoveNormalizedWeight
+    ) public returns (IERC20[] memory newTokens, uint256[] memory newWeights) {
```
nventuro, line 166:
```suggestion
    ) external returns (IERC20[] memory newTokens, uint256[] memory newWeights) {
```

nventuro (approved):

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
nventuro: The savings would not be large - I'd rather keep this explicit for now

PR #1877, opened 2022-09-30
Remove getInvariant

nventuro:
This saves ~150 bytes. Everything that is required to compute the invariant from the outside is already provided, and it'd be extremely rare for anyone to want to query it anyway. This also simplifies `getPoolTokens`, and lets us get rid of `dropBpt`.


nventuro:
We might actually not want to remove `dropBpt` as we'll potentially need it in joinswaps and exitswaps to get all balances so that we can check the circuit breakers.

TomAFrench (approved):
Happy to merge this once `dropBpt` is added back.

TomAFrench:
> We might actually not want to remove `dropBpt` as we'll potentially need it in joinswaps and exitswaps to get all balances so that we can check the circuit breakers.

This is the case in #1887 as we're using the returned `balances` array from `_getPoolTokens`

PR #1887, opened 2022-10-02
Apply Circuit Breakers

EndymionJkb:
For a regular token swap, we only need to check the breakers for the tokens involved in the swap (if they're set). With the supply constant (by definition in a swap), the BPT prices of all other tokens are unchanged.

Anything else - swap with BPT or join/exit - *does* change the supply (= virtualSupply, the BPT tokens in circulation), so all BPT prices change, and all breakers must be checked.

For these, we know the final virtualSupply (net the exchange value and any pending fees), the current weights, and the final balances of all tokens: which is enough information to compute the BPT prices and check the bounds. For a join or exit, we have the amountsIn/amountsOut deltas to the current balances. (For a swap with BPT, the amountsIn/Out will just be zero except for the token involved.)

It short circuits as soon as possible, by reading the raw bound first. If it is zero, the limit does not exist. Otherwise, it proceeds to check the bound.

Ideally we would move the storage into TokenStorageLib, but it doesn't seem feasible: we would lose a probably unacceptable amount of precision. Even if we decreased the weight bits in TokenStorageLib to the absolute minimum (without compression) of 60 bits, we would have only 131 bits free to store everything the circuit breakers need. Unless we don't optimize for unchanging weights (i.e., store only the bounds and BPT price), that isn't enough. If we decreased the BPT price from 96 to 80 bits (capping BPT prices to ~1m), that leaves only 51 bits, which is not even enough for the weight, much less the other 4 things.

I also thought about adding a flag in token lib indicating whether circuit breakers were set (plenty of room for that), but it didn't seem to help. If I have to read the token flag, I might as well just read the token bound itself.

It might be worth it to have a global flag that would short circuit faster if no breakers were set at all. Maybe some managed pools don't want to use circuit breakers, and it they never set any, it would be faster. This flag could be maintained as part of the setCircuitBreakers() call. There could even be two flags - one for lower and one for upper bounds - since the consensus seems to be the first pools will only set lower bounds.


pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -587,6 +710,73 @@ contract ManagedPool is ManagedPoolSettings {
...
+        PoolDelta memory poolDelta,
+        bool isJoin
+    ) private view {
+        for (uint256 i = 0; i < poolDelta.tokens.length; i++) {
+            // If it is a join, tokens are coming in, and we're checking the lower bound, so isLowerBound = isJoin.
+            uint256 bptPriceBound = _getCurrentCircuitBreakerBound(
+                poolDelta.tokens[i],
+                poolDelta.normalizedWeights[i].complement(),
+                isJoin
+            );
```
TomAFrench, line 766: I don't think this is true. It's true that in token swaps we can reason about how the BPT price of different tokens will change, we can't in situations where the total supply is changing.

e.g. Consider a 50:50 pool with tokens A and B. Say token A has a tight upper circuit breaker so if the price increases a little bit then the breaker will trip. If I do a huge join with token B then the join will succeed (assuming I don't trip a breaker on token B) despite the fact that I've massively inflated the BPT price of token A

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -587,6 +710,73 @@ contract ManagedPool is ManagedPoolSettings {
...
+        uint256 upscaledBalance,
+        bool isLowerBound
+    ) private pure {
+        uint256 currentBptPrice = _getBptPrice(virtualSupply, normalizedWeight, upscaledBalance);
+
+        _require(
+            isLowerBound ? currentBptPrice >= bptPriceBound : currentBptPrice <= bptPriceBound,
+            Errors.CIRCUIT_BREAKER_TRIPPED
+        );
+    }
```
TomAFrench, line 744: We're essentially reimplementing `hasCircuitBreakerTripped` here. I think a better route would be to have a single sided variant of that function as ManagedPool shouldn't really need to know what a "bptPrice" is.

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -179,47 +179,46 @@ library CircuitBreakerLib {
      *
      * @param circuitBreakerState - The bytes32 state of the token of interest.
      * @param weightComplement - The complement of this token's weight, generally given by (1 - weight).
-     * @return - lower and upper BPT price bounds, which can be directly compared against the current BPT price.
+     * @param isLowerBound - Specify which bound to return. (We don't need both at the same time.)
```
TomAFrench, line 182:
```suggestion
     * @param isLowerBound - Specify which bound to return.
```

pkg/pool-weighted/contracts/managed/ManagedPool.sol
EndymionJkb: Yes - I noticed I wasn't using that function, and was thinking along the same lines (better encapsulation).

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -277,58 +326,97 @@ contract ManagedPool is ManagedPoolSettings {
...
+            ) ||
+                CircuitBreakerLib.hasCircuitBreakerTripped(
+                    _getCircuitBreakerState(request.tokenOut),
+                    virtualSupply,
+                    swapData.tokenOutWeight,
+                    endingBalanceTokenOut,
+                    false // check upper bound
+                )),
+            Errors.CIRCUIT_BREAKER_TRIPPED
+        );
```
TomAFrench, line 398:
```suggestion
        // If circuit breakers are set, check the lower bound on the tokenIn, and the upper bound on the tokenOut.
        uint256 virtualSupply = getActualSupply();
        _require(
            !(CircuitBreakerLib.hasCircuitBreakerTripped(
                _getCircuitBreakerState(request.tokenIn),
                virtualSupply,
                swapData.tokenInWeight,
                endingBalanceTokenIn,
                true // check lower bound
            ) ||
                CircuitBreakerLib.hasCircuitBreakerTripped(
                    _getCircuitBreakerState(request.tokenOut),
                    virtualSupply,
                    swapData.tokenOutWeight,
                    endingBalanceTokenOut,
                    false // check upper bound
                )),
            Errors.CIRCUIT_BREAKER_TRIPPED
        );
```

We should load the virtual supply closer to where we're using it.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -277,58 +326,97 @@ contract ManagedPool is ManagedPoolSettings {
...
+        bytes32 tokenOutState = _getTokenState(request.tokenOut);
+
+        uint256 weightChangeProgress = ManagedPoolStorageLib.getGradualWeightChangeProgress(poolState);
+        tokenInfo.tokenInWeight = ManagedPoolTokenStorageLib.getTokenWeight(tokenInState, weightChangeProgress);
+        tokenInfo.tokenOutWeight = ManagedPoolTokenStorageLib.getTokenWeight(tokenOutState, weightChangeProgress);
+
+        tokenInfo.scalingFactorTokenIn = ManagedPoolTokenStorageLib.getTokenScalingFactor(tokenInState);
+        tokenInfo.scalingFactorTokenOut = ManagedPoolTokenStorageLib.getTokenScalingFactor(tokenOutState);
+
+        tokenInfo.swapFeeComplement = ManagedPoolStorageLib.getSwapFeePercentage(poolState).complement();
```
TomAFrench, line 419: It looks like we don't need this to be included in the struct but we can instead just load this inside `_onTokenSwap`. I'd be inclined to move it back so this struct only holds token related information (return variable name hints at this).

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -277,58 +326,97 @@ contract ManagedPool is ManagedPoolSettings {
...
-        uint256 scalingFactorTokenIn;
-        uint256 scalingFactorTokenOut;
-        uint256 swapFeeComplement;
-        {
-            uint256 weightChangeProgress = ManagedPoolStorageLib.getGradualWeightChangeProgress(poolState);
-            (tokenInWeight, scalingFactorTokenIn) = _getTokenInfo(request.tokenIn, weightChangeProgress);
-            (tokenOutWeight, scalingFactorTokenOut) = _getTokenInfo(request.tokenOut, weightChangeProgress);
+    ) internal view returns (uint256 amountCalculated) {
+        SwapData memory swapData = _getSwapData(request, poolState);
+        uint256 virtualSupply = getActualSupply();
```
TomAFrench, line 331: I think we agreed to use the virtual rather than actual supply here as it's not worth the extra gas and will affect bpt prices minimally.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
EndymionJkb: It was another stack-too-deep thing. I can try moving some stuff back.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
EndymionJkb: I remember asking about this and at one time it was actual, but will change it if that's the consensus.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -158,6 +158,17 @@ contract ManagedPool is ManagedPoolSettings {
...
 
+    // We need to check the breakers for all tokens on joins and exits (including join and exit swaps), since any
+    // change to the virtualSupply affects all BPT prices. For a multi-token join or exit, we will have a set of
+    // balances and delta amounts. For a join/exitSwap, only one token balance is changing. We can use the same
+    // data for both: in the single token swap case, the other token `amounts` will be zero.
+    struct PoolDelta {
+        IERC20[] tokens;
+        uint256[] normalizedWeights;
+        uint256[] balances;
+        uint256[] amounts;
```
TomAFrench, line 169:
```suggestion
        uint256[] finalBalances;
```
Not required but we may want to mutate the balances into the post join/exit balances and just pass that.

TomAFrench (approved):
Very sad that we need to do so much checking but... thems the breakers.

We'll probably need some smoke test tests for triggering the circuit breakers on swaps/joins/exits.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -208,7 +212,7 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, IControlled
      *
      * In the vast majority of cases, this function should be used instead of `totalSupply()`.
      */
-    function getActualSupply() external view returns (uint256) {
+    function getActualSupply() public view returns (uint256) {
```
TomAFrench, line 215:
```suggestion
    function getActualSupply() external view returns (uint256) {
```

EndymionJkb:
Superseded by 1904-1908

PR #1891, opened 2022-10-03
Bytecode improvement: ProtocolFeeCache structs

jubeira:
This PR replaces structs with bytes32 + word codec in protocol fee cache.

`yarn hardhat compile && cat artifacts/contracts/managed/ManagedPool.sol/ManagedPool.json | jq -r '.deployedBytecode' | wc --chars | awk '{print (($1-2)/2)}'` prints 29262,5 in master and 28815,5 with this change.

Savings: 447.

### DIFF
--- a/pkg/pool-utils/contracts/external-fees/ProtocolFeeCache.sol
+++ b/pkg/pool-utils/contracts/external-fees/ProtocolFeeCache.sol
@@ -18,6 +18,7 @@ pragma experimental ABIEncoderV2;
 import "@balancer-labs/v2-interfaces/contracts/solidity-utils/helpers/BalancerErrors.sol";
 import "@balancer-labs/v2-interfaces/contracts/standalone-utils/IProtocolFeePercentagesProvider.sol";
 
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/WordCodec.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/SafeCast.sol";
 
 import "../RecoveryMode.sol";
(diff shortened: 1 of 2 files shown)


jubeira:
Side comment: we could even get rid of the `if-else` chains in `getProtocolFeePercentageCache` and `getProviderFeeId` by calculating the offsets given the fee type, but the code would be somewhat less future-proof.

pkg/pool-utils/contracts/external-fees/ProtocolFeeCache.sol
```diff
@@ -84,11 +92,11 @@ abstract contract ProtocolFeeCache is RecoveryMode {
...
         if (feeType == ProtocolFeeType.SWAP) {
-            return _cache.swapFee;
+            return _feeCache.decodeUint(_SWAP_FEE_OFFSET, _FEE_TYPE_CACHE_WIDTH);
         } else if (feeType == ProtocolFeeType.YIELD) {
-            return _cache.yieldFee;
+            return _feeCache.decodeUint(_YIELD_FEE_OFFSET, _FEE_TYPE_CACHE_WIDTH);
         } else if (feeType == ProtocolFeeType.AUM) {
-            return _cache.aumFee;
+            return _feeCache.decodeUint(_AUM_FEE_OFFSET, _FEE_TYPE_CACHE_WIDTH);
         } else {
```
nventuro, line 100: If the if-else chain looked for the offset and you then had a single `decode` call, that'd probably save more bytecode.

pkg/pool-utils/contracts/external-fees/ProtocolFeeCache.sol
```diff
@@ -99,11 +107,11 @@ abstract contract ProtocolFeeCache is RecoveryMode {
...
         if (feeType == ProtocolFeeType.SWAP) {
-            return _swapFeeId;
+            return _feeIds.decodeUint(_SWAP_FEE_ID_OFFSET, _FEE_TYPE_ID_WIDTH);
         } else if (feeType == ProtocolFeeType.YIELD) {
-            return _yieldFeeId;
+            return _feeIds.decodeUint(_YIELD_FEE_ID_OFFSET, _FEE_TYPE_ID_WIDTH);
         } else if (feeType == ProtocolFeeType.AUM) {
-            return _aumFeeId;
+            return _feeIds.decodeUint(_AUM_FEE_ID_OFFSET, _FEE_TYPE_ID_WIDTH);
         } else {
```
nventuro, line 115: Same here.

pkg/pool-utils/contracts/external-fees/ProtocolFeeCache.sol
```diff
@@ -131,18 +136,23 @@ abstract contract ProtocolFeeCache is RecoveryMode {
...
-
-        _cache = FeeTypeCache({
-            swapFee: swapFee.toUint64(),
-            yieldFee: yieldFee.toUint64(),
-            aumFee: aumFee.toUint64()
-        });
-        emit ProtocolFeePercentageCacheUpdated(_cache);
+    function _updateProtocolFeeCache(IProtocolFeePercentagesProvider protocolFeeProvider, bytes32 feeIds) private {
+        uint256 swapFee = protocolFeeProvider.getFeeTypePercentage(
+            feeIds.decodeUint(_SWAP_FEE_ID_OFFSET, _FEE_TYPE_ID_WIDTH)
```
nventuro, line 141:
```suggestion
            getProviderFeeId(ProtocolFeeType.SWAP)
```

This likely saves bytecode as well

pkg/pool-utils/contracts/external-fees/ProtocolFeeCache.sol
jubeira: Done.

pkg/pool-utils/contracts/external-fees/ProtocolFeeCache.sol
jubeira: Done.

pkg/pool-utils/contracts/external-fees/ProtocolFeeCache.sol
jubeira: This doesn't work because that would read an immutable member during construction.
Adding a version of `getProviderFeeId` that takes the IDs as parameter doesn't save anything.

pkg/pool-utils/contracts/external-fees/ProtocolFeeCache.sol
```diff
@@ -34,19 +35,22 @@ import "../RecoveryMode.sol";
...
 
-    IProtocolFeePercentagesProvider private immutable _protocolFeeProvider;
-    uint256 private immutable _swapFeeId;
-    uint256 private immutable _yieldFeeId;
-    uint256 private immutable _aumFeeId;
+    // Protocol Fee IDs represent fee types; 8 bits is enough to represent all of them.
+    uint256 private constant _FEE_TYPE_ID_WIDTH = 8;
+    uint256 private constant _SWAP_FEE_ID_OFFSET = 0;
+    uint256 private constant _YIELD_FEE_ID_OFFSET = _SWAP_FEE_ID_OFFSET + _FEE_TYPE_ID_WIDTH;
+    uint256 private constant _AUM_FEE_ID_OFFSET = _YIELD_FEE_ID_OFFSET + _FEE_TYPE_ID_WIDTH;
```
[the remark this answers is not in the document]
jubeira: Done

PR #1892, opened 2022-10-03
Merge AUM fee getters

TomAFrench:
Builds on #1890 

This PR merges the getters for the AUM fee percentage and timestamp while packing the two values in storage. For bytecode reasons I've introduced (another!) library for managing this state vs using a struct.

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
@@ -86,12 +86,14 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, IControlled
     // If mustAllowlistLPs is enabled, this is the list of addresses allowed to join the pool
     mapping(address => bool) private _allowedAddresses;
 
-    // Percentage of the pool's TVL to pay as management AUM fees over the course of a year.
-    uint256 private _managementAumFeePercentage;
+    struct AumFeeParams {
+        // Percentage of the pool's TVL to pay as management AUM fees over the course of a year.
+        uint64 aumFeePercentage;
+        // Timestamp of the most recent collection of management AUM fees.
+        uint64 lastCollectionTimestamp;
+    }
 
-    // Timestamp of the most recent collection of management AUM fees.
-    // Note that this is only initialized the first time fees are collected.
-    uint256 internal _lastAumFeeCollectionTimestamp;
+    AumFeeParams internal _aumFeeParams;
 
     // Event declarations
 
@@ -215,11 +217,12 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, IControlled
     }
 
     function _getActualSupply(uint256 virtualSupply) internal view returns (uint256) {
+        AumFeeParams memory aumFeeParams = getManagementAumFeeParams();
         uint256 aumFeesAmount = ExternalAUMFees.getAumFeesBptAmount(
             virtualSupply,
             block.timestamp,
-            _lastAumFeeCollectionTimestamp,
-            getManagementAumFeePercentage()
+            aumFeeParams.lastCollectionTimestamp,
+            aumFeeParams.aumFeePercentage
         );
         return virtualSupply.add(aumFeesAmount);
     }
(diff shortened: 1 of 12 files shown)


pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -582,17 +576,18 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, IControlled
...
+    {
+        (aumFeePercentage, lastCollectionTimestamp) = ManagedPoolAumStorageLib.getAumFeeFields(_aumState);
 
-    /**
-     * @notice Returns the timestamp of the last collection of AUM fees.
-     */
-    function getLastAumFeeCollectionTimestamp() external view returns (uint256) {
-        return _lastAumFeeCollectionTimestamp;
+        // If we're in recovery mode then we bypass any fee logic by setting the fee percentage to zero.
+        aumFeePercentage = ManagedPoolStorageLib.getRecoveryModeEnabled(_poolState) ? 0 : aumFeePercentage;
```
nventuro, line 590:
```suggestion
        // If we're in recovery mode then we bypass any fee logic by setting the fee percentage to zero.
        if (ManagedPoolStorageLib.getRecoveryModeEnabled(_poolState)) {
            aumFeePercentage = 0;
        }
```

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -667,10 +671,9 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, IControlled
             return 0;
         }
 
-        // As we update `_lastAumFeeCollectionTimestamp` when updating `_managementAumFeePercentage`, we only need to
-        // update `_lastAumFeeCollectionTimestamp` when non-zero AUM fees are paid. This avoids an SSTORE on zero-length
-        // collections.
-        _lastAumFeeCollectionTimestamp = block.timestamp;
+        // As we update the AUM fee collection timestamp when updating the AUM fee percentage, we only need to
+        // update the timestamp when non-zero AUM fees are paid. This avoids an SSTORE on zero-length collections.
+        _updateAumFeeCollectionTimestamp();
```
nventuro, line 676: `getManagementAumFeeParams()` reads `_aumState`, and here we read it again. Bit unfortunate that we need to do the warm sload, but that's likely better than passing state around, or having duplicate functions.

pkg/pool-weighted/test/foundry/ManagedPoolAumStorageLib.t.sol
```diff
@@ -0,0 +1,73 @@
...
+
+        bytes32 newAumState = ManagedPoolAumStorageLib.setAumFeePercentage(aumState, expectedAumFeePercentage);
+        assertOtherStateUnchanged(aumState, newAumState, _AUM_FEE_PERCENTAGE_OFFSET, _AUM_FEE_PCT_WIDTH);
+
+        (uint256 actualAumFeePercentage, ) = ManagedPoolAumStorageLib.getAumFeeFields(newAumState);
+        assertEq(actualAumFeePercentage, expectedAumFeePercentage);
+    }
+
+    function testSetLastCollectionTimestamp(bytes32 aumState, uint256 expectedLastCollectionTimestamp) public {
+        vm.assume(expectedLastCollectionTimestamp < _MAX_TIMESTAMP);
```
nventuro, line 62:
```suggestion
        vm.assume(expectedLastCollectionTimestamp <= _MAX_TIMESTAMP);
```

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
TomAFrench: Huh, I had this as it saved bytecode in #1890. Seems like the opposite here though.

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
TomAFrench: uno_reverse_card.jpg

PR #1894, opened 2022-10-03
Store weight directly

EndymionJkb:
Previously we were passing weightComplement (1 - weight) into the library, which:
1) is encoding implementation details in the interface - normalizedWeight is well understood, and what we already have
2) is a bit messy, having to remember to take weight complements in various places
3) the complement operations waste gas if it turns out we don't need to calculate the bound anyway

I think it's much more intuitive to store and retrieve the normalized weight directly, and hide all the complement stuff inside the library. Just didn't want to whomp all this into the same PR as apply_breakers

### DIFF
--- a/pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
+++ b/pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
@@ -109,7 +109,7 @@ library CircuitBreakerLib {
     uint256 private constant _BOUND_SHIFT_BITS = 64 - _BOUND_WIDTH;
 
     /**
-     * @notice Returns the BPT price and normalized weight values, and lower and upper bounds for a given token.
+     * @notice Returns the BPT price and weight stored when the breaker was set, plus the upper and lower bounds.
      * @dev If an upper or lower bound value is zero, it means there is no circuit breaker in that direction for the
      * given token.
      * @param circuitBreakerState - The bytes32 state of the token of interest.
@@ -118,14 +118,14 @@ library CircuitBreakerLib {
         internal
         pure
         returns (
-            uint256 bptPrice,
-            uint256 normalizedWeight,
+            uint256 referenceBptPrice,
+            uint256 referenceNormalizedWeight,
             uint256 lowerBound,
             uint256 upperBound
         )
     {
-        bptPrice = circuitBreakerState.decodeUint(_BPT_PRICE_OFFSET, _BPT_PRICE_WIDTH);
-        normalizedWeight = circuitBreakerState.decodeUint(_NORMALIZED_WEIGHT_OFFSET, _NORMALIZED_WEIGHT_WIDTH);
+        referenceBptPrice = circuitBreakerState.decodeUint(_BPT_PRICE_OFFSET, _BPT_PRICE_WIDTH);
+        referenceNormalizedWeight = circuitBreakerState.decodeUint(_NORMALIZED_WEIGHT_OFFSET, _NORMALIZED_WEIGHT_WIDTH);
         // Decompress the bounds by shifting left.
         lowerBound = circuitBreakerState.decodeUint(_LOWER_BOUND_OFFSET, _BOUND_WIDTH) << _BOUND_SHIFT_BITS;
         upperBound = circuitBreakerState.decodeUint(_UPPER_BOUND_OFFSET, _BOUND_WIDTH) << _BOUND_SHIFT_BITS;
@@ -178,13 +178,13 @@ library CircuitBreakerLib {
      * price changes and BPT price changes. This calculation transforms one into the other.
      *
      * @param circuitBreakerState - The bytes32 state of the token of interest.
-     * @param currentNormalizedWeight - The normalized weight of the token.
+     * @param normalizedWeight - The normalized weight of the token.
      * @param isLowerBound - Specify which bound to return. (We don't need both at the same time.)
      * @return - lower or upper BPT price bound, which can be directly compared against the current BPT price.
      */
     function getCurrentCircuitBreakerBound(
         bytes32 circuitBreakerState,
-        uint256 currentNormalizedWeight,
+        uint256 normalizedWeight,
         bool isLowerBound
     ) internal pure returns (uint256) {
         uint256 bound = circuitBreakerState.decodeUint(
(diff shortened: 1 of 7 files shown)


pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -67,8 +67,8 @@ library CircuitBreakerLib {
 
     // Store circuit breaker information per token
     // When the circuit breaker is set, the caller passes in the lower and upper bounds (expressed as percentages),
-    // and the current weight complement (1 - weight). Since this value is bounded by 1e18, which fits in ~60 bits,
-    // there is no need for compression.
+    // and the current normalized weight. We will need the weight complement for the calculation: (1 - weight).
```
nventuro, line 70:
```suggestion
    // and the current normalized weight.
```

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -109,7 +109,7 @@ library CircuitBreakerLib {
     uint256 private constant _BOUND_SHIFT_BITS = 64 - _BOUND_WIDTH;
 
     /**
-     * @notice Returns the BPT price and weight complement values, and lower and upper bounds for a given token.
+     * @notice Returns the BPT price and normalized weight values, and lower and upper bounds for a given token.
```
nventuro, line 112: This makes it clear that we're missing some term here, since this is most definitely _not_ the token's weight.

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -197,22 +197,25 @@ library CircuitBreakerLib {
...
+        uint256 initialNormalizedWeight = circuitBreakerState.decodeUint(
+            _NORMALIZED_WEIGHT_OFFSET,
+            _NORMALIZED_WEIGHT_WIDTH
         );
 
+        //uint256 weightComplement = normalizedWeight.complement();
         uint256 boundRatio;
 
-        if (initialWeightComplement == weightComplement) {
+        if (initialNormalizedWeight == normalizedWeight) {
```
nventuro, line 210: See how here we need to add the 'initial' prefix to avoid confusion. We similarly shouldn't call this the 'normalized weight' anywhere.

nventuro (commented):
Great initiative - the old name was indeed bad. But this unveils the need for a more specific name for this value, though I don't know what it could be.

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
EndymionJkb: It is though - I'm directly storing and retrieving the normalized weight now, and only taking the complement inside when I compute the ratio.

TomAFrench: I think Nico's point is that the value we're returning here doesn't have to be the current weight of the token. It's the weight of the token at the point at which we cached the bound ratios, i.e. rather than `normalizedWeight` it should be `<something>Weight`

TomAFrench: `cacheWeight` maybe? Something to denote that it's "weighty" but is to test for whether we hit or miss the cache.

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
EndymionJkb: bptPrice and normalizedWeight are stored when it's set. There is no prefix on bptPrice, so I also removed the prefix from normalizedWeight (we were getting away from "reference", "cached", etc.)

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
EndymionJkb: Ok; was focused on complement vs weight, not stored weight vs "current" weight. I can standardize on "reference".

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -78,16 +78,16 @@ library CircuitBreakerLib {
...
     //
     // [      32 bits      |      32 bits      |  96 bits  |   64 bits    |   16 bits   |   16 bits   |
-    // [ upper bound ratio | lower bound ratio | BPT price | weight comp. | upper bound | lower bound |
+    // [ upper bound ratio | lower bound ratio | BPT price | norm. weight | upper bound | lower bound |
     // |MSB                                                                                        LSB|
     uint256 private constant _LOWER_BOUND_OFFSET = 0;
     uint256 private constant _UPPER_BOUND_OFFSET = _LOWER_BOUND_OFFSET + _BOUND_WIDTH;
-    uint256 private constant _WEIGHT_COMPLEMENT_OFFSET = _UPPER_BOUND_OFFSET + _BOUND_WIDTH;
-    uint256 private constant _BPT_PRICE_OFFSET = _WEIGHT_COMPLEMENT_OFFSET + _WEIGHT_COMPLEMENT_WIDTH;
+    uint256 private constant _NORMALIZED_WEIGHT_OFFSET = _UPPER_BOUND_OFFSET + _BOUND_WIDTH;
```
nventuro, line 85: Shouldn't this be called the reference weight?

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -67,8 +67,8 @@ library CircuitBreakerLib {
 
     // Store circuit breaker information per token
     // When the circuit breaker is set, the caller passes in the lower and upper bounds (expressed as percentages),
-    // and the current weight complement (1 - weight). Since this value is bounded by 1e18, which fits in ~60 bits,
-    // there is no need for compression.
+    // and the current normalized weight. Since this value is bounded by 1e18, which fits in ~60 bits, there is no
+    // need for compression.
     //
     // We then compute and store the current BPT price, and the lower and upper bound conversion ratios, used to
```
nventuro, line 73: Unrelated, but we don't compute the BPT price

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -78,16 +78,16 @@ library CircuitBreakerLib {
     // the circuit breaker was set, these stored values can still be used, avoiding a heavy computation.
```
nventuro, line 78: This is where we should say e.g. 'we store a reference weight, so that if the current weight equals the reference weight we can skip the expensive computation of the current ratios

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -78,16 +78,16 @@ library CircuitBreakerLib {
...
+    // [ upper bound ratio | lower bound ratio | BPT price | norm. weight | upper bound | lower bound |
     // |MSB                                                                                        LSB|
     uint256 private constant _LOWER_BOUND_OFFSET = 0;
     uint256 private constant _UPPER_BOUND_OFFSET = _LOWER_BOUND_OFFSET + _BOUND_WIDTH;
-    uint256 private constant _WEIGHT_COMPLEMENT_OFFSET = _UPPER_BOUND_OFFSET + _BOUND_WIDTH;
-    uint256 private constant _BPT_PRICE_OFFSET = _WEIGHT_COMPLEMENT_OFFSET + _WEIGHT_COMPLEMENT_WIDTH;
+    uint256 private constant _NORMALIZED_WEIGHT_OFFSET = _UPPER_BOUND_OFFSET + _BOUND_WIDTH;
+    uint256 private constant _BPT_PRICE_OFFSET = _NORMALIZED_WEIGHT_OFFSET + _NORMALIZED_WEIGHT_WIDTH;
     uint256 private constant _LOWER_BOUND_RATIO_OFFSET = _BPT_PRICE_OFFSET + _BPT_PRICE_WIDTH;
     uint256 private constant _UPPER_BOUND_RATIO_OFFSET = _LOWER_BOUND_RATIO_OFFSET + _BOUND_RATIO_WIDTH;
```
nventuro, line 88: These are also references

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -197,28 +197,29 @@ library CircuitBreakerLib {
             return 0;
         }
 
-        // Retrieve the weight complement passed in and bptPrice computed when the circuit breaker was set.
-        uint256 bptPrice = circuitBreakerState.decodeUint(_BPT_PRICE_OFFSET, _BPT_PRICE_WIDTH);
-        uint256 initialWeightComplement = circuitBreakerState.decodeUint(
-            _WEIGHT_COMPLEMENT_OFFSET,
-            _WEIGHT_COMPLEMENT_WIDTH
+        // Retrieve the weight passed in and bptPrice computed when the circuit breaker was set.
+        uint256 referenceBptPrice = circuitBreakerState.decodeUint(_BPT_PRICE_OFFSET, _BPT_PRICE_WIDTH);
```
nventuro, line 201: If we're calling the other 'references', I'd not call this one a ref. It's something else, not a precomputed value.

nventuro: This seems more like a true ref than the  others though.

nventuro (commented):
It feels like we're going aimlessly back and forth here. I strongly believe we need a clear naming convention that is consistently applied, given there's multiple concepts involved and this is a brand new idea, but we don't seem to be making much progress in any direction.

Here's an attempt to make sense of things.

----
The circuit breaker is composed of a reference BPT price (typically the price at the time the breaker was set), and upper and lower bounds (how far the price can move).

In order to determine if the breaker has tripped at a given bpt price, the current token weight is required since it may have changed since the time the reference bpt price was computed, and it is needed to adjust it. 

This adjustment however involves a power function, which is costly. We therefore also store the original weight at the time the breaker was set, along with the precomputed bounds for that weight. If the current weight equals the original weight, then we can use the precomputed values. Otherwise, they need to be calculated on the spot.

At any point the precomputed values and original weight can be updated to reflect the current state (typically after a weight change, when the weights will remain stable for a while).

EndymionJkb:
Superseded by #1904

PR #1895, opened 2022-10-03
Updating fee collection timestamp every time aum fees are collected

jubeira:
Codename: 4D chess.

Also adding some small code reuse when collecting fees while not initialized to revert in add / remove / external collect.

### DIFF
--- a/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
+++ b/pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
@@ -922,7 +922,7 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, IControlled
         // not paused.
         _ensureNotPaused();
 
-        // No need to collect fees if we are not initialized yet.
+        // If the Pool doesn't hold any tokens due to being uninitialized then we skip fee collection.
         uint256 supplyBeforeFeeCollection = _getVirtualSupply();
         if (supplyBeforeFeeCollection > 0) {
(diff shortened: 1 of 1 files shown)


jubeira:
CI failure [here](https://github.com/balancer-labs/balancer-v2-monorepo/actions/runs/3177401006/jobs/5177797689) doesn't seem to be related with these changes, but with random inputs in the test.

Suggestion by @nventuro: reducing the variance in `AddRemove` test, and decreasing the new weight from 0.1 to e.g. 0.05.

nventuro (commented):
Not sure if the changes and addition of _collectAumManagementFeesWhenInitialized are worth it tbh - they make the logic a bit more complicated with the introduction of this notion. The embedded `require` can be a bit surprising.

nventuro (approved):

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -924,7 +922,11 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, IControlled
         // not paused.
         _ensureNotPaused();
 
-        _collectAumManagementFees(_getVirtualSupply());
+        // No need to collect fees if we are not initialized yet.
```
TomAFrench, line 925:
```suggestion
        // If the Pool doesn't hold any tokens (as shown by being uninitialized) then we skip fee collection.
```

TomAFrench (approved):

PR #1896, opened 2022-10-03
Join exit rounding fuzz tests

nventuro:
Like #1875, but for joins and exits.

This is actually a bit different as there's only 2 swap functions (given in and given out), and 4 join/exit functions (given in and out for join and exit). We're only testing half the scenarios by first joining and then exiting, but I think this should be valid.

I have some improvements to add that remove some duplication, but wanted to keep things simple for now.


nventuro:
I ended up adding more tests in a20bf41 that use no swap fee. While the fuzzer will likely try the zero input at least once, by adding an entire new test we get a full set of runs all using that special value.

jubeira (approved):
Test looks pretty good! I've left a small comment and a suggestion.

PR #1898, opened 2022-10-04
Pure virtuals in ManagedPoolSettings

EndymionJkb:
Make virtual supply and token/balance getters pure virtual in ManagedPoolSettings; dangerous if not overridden.

Unaddressed things: "no default impls (virt supp)" and "get pool tokens no impl"


TomAFrench (approved):

nventuro (approved):
The comments in ManagedPoolSettings need to be updated, as there's no longer a default impl.

nventuro (approved):

PR #1903, opened 2022-10-04
Move managed docs to interface

nventuro:
I didn't really add anything - simply moved things around.


pkg/interfaces/contracts/pool-utils/IManagedPool.sol
```diff
@@ -0,0 +1,316 @@
...
+    );
+    event TokenAdded(IERC20 indexed token, uint256 normalizedWeight);
+    event TokenRemoved(IERC20 indexed token);
+
+    /**
+     * @notice Returns the effective BPT supply.
+     *
+     * @dev The Pool owes debt to the Protocol and the Pool's owner in the form of unminted BPT, which will be minted
+     * immediately before the next join or exit. We need to take these into account since, even if they don't yet exist,
+     *  they will effectively be included in any Pool operation that involves BPT.
```
EndymionJkb, line 53:
```suggestion
     * they will effectively be included in any Pool operation that involves BPT.
```

pkg/interfaces/contracts/pool-utils/IManagedPool.sol
```diff
@@ -0,0 +1,316 @@
...
+
+    // Swap fee percentage
+
+    /**
+     * @notice Schedule a gradual swap fee update.
+     * @dev The swap fee will change from the given starting value (which may or may not be the current
+     * value) to the given ending fee percentage, over startTime to endTime.
+     *
+     * Note that calling this with a starting swap fee different from the current value will immediately change the
+     * current swap fee to `startSwapFeePercentage`, before commencing the gradual change at `startTime`.
```
EndymionJkb, line 67: It does change immediately, because the interpolation will now return the start value. It used to set it in BasePool and emit SwapFeePercentageChanged, but since all that no longer exists, no need to talk about past functionality.

pkg/interfaces/contracts/pool-utils/IManagedPool.sol
```diff
@@ -0,0 +1,316 @@
...
+    // LP Allowlist
+
+    /**
+     * @notice Enable or disable the LP allowlist.
+     * @dev Note that any addresses added to the allowlist will be retained if the allowlist is toggled off and
+     * back on again, because this action does not affect the list of LP addresses.
+     * Emits the MustAllowlistLPsSet event. This is a permissioned function.
+     * @param mustAllowlistLPs - The new value of the mustAllowlistLPs flag.
+     */
+
```
EndymionJkb, line 171:
```suggestion
```

pkg/interfaces/contracts/pool-utils/IManagedPool.sol
```diff
@@ -0,0 +1,316 @@
...
+     */
+    function addAllowedAddress(address member) external;
+
+    /**
+     * @notice Removes an address from the LP allowlist.
+     * @dev Will fail if the address was not previously allowlisted.
+     * Emits the AllowlistAddressRemoved event. This is a permissioned function.
+     * @param member - The address to be removed from the allowlist.
+     */
+
```
EndymionJkb, line 188:
```suggestion
```

pkg/interfaces/contracts/pool-utils/IManagedPool.sol
```diff
@@ -0,0 +1,316 @@
...
+     * @return true if the given address is on the allowlist.
+     */
+    function isAddressOnAllowlist(address member) external view returns (bool);
+
+    // Management fees
+
+    /**
+     * @notice Collect any accrued AUM fees and send them to the pool manager.
+     * @dev This can be called by anyone to collect accrued AUM fees - and will be called automatically on
+     * joins and exits.
```
EndymionJkb, line 209:
```suggestion
     * @dev This can be called by anyone to collect accrued AUM fees - and will be called automatically
     * whenever the supply changes (e.g., joins and exits, add and remove token), and before the fee
     * percentage is changed by the manager, to prevent fees from being applied retroactively.
```

pkg/interfaces/contracts/pool-utils/IManagedPool.sol
```diff
@@ -0,0 +1,316 @@
...
+    function getManagementAumFeeParams()
+        external
+        view
+        returns (uint256 aumFeePercentage, uint256 lastCollectionTimestamp);
+
+    // Circuit Breakers
+
+    /**
+     * @notice Set a circuit breaker for one or more tokens.
+     * @dev This is a permissioned function, and disabled if the pool is paused. The lower and upper bounds
```
EndymionJkb, line 237:
```suggestion
     * @dev This is a permissioned function. The lower and upper bounds
```
It is disabled when the pool is paused, like all current permissioned functions, but we don't say this in any of the other permissioned function comments. (Should be all or none, to avoid the impression that this one is different.)

pkg/interfaces/contracts/pool-utils/IManagedPool.sol
```diff
@@ -0,0 +1,316 @@
...
+     * @dev These are the reference values (BPT price and weight complement) computed when the breaker was set,
+     * along with the percentage bounds. It also returns the current BPT price bounds, needed to check whether
+     * the circuit breaker should trip.
+     */
+    function getCircuitBreakerState(IERC20 token)
+        external
+        view
+        returns (
+            uint256 bptPrice,
+            uint256 weightComplement,
```
EndymionJkb, line 260:
```suggestion
            uint256 referenceWeight,
```

pkg/interfaces/contracts/pool-utils/IManagedPool.sol
```diff
@@ -0,0 +1,316 @@
...
+    function setCircuitBreakers(
+        IERC20[] memory tokens,
+        uint256[] memory bptPrices,
+        uint256[] memory lowerBoundPercentages,
+        uint256[] memory upperBoundPercentages
+    ) external;
+
+    /**
+     * @notice Return the full circuit breaker state for the given token.
+     * @dev These are the reference values (BPT price and weight complement) computed when the breaker was set,
```
EndymionJkb, line 251:
```suggestion
     * @dev These are the reference values (BPT price and reference weight) passed in when the breaker was set,
```

EndymionJkb (approved):
Just a few comment/name changes.

pkg/interfaces/contracts/pool-utils/IManagedPool.sol
nventuro: I'm not sure I follow, isn't that what the docs say?

EndymionJkb: Yes; I was just arguing with myself about whether it should explain how it changes immediately. I think it's fine as is.

pkg/interfaces/contracts/pool-utils/IManagedPool.sol
```diff
@@ -0,0 +1,316 @@
...
+        returns (uint256 aumFeePercentage, uint256 lastCollectionTimestamp);
+
+    // Circuit Breakers
+
+    /**
+     * @notice Set a circuit breaker for one or more tokens.
+     * @dev This is a permissioned function. The lower and upper bounds
+     * are percentages, corresponding to a *relative* change in the token's spot price: e.g., a lower bound
+     * of 0.8 means the breaker should prevent trades that result in the value of the token dropping 20% or
+     * more relative to the rest of the pool.
```
EndymionJkb, line 239:
```suggestion
     * @dev This is a permissioned function. The lower and upper bounds are percentages, corresponding to a
     * relative change in the token's spot price: e.g., a lower bound of 0.8 means the breaker should prevent
     * trades that result in the value of the token dropping 20% or more relative to the rest of the pool.
```
Would just format this so there isn't a big gap

EndymionJkb (approved):
Looks like everything was addressed - just one little format tweak (see comment)

EndymionJkb:
And there's a lint issue in IManagedPool.sol

PR #1904, opened 2022-10-04
Use weight instead of weight complement (CB)

EndymionJkb:
Previously we were passing weightComplement (1 - weight) into the library, which:

* is encoding implementation details in the interface - normalizedWeight is well understood, and what we already have
* is a bit messy, having to remember to take weight complements in various places
* the complement operations waste gas if it turns out we don't need to calculate the bound anyway

I think it's much more intuitive to store and retrieve the normalized weight directly, and hide all the complement stuff inside the library. This also uses the agreed-upon names, and builds on the previous work separating storage and computation.


pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -51,34 +51,40 @@ library CircuitBreakerLib {
     }
 
     /**
-     * @notice Convert bounds to BPT price ratios
+     * @notice Convert bounds to adjusted bounds (apply non-linear adjustment for weights)
```
TomAFrench, line 54:
```suggestion
     * @notice Convert bounds to adjusted bounds (apply non-linear adjustment for weight changes)
```

pkg/pool-weighted/contracts/lib/CircuitBreakerLib.sol
```diff
@@ -51,34 +51,40 @@ library CircuitBreakerLib {
...
      * @param lowerBound - the lower bound percentage; 0.8 means tolerate a 20% relative drop.
      * @param upperBound - the upper bound percentage; 5.0 means tolerate a 5x increase.
-     * @param weightComplement - the complement of the normalized token weight: 1 - weight.
+     * @param weight - the current normalized token weight.
+     * @return adjustedLowerBound - the final lower bound, adjusted for any weight changes.
+     * @return adjustedUpperBound - the final upper bound, adjusted for any weight changes.
+     * At any given time, the BPT price trading range is defined by the BPT price at the time
+     * the circuit breaker was set, multiplied by the weight-adjusted bounds.
```
TomAFrench, line 61: This seems like it should sit above the `@param` tags and be in a `@dev` tag.

pkg/pool-weighted/contracts/managed/CircuitBreakerStorageLib.sol
```diff
@@ -168,72 +170,67 @@ library CircuitBreakerStorageLib {
      * given by: B1 = k * B0. Since the BPTPrice0 = totalSupply*weight/B0, and BPTPrice1 = totalSupply*weight/B1,
      * we can combine these equations to compute the BPT price ratio BPTPrice1/BPTPrice0 = 1/k; BPT1 = BPT0/k.
      * So we see that the "conversion factor" between the spot price ratio and BPT Price ratio can be written
-     * as above BPT1 = BPT0 * (1/k), or more simply: (reference BPT price) * (priceRatio)**(1 - weight).
+     * as above BPT1 = BPT0 * (1/k), or more simply: (BPT price) * (priceRatio)**(1 - weight).
      *
-     * If the value of the weight complement has not changed, we can use the reference conversion ratios stored
+     * If the value of the weight complement has not changed, we can use the cached weight adjustments stored
```
TomAFrench, line 175:
```suggestion
     * If the value of the weight complement has not changed, we can use the cached adjusted bounds stored
```

"weight adjustments" to me implies a delta rather than the final value.

pkg/pool-weighted/contracts/managed/CircuitBreakerStorageLib.sol
```diff
@@ -168,72 +170,67 @@ library CircuitBreakerStorageLib {
...
+        uint256 referenceWeight = circuitBreakerState.decodeUint(_REFERENCE_WEIGHT_OFFSET, _REFERENCE_WEIGHT_WIDTH);
 
-        uint256 lowerBoundRatio;
-        uint256 upperBoundRatio;
+        uint256 adjustedLowerBound;
+        uint256 adjustedUpperBound;
 
-        if (initialWeightComplement == weightComplement) {
+        if (currentWeight == referenceWeight) {
             // If the weight complement hasn't changed since the circuit breaker was set, we can use the precomputed
```
TomAFrench, line 200:
```suggestion
            // If the weight complement hasn't changed since the circuit breaker was set, we can use the cached
```

pkg/pool-weighted/contracts/managed/CircuitBreakerStorageLib.sol
```diff
@@ -168,72 +170,67 @@ library CircuitBreakerStorageLib {
...
-     * @param weightComplement: This is (1 - currentWeight).
+     * of a token is generally given by: supply * weight / balance.
+     * @param weight: This is the current normalized weight of the token.
      * @param lowerBound: The value of the lower bound, expressed as a percentage.
      * @param upperBound: The value of the upper bound, expressed as a percentage.
      */
     function setCircuitBreaker(
         uint256 bptPrice,
-        uint256 weightComplement,
+        uint256 weight,
```
TomAFrench, line 233:
```suggestion
        uint256 referenceWeight,
```

pkg/pool-weighted/contracts/managed/CircuitBreakerStorageLib.sol
```diff
@@ -247,75 +244,75 @@ library CircuitBreakerStorageLib {
...
+                _ADJUSTED_UPPER_BOUND_OFFSET,
+                _ADJUSTED_BOUND_WIDTH
             );
     }
 
     /**
-     * @notice Update the cached ratios given a new weight complement.
+     * @notice Update the cached adjusted bounds, given a new weight.
      * @dev This might be used when weights are adjusted, pre-emptively updating the cache to improve performance
      * of operations after the weight change completed. Note that this does not update the BPT price: this is still
```
TomAFrench, line 286:
```suggestion
     * of operations after the weight change completes. Note that this does not update the BPT price: this is still
```

pkg/pool-weighted/contracts/managed/CircuitBreakerStorageLib.sol
```diff
@@ -247,75 +244,75 @@ library CircuitBreakerStorageLib {
...
      * relative to the last call to `setCircuitBreaker`. The intent is only to optimize the automatic bounds
      * adjustments due to changing weights.
      */
-    function updateBoundRatios(bytes32 circuitBreakerState, uint256 weightComplement) internal pure returns (bytes32) {
-        (uint256 lowerBoundRatio, uint256 upperBoundRatio) = CircuitBreakerLib.calcBoundaryConversionRatios(
+    function updateAdjustedBounds(bytes32 circuitBreakerState, uint256 currentWeight) internal pure returns (bytes32) {
```
TomAFrench, line 290:
```suggestion
    function updateAdjustedBounds(bytes32 circuitBreakerState, uint256 newReferenceWeight) internal pure returns (bytes32) {
```

Note that this weight is not going to be the current weight as we'll pass in the weight as of **after** the weight change so it will be a future weight.

pkg/pool-weighted/contracts/managed/CircuitBreakerStorageLib.sol
```diff
@@ -68,39 +68,41 @@ library CircuitBreakerStorageLib {
...
-    // bound to a minimum value > 0.
+    // bound to a minimum value >> 0.
     //
-    // Since the bound ratios are (bound percentage)**(weightComplement), and weights are stored normalized, the maximum
-    // weight complement is 1 - minimumWeight, which is 0.99 ~ 1. Therefore the ratio bounds are likewise constrained to
-    // 10**1 ~ 10. So we can use this as the maximum value of both the percentage and ratio values.
+    // Since the adjusted bounds are (bound percentage)**(1 - weight), and weights are stored normalized, the
+    // maximum normalized weight is 1 - minimumWeight, which is 0.99 ~ 1. Therefore the adjusted bounds are likewise
+    // constrained to 10**1 ~ 10. So we can use this as the maximum value of both the raw percentage and
+    // weight-adjusted bounds.
```
TomAFrench, line 105:
```suggestion
    // constrained to 10**1 ~ 10. So we can use this as the maximum value of both the raw and
    // weight-adjusted percentage bounds.
```

TomAFrench (dismissed):

pkg/pool-weighted/contracts/managed/CircuitBreakerStorageLib.sol
```diff
@@ -68,39 +68,41 @@ library CircuitBreakerStorageLib {
...
     //
-    // We then store the current BPT price, and the lower and upper bound conversion ratios, used to
-    // convert the percentage bounds into BPT prices that can be directly compared to the "runtime" BPT prices.
+    // We then store the current BPT price, and compute and cache the adjusted lower and upper bounds at the current
+    // weight. When multiplied by the stored BPT price, the adjusted bounds define the BPT price trading range: the
+    // "runtime" BPT prices can be directly compared to these BPT price bounds.
     //
-    // Since the price bounds shift along with the token weight, in general these bound ratios would need to be
+    // Since the price bounds need to be adjusted for the token weight, in general these adjusted bounds would be
     // computed every time. However, if the weight of the token and composition of the pool have not changed since
```
TomAFrench, line 79:
```suggestion
    // computed every time. However, if the weight of the token has not changed since
```

nventuro (approved):

PR #1906, opened 2022-10-05
Support incoming unscaled bptPrice

EndymionJkb:
Assume the incoming bptPrice was calculated with an unscaled token balance. Since our calculations assume all balances are scaled, we must adjust it.


pkg/pool-weighted/test/ManagedPoolSettings.test.ts
```diff
@@ -826,7 +826,13 @@ describe('ManagedPoolSettings', function () {
   });
 
   describe('circuit breakers', () => {
-    async function getBptPrice(): Promise<BigNumber> {
+    async function getUnscaledBptPrice(): Promise<BigNumber> {
+      const totalSupply = await pool.getActualSupply();
+
+      return fpDiv(fpMul(totalSupply, poolWeights[0]), initialBalances[0]);
```
nventuro, line 832: The first token has 18 decimals, so we're actually not testing anything

pkg/pool-weighted/contracts/managed/ManagedPoolSettings.sol
```diff
@@ -1027,14 +1027,21 @@ abstract contract ManagedPoolSettings is BasePool, ProtocolFeeCache, IControlled
...
+
+        _circuitBreakerState[token] = CircuitBreakerStorageLib.setCircuitBreaker(
+            scaledBptPrice,
             normalizedWeight.complement(),
             lowerBoundPercentage,
             upperBoundPercentage
         );
 
-        emit CircuitBreakerSet(token, bptPrice, lowerBoundPercentage, upperBoundPercentage);
+        emit CircuitBreakerSet(token, scaledBptPrice, lowerBoundPercentage, upperBoundPercentage);
```
nventuro, line 1044: I think it'd be good to have both the getter and this event return the unscaled bpt price, to make things easier for the outside world

pkg/pool-weighted/test/ManagedPoolSettings.test.ts
EndymionJkb: For now, making it use a dynamic tokenIndex.

pkg/pool-weighted/test/ManagedPoolSettings.test.ts
```diff
@@ -951,15 +969,12 @@ describe('ManagedPoolSettings', function () {
...
+            const expectedWeightComplement = FP_ONE.sub(poolWeights[tokenIndex]);
             const totalSupply = await pool.getActualSupply();
-            const scalingFactors = await pool.getScalingFactors();
 
-            const expectedBptPrice = fpDiv(
-              fpMul(totalSupply, poolWeights[0]),
-              fpMul(initialBalances[0], scalingFactors[0])
-            );
+            // Don't scale; the getter will return the unscaled value.
+            const expectedBptPrice = fpDiv(fpMul(totalSupply, poolWeights[tokenIndex]), initialBalances[tokenIndex]);
```
nventuro, line 977: Shouldn't this test that the pool returns the bpt price that we set?

nventuro: I guess the computation here returns the same thing. We could replace it with `getUnscaledBptPrice()` though

nventuro (approved):

PR #1907, opened 2022-10-05
Check circuit breakers on a regular swap

EndymionJkb:
Check the circuit breakers on a regular swap: _onTokenSwap. Since these do not change the supply, the only BPT prices that can change are the tokens involved in the swap. Not only that, we know the direction of the change.

The balance of tokenIn is increasing, which means the BPT price decreases (BPT price = supply*weight/balance), so we need to check the lower bound. For tokenOut, we need to check the upper bound.


pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -277,58 +285,101 @@ contract ManagedPool is ManagedPoolSettings {
...
+                CircuitBreakerLib.hasCircuitBreakerTripped(
+                    actualSupply,
+                    swapTokenData.tokenOutWeight,
+                    endingBalanceTokenOut,
+                    upperBoundBptPrice,
+                    false // check upper bound
+                )),
+            Errors.CIRCUIT_BREAKER_TRIPPED
+        );
+    }
```
nventuro, line 364: I'd put this inside a function

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -277,58 +285,101 @@ contract ManagedPool is ManagedPoolSettings {
...
 
-            swapFeeComplement = ManagedPoolStorageLib.getSwapFeePercentage(poolState).complement();
-        }
+        balanceTokenIn = _upscale(balanceTokenIn, swapTokenData.scalingFactorTokenIn);
+        balanceTokenOut = _upscale(balanceTokenOut, swapTokenData.scalingFactorTokenOut);
 
-        balanceTokenIn = _upscale(balanceTokenIn, scalingFactorTokenIn);
-        balanceTokenOut = _upscale(balanceTokenOut, scalingFactorTokenOut);
+        uint256 endingBalanceTokenIn;
+        uint256 endingBalanceTokenOut;
```
nventuro, line 296: These are never set

nventuro: Since the block below sets `amountCalculated` (and we know amount given - `request.amount`), we could pass both to the `checkBreaker` function I suggest and adjust theb alances with something like the following

```
    function _getAmounts(
        SwapKind kind,
        uint256 amountGiven,
        uint256 amountCalculated
    ) private pure returns (uint256 amountIn, uint256 amountOut) {
        if (kind == SwapKind.GIVEN_IN) {
            (amountIn, amountOut) = (amountGiven, amountCalculated);
        } else {
            // SwapKind.GIVEN_OUT
            (amountIn, amountOut) = (amountCalculated, amountGiven);
        }
    }
```

and then do

```
balance token in += amount in
balance token put -= amount out
```

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -277,58 +285,101 @@ contract ManagedPool is ManagedPoolSettings {
...
+        );
+        uint256 upperBoundBptPrice = CircuitBreakerStorageLib.getBptPriceBound(
+            _getCircuitBreakerState(request.tokenOut),
+            swapTokenData.tokenOutWeight,
+            false
+        );
+        uint256 actualSupply = _getActualSupply(_getVirtualSupply());
+
+        _require(
+            !(CircuitBreakerLib.hasCircuitBreakerTripped(
```
nventuro, line 348: I'd have this as `require(!tripped && !tripped)`, or even better two separate requires (to avoid so much indentation).

nventuro: These changes + separate function might let us remove the struct btw

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -277,58 +285,101 @@ contract ManagedPool is ManagedPoolSettings {
...
+                actualSupply,
+                swapTokenData.tokenInWeight,
+                endingBalanceTokenIn,
+                lowerBoundBptPrice,
+                true // check lower bound
```
nventuro, line 353: I think i'd defintely split this into two requires, explaining we're checking the lower bound of token in (to make sure we don't buy too much of a bad thing), and the upper bound of the token out

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -277,58 +285,101 @@ contract ManagedPool is ManagedPoolSettings {
...
+        uint256 lowerBoundBptPrice = CircuitBreakerStorageLib.getBptPriceBound(
+            _getCircuitBreakerState(request.tokenIn),
+            swapTokenData.tokenInWeight,
+            true
+        );
+        uint256 upperBoundBptPrice = CircuitBreakerStorageLib.getBptPriceBound(
+            _getCircuitBreakerState(request.tokenOut),
+            swapTokenData.tokenOutWeight,
+            false
+        );
```
nventuro, line 344: If splitting into two requires, I'd have this be interweaves with the requires, to make it clear the prices are with different tokens, and likely rename them to e.g. `lowerBoundBptPriceTokenIn`.

nventuro (approved):

PR #1908, opened 2022-10-05
Check circuit breakers on join/exits

EndymionJkb:
Since the virtual supply will change on all joins or exits (including join/exitSwaps), *all* BPT prices will change: and unpredictably at that. So we need to check both directions on all tokens in these cases.


pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -414,16 +454,29 @@ contract ManagedPool is ManagedPoolSettings {
...
             balances,
-            _getNormalizedWeights(tokens),
+            normalizedWeights,
             scalingFactors,
             actualSupply,
             userData
         );
 
+        // Do not check circuit breakers on proportional joins, which do not change BPT prices.
+        if (userData.joinKind() != WeightedPoolUserData.JoinKind.ALL_TOKENS_IN_FOR_EXACT_BPT_OUT) {
```
nventuro, line 469: I'd not have this special case to reduce the number of edge cases we support

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -603,6 +670,66 @@ contract ManagedPool is ManagedPoolSettings {
...
+
+            // Check both upper and lower bounds.
+            _require(
+                !(CircuitBreakerLib.hasCircuitBreakerTripped(
+                    virtualSupply,
+                    poolDelta.normalizedWeights[i],
+                    finalBalance,
+                    lowerBoundBptPrice,
+                    true // check lower bound
+                ) ||
```
nventuro, line 720: I'd replicate the new strcuture from #1907 once we have it

pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol
```diff
@@ -118,6 +125,14 @@ contract CircuitBreakerLibTest is Test {
             uint256(bptPrice).mulUp(expectedUpperBptPrice),
             _MAX_RELATIVE_ERROR
         );
+
+        // Single-sided
+
```
TomAFrench, line 130:
```suggestion
```

pkg/pool-weighted/contracts/managed/ManagedPool.sol
EndymionJkb: Ok; my only reservation is that we're greatly increasing the gas cost and therefore "punishing" proportional joins. If a proportional join were much cheaper than an "exact tokens in" join (even a proportional one), it would incentivize people to join this way, which helps keep the pool well-behaved: especially for pools with higher token counts.

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -659,6 +652,162 @@ contract ManagedPool is ManagedPoolSettings {
...
+            actualSupply,
+            tokenData.tokenInWeight,
+            balanceTokenIn.add(amountIn)
+        );
+
+        // Since the balance of tokenOut is decreasing, its BPT price will increase,
+        // so we need to check the upper bound.
+        _checkCircuitBreaker(
+            BoundCheckKind.UPPER,
+            _getCircuitBreakerState(request.tokenOut),
```
TomAFrench, line 690: We could move this call inside `_checkCircuitBreaker`

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -659,6 +652,162 @@ contract ManagedPool is ManagedPoolSettings {
...
+        }
+    }
+
+    // Check the appropriate circuit breaker(s) according to the BoundCheckKind.
+    function _checkCircuitBreaker(
+        BoundCheckKind checkKind,
+        bytes32 circuitBreakerState,
+        uint256 actualSupply,
+        uint256 weight,
+        uint256 balance
```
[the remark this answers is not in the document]
TomAFrench: This improves consistency with the calling function (where args are ordered `actualSupply, `balance`, `weight`)

pkg/pool-weighted/contracts/managed/ManagedPool.sol
```diff
@@ -659,6 +652,158 @@ contract ManagedPool is ManagedPoolSettings {
...
+                : (actualSupply.sub(amountCalculated), request.amount);
+        }
+
+        // Since this is a swap, we do not have all the tokens, balances, or weights, and need to fetch them.
+        (IERC20[] memory tokens, uint256[] memory balances) = _getPoolTokens();
+        uint256[] memory normalizedWeights = _getNormalizedWeights(tokens);
+        uint256[] memory scalingFactors = _scalingFactors(tokens);
+        _upscaleArray(balances, scalingFactors);
+
+        // Initialize to all zeros, and set the deltaAmount associated with the swap.
```
TomAFrench, line 730: We don't make use of the term `deltaAmount` anywhere else. We should update this comment.

TomAFrench (approved):

PR #1912, opened 2022-10-09
Add proportional joins/exits to ComposableStablePool

### LINKED ISSUE #1911: Add proportional joins and exits to StablePool
Not having these is problematic because it makes it hard to initiate an exit that has no price impact without at the same time not having any dust (due to slippage protection). All Pools should have this capability.

EndymionJkb:
Duplicated proportional amountsIn calculator from WeightedMath in StableMath (for now).

For consistency with exits, renamed joinGivenOut in the Stable model to singleJoinGivenOut, using joinGivenOut for the multi-token proportional.

Closes #1911


pkg/balancer-js/src/pool-stable/encoder.ts
```diff
@@ -5,11 +5,13 @@ export enum StablePoolJoinKind {
   INIT = 0,
   EXACT_TOKENS_IN_FOR_BPT_OUT,
   TOKEN_IN_FOR_EXACT_BPT_OUT,
+  ALL_TOKENS_IN_FOR_EXACT_BPT_OUT,
 }
 
 export enum StablePoolExitKind {
   EXACT_BPT_IN_FOR_ONE_TOKEN_OUT = 0,
   BPT_IN_FOR_EXACT_TOKENS_OUT,
+  EXACT_BPT_IN_FOR_TOKENS_OUT,
```
nventuro, line 14:
```suggestion
  EXACT_BPT_IN_FOR_TOKENS_OUT,
  BPT_IN_FOR_EXACT_TOKENS_OUT,
```

We should use the same ids we used in the past

nventuro: (This change needs to be done in multiple places)

pkg/pool-stable/contracts/StableMath.sol
```diff
@@ -447,4 +447,29 @@ library StableMath {
 
         _revert(Errors.STABLE_GET_BALANCE_DIDNT_CONVERGE);
     }
+
+    function _calcAllTokensInGivenExactBptOut(
```
nventuro, line 451: Why not have `_computeProportionalAmountsIn`?

EndymionJkb: I'm using the same name as WeightedMath. Change that one too?

nventuro: Yes!

pkg/balancer-js/src/pool-stable/encoder.ts
```diff
@@ -5,10 +5,12 @@ export enum StablePoolJoinKind {
   INIT = 0,
   EXACT_TOKENS_IN_FOR_BPT_OUT,
   TOKEN_IN_FOR_EXACT_BPT_OUT,
+  ALL_TOKENS_IN_FOR_EXACT_BPT_OUT,
 }
 
 export enum StablePoolExitKind {
   EXACT_BPT_IN_FOR_ONE_TOKEN_OUT = 0,
+  EXACT_BPT_IN_FOR_TOKENS_OUT,
```
nventuro, line 13: I'd rename this to ALL_TOKENS_OUT, for simmetry

pkg/balancer-js/src/pool-stable/encoder.ts
```diff
@@ -70,4 +79,11 @@ export class StablePoolEncoder {
       ['uint256', 'uint256[]', 'uint256'],
       [StablePoolExitKind.BPT_IN_FOR_EXACT_TOKENS_OUT, amountsOut, maxBPTAmountIn]
     );
+
+  /**
+   * Encodes the userData parameter for exiting a StablePool proportionally
+   * @param bptAmountIn - the amount of BPT to burn in exchange for withdrawn tokens
+   */
+  static exitExactBptInForTokensOut = (bptAmountIn: BigNumberish): string =>
+    defaultAbiCoder.encode(['uint256', 'uint256'], [StablePoolExitKind.EXACT_BPT_IN_FOR_TOKENS_OUT, bptAmountIn]);
```
nventuro, line 88: Nit, but can you move this so the order in the enum and the functions matches?

pkg/pool-stable/contracts/ComposableStablePool.sol
```diff
@@ -733,13 +733,29 @@ contract ComposableStablePool is
...
+            return _joinAllTokensInForExactBptOut(preJoinExitSupply, balances, userData);
         } else if (kind == StablePoolUserData.JoinKind.TOKEN_IN_FOR_EXACT_BPT_OUT) {
             return _joinTokenInForExactBPTOut(preJoinExitSupply, preJoinExitInvariant, currentAmp, balances, userData);
         } else {
             _revert(Errors.UNHANDLED_JOIN_KIND);
         }
     }
 
+    /**
+     * @dev Proportional join. Pays no protocol fees.
```
nventuro, line 746:
```suggestion
     * @dev Proportional join. Pays no swap fees.
```

pkg/pool-stable/contracts/ComposableStablePool.sol
```diff
@@ -830,13 +846,29 @@ contract ComposableStablePool is
...
+            return _exitExactBPTInForTokensOut(preJoinExitSupply, balances, userData);
         } else if (kind == StablePoolUserData.ExitKind.EXACT_BPT_IN_FOR_ONE_TOKEN_OUT) {
             return _exitExactBPTInForTokenOut(preJoinExitSupply, preJoinExitInvariant, currentAmp, balances, userData);
         } else {
             _revert(Errors.UNHANDLED_EXIT_KIND);
         }
     }
 
+    /**
+     * @dev Proportional exit. Pays no protocol fees.
```
nventuro, line 859:
```suggestion
     * @dev Proportional exit. Pays no swap fees. This is functionally equivalent to the recovery mode exit, except this
     * doesn't skip protocol fee collection, calling rate providers, etc., and doesn't require recovery mode to be enabled.
```

pkg/balancer-js/src/pool-stable/encoder.ts
EndymionJkb: Well, then it will be different from WeightedPool. As before, change that one too?

pkg/interfaces/contracts/pool-stable/StablePoolUserData.sol
```diff
@@ -15,8 +15,8 @@
 pragma solidity ^0.7.0;
 
 library StablePoolUserData {
-    enum JoinKind { INIT, EXACT_TOKENS_IN_FOR_BPT_OUT, TOKEN_IN_FOR_EXACT_BPT_OUT }
-    enum ExitKind { EXACT_BPT_IN_FOR_ONE_TOKEN_OUT, BPT_IN_FOR_EXACT_TOKENS_OUT }
+    enum JoinKind { INIT, EXACT_TOKENS_IN_FOR_BPT_OUT, TOKEN_IN_FOR_EXACT_BPT_OUT, ALL_TOKENS_IN_FOR_EXACT_BPT_OUT }
+    enum ExitKind { EXACT_BPT_IN_FOR_ONE_TOKEN_OUT, EXACT_BPT_IN_FOR_ALL_TOKENS_OUT, BPT_IN_FOR_EXACT_TOKENS_OUT }
```
danielmkm, line 19: Hey guys! First of all thanks so much for adding these operations back in, was in the process of doing it myself and preparing to beg when I saw this PR 😅.

Adding this exit option in the middle here aligns with the previous config, but will now break backwards compatibility with the exiting composable stable. It's a trade off, but I would say that it might be better to not break backwards compatibility as it increases implementation complexity.

EndymionJkb: I had it that way, but flipped it because of https://github.com/balancer-labs/balancer-v2-monorepo/pull/1912#discussion_r1006137221. Sounds like I should change it back then. @nventuro ?

pkg/balancer-js/src/pool-stable/encoder.ts
nventuro: Yes

pkg/interfaces/contracts/pool-stable/StablePoolUserData.sol
nventuro: @danielmkm what approach do you think would be better? If done as in this PR, we only have two variants: those with and those without proportional exits (same as we have today).

danielmkm: I think the best approach would be to preserve the existing ComposableStable ordering, so swap `BPT_IN_FOR_EXACT_TOKENS_OUT` and `EXACT_BPT_IN_FOR_ALL_TOKENS_OUT` in the enum definition.

```
enum ExitKind { EXACT_BPT_IN_FOR_ONE_TOKEN_OUT, BPT_IN_FOR_EXACT_TOKENS_OUT, EXACT_BPT_IN_FOR_ALL_TOKENS_OUT}
```

By doing this, any integration that gets built against the current ComposableStable will be able to support the new pool version without making any code changes. They can then choose to add proportional join/exit support on their own timeline.

With the PR as it is, the enum reordering is a breaking change. So any existing integration cannot just use the new version without also making code changes, usually meaning a longer timeline for adoption as additional systems need to be updated to account for the changes.

When doing the first integration I lost some time not realizing the enum value of `BPT_IN_FOR_EXACT_TOKENS_OUT` had changed, and the error code you get is the dreaded 430 😅.

PR #1916, opened 2022-10-10
Use SafeERC20 for approvals

jubeira:
This PR includes:
- An implementation for `safeApproval`, with the following behavior:
 - Works with standard ERC20 implementations returning `true`.
 - Reverts if `approve` returns false.
 - Works if `approve` does not return anything.
- A sanity check over `transfer`, `transferFrom` and `approve` throughout the repository. As a general rule, safe versions are preferred unless we are dealing with tokens that are known to be compliant.
- A test for `safeApproval`.


pkg/liquidity-mining/contracts/gauges/arbitrum/ArbitrumRootGauge.sol
```diff
@@ -45,7 +45,7 @@ contract ArbitrumRootGauge is StakelessGauge {
 
     function _postMintAction(uint256 mintAmount) internal override {
         // Token needs to be approved on the gateway NOT the gateway router
-        _balToken.approve(_gateway, mintAmount);
+        _balToken.approve(_gateway, mintAmount); // ERC20 compliant; no need to use safeApproval.
```
nventuro, line 48: I'd just use `safeApprove` everywhere

pkg/solidity-utils/contracts/openzeppelin/SafeERC20.sol
```diff
@@ -19,6 +19,20 @@ import "@balancer-labs/v2-interfaces/contracts/solidity-utils/openzeppelin/IERC2
  * which allows you to call the safe operations as `token.safeTransfer(...)`, etc.
  */
 library SafeERC20 {
+
+    function safeApproval(
```
nventuro, line 23: I'd rename this to `safeApprove`

pkg/solidity-utils/contracts/openzeppelin/SafeERC20.sol
```diff
@@ -19,6 +19,20 @@ import "@balancer-labs/v2-interfaces/contracts/solidity-utils/openzeppelin/IERC2
...
+        IERC20 token,
+        address to,
+        uint256 value
+    ) internal {
+        // Some contracts need their allowance reduced to 0 before setting it to an arbitrary amount.
+        if (token.allowance(address(this), address(to)) > 0) {
```
nventuro, line 29:
```suggestion
        if (value != 0 && token.allowance(address(this), address(to)) > 0) {
```

pkg/liquidity-mining/contracts/gauges/arbitrum/ArbitrumRootGauge.sol
nventuro: It's easier to check and we can skip these comments

nventuro (commented):
For completeness you could also test that we bubble-up revert reasons correctly - though that's already covered by our helper

pkg/solidity-utils/contracts/openzeppelin/SafeERC20.sol
jubeira: I got `safeApproval` from somewhere I can't remember now, but it did sound weird :)
Done.

pkg/liquidity-mining/contracts/gauges/arbitrum/ArbitrumRootGauge.sol
jubeira: Done.

nventuro (approved):

PR #1917, opened 2022-10-12
Sturdy linear pool

iris112:


TomAFrench:
Hey Iris :wave:.  It looks like Sturdy is a fork of Aave so you should be able to just use the AaveLinearPoolFactory and just pass in sToken addresses in place of aTokens. This could be a simpler way for you to get started without having to deploy a new pool type.

nventuro:
@iris112 can you check if the solution proposed above would work? That'd indeed be a much simpler solution.

TomAFrench:
@iris112 The `AaveLinearPool` uses `StaticAToken`s as the Balancer vault isn't compatible with rebasing tokens. Any interest accrued by aTokens will be lost. This means that placing sTokens in the vault will cause any interest to be lost as well and so you would have to deploy a "StaticSToken" wrapper for your tokens to be compatible with Balancer.

If you reuse Aave's `StaticAToken` contracts as your wrapper you'll then be compatible with AaveLinearPool.

iris112:
@TomAFrench  
Thanks, I'm working for `StaticAToken` now. AFter deploy I'll update this PR.

TomAFrench:
Great, if you end up using Aave's `StaticAToken` contract then you should be compatible with the existing factory.

I'm going to close this PR as it seems like it's unneeded but feel free to reopen it / reach out if you can't use the `AaveLinearPoolFactory` deployment for some reason.

iris112:
@TomAFrench 
I deployed `StaticAToken`.
Let me know the next step.
Thanks.

PR #1918, opened 2022-10-13
ProtocolFeeSplitter

bxmmm1:
This PR is an implementation of the [Protocol revenue sharing for pool creators](https://forum.balancer.fi/t/proposal-protocol-revenue-sharing-for-pool-creators/2539) proposal.

A formal verification report by Certora is available [here](https://github.com/balancer-labs/balancer-v2-monorepo/files/9777709/Certora_report.pdf).


pkg/standalone-utils/contracts/ProtocolFeeSplitter.sol
```diff
@@ -0,0 +1,152 @@
...
+        protocolFeesCollector = _protocolFeesCollector;
+        treasury = _treasury;
+    }
+
+    function setRevenueSharingFeePercentage(bytes32 poolId, uint256 newSwapFeePercentage)
+        external
+        override
+        authenticate
+    {
+        _require(newSwapFeePercentage >= _MIN_REVENUE_SHARING_FEE_PERCENTAGE, Errors.SPLITTER_FEE_PERCENTAGE_TOO_LOW);
```
nventuro, line 75: Why is there a minimum? The default can be as low as zero, so it feels weird for the override to have a non-zero minimum. Can't the override be zero? What if fees have to be turned off for a Pool?

pkg/standalone-utils/contracts/ProtocolFeeSplitter.sol
```diff
@@ -0,0 +1,152 @@
...
+    // All fee percentages are 18-decimal fixed point numbers.
+    // Absolute maximum fee percentages (1e18 = 100%, 1e16 = 1%).
+    uint256 private constant _MIN_REVENUE_SHARING_FEE_PERCENTAGE = 1e16; // 1%
+    uint256 private constant _MAX_REVENUE_SHARING_FEE_PERCENTAGE = 50e16; // 50%
+    address private constant _DELEGATE_OWNER = 0xBA1BA1ba1BA1bA1bA1Ba1BA1ba1BA1bA1ba1ba1B;
+
+    IProtocolFeesCollector public immutable protocolFeesCollector;
+
+    // Balancer DAO Multisig
+    address public immutable treasury;
```
nventuro, line 53: Having this be `immutable` seems to be make this contract a bit too strict, surely we can make this mutable and have some future proofing?

pkg/standalone-utils/contracts/ProtocolFeeSplitter.sol
```diff
@@ -0,0 +1,152 @@
...
+
+        if (beneficiary == address(0)) {
+            uint256[] memory amounts = new uint256[](1);
+            amounts[0] = feeCollectorBptBalance;
+            protocolFeesCollectorMemory.withdrawCollectedFees(tokens, amounts, treasury);
+            emit FeesCollected(poolId, beneficiary, 0, treasury, feeCollectorBptBalance);
+        } else {
+            uint256[] memory amounts = new uint256[](1);
+
+            uint256 poolFeeOverride = poolSettings[poolId].revenueSharePercentageOverride;
```
nventuro, line 117: This has a critical bug. `VaultHelpers.toPoolAddress` will simply remove the lower bits, leaving just the address, but we never check for it to be valid in any way. This means anyone can call `collectFees` with any `poolId` that resolves to the same address, and e.g. force collection in a way in which there's no beneficiary.

Similarly, if governance were to reduce a Pool's share to be below the default, the Pool owner could similarly set a benecifiary using an alternative `poolId` and sidestep the fee override that way.

There's a couple ways to solve this. One would be to check that the PoolId is legitimate in the Vault by e.g. calling `getPool` (and using this instead of `VaultHelpers`). Another would be to simply pass the Pool's address. I prefer the former, since it remains closer to the Vault's Pool model and doesn't assume too much about the Pool's (though we already assume it features `getOwner()` and is a token).

pkg/standalone-utils/contracts/ProtocolFeeSplitter.sol
```diff
@@ -0,0 +1,152 @@
...
+            uint256 feePercentage = poolFeeOverride != 0 ? poolFeeOverride : defaultRevenueSharingFeePercentage;
+
+            (uint256 ownerAmount, uint256 treasuryAmount) = _computeAmounts(feeCollectorBptBalance, feePercentage);
+
+            // owner doesn't get the tokens always
+            if (ownerAmount > 0) {
+                // mutate array, and set the owner amount for owner withdrawal
+                amounts[0] = ownerAmount;
+                protocolFeesCollectorMemory.withdrawCollectedFees(tokens, amounts, beneficiary);
```
nventuro, line 126: Fees should be collected from the `ProtocolFeesWithdrawer`, not the collector directly. Regardless of that, is this contract expected to be the only account with permission to call `withdrawCollectedFees`?

nventuro (commented):
Hello @bxmmm1, thanks a lot for opening this PR!

I have multiple suggestions in terms of minor adjustments, etc., but first I wanted to address some design questions on the proposed contracts, including a critical bug in the fee collection function.

pkg/standalone-utils/contracts/ProtocolFeeSplitter.sol
bxmmm1: I wasn't aware that it should go through `ProtocolFeesWithdrawer`, I'll make the adjustments.
Yes, this contract is the only one that needs `withdrawCollectedFees` permission.

pkg/standalone-utils/contracts/ProtocolFeeSplitter.sol
BritikovKI: But if treasury changes I suppose a lot of things will change with it?
This immutability on the other hand enforces security, that DAO definitely won't be able to give an ovnership over this contract to some single address, even in the case of some intelligent attack

pkg/standalone-utils/contracts/ProtocolFeeSplitter.sol
```diff
@@ -0,0 +1,162 @@
...
+        tokens[0] = bpt;
+
+        address beneficiary = poolSettings[poolId].beneficiary;
+
+        if (beneficiary == address(0)) {
+            uint256[] memory amounts = new uint256[](1);
+            amounts[0] = feeCollectorBptBalance;
+            protocolFeesWithdrawerMemory.withdrawCollectedFees(tokens, amounts, treasury);
+            emit FeesCollected(poolId, beneficiary, 0, treasury, feeCollectorBptBalance);
+        } else {
```
nventuro, line 124: This duplication seems unnecessary - you could instead have the following function

```
fn getPoolOwnerFeePercentage() {
  if beneficiary == 0
    return 0

  if feeOverride != 0
    return fee override
  else
   return default
} 
```

then pass that to `_computeAmounts`, and have the suggested `withdraw` helper simply return immediately if `amount` is 0. You'll end up with fewer branches and less special-casing - plus, the above function is a good `public` getter anyway, since otherwise it's relatively hard to compute from the outside what those might be.

I'd consider even going one step further and providing a second getter that returns the amounts that could be collected to the owner and to the treasury if `collect` was called now.

pkg/standalone-utils/test/ProtocolFeeSplitter.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    });
+    it('sets the treasury', async () => {
+      expect(await protocolFeeSplitter.treasury()).to.be.eq(treasury.address);
+    });
+  });
+
+  it('changes the treasury', async () => {
+    const receipt = await (await protocolFeeSplitter.connect(admin).setTreasury(newTreasury.address)).wait();
+    expectEvent.inReceipt(receipt, 'TreasuryChanged', { newTreasury: newTreasury.address });
+  });
```
nventuro, line 133: There should be a test where an account with no permissions fails to change the treasury

pkg/standalone-utils/test/ProtocolFeeSplitter.test.ts
```diff
@@ -0,0 +1,229 @@
...
+      expectEvent.inReceipt(receipt, 'DefaultRevenueSharingFeePercentageChanged', { revenueSharePercentage: newFee });
+      expect(await protocolFeeSplitter.defaultRevenueSharingFeePercentage()).to.be.eq(newFee);
+    });
+  });
+
+  describe('revenue sharing override', async () => {
+    it('overrides revenue sharing percentage for a pool', async () => {
+      const newFee = bn(50e16); // 50%
+      const receipt = await (
+        await protocolFeeSplitter.connect(admin).setRevenueSharingFeePercentage(poolId, newFee)
```
nventuro, line 150: There should be a test where an account with no permissions fails to change the pool fee

pkg/standalone-utils/test/ProtocolFeeSplitter.test.ts
```diff
@@ -0,0 +1,229 @@
...
+      ).to.be.revertedWith('SPLITTER_FEE_PERCENTAGE_TOO_HIGH');
+    });
+  });
+
+  describe('pool beneficiary', async () => {
+    it('reverts if caller is not the pool owner', async () => {
+      await expect(
+        protocolFeeSplitter.connect(liquidityProvider).setPoolBeneficiary(poolId, liquidityProvider.address)
+      ).to.be.revertedWith('SENDER_NOT_ALLOWED');
+    });
```
nventuro, line 168: I'd add a test here where the owner successfully sets the beneficiary, testing the return value, event, etc. (in separate tests)

pkg/standalone-utils/test/ProtocolFeeSplitter.test.ts
```diff
@@ -0,0 +1,229 @@
...
+  });
+
+  describe('revenue sharing override', async () => {
+    it('overrides revenue sharing percentage for a pool', async () => {
+      const newFee = bn(50e16); // 50%
+      const receipt = await (
+        await protocolFeeSplitter.connect(admin).setRevenueSharingFeePercentage(poolId, newFee)
+      ).wait();
+      expectEvent.inReceipt(receipt, 'PoolRevenueShareChanged', { poolId: poolId, revenueSharePercentage: newFee });
```
nventuro, line 152: This is only testing that the event was emitted, but not the getter. There should be a separate test that checks that.

nventuro: The same thing happens in the default fee and the treasury setter tests

pkg/standalone-utils/test/ProtocolFeeSplitter.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    });
+  });
+
+  describe('collect fees', async () => {
+    it('reverts if no BPT collected', async () => {
+      expect(await pool.balanceOf(protocolFeesCollector.address)).to.equal(0);
+      await expect(protocolFeeSplitter.collectFees(poolId)).to.be.revertedWith('NO_BPT_FEES_COLLECTED');
+    });
+
+    it('distributes collected BPT fees to owner and treasury (fee percentage not defined)', async () => {
```
nventuro, line 177: I suggest using `context` blocks with a `sharedBeforeEach` to separate the setup (setting of fee percentages, etc) from the actual test.

Also, the shared setup (there being tokens to distribute) could be moved to a broader `context('when the fee collector holds BPT')`.

pkg/interfaces/contracts/standalone-utils/IProtocolFeeSplitter.sol
```diff
@@ -0,0 +1,71 @@
...
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
```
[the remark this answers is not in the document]
nventuro: (see #1947)

nventuro:
Thanks! I opened https://github.com/daoism-systems/balancer-v2-monorepo/pull/4 with some final adjustments: once that's merged, those changes will be reflected here and we can merge those.

bxmmm1:
@nventuro Np, please message me if I should do something else.

pkg/interfaces/contracts/standalone-utils/IProtocolFeeSplitter.sol
```diff
@@ -0,0 +1,110 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity >=0.7.0 <0.9.0;
+
+import "@balancer-labs/v2-interfaces/contracts/vault/IVault.sol";
+import "@balancer-labs/v2-interfaces/contracts/standalone-utils/IProtocolFeesWithdrawer.sol";
```
nventuro, line 18:
```suggestion
import "../vault/IVault.sol";
import "./IProtocolFeesWithdrawer.sol";
```

nventuro: This causes build issues

PR #1922, opened 2022-10-18
Scale balances to accommodate low-decimal tokens

EndymionJkb:
Addresses flaky tests failing intermittently in master. Some tests pass in FP values as balances, assuming tokens are 18 decimals. Since we recently changed varyDecimals to start at 0, FP_ONE is a really huge balance for a zero-decimal token, which was causing problems.

There is now a `scaleBalances` function in TokenList (same logic as `mintScaled`), which takes raw numbers (e.g, 8, 10, 12) and converts them to what a user would really pass in given the decimals. 100 USDC = 100000000 with 6 decimals.


pvt/helpers/src/models/tokens/TokenList.ts
```diff
@@ -160,4 +160,8 @@ export default class TokenList {
       this.tokens.sort((tokenA, tokenB) => (tokenA.address.toLowerCase() > tokenB.address.toLowerCase() ? 1 : -1))
     );
   }
+
+  scaleBalances(rawBalances: number[]): BigNumber[] {
```
nventuro, line 164: What if instead we called this `scaledBalances` and had it receive a function that returns a number, and then call it repeateadly and scale them? e.g.

```
scaledBalances(rawBalance: (number) => BigNumber)): BigNumber[] {
   return this.tokens.map((t, i) => rawBalance(i).mul(10 ** t.decimals));
}
```

We would then simplify the callsites to e.g. `initialBalances: poolTokens.scaledBalances(() => 10 + random(10))`

nventuro (commented):
Thanks! Left a suggestion for an alternative API that does more work for us, simplifying usage

nventuro (approved):

nventuro:
(pending CI passing)

PR #1923, opened 2022-10-19
Fix weight generation to prevent flaky tests

nventuro:
The old weight generation scheme could produce weights that would end up being invalid, specially for large token counts. This should address that. I also added some randomness to the new weight.


EndymionJkb (approved):
Some inline comments

PR #1926, opened 2022-10-19
Remove base recovery mode implementation

TomAFrench:
This saves almost 200 bytes from ManagedPool (and is a good practice anyway)

I haven't run tests on all the other contracts this touches so will likely do touch ups to this PR.


nventuro (approved):
Lovely

TomAFrench:
Please don't merge unless you like red crosses

TomAFrench:
@nventuro Do you want to have a quick check over this again as there have been a fair number of changes.

nventuro (approved):
We'll need to add tests to ManagedPool, ComposableStablePool, BaseWeightedPool and LinearPool, since those now have custom recovery mode exits.

PR #1930, opened 2022-10-20
Make calcAdjustedBound external

jubeira:
Split `pow` out of CircuitBreakerLib to reduce bytecode size.

On top of #1921 to keep track of bytecode improvements.

### Results

`hh compile && cat artifacts/contracts/managed/ManagedPool.sol/ManagedPool.json | jq -r '.deployedBytecode' | wc --chars | awk '{print (($1-2)/2)/1024}'`:

- Base: @ https://github.com/balancer-labs/balancer-v2-monorepo/commit/a2986eab1e8485dd57875f54ed608ae56b209fa8: 27,7603
- This change: 25,1626


TomAFrench:
Merging master into this now.

PR #1933, opened 2022-10-20
Documentation updates

EndymionJkb:
Spotted a typo in ManagedPoolSettings, which led to a general doc review. Adjusted a few things we'd talked about (e.g., audit comments, 60 vs 64 bits in reference weight storage, circuit breaker metaphors).

I'm still too close to the CB docs. We should probably "pick one" explanation of circuit breakers for the in-code docs (perhaps presenting other formulations externally); it's a bit of a shotgun blast at the moment.


pkg/pool-weighted/contracts/managed/CircuitBreakerStorageLib.sol
```diff
@@ -69,7 +69,8 @@ library CircuitBreakerStorageLib {
     // Store circuit breaker information per token
     // When the circuit breaker is set, the caller passes in the lower and upper bounds (expressed as percentages),
     // and the current BPT price and normalized weight. Since this value is bounded by 1e18, which fits in ~60 bits,
```
TomAFrench, line 71: I think is just wrong and we shouldn't have the section about `~60 bits` at all.

The max bound percentage is `10e18` not `1e18`. `2**63 = 9.2e18` and `2**64 = 1.8e19` so we need 64 bits, it's not that we're just having round numbers.

EndymionJkb: Well, the *weight* is bound by 1, but it is confusing (mixing two things). How about:

```
// Since this value is bounded by 10e18, it fits in 64 bits,
// there is no need for compression. The weight is bound by 1e18, and fits in ~60 bits. For a bit of extra
// precision (and to use round numbers), we store the weight in 64 bits.
```

TomAFrench: Ohhh, sorry I'm being dumb. I misread this as the bound.

TomAFrench: I'd drop the bit about an extra bit of precision as we don't gain any precision by providing the extra bits.

jubeira (commented):
I did one quick pass over the changes and their surroundings.

+1 to the examples; they really help to illustrate what's going on.

nventuro (approved):

PR #1936, opened 2022-10-21
Match `Swap` event from `MockVault` with `IVault` event.

jubeira:
Same as #1935, for swaps.

### DIFF
--- a/pkg/pool-utils/contracts/test/MockVault.sol
+++ b/pkg/pool-utils/contracts/test/MockVault.sol
@@ -36,7 +36,13 @@ contract MockVault is IPoolSwapStructs {
 
     mapping(bytes32 => Pool) private pools;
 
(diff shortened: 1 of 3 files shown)


pkg/pool-utils/contracts/test/MockVault.sol
```diff
@@ -112,8 +118,11 @@ contract MockVault is IPoolSwapStructs {
         uint256 balanceTokenIn,
         uint256 balanceTokenOut
     ) external {
-        uint256 amount = IMinimalSwapInfoPool(pool).onSwap(request, balanceTokenIn, balanceTokenOut);
-        emit Swap(request.poolId, request.tokenIn, request.tokenOut, amount);
+        uint256 swapAmount = IMinimalSwapInfoPool(pool).onSwap(request, balanceTokenIn, balanceTokenOut);
```
EndymionJkb, line 121: Not so important here, but we've called this amountCalculated elsewhere.

nventuro (approved):

PR #1937, opened 2022-10-21
Refactor: `BaseManagedPoolFactory` --> `ManagedPoolFactory`

jubeira:
And `ManagedPoolFactory` --> `ControlledManagedPoolFactory`, as suggested by @nventuro.


EndymionJkb:
Do we want to say OwnedManagedPoolFactory?

nventuro:
Tbh I don't think the name is very important - we're not deploying this yet, and can sort it out later.

nventuro (approved):

PR #1938, opened 2022-10-21
Increase max tokens

EndymionJkb:
One thing I realized giving this a last once-over is that the max token value (38) is completely arbitrary. It arose from gas limits on joins/exits in a very old version, and we just never changed it. It used to be 50 (got reduced because of the aforementioned gas issues). Looks like it once again works with 50.

### DIFF
--- a/pvt/benchmarks/multihop.ts
+++ b/pvt/benchmarks/multihop.ts
@@ -39,6 +39,11 @@ async function main() {
   await multihop((index: number) => getWeightedPool(vault, tokens, 38, index), false);
   await multihop((index: number) => getWeightedPool(vault, tokens, 38, index), true);
 
+  console.log(`\n# Managed Pool with 50 tokens`);
+
+  await multihop((index: number) => getWeightedPool(vault, tokens, 50, index), false);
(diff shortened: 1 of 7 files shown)


pvt/benchmarks/multihop.ts
```diff
@@ -39,6 +39,11 @@ async function main() {
   await multihop((index: number) => getWeightedPool(vault, tokens, 38, index), false);
   await multihop((index: number) => getWeightedPool(vault, tokens, 38, index), true);
```
jubeira, line 40: Do we want to keep the benchmark at 38?

jubeira (commented):
There's one stale comment about max tokens being 38 in `AddRemove` test; other than that LGTM.

pvt/benchmarks/multihop.ts
EndymionJkb: It does 50 right after that.

EndymionJkb: Good catch on the comment! Will fix that (and the calculation).

jubeira (approved):
LGTM!

nventuro (approved):

PR #1939, opened 2022-10-24
Add Beefy Linear Pool Support including tests and utils

MirthFutures:
Beefy is on 17 blockchains including Arbitrum, Optimism, and Fantom. Specifically linear pools that can be immediately used are ones utilizing mooAaveETH (OP), mooAaveWBTC (OP), mooAaveDAI (OP), mooAaveUSDC (OP), mooOptimismBIFI, mooArbitrumBIFI and mooFantomBIFI. The BeefyVaultV6 standard is the same across all vaults. This utilizes the manual calculation of balance() (balance of want in the strategy) divided by totalSupply (of mooTokens) instead of ppfs for getWrappedTokenRate(). The tests have been altered to align with this. 

Files added:
BeefyLinearPool
BeefyLinearPoolFactory
BeefyLinearPoolRebalancer
BeefyWrapping (BatchRelayer extension)

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203321720209162


danielmkm:
Hey everyone! We're super excited about leveraging beefy vaults for boosted pools on optimism. The beefy vault is very similar to the reaper vault. The main difference is that the beefy vault uses `want` instead of `token` as the identifier for the underlying token. This distinction makes it challenging to create a unified generic pool, so we opted for a second linear pool implementation to avoid any issues with trying to create a generic implementation that would sacrifice some useful safety checks internally.

Instances where `token()` (reaper) was changed to `want()`:
https://github.com/balancer-labs/balancer-v2-monorepo/blob/aaf035eabcb2e236492a2045f3bf861b39408ac7/pkg/pool-linear/contracts/beefy/BeefyLinearPool.sol#L59

https://github.com/balancer-labs/balancer-v2-monorepo/blob/aaf035eabcb2e236492a2045f3bf861b39408ac7/pkg/standalone-utils/contracts/relayer/BeefyWrapping.sol#L50

https://github.com/balancer-labs/balancer-v2-monorepo/blob/aaf035eabcb2e236492a2045f3bf861b39408ac7/pkg/standalone-utils/contracts/relayer/BeefyWrapping.sol#L77

As @MirthFutures noted above, the getWrappedTokenRate is using the totalSupply and balance instead of the ppfs as it is less prone to rounding errors.

https://github.com/balancer-labs/balancer-v2-monorepo/blob/aaf035eabcb2e236492a2045f3bf861b39408ac7/pkg/pool-linear/contracts/beefy/BeefyLinearPool.sol#L89

https://github.com/balancer-labs/balancer-v2-monorepo/blob/aaf035eabcb2e236492a2045f3bf861b39408ac7/pkg/pool-linear/contracts/beefy/BeefyLinearPoolRebalancer.sol#L58

pkg/pool-linear/contracts/beefy/BeefyLinearPool.sol
```diff
@@ -0,0 +1,91 @@
...
+
+        return assetManagers;
+    }
+
+    function _getWrappedTokenRate() internal view override returns (uint256) {
+        uint256 vaultTotalSupply = _tokenVault.totalSupply();
+        return
+            vaultTotalSupply == 0
+                ? _balanceScaleFactor
+                : (_tokenVault.balance() * _balanceScaleFactor) / vaultTotalSupply;
```
nventuro, line 89: This seems a bit odd, under which conditions will the supply be zero?

I can't quite comment on the correctness of the rest of this function, as that requires reading the Beefy code thoroughly. In particular, I'd be concerned with the possibility of reentrancy attacks manipulating the values of `totalSupply()` and `balance()` so that they're potentially out of sync.

MirthFutures: Ya, it should never be 0 unless everyone has left the vault. 

Not sure where the reentrancy would occur? We have protection against it on our deposits. PPFS is calculated in the same way though it ends up rounding. So using this method would have better precision.

nventuro: Not sure where reentrancy might happen as that'd depend on the details of the `tokenVault`, but any action that results in callbacks after either supply or balance is updated but before the other is also updated would result in issues. If you're convinced that can't happen, then things should be fine.

MirthFutures: Here is a link to the V6 code which would be the code for the live vaults. 

https://github.com/beefyfinance/beefy-contracts/blob/master/contracts/BIFI/vaults/BeefyVaultV6.sol

Balance refers to both the strategy balance and anything that remains in the vault unused. The deposit function uses a nonreentrant modifier, the only callback would have to be from the transfer of the want. The shares are minted at the end of the function, after the transfer which is what increases the total supply. 

The withdraw function burns and reduces shares first before transferring any want back to the user.

mkflow27:
Hey @MirthFutures,

future LinearPool integrations will live in a separate Repo for the Balancer protocol. This repo will be maintained by Orb collective - a Balancer Serivce Provider (https://github.com/orbcollective). We are in the process of setting this repo up. I wanted to already reach out and inform you about it. Once it is set up, I will reach out with more details

EndymionJkb:
Overtaken by events - moved to Orb repo

PR #1944, opened 2022-10-25
Update deployments readme

jubeira:
I've tried the API and it works for deprecated contracts by appending `deprecated/` to the task ID

### DIFF
--- a/pkg/deployments/README.md
+++ b/pkg/deployments/README.md
@@ -3,9 +3,9 @@
 # Balancer V2 Deployments
 
 [![NPM Package](https://img.shields.io/npm/v/@balancer-labs/v2-deployments.svg)](https://www.npmjs.org/package/@balancer-labs/v2-deployments)
-[![GitHub Repository](https://img.shields.io/badge/github-deployments-lightgrey?logo=github)](https://github.com/balancer-labs/balancer-v2-monorepo/tree/deployments-latest/pkg/deployments)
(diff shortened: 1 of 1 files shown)


pkg/deployments/README.md
```diff
@@ -25,7 +25,9 @@ $ npm install @balancer-labs/v2-deployments
 
 ### Usage
 
-Import `@balancer-labs/v2-deployments` to access the different ABIs and deployed addresses. To see all Task IDs and their associated contracts, head to [Past Deployments](#past-deployments).
+Import `@balancer-labs/v2-deployments` to access the different ABIs and deployed addresses. To see all current Task IDs and their associated contracts, head to [Active Deployments](#active-deployments).
+
+Past deployments that are currently not in use or have been superseded can be accessed in the [Deprecated Deployments](#deprecated-deployments) section. Use `deprecated/` as prefix when referring to a deprecated task ID.
```
nventuro, line 30:
```suggestion
Past deployments that are currently not in use or have been superseded can be accessed in the [Deprecated Deployments](#deprecated-deployments) section. Use `deprecated/` as prefix when referring to a deprecated task ID.

> ⚠️ Exercise care when interacting with deprecated deployments: there's often a very good reason why they're no longer active.
>
> You can find information on why each deployment has been deprecated in their corresponding readme file.
```

nventuro (approved):

PR #1947, opened 2022-10-25
Make interface pragmas more permissive

nventuro:
We don't care about this, but it will be very useful for third parties.


EndymionJkb (approved):
Presumably, since it's only interfaces, it doesn't matter if the contracts themselves compile with 0.8 or not then.

PR #1950, opened 2022-10-26
Check action ID uniqueness when creating new ones

### LINKED ISSUE #1502: Make save-action-ids also check them
This will highlight repeated ids and force the user to handle them.

jubeira:
If my understanding is correct, we don't need to check action IDs that we've just generated; we should only check for uniqueness after creation.

Note for the reviewer: hide whitespaces when checking diff.

Closes #1502.

### DIFF
--- a/pkg/deployments/hardhat.config.ts
+++ b/pkg/deployments/hardhat.config.ts
@@ -113,63 +113,71 @@ task('check-artifacts', `check that contract artifacts correspond to their build
     }
   });
 
-task('save-action-ids', `Print the action IDs for a particular contract`)
+task('save-action-ids', `Print the action IDs for a particular contract and checks their uniqueness`)
   .addOptionalParam('id', 'Specific task ID')
   .addOptionalParam('name', 'Contract name')
   .addOptionalParam('address', 'Address of Pool created from a factory')
   .setAction(
     async (args: { id: string; name: string; address?: string; verbose?: boolean }, hre: HardhatRuntimeEnvironment) => {
-      Logger.setDefaults(false, args.verbose || false);
+      async function saveActionIdsTask(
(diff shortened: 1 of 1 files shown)


pkg/deployments/hardhat.config.ts
```diff
@@ -113,63 +113,71 @@ task('check-artifacts', `check that contract artifacts correspond to their build
...
-        await generateActionIdsForTask(taskID);
+        // We're calculating action IDs for whichever contracts we can pull enough information from disk for.
+        // This will calculate action IDs for any contracts which are a named output from a task.
+        for (const taskID of Task.getAllTaskIds()) {
+          await generateActionIdsForTask(taskID);
+        }
       }
+
+      await saveActionIdsTask(args, hre);
+      checkActionIdUniqueness(hre.network.name);
```
nventuro, line 180: Is this because of the early returns?

nventuro (commented):
Looks simple enough! Have you tried this to make sure it works?

pkg/deployments/hardhat.config.ts
jubeira: yes, it is.

I considered creating subtasks, but it seemed an overkill. I think this is a better way to put it: first do the original task (whatever the execution path is), then check uniqueness.

jubeira:
I didn't force a collision, but it seems to work:

```
juan@M16:~/prj/bal/pkg/deployments$ hh save-action-ids --id 20221021-managed-pool --network goerli 
  Generating action IDs for ManagedPoolAddRemoveTokenLib of 20221021-managed-pool
  Generating action IDs for CircuitBreakerLib of 20221021-managed-pool
  Generating action IDs for ManagedPoolFactory of 20221021-managed-pool
✅  Verified that no contracts unexpectedly share action IDs
```

PR #1951, opened 2022-10-26
Provide a 'contract' argument to extract-artifacts 

### LINKED ISSUE #1504: Provide a 'contract' argument to extract-artifacts
Some deployments, like the batch relayers, deploy more than one contract from the same build info. We should fetch their ABI.

Similarly, factory deployments should extract the ABI of whatever it is they deploy.

jubeira:
# Description

This PR adds two optional parameters to `extract-artifacts`: contract name and file name.

This is intended to support 3 use cases:
- Basic case: we want to generate artifacts for a single contract whose name matches the file name. In this case, the usual command without extra options will work: `hh extract-artifacts --id <task-id>`.
- More than one contract to extract, single `build-info` bundle. In this case, we currently need create a new copy of the same build info file for each of the target contracts and rename them to match contract names. With this change, we can keep a single build info file, and extract many contracts by running `hh extract-artifacts --id <task-id> --name <contract-name>` repeatedly, changing the contract name.
- More than one contract to extract, multiple build info files with different settings. When changing compiler settings (e.g. optimization runs), more than one build info will be generated, but the same contract will be present in all of them. With the `file` option we can tell `extract-artifacts` to look only in a specific file. For example, considering the recent managed pool deployment:

```
hh extract-artifacts --id managed-pool --name ExternalWeightedMath --file ManagedPool9999Runs
hh extract-artifacts --id managed-pool --name ManagedPoolFactory --file ManagedPool200Runs
```

Finally, if no file nor name is specified, the task will generate artifacts for all the build info files, using their file names as contracts names as it was before this change.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

- [ ] Tests are included for all code paths ---> use cases tested manually.

## Issue Resolution

Closes #1504


pkg/deployments/package.json
```diff
@@ -25,6 +25,7 @@
     "check-artifacts": "hardhat check-artifacts",
     "check-deployments": "hardhat check-deployments --network mainnet && hardhat check-deployments --network polygon && hardhat check-deployments --network arbitrum && hardhat check-deployments --network optimism && hardhat check-deployments --network goerli",
     "check-action-ids": "hardhat check-action-ids --network mainnet && hardhat check-action-ids --network polygon && hardhat check-action-ids --network arbitrum && hardhat check-action-ids --network optimism",
+    "extract-artifacts": "hardhat extract-artifacts --id",
```
TomAFrench, line 28:
```suggestion
    "extract-artifacts": "hardhat extract-artifacts",
```

I don't like how we're making `id` into a positional parameter here. If we want it to be positional then we should modify the task but I would leave it as it is.

pkg/deployments/src/artifact.ts
```diff
@@ -48,11 +63,16 @@ export function checkArtifact(task: Task): void {
...
+  const buildInfo = task.buildInfo(fileName);
 
   // Read ABI and bytecode from build-info file.
-  const contractSourceName = findContractSourceName(buildInfo, contractName);
+  let contractSourceName: string;
+  try {
+    contractSourceName = findContractSourceName(buildInfo, contractName);
+  } catch (e) {
+    throw new Error(`Could not find a unique source for ${contractName} in the file ${fileName}`);
+  }
```
TomAFrench, line 75: We already raise similar errors inside `findContractSourceName` and this message would likely end up being confusing in the case where the contract isn't found (which is the more likely error) as it implies that there's more than one.

pkg/deployments/DEPLOYING.md
```diff
@@ -22,11 +22,32 @@ As part of our development process, we frequently alter the source code of contr
...
+The build information should be generated on a clean commit, so that it can be verified by other parties. On the project directory, delete the `artifacts` directory and run `yarn hardhat compile` for a clean compilation. The file will then be located at `artifacts/build-info`, and it'll contain the data for all the contracts in the workspace.
+
+### Generating ABI and bytecode files
+
+To generate the artifacts from build info files:
+- Copy the generated build info `json` inside `<repo-root>/pkg/<workspace>/artifacts/build-info` to the new `<repo-root>/pkg/deployments/tasks/<task-id>/build-info` directory.
+- From the `deployments` workspace, run `extract-artifacts` task using the task ID as parameter. Optionally, the contract name and the target file can be specified if e.g. the contract name doesn't match the file name, or if the contract is present in more than one file with different compilation settings. For example:
+
+```
+yarn extract-artifacts 20221021-managed-pool --name ManagedPool --file ManagedPoolFactory
```
TomAFrench, line 47:
```suggestion
yarn extract-artifacts --id 20221021-managed-pool --name ManagedPool --file ManagedPoolFactory
```

pkg/deployments/src/artifact.ts
jubeira: Thanks; I was a bit in doubt about this change too.

jubeira:
Thanks for the review Tom! Comments addressed.

TomAFrench (approved):

Issue #1953, opened 2022-10-27
Propagate simplified BasePool


EndymionJkb:
We now have two BasePools - the original in pool-utils, and a special ManagedPool one in pool-weighted/managed/vendor. We'd like to propagate the simplification and design choices from Managed Pool to the rest of the codebase.

PR #1954, opened 2022-10-27
Great Pool Migration (prep)

EndymionJkb:
# Description

In preparation for propagating the leaner, meaner BasePool of ManagedPool fame to the rest of the code base, move it from the /vendor directory inside pool-weighted, and rename the old one LegacyBasePool (similar to how we did it when switching from token to BPT protocol fees).

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [X] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

- [~] The diff is legible and has no extraneous changes
          (It got a bit confused in some places. And there is one extra line, to silence lint errors.)
- [X] Tests are included for all code paths (no test changes)

## Issue Resolution

Advances #1953


nventuro (commented):
The 'legacy' approach unfortunately makes this very hard to review, since git treats this as LegacyBasePool being a spinoff of the new BasePool, with the old stuff added back, and the new BasePool being a spinoff of the old one, with the old stuff removed.

I suggest we don't rename the new one to BasePool yet, and only do that once the legacy one has been deleted from the repo. Instead of calling the old one legacy and new one base, we could keep the old one and call the new one NewBasePool. Alternatively, we could rename both - but I'm not sure this will fix the diff issues.

EndymionJkb:
> The 'legacy' approach unfortunately makes this very hard to review, since git treats this as LegacyBasePool being a spinoff of the new BasePool, with the old stuff added back, and the new BasePool being a spinoff of the old one, with the old stuff removed.
> 
> I suggest we don't rename the new one to BasePool yet, and only do that once the legacy one has been deleted from the repo. Instead of calling the old one legacy and new one base, we could keep the old one and call the new one NewBasePool. Alternatively, we could rename both - but I'm not sure this will fix the diff issues.

I tried it a couple ways; keeping the old one leads to a smaller diff at least.

nventuro (approved):

PR #1955, opened 2022-10-28
Rebase LinearPool

EndymionJkb:
# Description

With the infrastructure in place (NewBasePool in pool-utils, used by ManagedPool), make LinearPool also use NewBasePool.

It required minor changes to internal functions in NewBasePool: initialization required another argument, and we need to override `_getMinimumBpt`. I put in a check to not mint 0 tokens (and emit an event with 0); could remove if we don't mind that. (Not an optimization concern, since it's only called once; I mainly thought the event might be confusing.)

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

- [X] Tests are included for all code paths - same tests

## Issue Resolution

Advances #1953


pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -540,14 +515,17 @@ abstract contract LinearPool is ILinearPool, IGeneralPool, IRateProvider, BasePo
         }
     }
 
-    function _scalingFactors() internal view virtual override returns (uint256[] memory) {
+    /**
+     * @notice Return the scaling factors for all tokens, including the BPT.
+     */
+    function getScalingFactors() public view virtual override returns (uint256[] memory) {
```
nventuro, line 521: We can leave this as-is for now, but moving forward the scaling factors should not include BPT. We can create an issue for this and address it after merging here, to make it so this PR simply moves from one base pool to another.

pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -621,39 +618,36 @@ abstract contract LinearPool is ILinearPool, IGeneralPool, IRateProvider, BasePo
         _require(upperTarget % _TARGET_SCALING == 0, Errors.FRACTIONAL_TARGET);
         _require(lowerTarget % _TARGET_SCALING == 0, Errors.FRACTIONAL_TARGET);
 
-        _setMiscData(
+        _poolState =
             WordCodec.encodeUint(lowerTarget / _TARGET_SCALING, _LOWER_TARGET_OFFSET, _TARGET_BITS) |
-                WordCodec.encodeUint(upperTarget / _TARGET_SCALING, _UPPER_TARGET_OFFSET, _TARGET_BITS)
-        );
+            WordCodec.encodeUint(upperTarget / _TARGET_SCALING, _UPPER_TARGET_OFFSET, _TARGET_BITS);
```
nventuro, line 623: This is wrong - you're overriding the state entirely by only writing the new targets. You should instead use `insert` (which is what `setMiscData()` does).

pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -664,21 +658,35 @@ abstract contract LinearPool is ILinearPool, IGeneralPool, IRateProvider, BasePo
...
 
-    function _isOwnerOnlyAction(bytes32 actionId) internal view virtual override returns (bool) {
-        return actionId == getActionId(this.setTargets.selector) || super._isOwnerOnlyAction(actionId);
+    function _getMinSwapFeePercentage() internal pure virtual returns (uint256) {
+        return _MIN_SWAP_FEE_PERCENTAGE;
+    }
+
+    function _getMaxSwapFeePercentage() internal pure virtual returns (uint256) {
+        return _MAX_SWAP_FEE_PERCENTAGE;
     }
```
nventuro, line 686: I'd just inline these constants and remove this functions

pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -512,21 +493,15 @@ abstract contract LinearPool is ILinearPool, IGeneralPool, IRateProvider, BasePo
...
 
-    function _getMaxTokens() internal pure override returns (uint256) {
-        return _TOTAL_TOKENS;
-    }
-
     function _getMinimumBpt() internal pure override returns (uint256) {
         // Linear Pools don't lock any BPT, as the total supply will already be forever non-zero due to the preminting
         // mechanism, ensuring initialization only occurs once.
         return 0;
     }
```
nventuro, line 500: I know this was part of the original design, but would consider opening an issue to remove this. It feels like an unnecessary optimization that adds complexity.

pkg/pool-utils/contracts/NewBasePool.sol
```diff
@@ -214,13 +214,17 @@ abstract contract NewBasePool is
...
 
             // On initialization, we lock _getMinimumBpt() by minting it for the zero address. This BPT acts as a
             // minimum as it will never be burned, which reduces potential issues with rounding, and also prevents the
             // Pool from ever being fully drained.
+            // Some pool types do not require this mechanism, and the minimum BPT might be zero.
             _require(bptAmountOut >= _getMinimumBpt(), Errors.MINIMUM_BPT);
-            _mintPoolTokens(address(0), _getMinimumBpt());
+            if (_getMinimumBpt() > 0) {
+                _mintPoolTokens(address(0), _getMinimumBpt());
+            }
```
nventuro, line 226: I'd remove this since we're considering making the minimum be non-zero. This is another point of extra complexity.

nventuro (commented):
I *love* that this PR touches no tests.

pkg/pool-linear/contracts/LinearPool.sol
EndymionJkb: I went back and forth on that.

nventuro: I doubt that will be required, but if so we can always add it later on.

pkg/pool-linear/contracts/LinearPool.sol
nventuro: Created https://github.com/balancer-labs/balancer-v2-monorepo/issues/1960

pkg/pool-linear/contracts/LinearPool.sol
nventuro: Created https://github.com/balancer-labs/balancer-v2-monorepo/issues/1958

nventuro (approved):

PR #1959, opened 2022-10-28
Start using yarn versioning

nventuro:
# Description

This fixes the old versioning setup, and marks the `interfaces` package as being pending a minor release.

From now on, any PR that introduces changes to a component in a package should be accompanied with a run of the command `yarn version <major/minor/patch> --deferred` in the corresponding package. This will create an entry in the `.yarn/versions` directory (which must be committed to the repository), marking the packages that are due a version bump. Once a new release is ready, we can then run `yarn version apply`, which will update all of the package.json version entries on all related packages.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other


EndymionJkb (commented):
I tested it, and it works. The only weird thing is it also adds "undecided" entries, which I don't entirely understand (and it doesn't seem to be documented well).

Also, I get a long page of warnings whenever I run yarn, which I've always ignored because it seems to work, but maybe my system has some peculiarities/issues?

Finally, we should have a document describing when to bump each kind of version. Major versions are pretty clear: a new production release/product launch. But is every PR a patch? Is a minor version anything that changes an external interface?

This is a lot of work if literally every change to any file in a package is a patch. If you make a change in pool-weighted that also touches pool-utils and pvt/benchmarks, you'd have to run `yarn version patch --deferred` three times, and the reviewer would need to check that it was done, and then run `yarn version apply` three times in a final commit before merging? I have some doubt that we would be able to keep up with this.

It would be less work if we didn't worry about patches, and defined minor versions as interface changes. Then it would be fairly infrequent.

```
undecided:
  - "@balancer-labs/v2-benchmarks"
```

nventuro:
> If you make a change in pool-weighted that also touches pool-utils and pvt/benchmarks, you'd have to run yarn version patch --deferred three times, and the reviewer would need to check that it was done, and then run yarn version apply three times in a final commit before merging?

No, if there's a patch change to pool-weighted then that's it's just that. There'd be a deferred patch and a changelog entry.

We would only run `apply` before publishing to npm, which is when bumping matters. That's the whole point - we're deferring bumping until we publish something, and we know whether we need to do a major, minor or patch release (for each package) depending on the deferred status.

EndymionJkb:
> We would only run `apply` before publishing to npm, which is when bumping matters. That's the whole point - we're deferring bumping until we publish something, and we know whether we need to do a major, minor or patch release (for each package) depending on the deferred status.

Ok; I was wondering what would happen if we didn't publish to NPM for a while, and had multiple changes. Apparently, it just takes the last one. So we'd have to be careful not to do a minor and then a patch after it, without updating in between.

jubeira:
Please dismiss the force pushes; the diff should be unchanged.

nventuro:
This seems to not do what we wanted :(

PR #1961, opened 2022-10-31
Rebase LBP

EndymionJkb:
# Description

We're likely going to change this in other ways, but rebase it as a start.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Advances #1953


pkg/pool-weighted/contracts/lbp/LiquidityBootstrappingPool.sol
```diff
@@ -61,21 +73,27 @@ contract LiquidityBootstrappingPool is BaseWeightedPool, ReentrancyGuard {
...
     bytes32 private _poolState;
 
     // Offsets for data elements in _poolState
     uint256 private constant _SWAP_ENABLED_OFFSET = 0;
-    uint256 private constant _START_WEIGHT_OFFSET = 4;
-    uint256 private constant _END_WEIGHT_OFFSET = 128;
-    uint256 private constant _START_TIME_OFFSET = 192;
-    uint256 private constant _END_TIME_OFFSET = 224;
+    uint256 private constant _RECOVERY_MODE_BIT_OFFSET = 1;
+    uint256 private constant _START_WEIGHT_OFFSET = _RECOVERY_MODE_BIT_OFFSET + 2;
```
TomAFrench, line 85:
```suggestion
    uint256 private constant _START_WEIGHT_OFFSET = _RECOVERY_MODE_BIT_OFFSET + 3;
```

We haven't moved the start weights so they should still start 4 bits in.

pkg/pool-weighted/contracts/lbp/LiquidityBootstrappingPool.sol
```diff
@@ -387,4 +733,43 @@ contract LiquidityBootstrappingPool is BaseWeightedPool, ReentrancyGuard {
...
+    }
+
+    /**
+     * @dev Sets the recoveryMode state, and emits the corresponding event.
+     */
+    function _setRecoveryMode(bool enabled) internal virtual override {
+        _poolState = _poolState.insertBool(enabled, _RECOVERY_MODE_BIT_OFFSET);
+
+        emit RecoveryModeStateChanged(enabled);
+    }
```
TomAFrench, line 753:
```suggestion
    /**
     * @dev Sets the recoveryMode state.
     */
    function _setRecoveryMode(bool enabled) internal virtual override {
        _poolState = _poolState.insertBool(enabled, _RECOVERY_MODE_BIT_OFFSET);
    }
```

`RecoveryModeStateChanged` is emitted by `RecoveryMode` now so we don't need to emit it here.

TomAFrench (commented):
Agreed that there's quite a few things we'll want to change to match the new architecture. I've found a couple of correctness issues but other than those we can merge and migrate stuff over time.

pkg/pool-weighted/contracts/lbp/LiquidityBootstrappingPool.sol
EndymionJkb: Ha! Good catch, though it would still work (overwriting one of the unused bits)

pkg/pool-weighted/contracts/lbp/LiquidityBootstrappingPool.sol
```diff
@@ -240,79 +278,389 @@ contract LiquidityBootstrappingPool is BaseWeightedPool, ReentrancyGuard {
...
 
     // Swap overrides - revert unless swaps are enabled
 
+    function _onSwapGeneral(
+        SwapRequest memory,
+        uint256[] memory,
+        uint256,
+        uint256
+    ) internal virtual override returns (uint256) {
+        _revert(Errors.UNHANDLED_BY_LBP);
```
TomAFrench, line 553:
```suggestion
        _revert(Errors.UNIMPLEMENTED);
```

I don't think we need to add a new error type for this.

EndymionJkb: It was patterned off Linear Pool and others, that used "UNHANDLED_BY" for things like joins and exits - and unsupported swap types. Should we replace all the "UNHANDLED" ones with UNIMPLEMENTED then?

TomAFrench: Are they actually used? I can't find any usage of `UNHANDLED_BY_PHANTOM_POOL`, `UNHANDLED_BY_MANAGED_POOL`.

I'm not a fan of these message generally as they're all reporting the same underlying issue, "You're asking me to perform an action for which I don't have code to handle".

EndymionJkb: They *were* used, but it looks like now they're only in Linear. Ok; I'm not a big fan either - especially having multiple "versions" of these kinds of messages. UNHANDLED also has the sense of "I should be able to do this, but can't" - while UNIMPLEMENTED seems more intentional.

I'll change this one then, and make a new PR to change Linear - and we'll not introduce any further UNHANDLED ones.

TomAFrench (approved):

PR #1968, opened 2022-11-01
Update hardhat verifier

nventuro:
We need this new version to be able to deploy to gnosis and bsc.


jubeira (approved):
I can confirm I could deploy / verify a test contract in Goerli with these changes.

PR #1969, opened 2022-11-01
Add Gnosis and BNB deployments

nventuro:
The BNB deployment should be complete for now, Gnosis is lacking the core pool factories as the RPC errors out when I try to deploy those. I get the same error reported [here](https://github.com/NethermindEth/nethermind/issues/4276), so that might be related.


TomAFrench (approved):
CI's happy so I'm happy.

PR #1601, opened 2022-08-11
Deployment preparation: L2 gauge checkpointer

jubeira:
Adding fork tests and deployment artifacts / scripts for L2 gauge checkpointer.

Fork test block # was set close to the time where the fork test was last updated.

Build info from https://github.com/balancer-labs/balancer-v2-monorepo/pull/1601/commits/db9ab19541945ed8419cdf2f716b22e8b5f31bd8.

Note that at https://github.com/balancer-labs/balancer-v2-monorepo/pull/1601/commits/1895091f1dab5d4b67f36d72fa44edc05309b92a the updates from master already have the changes from the entrypoint, so the build info generated at that point will be different and incompatible with the fork test.

PR #1974, opened 2022-11-01
Upgrade hardhat

nventuro:
# Description

This latest release fixes the issue we found where the build infos were not always the same while at the same commit. We should merge this and rebuild outstanding compilation artifacts (e.g. #1601) using this new version.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other

### DIFF
--- a/pkg/deployments/package.json
+++ b/pkg/deployments/package.json
@@ -45,7 +45,7 @@
     "eslint": "^7.9.0",
     "eslint-plugin-prettier": "^3.1.4",
     "ethers": "^5.4.1",
-    "hardhat": "^2.11.1",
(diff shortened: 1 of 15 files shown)


EndymionJkb (approved):
I built with this; seems to work.

EndymionJkb:
Is there any interaction with #1968?

TomAFrench (approved):

TomAFrench:
> Is there any interaction with #1968?

There shouldn't be but I think we can just merge this and recheck #1968 after merging master into it.

PR #1979, opened 2022-11-03
Split LBP contract into LBP and LBPSettings

TomAFrench:
# Description

This PR ports the architecture changes we made to ManagedPool over to LBPs.

1. Swap, join, exit hooks are moved to their own contract.
2. We make use of `WeightedJoinsLib` and `WeightedExitsLib`.
3. Some inlining of functions which don't make sense in a flat contract.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/pool-weighted/contracts/lbp/LiquidityBootstrappingPool.sol
```diff
@@ -123,166 +61,96 @@ contract LiquidityBootstrappingPool is IMinimalSwapInfoPool, NewBasePool {
...
         }
     }
 
     /**
-     * @notice Pause/unpause trading.
+     * @dev Adds swap fee amount to `amount`, returning a higher value.
      */
-    function setSwapEnabled(bool swapEnabled) external authenticate whenNotPaused {
-        _setSwapEnabled(swapEnabled);
+    function _addSwapFeeAmount(uint256 amount) internal view returns (uint256) {
```
EndymionJkb, line 140: Seems we could also inline _addSwapFeeAmount and _subtractSwapFeeAmount, since they're only used here now.

pkg/pool-weighted/contracts/lbp/LiquidityBootstrappingPool.sol
```diff
@@ -123,166 +61,96 @@ contract LiquidityBootstrappingPool is IMinimalSwapInfoPool, NewBasePool {
...
-        return normalizedWeights;
+    function _subtractSwapFeeAmount(uint256 amount) internal view returns (uint256) {
+        // This returns amount - fee amount, so we round up (favoring a higher fee amount).
+        uint256 feeAmount = amount.mulUp(getSwapFeePercentage());
+        return amount.sub(feeAmount);
     }
 
-    // Pool callback functions
-
     // Prevent any account other than the owner from joining the pool
```
EndymionJkb, line 154:
```suggestion
```
This comment was here before, but it's describing an implementation detail of the function, vs what the function does - and the same comment appears later, where it should be - so I don't think we need this here.

EndymionJkb: Also, I can't reach this in the diff, but I would also reword this:

```
// Set the initial BPT to the value of the invariant times the number of tokens. This makes BPT supply more
// consistent in Pools with similar compositions but different number of tokens.
```

To:
```
// Set the initial BPT to the value of the invariant times the number of tokens. This makes the BPT supply
// more consistent in Pools with similar token composition, but a different number of tokens.
```

pkg/pool-weighted/contracts/lbp/LiquidityBootstrappingPoolSettings.sol
```diff
@@ -0,0 +1,397 @@
...
+
+import "../lib/GradualValueChange.sol";
+import "../lib/ValueCompression.sol";
+import "../WeightedMath.sol";
+
+/**
+ * @dev Weighted Pool with mutable weights, designed to support V2 Liquidity Bootstrapping.
+ */
+abstract contract LiquidityBootstrappingPoolSettings is IMinimalSwapInfoPool, NewBasePool {
+    // LiquidityBootstrappingPool change their weights over time: these periods are expected to be long enough (e.g.
```
EndymionJkb, line 34:
```suggestion
    // LiquidityBootstrappingPools change their weights over time: these periods are expected to be long enough (e.g.
```

EndymionJkb (approved):
Looks much cleaner now.

pkg/pool-weighted/contracts/lbp/LiquidityBootstrappingPool.sol
TomAFrench: Yep, we also need to update the swap/join/exit hooks to pass an explicit `poolState` parameter around which should be handled at the same time imo. My thinking is that this falls out of scope for this PR and we can clean this up later.

PR #1981, opened 2022-11-03
Helpers to deploy with authorizer adaptor entrypoint

EndymionJkb:
# Description

Do a little dance when deploying the Vault in tests so that we end up with an authorizer that routes all adaptor calls through the new adaptor entrypoint. This should be merged into #1977, the PR that introduced the adaptor entrypoint.

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/liquidity-mining/test/AuthorizerAdaptorEntrypoint.test.ts
+++ b/pkg/liquidity-mining/test/AuthorizerAdaptorEntrypoint.test.ts
@@ -6,7 +6,6 @@ import { expect } from 'chai';
 import { defaultAbiCoder } from 'ethers/lib/utils';
 import { sharedBeforeEach } from '@balancer-labs/v2-common/sharedBeforeEach';
 import { ANY_ADDRESS } from '@balancer-labs/v2-helpers/src/constants';
-import { deploy } from '@balancer-labs/v2-helpers/src/contract';
 import { actionId } from '@balancer-labs/v2-helpers/src/models/misc/actions';
 import Vault from '@balancer-labs/v2-helpers/src/models/vault/Vault';
 
@@ -22,11 +21,12 @@ describe('AuthorizerAdaptorEntrypoint', () => {
   });
 
   sharedBeforeEach('deploy vault with entrypoint', async () => {
-    ({ instance: vault, authorizer, authorizerAdaptor: adaptor } = await Vault.create({ admin }));
-
-    // TODO(@jubeira): initialize entrypoint and adaptor inside helpers.
-    entrypoint = await deploy('AuthorizerAdaptorEntrypoint', { args: [adaptor.address] });
-    await authorizer.setAdaptorEntrypoint(entrypoint.address);
+    ({
+      instance: vault,
+      authorizer,
+      authorizerAdaptor: adaptor,
+      authorizerAdaptorEntrypoint: entrypoint,
+    } = await Vault.create({ admin }));
   });
 
   describe('constructor', () => {
@@ -55,7 +55,7 @@ describe('AuthorizerAdaptorEntrypoint', () => {
     let expectedResult: string;
 
     sharedBeforeEach('prepare action', async () => {
-      action = await actionId(adaptor, 'getProtocolFeesCollector', vault.interface);
+      action = await actionId(entrypoint, 'getProtocolFeesCollector', vault.interface);
 
       target = vault.address;
       calldata = vault.interface.encodeFunctionData('getProtocolFeesCollector');
@@ -96,6 +96,10 @@ describe('AuthorizerAdaptorEntrypoint', () => {
           'SENDER_NOT_ALLOWED'
         );
       });
+    });
+
+    it('rejects direct calls from the adaptor', async () => {
+      await expect(adaptor.connect(other).performAction(target, calldata)).to.be.revertedWith('SENDER_NOT_ALLOWED');
     });
 
     context('when caller is not authorized', () => {
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -107,9 +107,8 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
 
     TimelockExecutor private immutable _executor;
     IAuthentication private immutable _vault;
-    // TODO(@jubeira): make adaptor and entrypoint immutable and set in the constructor.
-    IAuthorizerAdaptorEntrypoint private _authorizerAdaptorEntrypoint;
-    IAuthorizerAdaptor private _authorizerAdaptor;
+    IAuthorizerAdaptorEntrypoint private immutable _authorizerAdaptorEntrypoint;
+    IAuthorizerAdaptor private immutable _authorizerAdaptor;
     uint256 private immutable _rootTransferDelay;
 
     address private _root;
@@ -166,10 +165,13 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
     constructor(
         address admin,
         IAuthentication vault,
+        IAuthorizerAdaptorEntrypoint authorizerAdaptorEntrypoint,
         uint256 rootTransferDelay
     ) {
         _setRoot(admin);
         _vault = vault;
+        _authorizerAdaptorEntrypoint = authorizerAdaptorEntrypoint;
+        _authorizerAdaptor = authorizerAdaptorEntrypoint.getAuthorizerAdaptor();
         _executor = new TimelockExecutor();
         _rootTransferDelay = rootTransferDelay;
 
@@ -190,12 +192,6 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
         SCHEDULE_DELAY_ACTION_ID = getActionId(TimelockAuthorizer.scheduleDelayChange.selector);
         _GENERAL_GRANT_ACTION_ID = generalGrantActionId;
         _GENERAL_REVOKE_ACTION_ID = generalRevokeActionId;
-    }
-
-    // TODO(@jubeira): remove this; send entrypoint in the constructor.
-    function setAdaptorEntrypoint(IAuthorizerAdaptorEntrypoint authorizerAdaptorEntrypoint) external {
-        _authorizerAdaptorEntrypoint = authorizerAdaptorEntrypoint;
-        _authorizerAdaptor = authorizerAdaptorEntrypoint.getAuthorizerAdaptor();
     }
 
     /**
--- a/pvt/helpers/src/models/vault/VaultDeployer.ts
+++ b/pvt/helpers/src/models/vault/VaultDeployer.ts
@@ -4,12 +4,13 @@ import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/dist/src/signer-wit
 
 import { deploy } from '../../contract';
 import { MONTH } from '../../time';
-import { ZERO_ADDRESS } from '../../constants';
+import { ANY_ADDRESS, ZERO_ADDRESS } from '../../constants';
 import { RawVaultDeployment, VaultDeployment } from './types';
 
 import Vault from './Vault';
 import TypesConverter from '../types/TypesConverter';
 import TokensDeployer from '../tokens/TokensDeployer';
+import { actionId } from '../misc/actions';
 
 export default {
   async deploy(params: RawVaultDeployment): Promise<Vault> {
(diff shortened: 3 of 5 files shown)


pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -166,10 +165,13 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
     constructor(
         address admin,
         IAuthentication vault,
```
jubeira, line 167: You can get the vault from the entrypoint, and it should always be the same one. Maybe we can remove this constructor argument?

pvt/helpers/src/models/vault/VaultDeployer.ts
```diff
@@ -42,14 +68,30 @@ export default {
...
-  async _deployAuthorizer(admin: SignerWithAddress, from?: SignerWithAddress): Promise<Contract> {
-    return deploy('v2-vault/TimelockAuthorizer', { args: [admin.address, ZERO_ADDRESS, MONTH], from });
+  async _deployAuthorizer(
+    admin: SignerWithAddress,
+    authorizerAdaptorEntrypoint: Contract,
+    from?: SignerWithAddress,
+    vault?: string
+  ): Promise<Contract> {
+    return deploy('v2-vault/TimelockAuthorizer', {
+      args: [admin.address, vault || ZERO_ADDRESS, authorizerAdaptorEntrypoint.address ?? ZERO_ADDRESS, MONTH],
```
jubeira, line 78: the entrypoint gets the adaptor in the `TimelockAuthorizer` constructor; doesn't this revert if we send it the `ZERO_ADDRESS`? Why would the address be undefined if the entrypoint is a `Contract`?

jubeira (commented):
Two minor early comments before I forget.

pvt/helpers/src/models/vault/VaultDeployer.ts
EndymionJkb: Correct - this is leftover from when it was optional. It may change again, but yes, I can remove that.

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
EndymionJkb: I think I can, and it actually makes some downstream stuff (deployer helpers) a bit cleaner. Will try.

pkg/liquidity-mining/contracts/admin/GaugeAdder.sol
```diff
@@ -37,22 +37,24 @@ contract GaugeAdder is IGaugeAdder, SingletonAuthentication, ReentrancyGuard {
...
         _balWethBpt = gaugeController.token();
     }
 
     /**
-     * @notice Returns the address of the Authorizer adaptor contract.
+     * @notice Returns the address of the Authorizer adaptor entrypoint contract.
      */
-    function getAuthorizerAdaptor() external view override returns (IAuthorizerAdaptor) {
-        return _authorizerAdaptor;
+    function getAuthorizerAdaptorEntrypoint() external view override returns (IAuthorizerAdaptorEntrypoint) {
```
jubeira, line 56: I was in doubt yesterday when I did this fix: should we also add a getter for the adaptor?

I think it's fine as it is, since you could also get the adaptor from the entrypoint's getter, but I wanted to double check JIC.

pkg/pool-utils/contracts/test/MockVault.sol
```diff
@@ -204,4 +204,13 @@ contract MockVault is IPoolSwapStructs {
 
         emit PoolBalanceChanged(poolId, msg.sender, tokens, deltas, dueProtocolFeeAmounts);
     }
+
+    // Needed to support authorizer adapter entrypoint
```
jubeira, line 208:
```suggestion
    // Needed to support authorizer adaptor entrypoint
```

pkg/vault/test/VaultAuthorization.test.ts
```diff
@@ -243,10 +245,11 @@ describe('VaultAuthorization', function () {
     const BUFFER_PERIOD_DURATION = MONTH;
 
     sharedBeforeEach(async () => {
-      authorizer = await deploy('TimelockAuthorizer', { args: [admin.address, ZERO_ADDRESS, MONTH] });
-      vault = await deploy('Vault', {
-        args: [authorizer.address, ZERO_ADDRESS, PAUSE_WINDOW_DURATION, BUFFER_PERIOD_DURATION],
```
jubeira, line 248: Question: what was this `ZERO_ADDRESS` that is now gone?

pvt/helpers/src/models/vault/VaultDeployer.ts
```diff
@@ -19,15 +20,40 @@ export default {
     const { from, mocked } = deployment;
     if (!admin) admin = from || (await ethers.getSigners())[0];
 
-    const authorizer = await this._deployAuthorizer(admin, from);
+    // Needed so that the getAuthorizerAdaptor() call in the Authorizer constructor will not revert
+    const mockEntrypoint = await deploy('v2-vault/MockAuthorizerAdaptorEntrypoint');
+
+    // Deploy the Vault with a placeholder authorizer
+    const authorizer = await this._deployAuthorizer(admin, mockEntrypoint, from);
```
jubeira, line 27: While this sequence is fine and it'll work, it's a bit different from what we'll do in production.

Namely, the entrypoint came afterwards, and in the starting point we had an old authorizer that didn't need the entrypoint.

Again, this is not a strong opinion, but we could also consider deploying a mock authorizer first with no entrypoint, and then change it to the actual authorizer with the actual entrypoint. With that approach, the first mock authorizer would be analogous to the current `Authorizer` used by the Vault in mainnet.

Something like this https://github.com/balancer-labs/balancer-v2-monorepo/pull/1977/commits/6d8f3e2805321ca0f362bb07534f7c80bd9ef178#diff-d1f1f09715847682508eabc6c0e80ebd2772af929faeacaabf0e1382e242cc9f.

pkg/vault/test/authorizer/TimelockAuthorizerRoot.test.ts
```diff
@@ -0,0 +1,172 @@
...
+    [, root, grantee] = await ethers.getSigners();
+  });
+
+  const WHERE_1 = ethers.Wallet.createRandom().address;
+  const WHERE_2 = ethers.Wallet.createRandom().address;
+
+  const GENERAL_PERMISSION_SPECIFIER = TimelockAuthorizer.GENERAL_PERMISSION_SPECIFIER;
+  const EVERYWHERE = TimelockAuthorizer.EVERYWHERE;
+
+  describe('root', () => {
```
jubeira, line 25: So if I got this correctly, this is the exact same test that applies to everything related with the authorizer without the adaptor / entrypoint. Is that correct?

jubeira (commented):
Looks pretty good; it's taking shape!
I've left a few minor comments and some questions.

pkg/governance-scripts/contracts/TimelockAuthorizerMigrator.sol
```diff
@@ -53,6 +53,7 @@ contract TimelockAuthorizerMigrator {
      */
     constructor(
         IVault _vault,
```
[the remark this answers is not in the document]
EndymionJkb: I went back and forth on this. While in general we don't want to get things from other things - doing that is why we need to redeploy the Gauge and TokenAdders - the Vault is pretty safe to do that with (and we certainly do it elsewhere). And you could argue it makes it more consistent with the TimelockAuthorizer, which is relevant to the migrator.

I'll give it a try. Mainly it just got too late last night and I didn't want to change anything else :)

pkg/liquidity-mining/contracts/admin/GaugeAdder.sol
EndymionJkb: I also think it's fine as is - I don't think there's a need to have it externally (and could be confusing). Plus as you point out, there's another way to get it if necessary.

pkg/pool-utils/contracts/test/MockVault.sol
EndymionJkb: Having to speak British trips me up. Blimey!

pkg/vault/test/VaultAuthorization.test.ts
EndymionJkb: The Vault

pkg/vault/test/authorizer/TimelockAuthorizerRoot.test.ts
EndymionJkb: Yes - I separated it because it had to initialize in a separate and weird way that interfered with the other tests unless I renamed stuff, which would have made the diff harder to read.

pvt/helpers/src/models/vault/VaultDeployer.ts
EndymionJkb: That's fine, and I like the comments too. If you can run the full suite and everything passes, I'd say you could just push that.

pkg/vault/test/authorizer/TimelockAuthorizerRoot.test.ts
jubeira: :+1: to that

pkg/liquidity-mining/test/ChildChainGaugeTokenAdder.test.ts
```diff
@@ -35,6 +36,7 @@ describe('ChildChainGaugeTokenAdder', () => {
   sharedBeforeEach('deploy token', async () => {
     vault = await Vault.create({ admin });
     adaptor = vault.authorizerAdaptor;
```
TomAFrench, line 38:
```suggestion
    const adaptor = vault.authorizerAdaptor;
```

ditto here as well. We could likely make this change in a number of the other test files.

pkg/governance-scripts/contracts/deprecated/20220704-timelock-authorizer-migrator/README.md
```diff
@@ -0,0 +1,12 @@
+# Timelock Authorizer Migrator
```
TomAFrench, line 1: As this was never executed I think we can just delete this and the deployment task rather than deprecating it.

pkg/liquidity-mining/contracts/admin/ChildChainGaugeTokenAdder.sol
```diff
@@ -32,21 +32,21 @@ contract ChildChainGaugeTokenAdder is SingletonAuthentication {
...
         _gaugeFactory = gaugeFactory;
     }
 
     /**
      * @notice Returns the address of the Authorizer adaptor contract.
      */
     function getAuthorizerAdaptor() external view returns (IAuthorizerAdaptor) {
-        return _authorizerAdaptor;
+        return _authorizerAdaptorEntrypoint;
     }
```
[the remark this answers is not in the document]
TomAFrench: There is also natspec to be updated as well.

pkg/liquidity-mining/test/AuthorizerAdaptorEntrypoint.test.ts
```diff
@@ -151,6 +152,10 @@ describe('AuthorizerAdaptorEntrypoint', () => {
       });
```
TomAFrench, line 152: I'd wrap

```
      it('performs the expected function call', async () => {
        const value = await entrypoint.connect(grantee).callStatic.performAction(target, calldata);
        expect(value).to.be.eq(expectedResult);
      });
      it('sends value to target contract correctly', async () => {
        const value = await entrypoint
          .connect(grantee)
          .callStatic.performAction(payableTarget, payableCalldata, { value: payment });
        expect(value).to.be.eq(payableExpectedResult);
      });
      it('rejects direct calls from the adaptor', async () => {
        // The authorizer will reject calls that are not initiated in the adaptor entrypoint.
        await expect(adaptor.connect(grantee).performAction(target, calldata)).to.be.revertedWith('SENDER_NOT_ALLOWED');
      });
```

into a `itDoesXCorrectly` function to reduce some repetition here.

pkg/liquidity-mining/test/AuthorizerAdaptorEntrypoint.test.ts
```diff
@@ -151,6 +152,10 @@ describe('AuthorizerAdaptorEntrypoint', () => {
...
     });
 
+    it('rejects direct calls from the adaptor', async () => {
```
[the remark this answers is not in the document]
EndymionJkb: No - but it is still wrong, because the test above it changed. It's trying to test that an otherwise correct call - that succeeds when called on the entrypoint - fails when called on the adaptor.

pkg/vault/test/authorizer/TimelockAuthorizerRoot.test.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    sharedBeforeEach('deploy authorizer', async () => {
+      const oldAuthorizer = await TimelockAuthorizer.create({ root });
+
+      vault = await deploy('Vault', { args: [oldAuthorizer.address, ZERO_ADDRESS, 0, 0] });
+      authorizer = await TimelockAuthorizer.create({ root, vault });
+
+      const setAuthorizerAction = await actionId(vault, 'setAuthorizer');
+      await oldAuthorizer.grantPermissions(setAuthorizerAction, root, vault, { from: root });
+      await vault.connect(root).setAuthorizer(authorizer.address);
```
TomAFrench, line 36: See https://github.com/balancer-labs/balancer-v2-monorepo/pull/1977#discussion_r1015654675

pkg/liquidity-mining/contracts/admin/ChildChainGaugeTokenAdder.sol
```diff
@@ -32,25 +32,30 @@ contract ChildChainGaugeTokenAdder is SingletonAuthentication {
...
     {
-        _authorizerAdaptor = authorizerAdaptor;
+        _authorizerAdaptorEntrypoint = authorizerAdaptorEntrypoint;
         _gaugeFactory = gaugeFactory;
     }
 
     /**
-     * @notice Returns the address of the Authorizer adaptor contract.
+     * @notice Returns the address of the Authorizer adaptor entrypoint contract.
+     * @return - address of the entrypoint.
```
TomAFrench, line 47:
```suggestion
```

pkg/liquidity-mining/test/AuthorizerAdaptorEntrypoint.test.ts
```diff
@@ -112,6 +103,15 @@ describe('AuthorizerAdaptorEntrypoint', () => {
         // The authorizer will reject calls that are not initiated in the adaptor entrypoint.
         await expect(adaptor.connect(grantee).performAction(target, calldata)).to.be.revertedWith('SENDER_NOT_ALLOWED');
       });
+    }
+
+    context('when caller is authorized globally', () => {
+      sharedBeforeEach('authorize caller globally', async () => {
+        await authorizer.connect(admin).grantPermissions([action], grantee.address, [ANY_ADDRESS]);
+        await authorizer.connect(admin).grantPermissions([payableAction], grantee.address, [ANY_ADDRESS]);
```
TomAFrench, line 111:
```suggestion
        await authorizer.connect(admin).grantPermissions([action, payableAction], grantee.address, [ANY_ADDRESS]);
```

TomAFrench (approved):
Looks good. There's just a couple of nits left over.

I'd like to rename the `entrypoint` variable to `adaptorEntrypoint in some of the tests as `entrypoint` feels a bit vague. Happy to handle this myself though.

PR #1992, opened 2022-11-08
Swap out `ln` and `exp` for Solmate versions

TomAFrench:
# Description

This swaps out the exp and ln implementations for the solmate version. Draft as fuzzing of `pow` is tricky.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/solidity-utils/test/foundry/LogExpMath.t.sol
+++ b/pkg/solidity-utils/test/foundry/LogExpMath.t.sol
@@ -1,0 +1,28 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.7.0;
+
+import "forge-std/Test.sol";
+
+import "../../contracts/math/LogExpMath.sol";
+import "../../contracts/math/ReferenceLogExpMath.sol";
+
+contract LogExpMathsTest is Test {
+    function testExpEquivalence(int256 a) external {
+        vm.assume(a > ReferenceLogExpMath.MIN_NATURAL_EXPONENT);
+        vm.assume(a < ReferenceLogExpMath.MAX_NATURAL_EXPONENT);
+
+        int256 expectedExp = ReferenceLogExpMath.exp(a);
+        int256 actualExp = LogExpMath.exp(a);
+
+        assertApproxEqRel(actualExp, expectedExp, 1e2);
+    }
+
+    function testLnEquivalence(int256 a) external {
+        vm.assume(a > 0);
+
+        int256 expectedLn = ReferenceLogExpMath.ln(a);
+        int256 actualLn = LogExpMath.ln(a);
+
+        assertApproxEqAbs(actualLn, expectedLn, 2);
+    }
+}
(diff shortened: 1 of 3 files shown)


TomAFrench:
An alternative to explore, we could use [PRBMath's](https://github.com/paulrberg/prb-math/blob/88d0815baef78c0699fec1ff10b34a35903e110f/contracts/PRBMathUD60x18.sol#L432-L438) implementation. It doesn't have the crazy optimisation theory maths but switching from using base e to base 2 should definitely save gas.

pkg/solidity-utils/test/foundry/LogExpMath.t.sol
```diff
@@ -0,0 +1,28 @@
...
+        assertApproxEqRel(actualExp, expectedExp, 1e2);
+    }
+
+    function testLnEquivalence(int256 a) external {
+        vm.assume(a > 0);
+
+        int256 expectedLn = ReferenceLogExpMath.ln(a);
+        int256 actualLn = LogExpMath.ln(a);
+
+        assertApproxEqAbs(actualLn, expectedLn, 2);
```
nventuro, line 26: Wow, I'm surprised this passes with such little difference.

pkg/solidity-utils/test/foundry/LogExpMath.t.sol
```diff
@@ -0,0 +1,28 @@
...
+
+contract LogExpMathsTest is Test {
+    function testExpEquivalence(int256 a) external {
+        vm.assume(a > ReferenceLogExpMath.MIN_NATURAL_EXPONENT);
+        vm.assume(a < ReferenceLogExpMath.MAX_NATURAL_EXPONENT);
+
+        int256 expectedExp = ReferenceLogExpMath.exp(a);
+        int256 actualExp = LogExpMath.exp(a);
+
+        assertApproxEqRel(actualExp, expectedExp, 1e2);
```
nventuro, line 17: It's good to test that they are quite close, but we also definitely want some ffi testing to test that the maximum error remains what it used to be.

nventuro (commented):
I'm in favor of this renaming, but I'd do this in two steps. On the first PR, we create `NewLogExpMath` and rename the existing one to `ReferenceLogExpMath` (changing the imports as well). We have equivalence tests etc. Then, on a second PR (once we have confidence in the new impl), we rename `NewLogExpMath` to `LogExpMath`, and change the imports from `Reference` to `LogExpMath`. 

This may be slightly annoying as we'll need to touch the imports twice, but it clearly defines the point in which we change impls, and provides a clean diff (since we'll just have renames, unlike in this PR where the reference is a seemingly new file, and the new impl is also hard to review since the diff is messy (because it replaces old code).

nventuro:
Have you performed any gas measurements to compare them?

TomAFrench:
> Have you performed any gas measurements to compare them? Also bytecode size differences.

Not as of yet. My major concern is the removal of the 36 digit variant of `ln` so I was wanting to test `pow` against the TS implementation.

PR #1996, opened 2022-11-09
Deployment preparation: Gauge Adder v3

EndymionJkb:
# Description

Building on the TimelockAuthorizer, this is the new version of the GaugeAdder that supports the AuthorizerAdaptorEntrypoint. After discussion, it does use the new authorizer.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira:
> Needs discussion on whether we can still use the old authorizer for this - as it is now, in which case we could rebase off the entrypoint branch - or need to fake out the migration to the new one, or want to deploy the new authorizer first, then use it directly to simplify the tests.

My 2c: I'd try to write fork tests that work with the components that will actually be deployed to mainnet.
We don't necessarily need to actually deploy the dependencies to do so (in this case, the entrypoint and the new authorizer), but I'll need a bit of work since we usually use deployed dependencies in these tests.

pkg/deployments/tasks/20220628-gauge-adder-v2/readme.md
```diff
@@ -1,4 +1,6 @@
-# 2022-06-28 - GaugeAdder
+# 2022-06-28 - GaugeAdder V2
+
+> This Gauge Adder has been deprecated in favor of an [updated version](../../20221111-gauge-adder-v3), which uses the Authorizer Adaptor Entrypoint.
```
TomAFrench, line 3: We should potentially wait until the migration is complete before marking this as deprecated? If we do want to deprecate this now then we should move it to the `deprecated` folder (and update readme).

EndymionJkb:
We can't deprecate it now, since it's still used. (I'd tried to, hence this
somewhat leftover comment.) Maybe just say "there is a more recent
version," without calling it deprecated.

EndymionJkb:
Superseded by #2160

PR #2004, opened 2022-11-11
Add ExternalCallLib which checks revert data for malicious data

### LINKED ISSUE #1973: Wrap external view calls in LinearPool
LinearPools call `view` functions in external contracts in order to get the wrapped token rate. This is dangerous because of the `LinearPoolRebalancer`, which uses queries to manage the Pool. A malicious or vulnerable rate provider could revert with a revert reason that matches the one used in `BasePool`'s query mechanism, causing the rebalancer to perform an incorrect operation, potentially resulting in theft of Pool funds. Note that no Pool currently deployed is affected by this since Aave's system is not vulnerable- this is so far simply a hypothetical issue.

In order to prevent this, we need to wrap all external calls in a `try/catch` block and prevent a malicious revert (one that will be caught by `BasePool`'s queries) from bubbling up: an easy way to achieve this would be to bubble up a different reason. We could also perform the call manually insted of using `try/catch`.

Once we have #1941, we can have custom detectors that check that no codepaths result in external calls, except those that are covered by the guard described above.

TomAFrench:
# Description

This PR introduces the `ExternalCallLib` which exposes a function to check if revert data from external calls will be picked up by BasePool's `_queryAction` function.

~~Note that this currently doesn't fix the issue described in #1973. We must also protect the error signature used by `queryBatchSwap` as well do to this. Currently we only protect joins and exits but not swaps.~~ Fixed now.

We should also include a test of this protection when used within a pool.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

closes #1973

### DIFF
--- a/pkg/pool-linear/test/AaveLinearPool.test.ts
+++ b/pkg/pool-linear/test/AaveLinearPool.test.ts
@@ -15,6 +15,13 @@ import { deploy } from '@balancer-labs/v2-helpers/src/contract';
 import Vault from '@balancer-labs/v2-helpers/src/models/vault/Vault';
 import { ZERO_ADDRESS } from '@balancer-labs/v2-helpers/src/constants';
 
+enum RevertType {
+  DoNotRevert,
+  NonMalicious,
+  MaliciousSwapQuery,
+  MaliciousJoinExitQuery,
+}
+
 describe('AaveLinearPool', function () {
   let vault: Vault;
   let pool: LinearPool, tokens: TokenList, mainToken: Token, wrappedToken: Token;
@@ -98,9 +105,19 @@ describe('AaveLinearPool', function () {
       });
     });
 
-    context('when Aave reverts maliciously', () => {
+    context('when Aave reverts maliciously to impersonate a swap query', () => {
       sharedBeforeEach('make Aave lending pool start reverting', async () => {
-        await mockLendingPool.setRevertMaliciously(true);
+        await mockLendingPool.setRevertType(RevertType.MaliciousSwapQuery);
+      });
+
+      it('reverts with MALICIOUS_QUERY_REVERT', async () => {
+        await expect(pool.getWrappedTokenRate()).to.be.revertedWith('MALICIOUS_QUERY_REVERT');
+      });
+    });
+
+    context('when Aave reverts maliciously to impersonate a join/exit query', () => {
+      sharedBeforeEach('make Aave lending pool start reverting', async () => {
+        await mockLendingPool.setRevertType(RevertType.MaliciousJoinExitQuery);
       });
 
       it('reverts with MALICIOUS_QUERY_REVERT', async () => {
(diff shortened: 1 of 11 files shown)


pkg/pool-utils/contracts/test/MaliciousQueryReverter.sol
```diff
@@ -0,0 +1,79 @@
...
+    function setRevertMaliciously(bool enabled) external {
+        revertMaliciously = enabled;
+    }
+
+    function spoofJoinExitQueryRevert() public pure {
+        uint256[] memory tokenAmounts = new uint256[](2);
+        tokenAmounts[0] = 1;
+        tokenAmounts[1] = 2;
+
+        uint256 bptAmount = 420;
```
nventuro, line 29: _the number_

pkg/pool-linear/test/AaveLinearPool.test.ts
```diff
@@ -82,15 +85,27 @@ describe('AaveLinearPool', function () {
...
+        // We now double the reserve's normalised income to change the exchange rate to 2:1
+        await mockLendingPool.setReserveNormalizedIncome(bn(2e27));
+        expect(await pool.getWrappedTokenRate()).to.be.eq(fp(2));
+      });
+    });
+
+    context('when Aave reverts maliciously', () => {
+      sharedBeforeEach('make Aave lending pool start reverting', async () => {
+        await mockLendingPool.setRevertMaliciously(true);
+      });
```
nventuro, line 104: It'd be good to also have a rebalancer test that fails if the lending pool is malicious. 

Also, it'd be great if we tested both swap and joinexit queries. I know we test both at the lib, but it shouldn't be too much work to have both, and it'll cover all cases.

pkg/pool-utils/test/ExternalCallLib.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    maliciousReverter = await deploy('MaliciousQueryReverter');
+    caller = await deploy('MockExternalCaller', { args: [maliciousReverter.address] });
+  });
+
+  describe('when calling into a malicious contract in a swap', () => {
+    context('when call is protected', () => {
+      it('reverts with MALICIOUS_QUERY_REVERT', async () => {
+        await expect(caller.protectedSwapExternalCall()).to.be.revertedWith('MALICIOUS_QUERY_REVERT');
+      });
+    });
```
nventuro, line 21: We're missing tests where we check we bubble up the original revert reason for non malicious callees. The bubble up tests we have here are all of unprotected calls, meaning our bubble up code is not being executed.

nventuro (approved):
Very clean, loved the approach!

pkg/pool-utils/contracts/test/MaliciousQueryReverter.sol
TomAFrench: It's like leaving known bugs in a commit for audit, gotta know if the review is thorough. :smile:

PR #2010, opened 2022-11-14
Migrate to `forge-std` v1

### LINKED ISSUE #1999: Update `forge-std` to 1.0.0
We should update our version of `forge-std` to 1.0.0. I've updated locally and ran into the error described here: https://github.com/foundry-rs/forge-std/issues/215.

Once this is fixed then we can update.

Update as of 5/2/23: We have updated forge in #2448... but some (different) issues remain. We are getting intermittent failures now on the same two pesky join/exit tests.

TomAFrench:
# Description

This PR updates us to use forge-std as of commit https://github.com/foundry-rs/forge-std/commit/33d4895fb629aabf12e0f60740367bd03e702bd1

~~It seems like the majority of the issue was us using the old abi encoder. I'm not really sure how we didn't run into this issue early but :shrug:. I had to update my foundry version as well to get fuzzing working, the recent changes to the fuzzer meant we had a couple of tests failing due to too many `vm.assume` calls so I've switch them to use `bound`.~~ Fixed in other PRs

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [x] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

closes #1999


TomAFrench:
I'm running into an issue where `WeightedMathTest.testExitSwaps` consistently fails locally due to rejecting too many inputs. Seems like some of the recent changes to the fuzzer means that we're sampling inputs which fail more ofter. Running on the intense profile seems to work consistently however.

In order to get the default profile to succeed I need to bump the max rejections to 100k which is annoying and I'd prefer not to do that across the board. I'd like to either tweak the max runs for this test specifically or adjust it to avoid rejecting so many inputs.

nventuro:
I replaced one of the two `assume`'s with `bound`, but still get the same error. This is bizarre, since `testJoinSwaps` has the exact same logic for discarding runs based on weights, but that one doesn't fail.

TomAFrench:
Checked on latest forge version and this is still an issue.

https://github.com/foundry-rs/foundry/issues/3740

EndymionJkb:
Encompassed by #2448

PR #2021, opened 2022-11-17
Lint fix

jubeira:
Broken in master: https://github.com/balancer-labs/balancer-v2-monorepo/actions/runs/3488975155/jobs/5838481403; not sure why the linter complains about this.


TomAFrench:
> not sure why the linter complains about this

There's supposed to be something like `returns (uint256 rate)` before the `{` for if we needed to use the rate inside the try block. I imagine it's not smart enough to only use one space if we don't have that bit of code.

TomAFrench (approved):

PR #2015, opened 2022-11-15
Move LBP storage handling to library

TomAFrench:
# Description

This PR builds on #1979 to abstract all the logic interacting with pool state into a library.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Documentation or wording changes
- [ ] Other

## Checklist:

- [x] The diff is legible and has no extraneous changes
- [x] Complex code has been commented, including external interfaces
- [x] Tests are included for all code paths
- [x] The base branch is either `master`, or there's a description of how to merge

PR #2022, opened 2022-11-17
Fuzz helpers

EndymionJkb:
# Description

Reviewing #2015, I realized we now had multiple copies of some helper routines used in fuzz tests. They were very WordCodec-y, so the library is called WordCodecHelpers.

It's off master, rather than 2015's branch (though it also applies there), to not introduce extraneous changes.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A


nventuro (commented):
Hm, I'm all for having test helpers, but I wouldn't add them to WordCodec if we're not actually using them. That'd mean we start using lib functions to test other lib functions, and adds more production code.

I'd consider instead having test helpers (that are not part of our prod code, e.g. they live in the `test` directory or w/e, or perhaps we even have a test helpers package), and use those in our tests.

EndymionJkb:
> I'd consider instead having test helpers (that are not part of our prod code, e.g. they live in the `test` directory or w/e, or perhaps we even have a test helpers package), and use those in our tests.

Simplest might be a parallel WordCodecHelpers in the same directory? Those functions are mainly testing that WordCodec operations don't stomp on each other. They aren't tests themselves; `clearWordAtPosition` is a reasonable WordCodec function, though it isn't currently used in production. `isOtherStateUnchanged` is basically an invariant we are checking in the tests.

TomAFrench (approved):
Looks good. Agreed on WordCodec not being the place for this but I'm happy with a parallel library.

PR #1993, opened 2022-11-08
Deployment preparation: TimelockAuthorizer (with Migrator)

jubeira:
# Description

This PR adds the deployment task for the `TimelockAuthorizer` and its migrator.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Documentation or wording changes
- [x] Other

## Checklist:

- [ ] The diff is legible and has no extraneous changes
- [ ] Complex code has been commented, including external interfaces
- [ ] Tests are included for all code paths
- [ ] The base branch is either `master`, or there's a description of how to merge

## Issue Resolution

See #1990

PR #2027, opened 2022-11-18
Begin adding tracking for "real world" state of permissions granted onchain

TomAFrench:
# Description

This PR starts the creation of the `v2-permissions` package which includes various scripts which aim to pull the current state of the Authorizer from onchain (through a subgraph) and reconstruct a human readable representation of the permissions which have been granted.

Using the subgraph we can get a list of addresses and for each address a list of action ids which they have permissions over. We can then map from action ids to a list of contract-function pairs which it allows them to call, and from addresses to a human readable name, e.g.

```
{
  "GNOSIS_PROTOCOL_RELAYER": [
    {
      "taskId": "20210418-vault",
      "contractName": "Vault",
      "signature": "manageUserBalance((uint8,address,uint256,address,address)[])",
      "useAdaptor": false
    },
    {
      "taskId": "20210418-vault",
      "contractName": "Vault",
      "signature": "batchSwap(uint8,(bytes32,uint256,uint256,uint256,bytes)[],address[],(address,bool,address,bool),int256[],uint256)",
      "useAdaptor": false
    }
  ],
  "BLABS_VEBAL_MULTISIG": [
    {
      "taskId": "20220325-bal-token-holder-factory",
      "contractName": "BALTokenHolder",
      "signature": "function withdrawFunds(address,uint256)",
      "useAdaptor": false
    }
  ],
}
```

The state of this package is very rough and will be refined but I want to share it early to help with verifying #1993 

1. Verify that all action ids in `permissions/unrecognised.json` are unnecessary and then renounce/revoke them. 
2. Sanity check that `permissions/functions.json` is a desirable end state for the system to be in post-migration.
3. Ensure that the `permissions/actionIds.json` matches the input to the authorizer migration deployment task
    - This isn't 1:1 as the authorizer doesn't currently have a concept of `where` so we need to check that this is set properly for each action ID in the deployment task.
    - This can also be done less rigorously by just checking `permissions/functions.json` against the deployment task. This is likely good enough. 

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


TomAFrench:
Waiting for #2031 to be merged so we can use it here.

pkg/permissions/labelled-accounts/mainnet.json
```diff
@@ -0,0 +1,13 @@
...
+  "0x166f54f44f271407f24aa1be415a730035637325": "BALLER_OPS_MULTISIG",
+  "0x75a52c0e32397a3fc0c052e2ceb3479802713cf4": "LINEAR_POOL_CONTROLLER_BALLER_MULTISIG",
+  "0xf4a80929163c5179ca042e1b292f5efbbe3d89e6": "SWAP_FEE_CONTROLLER_BALLER_MULTISIG",
+  "0xc38c5f97b34e175ffd35407fc91a937300e33860": "LM_MULTISIG",
+  "0x7c68c42de679ffb0f16216154c996c354cf1161b": "TREASURY_MULTISIG",
+  "0xa29f61256e948f3fb707b4b3b138c5ccb9ef9888": "EMERGENCY_SUBDAO_MULTISIG",
+  "0x02f35dA6A02017154367Bc4d47bb6c7D06C7533B": "BLABS_OPS_MULTISIG",
+  "0xd2eb7bd802a7ca68d9acd209bec4e664a9abdd7b": "BLABS_VEBAL_MULTISIG",
+  "0xe4a8ed6c1d8d048bd29a00946bfcf2db10e7923b": "GAUNTLET_FEE_SETTER",
+  "0xc92e8bdf79f0507f65a392b0ab4667716bfe0110": "GNOSIS_PROTOCOL_RELAYER"
```
nventuro, line 12: It'd be good if we could also have a short description of what these are, their scope, and potentially even link to gov discussion. We could add a metadata field, or make this a ts file and just write comments

pkg/permissions/scripts/getCallableFunctions.ts
```diff
@@ -0,0 +1,65 @@
...
+    {
+      taskId: '20220325-bal-token-holder-factory',
+      contractName: 'BALTokenHolder',
+      signature: 'function withdrawFunds(address,uint256)',
+      useAdaptor: false,
+    },
+  ],
+};
+
+const ignoredActionIds: string[] = [
```
nventuro, line 22: These are network specific, right?

pkg/permissions/src/accounts.ts
```diff
@@ -0,0 +1,35 @@
...
+  const response: Response = await fetch('https://api.thegraph.com/subgraphs/name/balancer-labs/balancer-authorizer', {
+    method: 'POST',
+
+    headers: {
+      'Content-Type': 'application/json',
+    },
+
+    body: JSON.stringify({
+      query: `{
+        accounts(first: 500) {
```
nventuro, line 14: What if 500 is not enough?

nventuro (commented):
Fantastic initiative!

pkg/permissions/scripts/getCallableFunctions.ts
TomAFrench: I'm making this package be mainnet only for the time being we're mostly interested in it for the authorizer migration.

pkg/permissions/src/accounts.ts
TomAFrench: We'd have to do pagination at some point as the graph limits how many responses is returns. We've got much fewer than 500 permissions atm (<50) so I'm not too concerned.

I'll add something to blow up this function if we actually receive 500 accounts.

TomAFrench:
As we're migrating the authorizer on mainnet but not on L2s we'll need to have support for both authorizers for the foreseeable future. Parking this PR until we update subgraph, etc. appropriately.

PR #2030, opened 2022-11-21
Make unnecessarily async functions sync

TomAFrench:
# Description

We've got a number of helper functions in the `v2-deployments` package as async whereas the implementations are sync. We can then simplify usage of these functions by marking them as sync.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Documentation or wording changes
- [ ] Other

## Checklist:

- [x] The diff is legible and has no extraneous changes
- [x] Complex code has been commented, including external interfaces
- [x] Tests are included for all code paths
- [x] The base branch is either `master`, or there's a description of how to merge

PR #2031, opened 2022-11-21
Add address lookup table to find contract task/name

TomAFrench:
# Description

builds on #2030 

Given a deployment task and contract name, it's currently easy to lookup the onchain address for that contract. What's more difficult however is given an onchain contract address, check to see if it's a Balancer-deployed contract and if so return the task id and contract name.

This PR generates a lookup table for this operation which makes it easier to generate human-readable identifiers for contracts when starting from onchain data. It currently only handles contracts which are directly deployed (i.e. Pool factories but not Pools themselves) as the initial usecase for this map (tracking addresses which hold permissions on the authorizer) doesn't require this. 

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


nventuro (commented):
I understand wanting to have this, but wouldn't a better way be to load all output files for all tasks in said network, and search for the address there? It'd mean we don't need to maintain these json files (which we'd need to check for correctness if we kept them).

nventuro:
We already have this task-output looping behavior when checking deployments for a network.

TomAFrench:
> I understand wanting to have this, but wouldn't a better way be to load all output files for all tasks in said network, and search for the address there? It'd mean we don't need to maintain these json files (which we'd need to check for correctness if we kept them).

This is a bit nasty when you're dealing with many addresses. For every address you want to check you need to find all the deployment tasks, iterate through them loading the output json files and then iterate through every key-value pair to find one with the correct value. We need to do a lot more work than if we just precompute the result of that and save it here so we're just looking up a key in a mapping.

That said, we could use that method to compute this mapping at build-time rather than at deployment like this PR currently does. I'd still want a CI action to ensure this is up to date though.

nventuro:
What is the use case for this? If we don't care _that much_ about performance, then having way simpler code seems like a good idea, instead of having to maintain the build list code, check list code, etc.

TomAFrench:
I think shifting this code to `index.ts` would _increase_ code complexity. We have a lot of helpers for manipulating the task data which we can use internally but whereas we'd get more brittle and harder to maintain code by redoing this in `index.ts`

pkg/deployments/index.ts
```diff
@@ -92,3 +103,11 @@ function getBalancerContractArtifactPath(task: string, contract: string): string
...
   return `@balancer-labs/v2-deployments/dist/tasks/${task}/output/${network}.json`;
 }
+
+/**
+ * @dev Returns the path for the list of Balancer contract addresses on a network
+ * @param network Name of the network looking the deployment path for (e.g. mainnet, polygon, goerli, etc)
+ */
+function getBalancerContractAddresses(network: string): string {
+  return `@balancer-labs/v2-deployments/dist/addresses/${network}.json`;
+}
```
nventuro, line 113: What's the purpose of this function?

pkg/deployments/index.ts
```diff
@@ -1,6 +1,17 @@
...
+/**
+ * @dev Returns the task id and contract name for a canonical contract deployed on a specific network
+ * @param address Address of the contract to be fetched
+ * @param network Name of the network looking the deployment for (e.g. mainnet,  polygon, goerli, etc)
+ */
+export function lookupBalancerContractByAddress(address: string, network: string): { task: string; name: string } {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const networkAddresses = require(getBalancerContractAddresses(network));
+  return networkAddresses[address];
+}
```
nventuro, line 13: Let's add this to the changelog

pkg/deployments/src/network.ts
```diff
@@ -39,3 +41,28 @@ export function getContractDeploymentTransactionHash(deployedAddress: string, ne
 
   return txHash;
 }
+
+export function saveContractDeploymentAddresses(task: Task): void {
+  if (task.network === 'hardhat') return;
```
nventuro, line 46: Do we need this? We'll pass an explicit `--network` flag

pkg/deployments/src/network.ts
```diff
@@ -39,3 +41,28 @@ export function getContractDeploymentTransactionHash(deployedAddress: string, ne
...
+
+  const newEntries = Object.fromEntries(
+    Object.entries(task.output({ ensure: false }))
+      .filter(([name]) => name !== 'timestamp')
```
nventuro, line 50: :sadge:
we should remove these timestamp fields

nventuro (approved):

pkg/deployments/src/network.ts
TomAFrench: Yeah, I don't think we use them at all and we can look up the deployment tx in a pinch.

pkg/deployments/src/network.ts
TomAFrench: I'm not sure. It's mostly there because we had it for `saveContractDeploymentTransactionHash` and we were building the lookup table in the same way as for deployment transactions initially. I'd say it doesn't hurt to keep it.

pkg/deployments/index.ts
TomAFrench: It returns the path of the file which contains the addresses on a particular network. It means so that if the path changes in future we only need to change it in a single place.

Issue #2026, opened 2022-11-18
Add Etherscan-visible versioning to pools


EndymionJkb:
We have a versioning system in the /deployments package and readme's that labels each deployment with a date, includes the complete build info, and often indicates a version in the name (e.g., gauge-adder-v2). Of course, none of this is known by Etherscan, so finding out which version a pool is could involve checking all the factories to find which one deployed it, then noting the date the factory was deployed, and looking it up in the repo docs, which is all manual and fairly cumbersome.

We could change the name of the contract itself, which would be visible in Etherscan - but we don't generally want to change the names of the contracts unless it is truly different (e.g., StablePool -> ComposableStablePool). If we're just fixing a bug or updating a dependency (e.g., Aave v2 to v3), the name should stay the same.

We could also have an external contract registry with metadata, including this information - and we may well do this at some point - but that's too heavyweight for now.

So the consensus solution is to add a `version` method to all pools that returns a string. Alternatively, if we want an easily machine-readable numeric value (vs parsing it out of a string), we could add a `versionNumber`: 1,2,3,4,... and a `versionName`, which might be a string identifying the task id (e.g., '20221111-authorizer-adaptor-entrypoint'). Then it would be trivial to find the correct deployment from Etherscan

PR #2032, opened 2022-11-22
Aave linear pool versioning

jubeira:
# Description

This PR adds an `version` getter for `AaveLinearPool` and `AaveLinearPoolFactory` (factory provides created pools with the actual version string).

Eventually the interface will apply to `BasePoolFactory` and `Version` to `BasePool` and `NewBasePool`. Since that change would require some refactoring across the codebase we're starting with this one since it's more urgent.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

See #2026 (this one shall be closed when this feature is implemented in every pool).

### DIFF
--- a/pkg/pool-linear/test/AaveLinearPoolFactory.test.ts
+++ b/pkg/pool-linear/test/AaveLinearPoolFactory.test.ts
@@ -15,6 +15,7 @@ import Token from '@balancer-labs/v2-helpers/src/models/tokens/Token';
 describe('AaveLinearPoolFactory', function () {
   let vault: Vault, tokens: TokenList, factory: Contract;
   let creationTime: BigNumber, owner: SignerWithAddress;
+  let version: string;
 
   const NAME = 'Balancer Linear Pool Token';
   const SYMBOL = 'LPT';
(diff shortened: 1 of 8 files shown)


pkg/pool-linear/test/AaveLinearPoolFactory.test.ts
```diff
@@ -72,6 +78,14 @@ describe('AaveLinearPoolFactory', function () {
       expect(await pool.getVault()).to.equal(vault.address);
     });
 
+    it('checks the factory version', async () => {
+      expect(await factory.version()).to.equal(version);
+    });
+
+    it('checks the pool version', async () => {
+      expect(await pool.version()).to.equal(version);
+    });
```
nventuro, line 87: Don't you think it's weird that the factory returns the name of the pool?

nventuro: If we're just storing values in storage, we could drop the whole version provider idea, and have the factory pass a version to the pool. The only issue is we'd either have to a) generate the json on the fly (and solidity is bad) or b) pass both the factory and pool version to the factory constructor.

jubeira: > Don't you think it's weird that the factory returns the name of the pool?

Hmm not really. When we do a deployment, we think of e.g. `ManagedPool`, `ComposableStablePool`, etc; the fact that the pools actually come from factories is sort of implicit.

Generating the JSON in solidity doesn't sound very good. 

I'd say that if pools of a given version will always come from the same factory, I don't see a reason to make a difference between factory / pool version. We could rename the method in the factory if it sounds misleading.

Otherwise, I guess we can give the factory the two names.

nventuro (approved):
Thanks a lot! Merging so that we can update the deployment, but can you separately add a simple unit test for `Version`?

PR #2034, opened 2022-11-22
Composable stable pool versioning

jubeira:
# Description

This PR adds an version getter for `ComposableStablePool` and `ComposableStablePoolFactory` (factory provides created pools with the actual version string).

Similar to #2032, but using composable pool's deployment branch as base (this is just for this deployment, not a final implementation of this feature).
This PR does not introduce a separate `Version` contract because that caused a 'stack too deep' error in `ComposableStablePoolContract`, so the functionality is implemented directly in the pool.

In this case, the TS helpers deploy the pool directly without using the factory, so this PR doesn't test the version that the factory should provide.

Merge before #2028 and update the artifacts in that PR.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

- [ ] Tests are included for all code paths --> **See note about TS helpers and composable pool factory above**

## Issue Resolution

See #2026.

### DIFF
--- a/pkg/interfaces/contracts/pool-utils/IVersionProvider.sol
+++ b/pkg/interfaces/contracts/pool-utils/IVersionProvider.sol
@@ -1,0 +1,28 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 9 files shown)


pkg/interfaces/contracts/pool-utils/IVersionProvider.sol
```diff
@@ -0,0 +1,28 @@
...
+ * @notice Simple interface to retrieve the version of a deployed contract.
+ *
+ * @dev The contract implementing the interface may provide the actual version, or forward the call to another
+ * version provider (e.g. a pool will forward the call to its respective pool factory).
+ */
+interface IVersionProvider {
+    /**
+     * @dev Returns a JSON representation of the contract version containing name, version number and task ID.
+     */
+    function version() external view returns (string memory);
```
nventuro, line 27: We're not using this, delete this file.

nventuro (approved):

PR #2041, opened 2022-11-22
Update Linear build info

nventuro:
# Description

This upgrades the linear deployment to include the latest changes, namely https://github.com/balancer-labs/balancer-v2-monorepo/pull/2032.

It will conflict with https://github.com/balancer-labs/balancer-v2-monorepo/pull/2040, which also touches the input and index files, but it should be relatively simple to address these issues.


jubeira (approved):
I've built the contracts and the artifact matches the contents of this branch @https://github.com/balancer-labs/balancer-v2-monorepo/pull/2041/commits/e00ac4b414eccdc39bfc667cb43c853985df179b.

Issue #2039, opened 2022-11-22
Add standardised fork test for deployments to check that it reports the correct task id.


TomAFrench:
We're going to be deploying contracts with a version string which contains the task id contained within it. It would be a pain to have this be incorrect so we should have a simple method to check beforehand that the task will pass the correct task id to the contract being deployed.

jubeira:
Moved to https://github.com/balancer/balancer-deployments/issues/41.

PR #2046, opened 2022-11-23
Finalize composable stable pool deployment

nventuro:
# Description

This updates the new composable stable pool deployment with the latest artifacts, and adjusts the deployment to account for the version. Once this is in, we can merge `composable-stable-pool-deployment-base` into `master` (since we'll have removed all dependencies on contract source code).

We should double-check that the branch that produces this build info is the exact same as the one used for the old deployment, plus: a) versioning and b) proportional exits and joins.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A


jubeira (commented):
- I've built the artifacts @https://github.com/balancer-labs/balancer-v2-monorepo/pull/2046/commits/a69da44ea07f8dd21a0e35da82a3cfa4e328df8c and they match the ones in this PR.
- The [diff wrt. master](https://github.com/balancer-labs/balancer-v2-monorepo/compare/composable-stable-pool-deployment-base#) so far looks good regarding the version. The modifications to the fork test look good as well.
- I've done a quick pass over proportional joins exits and it seems to be OK -there is some refactoring going on as well which I understand is also OK from the PRs-.

pkg/deployments/tasks/20221122-composable-stable-pool-v2/input.ts
```diff
@@ -3,12 +3,19 @@ import Task, { TaskMode } from '../../src/task';
...
   Vault: string;
   ProtocolFeePercentagesProvider: string;
+  FactoryVersion: string;
+  PoolVersion: string;
 };
 
 const Vault = new Task('20210418-vault', TaskMode.READ_ONLY);
 const ProtocolFeePercentagesProvider = new Task('20220725-protocol-fee-percentages-provider', TaskMode.READ_ONLY);
 
+const BaseVersion = { version: 2, deployment: '20221122-composable-stable-pool-v2' };
```
TomAFrench, line 13: Seems like we should be able to pull the name and deployment fields from the deployment task automatically. This would avoid us having to check for consistency. (Although we should anyway)

nventuro: Agreed. Let's wait until #2039 to add that though.

TomAFrench (approved):
I've built a69da44ea07f8dd21a0e35da82a3cfa4e328df8c and get the same build info as committed. I've also checked to make sure that the diff from 93af2fda1eefe1b7d103e6c54ce159951ff001fc is just adding proportional exits and version getters.

PR #2051, opened 2022-11-23
Adaptor entrypoint events

### LINKED ISSUE #1809: Add events to AuthorizerAdaptor
Some actions (such as killing gauges) don't emit events - it'd be useful to get the Adaptor to emit events with target and data when triggered.

This can of course not be done natively as we cannot modify the already deployed adaptors, but we could have an external layer (an entrypoint) that is interacted with instead of the adaptor, and have it emit said events. This could prove useful if we wanted to add more features to it (such as improving the permission scheme).

jubeira:
# Description

Add event to `performAction` in adaptor entrypoint.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #1809.

### DIFF
--- a/pkg/liquidity-mining/contracts/admin/AuthorizerAdaptorEntrypoint.sol
+++ b/pkg/liquidity-mining/contracts/admin/AuthorizerAdaptorEntrypoint.sol
@@ -29,7 +29,7 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/Address.sol";
  * permissions correctly.
  */
 contract AuthorizerAdaptorEntrypoint is IAuthorizerAdaptorEntrypoint {
-    event ActionPerformed(address indexed target, bytes data);
+    event ActionPerformed(address indexed caller, address indexed target, bytes data);
 
     using Address for address;
 
@@ -84,7 +84,7 @@ contract AuthorizerAdaptorEntrypoint is IAuthorizerAdaptorEntrypoint {
 
         _require(canPerform(getActionId(selector), msg.sender, target), Errors.SENDER_NOT_ALLOWED);
 
-        emit ActionPerformed(target, data);
+        emit ActionPerformed(msg.sender, target, data);
 
         // The `AuthorizerAdaptor` will not check for permissions: it is special-cased in the `TimelockAuthorizer` so
         // that all calls to it that are not made from this entrypoint fail, while those that originate in the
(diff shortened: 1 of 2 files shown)


pkg/liquidity-mining/contracts/admin/AuthorizerAdaptorEntrypoint.sol
```diff
@@ -29,6 +29,8 @@ import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/Address.sol";
  * permissions correctly.
  */
 contract AuthorizerAdaptorEntrypoint is IAuthorizerAdaptorEntrypoint {
+    event ActionPerformed(address indexed caller, address indexed target, bytes data);
```
jubeira, line 32: I guess indexing the addresses makes sense since both the authorized callers and the targets are not _that_ many.

Optionally we could also include the selector; even if it's contained in the data perhaps it could be useful?

nventuro: I like indexing the selector!

nventuro: It is possible we'll want to look to e.g. `checkpoint` calls in multiple targets

nventuro (approved):

pkg/liquidity-mining/contracts/admin/AuthorizerAdaptorEntrypoint.sol
jubeira: Done.

I've modified the test accordingly. I added some random bytes to the calldata in the test so that we can actually verify that the selector is where it's supposed to be, and it's not just a match by chance. If you think it's not too clear please let me know; holding the merge til then.

pkg/liquidity-mining/test/AuthorizerAdaptorEntrypoint.test.ts
```diff
@@ -103,6 +107,16 @@ describe('AuthorizerAdaptorEntrypoint', () => {
         // The authorizer will reject calls that are not initiated in the adaptor adaptorEntrypoint.
         await expect(adaptor.connect(grantee).performAction(target, calldata)).to.be.revertedWith('SENDER_NOT_ALLOWED');
       });
+
+      it('emits an event describing the performed action', async () => {
+        const tx = await adaptorEntrypoint.connect(grantee).performAction(target, calldata);
+        expectEvent.inReceipt(await tx.wait(), 'ActionPerformed', {
+          selector: calldata.slice(0, 10), // '0x' + 8 characters representing 4 bytes.
```
nventuro, line 114:
```suggestion
          selector: paymentReceiver.interface.getSigHash('receivePayment'),
```

PR #2052, opened 2022-11-23
Expose link references from deployments package

TomAFrench:
# Description

@gerrrg is having troubles deploying the Managed pool deployment as he cannot link the library bytecode into that of the factory. To do this he needs the link references which are contained in the build-info.

This PR writes these into the `bytecode` files we already export. This seems to be the most natural place to store these and I'd like to avoid creating yet another directory and file to hold these, especially as for 99% of contracts it will be empty.

We then also need to change the interface of the package to return an object rather than simply a string containing the contract's bytecode.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


TomAFrench:
Closing as we're going a different route.

PR #2058, opened 2022-11-24
Expose deployment artifacts from deployments package rather than ABI + bytecode

TomAFrench:
# Description

Rather than pulling out the ABI and bytecode for each contract and saving them to the `bytecode`/`abi` folders, we now have a single `artifact` folder per deployment task. This is then populated from the build-info file in the exact same way that hardhat generates its own artifact files so users of the npm package can use these as if they had compiled the contracts locally.

`getBalancerContractABI` is not deprecated as this is a useful helper for users who only want to be able to interact with existing contracts but not deploy them. For anyone deploying their own contracts then it's expected that they use `getBalancerContractArtifact`. `getBalancerContractBytecode` is deprecated.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


TomAFrench:
Adding changelog stuff now.

TomAFrench:
There's also a load of really mangled english in the jsdoc here. Will fix in a separate PR.

pkg/deployments/src/artifact.ts
```diff
@@ -44,69 +43,88 @@ export function checkArtifact(task: Task): void {
       const fileName = path.parse(buildInfoFileName).name;
       const contractName = fileName;
 
-      const expectedArtifact = extractContractArtifact(task, fileName, contractName);
-      const { abi, bytecode } = readContractABIAndBytecode(task, contractName);
+      const expectedArtifact = task.artifact(contractName, fileName);
+      const artifact = readContractArtifact(task, contractName);
```
nventuro, line 47:
```suggestion
      const actualArtifact = readContractArtifact(task, contractName);
```

pkg/deployments/src/artifact.ts
```diff
@@ -44,69 +43,88 @@ export function checkArtifact(task: Task): void {
...
 
-      const bytecodeMatch = bytecode === expectedArtifact.evm.bytecode.object;
-      const abiMatch = JSON.stringify(abi) === JSON.stringify(expectedArtifact.abi);
-      if (bytecodeMatch && abiMatch) {
-        logger.success(`Verified ABI and bytecode integrity of contract '${contractName}' of task '${task.id}'`);
+      const artifactMatch = JSON.stringify(artifact) === JSON.stringify(expectedArtifact);
```
nventuro, line 49:
```suggestion
      const artifactMatch = JSON.stringify(actualArtifact) === JSON.stringify(expectedArtifact);
```

nventuro (approved):

pkg/deployments/index.ts
```diff
@@ -18,28 +19,40 @@ export async function getBalancerContract(task: string, contract: string, networ
...
+  return require(getBalancerContractArtifactPath(task, contract));
 }
 
 /**
+ * @dev Returns the ABI for a contract from a specific task
+ * @param task ID of the task to look the ABI of the required contract
+ * @param contract Name of the contract to be fetched.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function getBalancerContractABI(task: string, contract: string): any[] {
```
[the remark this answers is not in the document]
TomAFrench: Ack, managed to delete this function and didn't notice this when I re-added it.

TomAFrench:
@nventuro Just realised that this PR means that we need to go update all the links in each deployment :sadge:

nventuro:
'we'

PR #1690, opened 2022-09-02
Prevent invariant overflow

nventuro:
Not quite sure what we were thinking with the lack of SafeMath here :shrug:

PR #2059, opened 2022-11-24
Rebased composable stable pool deployment base

jubeira:
# Description

#2053, but with base in https://github.com/balancer-labs/balancer-v2-monorepo/commit/93af2fda1eefe1b7d103e6c54ce159951ff001fc.

Build info added in commit https://github.com/balancer-labs/balancer-v2-monorepo/pull/2059/commits/a49fe13ff97d881ded2a7f9c4bf5dff2578ee5ad was created from commit https://github.com/balancer-labs/balancer-v2-monorepo/pull/2059/commits/454c740b1027a4996847b532361e3e217aa6c409.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other

## Checklist:


pkg/pool-stable/contracts/ComposableStablePool.sol
```diff
@@ -903,6 +946,31 @@ contract ComposableStablePool is
         return (bptAmountIn, amountsOut);
     }
 
+    function _computeProportionalAmountsOut(
```
jubeira, line 949: This should probably _not_ be here.

When merging master back to the branch to solve the conflicts, the build breaks without this.
This function was implemented in `RecoveryMode`, and is similar to `_doRecoveryModeExit` (except that the latter expects balances including BPT, and the API is different).

I just added it so that we can start the review, but this should be addressed before merging.

jubeira:
@TomAFrench I've pushed the artifacts and solved the conflicts so that we can start reviewing this.

@nventuro to summarize:
- This PR branches off https://github.com/balancer-labs/balancer-v2-monorepo/commit/93af2fda1eefe1b7d103e6c54ce159951ff001fc (which is the merge commit for the old composable stable pool deployment in https://github.com/balancer-labs/balancer-v2-monorepo/pull/1695), and then rebases everything that is in https://github.com/balancer-labs/balancer-v2-monorepo/pull/2053, solving conflicts in the middle.
- After the whole rebase, I've updated the artifacts with the tip of the rebased branch. Those are the artifacts that we should deploy, since they have composable V1 + all the changes needed for composable V2.
- Finally, I've merged `master` back into this branch to solve merge conflicts. About those:
  - I've taken [a shortcut](https://github.com/balancer-labs/balancer-v2-monorepo/pull/2059#discussion_r1031728955) in the process
  - The biggest conflict was in `_updateInvariantAfterJoinExit`; just for reference it was introduced [here](https://github.com/balancer-labs/balancer-v2-monorepo/pull/1690/files#).

jubeira:
Superseded by #2065

PR #2060, opened 2022-11-24
(Re) deployment preparation: `AuthorizerAdaptorEntrypoint`

jubeira:
# Description

Redeployment of `AuthorizerAdaptorEntrypoint`, deployed in https://github.com/balancer-labs/balancer-v2-monorepo/pull/2009 (not yet in use), but now including the new event feature introduced in https://github.com/balancer-labs/balancer-v2-monorepo/pull/2051.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Documentation or wording changes
- [x] Other

## Checklist:

- [x] The diff is legible and has no extraneous changes
- [x] Complex code has been commented, including external interfaces
- [x] Tests are included for all code paths
- [x] The base branch is either `master`, or there's a description of how to merge

## Issue Resolution

See #2057 (will close after deployment).

PR #2062, opened 2022-11-24
(re)deploy entrypoint

nventuro:
# Description

From https://github.com/balancer-labs/balancer-v2-monorepo/pull/2060


jubeira (commented):
LGTM!

Do we care about the deployed address in this case? I could deploy with a wallet that would produce the same address in every network.

nventuro:
No - actually better that the addresses _don't_ match, so that we have to check they're the right one. Addresses being the same in different networks means nothing, we should not rely on that.

jubeira (approved):

Issue #2042, opened 2022-11-23
Make RecoveryMode work with cash amounts instead of total


nventuro:
The Recovery Mode exit currently lets LPs exit with their share of the Pool's `total` balance for each token. However, this could fail in Pools that have Asset Managers, since `managed` will be non-zero: if all LPs tried to exit, the Pool would at some point end up with negative `cash`, causing reverts.

To avoid this, we need to have Recovery Mode use `cash` instead of `total`. This would be achieved by calling back into the Vault with the `getPoolTokenInfo` function.

We may not want to do this for _all_ Pools, but it is clear that at least Managed Pools should work this way.

PR #2068, opened 2022-11-25
Introduce RecoveryModeHelper

nventuro:
# Description

This introduces an external helper contract that can be used to compute recovery mode exits. Unlike all exits we've had so far, it uses the 'cash' balance of the Pool exclusively. By placing this code outside of the Pool, we can reduce its bytecode size.

This initial version only works with Composable Pools, but it's easily extendable to also work with regular Pools. On a later PR I'll modify ManagedPool so that it uses this helper.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Part of #2042

### DIFF
--- a/pkg/interfaces/contracts/pool-utils/IRecoveryModeHelper.sol
+++ b/pkg/interfaces/contracts/pool-utils/IRecoveryModeHelper.sol
@@ -1,0 +1,35 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity >=0.7.0 <0.9.0;
+
+/**
+ * Interface for an auxiliary contract that computes Recovery Mode exits, removing logic from the core Pool contract
+ * that would otherwise take up a lot of bytecode size at the cost of some slight gas overhead. Since Recovery Mode
+ * exits are expected to be highly infrequent (and ideally never occur), this tradeoff makes sense.
+ */
+interface IRecoveryModeHelper {
+    /**
+     * @dev Computes a Recovery Mode Exit BPT and token amounts for a Pool. Only 'cash' balances are considered, to
+     * avoid scenarios where the last LPs to attempt to exit the Pool cannot because only 'managed' balance remains.
+     *
+     * The Pool is assumed to be a Composable Pool that uses ComposablePoolLib, meaning BPT will be its first token. It
+     * is also assumed that there is no 'managed' balance for BPT.
+     */
+    function calcComposableRecoveryAmountsOut(
+        bytes32 poolId,
+        bytes memory userData,
+        uint256 totalSupply
+    ) external view returns (uint256 bptAmountInt, uint256[] memory amountsOut);
+}
--- a/pkg/interfaces/contracts/pool-utils/IRecoveryModeHelper.sol
+++ b/pkg/interfaces/contracts/pool-utils/IRecoveryModeHelper.sol
@@ -31,5 +31,5 @@ interface IRecoveryModeHelper {
         bytes32 poolId,
         bytes memory userData,
         uint256 totalSupply
-    ) external view returns (uint256 bptAmountInt, uint256[] memory amountsOut);
+    ) external view returns (uint256 bptAmountIn, uint256[] memory amountsOut);
 }
--- a/pkg/pool-utils/contracts/RecoveryModeHelper.sol
+++ b/pkg/pool-utils/contracts/RecoveryModeHelper.sol
@@ -42,9 +42,9 @@ contract RecoveryModeHelper is IRecoveryModeHelper {
         bytes memory userData,
         uint256 totalSupply
     ) external view override returns (uint256 bptAmountIn, uint256[] memory amountsOut) {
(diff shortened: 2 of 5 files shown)


pkg/pool-utils/contracts/RecoveryModeHelper.sol
```diff
@@ -0,0 +1,69 @@
...
+    function getVault() public view returns (IVault) {
+        return _vault;
+    }
+
+    function calcComposableRecoveryAmountsOut(
+        bytes32 poolId,
+        bytes memory userData,
+        uint256 totalSupply
+    ) external view override returns (uint256 bptAmountIn, uint256[] memory amountsOut) {
+        // As ManagedPool is a composable Pool, `_doRecoveryModeExit()` must use the virtual supply rather than the
```
EndymionJkb, line 45:
```suggestion
        // As this is a composable Pool, `_doRecoveryModeExit()` must use the virtual supply rather than the
```
If the intent is to use this for all kinds of composable pools (linear, stable as well), we should avoid mentions of explicit pool types

pkg/pool-utils/contracts/RecoveryModeHelper.sol
```diff
@@ -0,0 +1,69 @@
...
+            (uint256 cash, , , ) = vault.getPoolTokenInfo(poolId, registeredTokens[i]);
+            cashBalances[i] = cash;
+        }
+
+        uint256 virtualSupply;
+        (virtualSupply, cashBalances) = ComposablePoolLib.dropBptFromBalances(totalSupply, cashBalances);
+
+        bptAmountIn = userData.recoveryModeExit();
+
+        amountsOut = WeightedMath._calcTokensOutGivenExactBptIn(cashBalances, bptAmountIn, virtualSupply);
```
EndymionJkb, line 64: In the same vein as the above comment about overfitting to pool types, we have several versions of the `_calcTokensOutGivenExactBptIn` function (and with stable it's embedded in the recovery mode exit). Since it's external anyway, maybe implement a version locally here (like the old _computeProportionalAmountsOut) for all pool types to use, so that we avoid using libraries specific to weighted pools?

EndymionJkb (commented):
Using `cash` is good future-proofing for when we start using asset managers for various unseeded things (managed add/remove, LBPs, etc.) Comments inline.

This should be highlighted in the documentation though, since recovery UIs or users on Etherscan would likely not think of this, and expect values proportional to the total balances

pkg/pool-utils/contracts/RecoveryModeHelper.sol
nventuro: Ah yes, i forgot to change this. Thanks!

pkg/pool-utils/contracts/RecoveryModeHelper.sol
nventuro: Yes, I want to have those in the library for recovery mode user data decoding. I didn't do that yet because it'll cause conflicts with the stable composable branch merge, but we should do it once that's in.

pkg/interfaces/contracts/pool-utils/IRecoveryModeHelper.sol
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Interface for an auxiliary contract that computes Recovery Mode exits, removing logic from the core Pool contract
+ * that would otherwise take up a lot of bytecode size at the cost of some slight gas overhead. Since Recovery Mode
+ * exits are expected to be highly infrequent (and ideally never occur), this tradeoff makes sense.
+ */
+interface IRecoveryModeHelper {
+    /**
+     * @dev Computes a Recovery Mode Exit BPT and token amounts for a Pool. Only 'cash' balances are considered, to
+     * avoid scenarios where the last LPs to attempt to exit the Pool cannot because only 'managed' balance remains.
```
jubeira, line 25:
```suggestion
     * avoid scenarios where the last LPs to attempt to exit the Pool cannot do it because only 'managed' balance
     * remains.
```

pkg/pool-utils/contracts/RecoveryModeHelper.sol
```diff
@@ -0,0 +1,69 @@
...
+contract RecoveryModeHelper is IRecoveryModeHelper {
+    using BasePoolUserData for bytes;
+
+    IVault private immutable _vault;
+
+    constructor(IVault vault) {
+        _vault = vault;
+    }
+
+    function getVault() public view returns (IVault) {
```
jubeira, line 36: Maybe consider adding this to the interface?

Perhaps it's not necessary, but there are other interfaces where we do include it.

EndymionJkb (approved):
Looks good then, given that the rest of the comments will be addressed in future PRs (after the stable deployment)

pkg/pool-utils/contracts/RecoveryModeHelper.sol
jubeira: https://github.com/balancer-labs/balancer-v2-monorepo/issues/2072

pvt/helpers/src/models/vault/Vault.ts
```diff
@@ -73,12 +73,12 @@ export default class Vault {
...
 
-  async updateCash(poolId: string, cash: BigNumber[]): Promise<ContractTransaction> {
+  async updateCash(poolId: string, cash: BigNumberish[]): Promise<ContractTransaction> {
     return this.instance.updateCash(poolId, cash);
   }
 
-  async updateManaged(poolId: string, managed: BigNumber[]): Promise<ContractTransaction> {
-    return this.instance.updateManaged(poolId, managed);
+  async updateManaged(poolId: string, managedl: BigNumberish[]): Promise<ContractTransaction> {
+    return this.instance.updateManaged(poolId, managedl);
```
jubeira, line 81:
```suggestion
  async updateManaged(poolId: string, managed: BigNumberish[]): Promise<ContractTransaction> {
    return this.instance.updateManaged(poolId, managed);
```

jubeira (approved):
I was 30 secs late :skull: 
LGTM though; just a few nits / suggestions.

pvt/helpers/src/models/vault/Vault.ts
EndymionJkb: Good catch! I didn't see that one

PR #2077, opened 2022-11-29
Remove timestamp from deployment task output

TomAFrench:
# Description

We never use this field and it just causes issues where we need to filter it out to use the contract addresses. If we do need to know when a contract was deployed then we can look up its transaction hash and get the block timestamp which is more meaningful.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other

## Checklist:


EndymionJkb:
Is there any value to replacing it with the blocknumber? Of course we can just look either of them up - and we'd be replacing one filter for another - but the blocknumber is at least useful for knowing when to start downstream/dependent fork tests.

I'm thinking this is fine and we should just remember that and look up whatever we need on Etherscan, but thought I'd throw the idea out there.

TomAFrench:
My issue with this is that we can't trust these timestamp/block numbers for any multi-contract deployment ("which contract is this the block number for?").

re fork test blocks: I think deployments are infrequent enough that we're unlikely going to have a fork test depending on many recent deployments such that "go on etherscan and pick a block sometime earlier today" isn't a good metric.

nventuro (approved):
The CI checks mean we don't need to bother checking if you broke any of the output files. Lovely.

PR #2080, opened 2022-11-30
Add version to base pools and refactor pool constructors

### LINKED ISSUE #2026: Add Etherscan-visible versioning to pools

jubeira:
# Description

This PR moves the version to the base pool layer (both `BasePool` and `NewBasePool`), and refactors the constructors for every pool to reduce the stack pressure and arrange the arguments in a way that makes more sense and is more consistent.

The idea behind the change is the following:
- We are passing around many parameters that are always the same regardless the pool type. Those are bundled in this PR as `BasePoolParams`, and every pool has them. There are two benefits here: as a user you know that you'll need at least those in any pool, and also the parameters are routed all the way down to the base layer directly. The version is actually one of those.
- Besides the base params, every pool type has its own required params. Those are generally bundled in `NewPoolParams` structs, which are basically the current ones in master when extracting the base ones.
  - One exception to this rule is `LinearPool`, since every linear pool type has its own nuances and generalizing the 'non-base' params doesn't add up that much.
- I've also bundled the pool registration params. I was on the fence with this one, but in the end this also reduces stack pressure, and the structs are only bundled / unbundled in the constructors, so it shouldn't add much overhead.

This PR is expected to compile, but tests have not been refactored yet.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #2026, supersedes #2075.
Alternative to #2078.

### DIFF
--- a/pkg/interfaces/contracts/pool-utils/IBasePoolFactory.sol
+++ b/pkg/interfaces/contracts/pool-utils/IBasePoolFactory.sol
@@ -15,9 +15,11 @@
 pragma solidity >=0.7.0 <0.9.0;
 pragma experimental ABIEncoderV2;
 
+import "./IFactoryCreatedPoolVersion.sol";
+import "./IVersion.sol";
 import "../solidity-utils/helpers/IAuthentication.sol";
 
-interface IBasePoolFactory is IAuthentication {
+interface IBasePoolFactory is IAuthentication, IVersion, IFactoryCreatedPoolVersion {
     /**
      * @dev Returns true if `pool` was created by this factory.
      */
--- a/pkg/interfaces/contracts/vault/IBasePool.sol
+++ b/pkg/interfaces/contracts/vault/IBasePool.sol
@@ -24,6 +24,17 @@ import "./IPoolSwapStructs.sol";
  * either IGeneralPool or IMinimalSwapInfoPool
  */
 interface IBasePool is IPoolSwapStructs {
+
+    struct BasePoolParams {
+        IVault vault;
+        string name;
+        string symbol;
+        uint256 pauseWindowDuration;
+        uint256 bufferPeriodDuration;
+        address owner;
+        string version;
+    }
+
     /**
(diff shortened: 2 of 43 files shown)


nventuro (commented):
This is definitely an interesting approach, and we may end up choosing this direction or a similar one in the future. However, as-is this changeset is exceedingly large (and doesn't include the corresponding changes in the tests and helpers, which are far from trivial).

I'd go with https://github.com/balancer-labs/balancer-v2-monorepo/pull/2078 for now to address the immediate situation, and then consider a route like this one in the future.

jubeira:
Sounds good. I'll reopen when the time comes.

Note that the large changeset is hard to avoid: the moment you touch base pools, half of the codebase goes upside down.

PR #2076, opened 2022-11-29
Deploy timelock authorizer

nventuro:
# Description

We're only deploying to Goerli and Mainnet for now.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Documentation or wording changes
- [ ] Other

## Checklist:

- [x] The diff is legible and has no extraneous changes
- [x] Complex code has been commented, including external interfaces
- [x] Tests are included for all code paths
- [x] The base branch is either `master`, or there's a description of how to merge

PR #2082, opened 2022-11-30
Redeploy TimelockAuthorizer

nventuro:
This replaces the deployment from #2076, with small changes to the inputs. I removed the delays related to adding gauges, as well as granting permission to add them.


pkg/deployments/deployment-txs/goerli.json
```diff
@@ -74,6 +74,8 @@
   "0x8aB784368A1883DA90D8513b48801e2Db1cb2D5D": "0xe908907a76ca9187729a78ef2814cee5481819a914340db4882522386c7890b9",
   "0x994086630773dC6cB54D3A5E0Ef0963532789E75": "0xb90d6cea041058605a596a7b988f3b0f08cfb457fb332e4cc53708f336ed86bc",
   "0x7C71313d3b84728cA0FD4cDaD8c86067E3548dCe": "0xc653da0876b1d35f922b211f07f092a3a9fda6d0547bfb729c6c9ec49055942d",
```
TomAFrench, line 76:
```suggestion
```

ditto

TomAFrench (approved):
Build info is the same with only changes to deployment args. CI checks this for us so this is good to go.

PR #2083, opened 2022-11-30
Deploy: composable v2

jubeira:
Outputs and action IDs.

Other changes:
- Add BAL to BSC in mock task.
- Tokens need to be sorted for the pool deployment to work (case insensitive; the checksum of the address still needs to be valid).


nventuro (approved):
We can merge as soon as CI passes, since it will check all addresses and IDs.

EndymionJkb (approved):
Checked that all contracts are verified on all chains

PR #2091, opened 2022-12-01
Remove usage of hardhat in deployments package to create a contract object

TomAFrench:
# Description

I'm not sure why we were using the ethers object provided by hardhat here. I've changed it to use vanilla ethers and changed the peer dependency to just be for `ethersproject/contracts`. I've set it to be anything from v5 to be permissive to whatever the user has installed.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other

## Checklist:

### DIFF
--- a/pkg/deployments/index.ts
+++ b/pkg/deployments/index.ts
@@ -1,4 +1,4 @@
-import { Contract } from 'ethers';
+import { Contract } from '@ethersproject/contracts';
 import { Artifact } from 'hardhat/types';
 
 /**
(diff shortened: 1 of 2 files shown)


nventuro (commented):
Doesn't this mean that hardhat should no longer be a peer dep then?

TomAFrench:
We still expose hardhat's types so I thought it makes sense to keep that. As it's a peerdep the user could just not install it however.

nventuro (approved):

PR #2093, opened 2022-12-01
Improve deployment address checking

nventuro:
# Description

This is a very simple way to add checks for contracts that are deployed by other contracts, either via `create` or `create2`, and therefore not possible to check via our standard mechanism (which expects contracts to be deployed by EOAs).

Since these contracts' addresses are stored via `save()`, we can introduce a check there to compare what is being saved to what is stored as the task's output.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A


pkg/deployments/src/task.ts
```diff
@@ -286,14 +286,39 @@ export default class Task {
   }
 
   save(output: RawOutput): void {
+    const parsedOutput = this._parseRawOutput(output);
+
+    if (this.mode === TaskMode.CHECK) {
+      // `save` is only called when by `deploy` (which only happens in LIVE mode), or manually for contracts that are
```
EndymionJkb, line 292:
```suggestion
      // `save` is called either by `deploy` (which only happens in LIVE mode), or manually for contracts that are
```

pkg/deployments/src/task.ts
```diff
@@ -286,14 +286,39 @@ export default class Task {
...
+      // deployed by other contracts (e.g. Batch Relayer Entrypoints). Therefore, by testing for CHECK mode we can
+      // indentify this second type of contracts, and check them by comparing the saved address to the address that the
+      // task would attempt to save.
```
EndymionJkb, line 295:
```suggestion
      // identify this second type of contracts, and check them by comparing the saved address to the address that the
      // task is attempting to save.
```

pkg/deployments/src/task.ts
```diff
@@ -286,14 +286,39 @@ export default class Task {
   }
 
   save(output: RawOutput): void {
```
TomAFrench, line 288: I'd break this up into two separate private functions which we select between using the taskmode.

```
 save(output: RawOutput): void {
   if (this.mode === TaskMode.LIVE) {
     this._save(output);
   } else if (this.mode === TaskMode.CHECK) {
     this._smortCheck(output);
   }
```

where `_save()` is the previous function and `_smortCheck()` is the new code.

pkg/deployments/src/task.ts
TomAFrench, line 295:
```suggestion
      // `save` is only called by `deploy` (which only happens in LIVE mode), or manually for contracts that are
      // deployed by other contracts (e.g. Batch Relayer Entrypoints). Therefore by testing for CHECK mode we can
      // identify this second type of contracts, and check them by comparing the saved address to the address that the
      // task would attempt to save.
```

This should also get moved up into the dispatcher function described above.

jubeira (commented):
Cool!
I agree with Tom's proposal about splitting into private functions :+1:

pkg/deployments/src/task.ts
```diff
@@ -286,14 +286,39 @@ export default class Task {
...
+    const parsedOutput = this._parseRawOutput(output);
+
+    if (this.mode === TaskMode.CHECK) {
+      // `save` is only called when by `deploy` (which only happens in LIVE mode), or manually for contracts that are
+      // deployed by other contracts (e.g. Batch Relayer Entrypoints). Therefore, by testing for CHECK mode we can
+      // indentify this second type of contracts, and check them by comparing the saved address to the address that the
+      // task would attempt to save.
+
+      for (let name of Object.keys(parsedOutput)) {
```
TomAFrench, line 297:
```suggestion
      for (const name of Object.keys(parsedOutput)) {
```

TomAFrench (approved):

PR #2098, opened 2022-12-01
Add totalSupply to VotingEscrowDelegationProxy

TomAFrench:
# Description

On L2s we're going to need some way of getting the total veBAL supply. On L1 the gauge can go directly to the veBAL contract but on L2 we may want to swap out the oracle contract entirely so we need to be able to update this after the fact. My solution to this is that we allow querying the veBAL total supply on the veBoost proxy, which then forwards on the call to the implementation (which would then query from the oracle).

We can then switch out the oracle by updating the veBoost implementation to a new one which will query the balance and total supply from the new oracle contract. This will clear out any existing boosts/delegations which is desirable as it's likely the oracle was reporting incorrect data in this scenario.

We will also need to deploy a "null implementation" of veBAL on each L2 which will report a zero total supply and balance for every user. The veBoost proxy can then fall back to this if we decide we want to remove the oracle without a replacement.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/liquidity-mining/contracts/VotingEscrowDelegationProxy.sol
+++ b/pkg/liquidity-mining/contracts/VotingEscrowDelegationProxy.sol
@@ -67,6 +67,14 @@ contract VotingEscrowDelegationProxy is SingletonAuthentication {
         return _adjustedBalanceOf(user);
     }
 
+    /**
+     * @notice Get the current veBAL total supply from the votingEscrow contract.
+     * @return The current veBAL total supply.
+     */
+    function totalSupply() external view returns (uint256) {
+        return _votingEscrow.totalSupply();
+    }
+
     // Internal functions
 
     function _adjustedBalanceOf(address user) internal view returns (uint256) {
@@ -75,6 +83,14 @@ contract VotingEscrowDelegationProxy is SingletonAuthentication {
             return IERC20(_votingEscrow).balanceOf(user);
         }
         return implementation.adjusted_balance_of(user);
+    }
+
+    function _totalSupply() internal view returns (uint256) {
+        IVeDelegation implementation = _delegation;
+        if (implementation == IVeDelegation(0)) {
+            return IERC20(_votingEscrow).totalSupply();
+        }
+        return implementation.totalSupply();
     }
 
     // Admin functions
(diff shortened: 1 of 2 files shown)


pkg/liquidity-mining/contracts/VotingEscrowDelegationProxy.sol
```diff
@@ -67,6 +67,14 @@ contract VotingEscrowDelegationProxy is SingletonAuthentication {
         return _adjustedBalanceOf(user);
     }
 
+    /**
+     * @notice Get the current veBAL total supply from the votingEscrow contract.
+     * @return The current veBAL total supply.
+     */
+    function totalSupply() external view returns (uint256) {
+        return _votingEscrow.totalSupply();
+    }
```
nventuro, line 76: This seems wrong, shouldn't this call `_totalSupply()`?

nventuro (commented):
The goal of this would then to not have the gauges depend on VotingEscrow, right?

pkg/liquidity-mining/contracts/VotingEscrowDelegationProxy.sol
TomAFrench: Whoops, yes that's right.

TomAFrench:
Yes, the goal of this is that on L2s we can have `_votingEscrow` return zero values for both balances and total supply such that we "turn off" boosts globally. If we have an oracle contract (which fulfils the role of veBoost) then it should report the total supply appropriately.

We can't read total supply directly from `_votingEscrow` as this would lock us in to a particular oracle reporting the total supply.

pkg/liquidity-mining/contracts/VotingEscrowDelegationProxy.sol
```diff
@@ -67,6 +67,14 @@ contract VotingEscrowDelegationProxy is SingletonAuthentication {
         return _adjustedBalanceOf(user);
     }
 
+    /**
+     * @notice Get the current veBAL total supply from the votingEscrow contract.
+     * @return The current veBAL total supply.
+     */
+    function totalSupply() external view returns (uint256) {
+        return _totalSupply();
```
nventuro, line 75: I'd just inline this, no need to make things extra complicated.

nventuro (approved):

PR #2099, opened 2022-12-02
Refactor WeightedPool test helpers

### LINKED ISSUE #2088: Refactor deployment/test helpers
As noted in the description of #2078 , the test helpers - WeightedPoolDeployer in particular - have, with the proliferation of pool types,  grown into "unholy abominations" that are doing too much. We have moved the solidity codebase from generic base contract with lots of special casing toward encapsulating pool-specific functionality within each contract, and could do much the same here.

EndymionJkb:
# Description

Get rid of WeightedPoolDeployer which, compared to how we were previously upset about some special casing in the code (e.g., using proportional exits pre-recovery mode vs pausing), is a whole other level of travesty.

All WeightedPools shared a single TS model, which was ok-ish in the beginning, but as the number and complexity of the pools has grown, it's becoming increasingly untenable to read and maintain.

Address this by splitting WeightedPool.ts into smaller models specific to each contract, and moving each deployment inside them. It does use inheritance to avoid massive duplication (e.g., there is a BaseWeightedPool model for all the join/exit/swap helpers), and could probably be refactored further, but this does most of the heavy lifting.

The diff of WeightedPool is mostly deletions, as it's going from "kitchen sink" to "just WeightedPool" - and the shared join/exit/swap helpers are in a new base class. The tests are unchanged, except to remove all the "pool type" arguments, and use the specific model instead of just WeightedPool.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

- [~] The diff is legible and has no extraneous changes

## Issue Resolution

Closes #2088


pkg/pool-weighted/test/ManagedPool.test.ts
```diff
@@ -68,18 +67,18 @@ describe('ManagedPool', function () {
...
-  async function deployPool(overrides: RawWeightedPoolDeployment = {}): Promise<WeightedPool> {
+  async function deployPool(overrides: RawManagedPoolDeployment = {}): Promise<ManagedPool> {
     const params = {
       vault,
       tokens: poolTokens,
       weights: poolWeights,
       owner: owner.address,
       aumFeeId: ProtocolFee.AUM,
-      poolType: WeightedPoolType.MOCK_MANAGED_POOL,
+      mockContractName: 'MockManagedPool',
```
jubeira, line 77: We moved away from the contract name in favor of an enum in https://github.com/balancer-labs/balancer-v2-monorepo/pull/2078.

This is not a very strong opinion, but I think the enum is a bit more accurate to select mock contracts (the comment applies here and wherever `mockContractName` is used). 

Given that the pool type is the object that we are creating, perhaps `poolType` is no longer the best name to select which contract to create.

pvt/helpers/src/models/pools/weighted/types.ts
```diff
@@ -8,14 +8,6 @@ import TokenList from '../../tokens/TokenList';
 import { Account, NAry } from '../../types/types';
 import Vault from '../../vault/Vault';
 
-export enum WeightedPoolType {
```
jubeira, line 11: given that we'll only use this to select different kinds of managed pools (considering mock settings as one of them), maybe the enum can be recycled as `ManagedPoolType` and just be used for the managed pool deployment (by default, `MANAGED_POOL`).

pvt/helpers/src/models/types/TypesConverter.ts
```diff
@@ -62,12 +65,68 @@ export default {
...
       pauseWindowDuration,
       bufferPeriodDuration,
+    } = params;
+    if (!params.owner) params.owner = ZERO_ADDRESS;
+    if (!tokens) tokens = new TokenList();
+    if (!weights) weights = Array(tokens.length).fill(fp(1));
+    weights = toNormalizedWeights(weights.map(bn));
+    if (!swapFeePercentage) swapFeePercentage = bn(1e16);
+    if (!pauseWindowDuration) pauseWindowDuration = 3 * MONTH;
+    if (!bufferPeriodDuration) bufferPeriodDuration = MONTH;
```
jubeira, line 75: I'd consider moving these defaults to consts at the top of the file, given that we now use the same values for different pool types. If that ever changes, we might want to change the value for all the pools at once.

jubeira (commented):
I've done a quick review skimming through the files; great job!
I've left a few small comments; overall I like where this is going.

pkg/pool-weighted/test/ManagedPool.test.ts
EndymionJkb: The downstream code is directly using that name in the `deploy` call, and it's not always the same. There's another mock for the settings, which gets passed elsewhere. This let me get rid of the enum completely. I thought it was ok to pass the string in this case, since it was semantic and directly used: literally the name of the mock contract to deploy.

TomAFrench: I'd agree that changing this to a string enum would be helpful.

TomAFrench (approved):
This all looks good to me. Will defer to @EndymionJkb on when you think would be the best time to merge.

pvt/helpers/src/models/pools/weighted/BaseWeightedPool.ts
```diff
@@ -0,0 +1,484 @@
...
+    const pool = this.instance.connect(from);
+    return pool.setSwapEnabled(swapEnabled);
+  }
+
+  async setSwapFeePercentage(from: SignerWithAddress, swapFeePercentage: BigNumberish): Promise<ContractTransaction> {
+    const pool = this.instance.connect(from);
+    return pool.setSwapFeePercentage(swapFeePercentage);
+  }
+
+  async updateWeightsGradually(
```
jubeira, line 469: Isn't this for LBPs / managed pools only?

pvt/helpers/src/models/pools/weighted/BaseWeightedPool.ts
```diff
@@ -0,0 +1,484 @@
...
+    const pool = this.instance.connect(from);
+    return pool.setJoinExitEnabled(joinExitEnabled);
+  }
+
+  async setSwapEnabled(from: SignerWithAddress, swapEnabled: boolean): Promise<ContractTransaction> {
+    const pool = this.instance.connect(from);
+    return pool.setSwapEnabled(swapEnabled);
+  }
+
+  async setSwapFeePercentage(from: SignerWithAddress, swapFeePercentage: BigNumberish): Promise<ContractTransaction> {
```
jubeira, line 464: This is not for managed pools right? Should this be moved up one level..?

pvt/helpers/src/models/pools/weighted/BaseWeightedPool.ts
```diff
@@ -0,0 +1,484 @@
...
+    startTime: BigNumberish,
+    endTime: BigNumberish,
+    endWeights: BigNumberish[]
+  ): Promise<ContractTransaction> {
+    const pool = this.instance.connect(from);
+
+    return await pool.updateWeightsGradually(startTime, endTime, endWeights);
+  }
+
+  async getGradualWeightUpdateParams(from?: SignerWithAddress): Promise<GradualWeightUpdateParams> {
```
jubeira, line 480: same as with `updateWeightsGradually`

pvt/helpers/src/models/pools/weighted/types.ts
```diff
@@ -11,12 +11,69 @@ import Vault from '../../vault/Vault';
...
-  MANAGED_POOL,
-  MOCK_MANAGED_POOL,
-  MOCK_MANAGED_POOL_SETTINGS,
+}
+
+export enum ManagedPoolType {
+  MANAGED_POOL = 'ManagedPool',
+  MOCK_MANAGED_POOL = 'MockManagedPool',
+  MOCK_MANAGED_POOL_SETTINGS = 'MockManagedPoolSettings',
 }
```
jubeira, line 20: Ah, nice move here using the contract name as the enum value :+1: 

Maybe I'd add a comment that this is on purpose to make deployments easier in the helpers, since these names are used directly to fetch the artifacts.

pvt/helpers/src/models/types/TypesConverter.ts
```diff
@@ -58,47 +69,106 @@ export default {
...
-    if (undefined == aumFeeId) aumFeeId = ProtocolFee.AUM;
+    if (!poolType) poolType = ManagedPoolType.MANAGED_POOL;
+    if (swapEnabledOnStart == undefined) swapEnabledOnStart = true;
     if (undefined == swapEnabledOnStart) swapEnabledOnStart = true;
     if (undefined == mustAllowlistLPs) mustAllowlistLPs = false;
     if (undefined == managementAumFeePercentage) managementAumFeePercentage = FP_ZERO;
     if (undefined == factoryVersion) factoryVersion = 'default factory version';
     if (undefined == poolVersion) poolVersion = 'default pool version';
+    if (undefined == aumFeeId) aumFeeId = ProtocolFee.AUM;
+    if (!pauseWindowDuration) pauseWindowDuration = 9 * MONTH;
```
jubeira, line 153: Is 9 months the default for managed pools? I can't find that before the refactor.

jubeira (commented):
Whoa this is a huge undertaking! 
Great job overall, I think this makes a lot of sense.

I've left a bunch of questions and observations, but other than that I think there's no point in keeping this open any longer. I'd approve and merge in the next pass.

pkg/standalone-utils/test/ProtocolFeeSplitter.test.ts
```diff
@@ -47,7 +47,7 @@ describe('ProtocolFeeSplitter', function () {
   });
 
   sharedBeforeEach('create and initialize pools', async () => {
-    pool = await WeightedPool.create({ vault, tokens, owner });
+    pool = await WeightedPool.create({ vault, tokens, owner: owner.address });
```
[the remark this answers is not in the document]
EndymionJkb: Weird; it already is an "Account" type, but owner by itself doesn't seem to work.

pvt/helpers/src/models/pools/weighted/ManagedPool.ts
```diff
@@ -0,0 +1,426 @@
...
+      canChangeMgmtFees: true,
+      canDisableJoinExit: true,
+    };
+
+    const tx = await controlledFactory
+      .connect(from || ZERO_ADDRESS)
+      .create(newPoolParams, basePoolRights, managedPoolRights, DAY, from?.address || ZERO_ADDRESS);
+    const receipt = await tx.wait();
+    const event = expectEvent.inReceipt(receipt, 'ManagedPoolCreated');
+    return deployedAt('v2-pool-weighted/ManagedPool', event.args.pool);*/
```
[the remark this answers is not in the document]
EndymionJkb: Yes; just leftover from the last big refactor after all the Managed Pool arguments changed.

pvt/helpers/src/models/pools/weighted/WeightedPool.ts
```diff
@@ -1,724 +1,105 @@
...
-  }
-
-  get normalizedWeights(): BigNumberish[] {
-    return this.weights;
-  }
-
-  get maxWeightIndex(): BigNumberish {
-    const maxIdx = this.weights.indexOf(this.maxWeight);
-    return bn(maxIdx);
-  }
```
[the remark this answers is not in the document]
EndymionJkb: Yes; these was used back in the days when we paid protocol fees in tokens, and we needed to know the index of the highest weighted token (to pay the fees in)

pvt/helpers/src/models/types/TypesConverter.ts
EndymionJkb: That's what I deployed it with. Yes; we wanted to make the pause window longer for Managed, since they're new, and without a controller, there won't be "production" pools for a while: we don't the pause window to expire before any pools are created.

pvt/helpers/src/models/pools/weighted/BaseWeightedPool.ts
EndymionJkb: There is still this intermediate layer for shared code. I wanted to strictly separate the deployment code, since those are all truly independent, but I think it's ok to share code like this, vs duplicating it in each model.

jubeira (approved):
:rocket: let's go!

PR #2101, opened 2022-12-02
remove signers.ts dependency on private v2-helpers/src/numbers

gerrrg:
# Description

signers.ts in the v2-deployments package depends on v2-helpers which is a package private to the monorepo. Removing this dependency and directly declaring `balance` as a BigNumber should alleviate this issue.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/deployments/src/signers.ts
+++ b/pkg/deployments/src/signers.ts
@@ -2,7 +2,6 @@ import { BigNumber } from 'ethers';
 import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/dist/src/signer-with-address';
 
 import { impersonateAccount, setBalance as setAccountBalance } from '@nomicfoundation/hardhat-network-helpers';
-import { fp } from '@balancer-labs/v2-helpers/src/numbers';
 
 export async function getSigners(): Promise<SignerWithAddress[]> {
   const { ethers } = await import('hardhat');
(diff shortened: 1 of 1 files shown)


TomAFrench:
@gerrrg Are you trying to use `signers.ts` in your repo? The `src` directory shouldn't really be included in the NPM package to my knowledge (I've been planning on removing it) and so shouldn't really be depended on. Or is this causing issues even without importing `signers.ts`?

gerrrg:
@TomAFrench I'm using contracts.ts which Nico pointed me to, and it uses signers.ts. Perhaps pointing me to that was intended to be an example snippet instead of "use this"? 

Either way, it's incredibly useful, so if it's removed, is it at least going to be put somewhere else? It kind of makes sense to me that the deployments repo would have handy tools to deploy the contracts

TomAFrench:
Yep, I just asked him and he meant as an example rather than suggesting you use it directly.

The issue with using the code in `src` is that we don't currently consider api stability at all so there's potential for many unannounced breaking changes if you import it. Once you have the artifacts though you should be able to deploy using just ethers or you could copy the `deploy` function into your codebase.

PR #2114, opened 2022-12-05
Deploy L2 Gauge Checkpointer

### LINKED ISSUE #1289: Build automation for L2 gauge checkpointing and bridging
This isn't much more than a contract that goes over all gauges and checkpoints them. Since there's no registry of gauges anywhere, we'd need to keep one in this checkpointer, which would then be permissionlessly updated (since we can check that the gauges are in the controller for the correct gauge type). There's no need for any permissions of any kind as checkpointing and bridging does not imply any decisions.

Keep in mind that Aribitrum bridging requires ETH to be sent. The AuthAdaptor unfortunately does not return unspent ETH, so if we wanted to could try to check beforehand if the gauge will actually mint tokens and skip it if not.

jubeira:
# Description

L2 Gauge checkpointer deployment. Verification went well.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other

## Checklist:

- [ ] Complex code has been commented, including external interfaces -- N/A
- [ ] Tests are included for all code paths -- N/A

## Issue Resolution

Closes #1289.


pkg/deployments/action-ids/mainnet/action-ids.json
```diff
@@ -968,5 +968,15 @@
...
+  },
+  "20221205-l2-gauge-checkpointer": {
+    "L2GaugeCheckpointer": {
+      "useAdaptor": true,
+      "actionIds": {
+        "addGauges(uint8,address[])": "0x843755caf8023fd0387ada7fcd6646d4eb56a0c4328a5000c3ed3262fb4494c7",
+        "checkpointGaugesAboveRelativeWeight(uint256)": "0xa5c8ab0718d9225bc79e7638a58f65954aa5ddd3201793bcd02dd8b54675dd41",
+        "removeGauges(uint8,address[])": "0x7ea6c79a37cf05e2a14a05a56412077358ecc45161015ad2be70f226012213cd"
+      }
+    }
```
nventuro, line 980: You can remove these, since this contract does not use the authorizer.

nventuro (approved):
Deploy to goerli as well!

PR #2115, opened 2022-12-05
Add protocol id disambiguator to AaveLinearPoolFactory

### LINKED ISSUE #2104: Linear Pool differentiation
Multiple protocols use aave linear pools (e.g. aave, beefy, sturdy), and we need a way to tell those pools apart. The proposed solution involves a set of valid Protocol Ids (probably maintained by a multisig account, granted permission through governance), which each factory would store, and emit whenever a pool was created using it.

EndymionJkb:
# Description

Multiple protocols use the AaveLinearPoolFactory to deploy pools (e.g., Beefy, Sturdy). Add a protocol identifier to protocol creation, and an event that emits it, so that off-chain tools can easily find AaveLinearPools associated with a given protocol.

For further convenience and transparency, there is a permissioned function to register protocol ids with associated names. (We do not require that a protocol id be registered in order to create a pool with it.)

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #2104


pkg/pool-linear/contracts/aave/AaveLinearPoolFactory.sol
```diff
@@ -108,21 +139,20 @@ contract AaveLinearPoolFactory is
...
+        args.symbol = symbol;
+        args.mainToken = mainToken;
+        args.wrappedToken = wrappedToken;
+        args.assetManager = expectedRebalancerAddress;
+        args.upperTarget = upperTarget;
+        args.swapFeePercentage = swapFeePercentage;
+        args.pauseWindowDuration = pauseWindowDuration;
+        args.bufferPeriodDuration = bufferPeriodDuration;
+        args.owner = owner;
+        args.version = getPoolVersion();
```
jubeira, line 155: One thing I like about initializing the struct inline is that it forces you to provide all the required fields. It is a bit more verbose / redundant, but a bit less error prone.

Is there a particular reason for this change?

pkg/pool-linear/contracts/aave/AaveLinearPoolFactory.sol
```diff
@@ -135,7 +165,24 @@ contract AaveLinearPoolFactory is
...
+     * @dev This is a permissioned function. Protocol ids cannot be deregistered.
+     */
+    function registerProtocolId(uint256 protocolId, string memory name) external authenticate {
+        require(!_protocolIds[protocolId].registered, "Protocol already registered");
+
+        _registerProtocolId(protocolId, name);
+    }
+
+    function _registerProtocolId(uint256 protocolId, string memory name) private {
+        _protocolIds[protocolId] = ProtocolIdData({ name: name, registered: true });
```
jubeira, line 186: do we need the `registered` member if protocol IDs cannot be deregistered?

pkg/pool-linear/contracts/aave/AaveLinearPoolFactory.sol
```diff
@@ -135,7 +165,24 @@ contract AaveLinearPoolFactory is
         address actualRebalancerAddress = Create2.deploy(0, rebalancerSalt, rebalancerCreationCode);
         require(expectedRebalancerAddress == actualRebalancerAddress, "Rebalancer deployment failed");
 
+        // Identify the protocolId associated with this pool. We do not require that the protocolId be registered.
+        emit AaveLinearPoolCreated(address(pool), protocolId);
```
jubeira, line 169: for my own understanding: can a pool be created with an incorrect / unregistered protocol ID? Is that expected?

pkg/pool-linear/contracts/aave/AaveLinearPoolFactory.sol
```diff
@@ -57,14 +70,31 @@ contract AaveLinearPoolFactory is
...
+     */
     function getPoolVersion() public view override returns (string memory) {
         return _poolVersion;
     }
 
+    /**
+     * @dev Return the name associated with the given protocolId, if registered.
+     */
+    function getProtocolName(uint256 protocolId) external view returns (string memory) {
+        ProtocolIdData storage protocolIdData = _protocolIds[protocolId];
```
jubeira, line 91: why not `memory`?

pkg/pool-linear/test/AaveLinearPoolFactory.test.ts
```diff
@@ -202,4 +219,41 @@ describe('AaveLinearPoolFactory', function () {
...
+
+      it('should register protocols', async () => {
+        expect(await factory.getProtocolName(AAVE_PROTOCOL_ID)).to.equal(AAVE_PROTOCOL_NAME);
+        expect(await factory.getProtocolName(BEEFY_PROTOCOL_ID)).to.equal(BEEFY_PROTOCOL_NAME);
+        expect(await factory.getProtocolName(STURDY_PROTOCOL_ID)).to.equal(STURDY_PROTOCOL_NAME);
+      });
+
+      it('should fail when a protocol is already registered', async () => {
+        await expect(
+          factory.connect(admin).registerProtocolId(STURDY_PROTOCOL_ID, 'Rando protocol')
```
jubeira, line 254:
```suggestion
          factory.connect(admin).registerProtocolId(STURDY_PROTOCOL_ID, 'Random protocol')
```

jubeira (commented):
Looks pretty good!
I've left a few comments / questions for my own understanding.

pkg/pool-linear/contracts/aave/AaveLinearPoolFactory.sol
EndymionJkb: Yes - got a stack-too-deep as soon as I added the new parameter. This solved that (and is also more memory efficient)

jubeira: ugh :sad: we are on the edge with the stack pressure. Then I guess it's fine!

pkg/pool-linear/contracts/aave/AaveLinearPoolFactory.sol
EndymionJkb: Yes; my interpretation of the conversation around this was that as long as a protocol (e.g., Beefy) can create pools with the correct id - so that all legit Beefy pools had that id - they weren't concerned about "false positives" (random non-Beefy pools using the same id).

It's all for the UIs: it's just emitting an event, not storing or validating anything. Maybe the "real" protocol IDs should be odd numbers like 42, 327, 8143 rather than 0,1,2, for that reason. Since "non-participating" protocols would probably just pass 0, we wouldn't want 0 to mean "AAVE"

pkg/pool-linear/contracts/aave/AaveLinearPoolFactory.sol
EndymionJkb: It could be memory, but we're not changing anything, so this should avoid a copy. (It already has to copy the string out.)

pkg/pool-linear/test/AaveLinearPoolFactory.test.ts
EndymionJkb: I actually did intend "rando" - it means random, but also has the connotation of being suspicious / potentially dangerous. Maybe that's too much slang :) I can see how people might think it's a misspelling of random.

pkg/pool-linear/contracts/aave/AaveLinearPoolFactory.sol
TomAFrench: We could read the stored value, hash it, and check against the hash of the empty string afaik. This isn't gas sensitive though so better to go for clarity.

pkg/pool-linear/contracts/aave/AaveLinearPoolFactory.sol
nventuro: It doesn't really matter, but I'd use `memory` here. We don't care about gas, and you typically see storage pointers when there's writes involved.

pkg/pool-linear/contracts/aave/AaveLinearPoolFactory.sol
```diff
@@ -57,14 +70,31 @@ contract AaveLinearPoolFactory is
...
+
+        require(protocolIdData.registered, "Protocol not registered");
```
nventuro, line 93:
```suggestion
        require(protocolIdData.registered, "Protocol ID not registered");
```

pkg/pool-linear/contracts/aave/AaveLinearPoolFactory.sol
```diff
@@ -108,21 +139,20 @@ contract AaveLinearPoolFactory is
...
-            assetManager: expectedRebalancerAddress,
-            upperTarget: upperTarget,
-            swapFeePercentage: swapFeePercentage,
-            pauseWindowDuration: pauseWindowDuration,
-            bufferPeriodDuration: bufferPeriodDuration,
-            owner: owner,
-            version: getPoolVersion()
-        });
+        AaveLinearPool.ConstructorArgs memory args;
+
```
nventuro, line 143:
```suggestion

        AaveLinearPool.ConstructorArgs memory args;
```

pkg/pool-linear/contracts/aave/AaveLinearPoolFactory.sol
```diff
@@ -135,7 +165,24 @@ contract AaveLinearPoolFactory is
...
         // We don't return the Rebalancer's address, but that can be queried in the Vault by calling `getPoolTokenInfo`.
         return pool;
     }
+
+    /**
+     * @notice Register an id (and name) to differentiate between multiple protocols using this factory.
+     * @dev This is a permissioned function. Protocol ids cannot be deregistered.
+     */
+    function registerProtocolId(uint256 protocolId, string memory name) external authenticate {
+        require(!_protocolIds[protocolId].registered, "Protocol already registered");
```
nventuro, line 180: This should emit an event, it'd otherwise be very hard to keep track of them

pkg/pool-linear/contracts/aave/AaveLinearPoolFactory.sol
nventuro: Correct, we don't care about pools with the wrong ID. It'll just mean those Pools have bad data.

pkg/pool-linear/test/AaveLinearPoolFactory.test.ts
```diff
@@ -202,4 +219,41 @@ describe('AaveLinearPoolFactory', function () {
...
+
+  describe('protocol id', () => {
+    context('with no registered protocols', () => {
+      it('should revert when asking for an unregistered protocol name', async () => {
+        await expect(factory.getProtocolName(AAVE_PROTOCOL_ID)).to.be.revertedWith('Protocol not registered');
+      });
+
+      it('should not allow adding protocols without permission', async () => {
+        await expect(factory.registerProtocolId(AAVE_PROTOCOL_ID, 'AAVE')).to.be.revertedWith('SENDER_NOT_ALLOWED');
+      });
```
nventuro, line 231: This does not belong in this `describe` block

nventuro (approved):

PR #2119, opened 2022-12-07
Configurable pause window

### LINKED ISSUE #2103: Allow alternate pause window periods
Currently the pause buffer is hard-coded to three months, with a one month buffer, and all factories are deployed with these settings.

Some new pool types, such as Managed, are so different from past efforts - and so highly configurable for each partner - that it might take longer for potential issues to arise. We would the ability to use a longer pause period for these factories.

EndymionJkb:
# Description

Currently the pause window is fixed at three months for all pool factories. Yet pool types have gotten markedly more complex than the Weighted Pools we launched with on V2. Newer pools are not only more complicated internally (e.g., the Composable logic used to construct Boosted Pools), but have ever greater and deeper interactions with external contracts: starting with Rate Providers in the original MetaStable pools, through relayers and re-balancers required to support Boosted Pools, and finally to Managed Pools: which have all that, plus sophisticated owner contracts.

This justifies a longer pause window for some pools, so this PR allows factories to pass in custom values for the pause window and buffer period, subject to longer maximums. (There still has to be a maximum to preserve the long-term permissionless nature of Balancer pools.)

Merge into base-pool-factory.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #2103


pkg/pool-utils/contracts/factories/FactoryWidePauseWindow.sol
```diff
@@ -15,25 +15,46 @@
...
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/TemporarilyPausable.sol";
+
 /**
  * @dev Utility to create Pool factories for Pools that use the `TemporarilyPausable` contract.
  *
  * By calling `TemporarilyPausable`'s constructor with the result of `getPauseConfiguration`, all Pools created by this
  * factory will share the same Pause Window end time, after which both old and new Pools will not be pausable.
  */
-contract FactoryWidePauseWindow {
+abstract contract FactoryWidePauseWindow {
```
nventuro, line 28: Why is this `abstract`?

pkg/pool-utils/contracts/test/MockFactoryWidePauseWindow.sol
```diff
@@ -0,0 +1,25 @@
...
+
+import "../factories/FactoryWidePauseWindow.sol";
+
+contract MockFactoryWidePauseWindow is FactoryWidePauseWindow {
+    constructor(uint256 initialPauseWindowDuration, uint256 bufferPeriodDuration)
+        FactoryWidePauseWindow(initialPauseWindowDuration, bufferPeriodDuration)
+    {
+        // solhint-disable-previous-line no-empty-blocks
+    }
+}
```
nventuro, line 25: Since `FactoryWidePauseWindow` doesn't need to be `abstract`, we can delete this.

nventuro (approved):

pkg/pool-utils/contracts/factories/FactoryWidePauseWindow.sol
EndymionJkb: Would this ever be instantiated directly? Just seemed like it should be an abstract base contract. Anyway I can put it back.

PR #2128, opened 2022-12-10
ProtocolFeeSplitter set beneficiary override

### LINKED ISSUE #2127: Pool beneficiary setting when the owner is a contract
The current design allows only the pool owner to designate a beneficiary. This works well for pools where the owner is an EOA or multisig. But if it is a Managed Pool, the (immutable) owner will be a contract. Unless the ability to call this function is built into the owner contract, managers would be unable to set a beneficiary.

EndymionJkb:
# Description

In the current design of ProtocolFeeSplitter, the beneficiary for a pool revenue share can only be set by the pool owner. This is fine for pools where the owner is an EOA or multisig, but could be problematic for pools that are owned by contracts (e.g., Managed Pools). The ability to call this would need to be anticipated in the controller, which might not be the case (and might be considered extraneous).

This adds the ability for governance to set a pool beneficiary in these cases. Of course, this also allows governance to override the owner's choice of beneficiary - but governance can also set the revenue share for a pool (or all pools) to 0, and withdraw protocol fees to anywhere it likes, so it seems this is not much of an escalation.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #2127


TomAFrench (approved):

TomAFrench:
We could merge `setPoolBeneficiary` and `setPoolBeneficiaryOverride` by only checking against the authorizer if the caller doesn't match the pool's owner. Something similar to https://github.com/balancer-labs/balancer-v2-monorepo/blob/c2e338938ae8b9b58d8a7d2204afd2bd4f09dcd8/pkg/pool-utils/contracts/BasePoolAuthorization.sol#L48-L56

PR #2146, opened 2023-01-04
Add solc+vyper compiler binaries to cache

TomAFrench:
# Description

This is a bit of a kludge-y solution to the vyper compiler issue. Rather than downloading it each time, we can just add the compiler binaries to the dependencies cache.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other

## Checklist:

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -11,7 +11,9 @@ runs:
       uses: actions/cache@v2
       id: cache
       with:
-        path: '**/node_modules'
+        path: |
+          **/node_modules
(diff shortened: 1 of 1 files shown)


TomAFrench:
I was thinking this was only going to take effect after the next cache miss but it looks like it never will as the linting job will write a cache entry without the compilers. Closing this as it's gonna be more ugly than I expected in order to get the cache working right.

Issue #2143, opened 2023-01-03
Prepare migration for permissions granted after `TimelockAuthorizer` deployment


jubeira:
Permissions granted after `TimelockAuthorizer` was deployed (https://github.com/balancer-labs/balancer-v2-monorepo/pull/2082) were not migrated automatically, so we should be prepared to re-grant them when the migration is finished.

jubeira:
We'll need to add some permissions that don't exist in the old authorizer as well as part of the migration.

One example is the `addGaugeFactory` permission for the upcoming gauge adder (#1996).

jubeira:
About the deployment task: the scripts should have an explicit list of the permissions being migrated. We can compare this list with the one we get querying the permissions granted to the old authorizer as a double check.

PR #2150, opened 2023-01-04
Timelock authorizer transition migrator

jubeira:
# Description

When the `TimelockAuthorizer` was deployed, existing permissions in the old authorizer were migrated as part of the deployment.

The new authorizer is still not connected to the vault because of the migration period, and some other permissions were granted in the old authorizer during this time.

This PR introduces a script to migrate the remaining permissions from the old authorizer to the new one when the time comes.
Some considerations:
- The permissions to migrate will be stored in the migration contract upon deployment.
- The permissions to migrate must exist in the old authorizer.
- The contract needs to be made a general granter by governance (when `TimelockAuthorizer`'s root migration is finished), and then anyone can trigger the migration.
- The migration can only be done once; after that the script becomes basically useless.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other

## Checklist:

## Issue Resolution

See #2143 (will be solved when this is deployed).

### DIFF
--- a/pkg/governance-scripts/contracts/TimelockAuthorizerTransitionMigrator.sol
+++ b/pkg/governance-scripts/contracts/TimelockAuthorizerTransitionMigrator.sol
@@ -30,49 +30,36 @@ contract TimelockAuthorizerTransitionMigrator {
         address target;
     }
 
-    RoleData[] internal rolesData;
+    RoleData[] private _rolesData;
 
     /**
-     * @dev Reverts if _rolesData contains a role for an account which doesn't hold the same role on the old Authorizer.
+     * @dev Reverts if rolesData contains a role for an account which doesn't hold the same role on the old Authorizer.
      */
     constructor(
-        IBasicAuthorizer _oldAuthorizer,
+        IBasicAuthorizer oldAuthorizer,
         TimelockAuthorizer _timelockAuthorizer,
-        RoleData[] memory _rolesData
+        RoleData[] memory rolesData
     ) {
         timelockAuthorizer = _timelockAuthorizer;
 
-        for (uint256 i = 0; i < _rolesData.length; i++) {
-            RoleData memory roleData = _rolesData[i];
+        for (uint256 i = 0; i < rolesData.length; i++) {
+            RoleData memory roleData = rolesData[i];
             // We require that any permissions being copied from the old Authorizer must exist on the old Authorizer.
             // This simplifies verification of the permissions being added to the new TimelockAuthorizer.
-            require(_oldAuthorizer.canPerform(roleData.role, roleData.grantee, roleData.target), "UNEXPECTED_ROLE");
-            rolesData.push(roleData);
+            require(oldAuthorizer.canPerform(roleData.role, roleData.grantee, roleData.target), "UNEXPECTED_ROLE");
+            _rolesData.push(roleData);
         }
     }
 
     /**
      * @notice Migrates permissions stored at contract creation time.
-     * @dev Permissions can only be granted by TimelockAuthorizer's root.
      */
     function migratePermissions() external {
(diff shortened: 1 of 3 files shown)


pkg/governance-scripts/contracts/TimelockAuthorizerTransitionMigrator.sol
```diff
@@ -0,0 +1,88 @@
...
+        TimelockAuthorizer _timelockAuthorizer,
+        RoleData[] memory _rolesData
+    ) {
+        timelockAuthorizer = _timelockAuthorizer;
+
+        for (uint256 i = 0; i < _rolesData.length; i++) {
+            RoleData memory roleData = _rolesData[i];
+            // We require that any permissions being copied from the old Authorizer must exist on the old Authorizer.
+            // This simplifies verification of the permissions being added to the new TimelockAuthorizer.
+            require(oldAuthorizer.canPerform(roleData.role, roleData.grantee, roleData.target), "UNEXPECTED_ROLE");
```
EndymionJkb, line 53: It means you'd have to store the old authorizer address, but it would be safer to check this in `migratePermissions` - in case there is a delay between deployment and whenever `migratePermissions` is called, and any of the old permissions are revoked. (It wouldn't revert, just skip any that had been revoked, maybe emitting an event.) Granting permissions will already emit an event.

pkg/governance-scripts/contracts/TimelockAuthorizerTransitionMigrator.sol
```diff
@@ -0,0 +1,88 @@
...
+        address target;
+    }
+
+    bool private _migrationDone;
+
+    /**
+     * @dev Reverts if rolesData contains a role for an account which doesn't hold the same role on the old Authorizer.
+     */
+    constructor(
+        IBasicAuthorizer oldAuthorizer,
```
EndymionJkb, line 43: Could even hard-code this. Or the script could pull it in from the Authorizer deployment task

pkg/governance-scripts/contracts/TimelockAuthorizerTransitionMigrator.sol
jubeira: Good point. I've added the check in `migratePermissions` in https://github.com/balancer-labs/balancer-v2-monorepo/pull/2150/commits/615be11bd84e8df3ed57cbaaccb7a15a19d684c6.
I'd keep this require anyways, since it's pointless to store permissions that are not present in the old authorizer at creation time -- it would mean we are doing something wrong during the deployment.

pkg/governance-scripts/contracts/TimelockAuthorizerTransitionMigrator.sol
jubeira: That's right; I'm pulling the address from the task in the deployment script and fork test in #2151.

EndymionJkb (approved):
Looks better now, with some nits

pkg/governance-scripts/test/TimelockAuthorizerTransitionMigrator.test.ts
```diff
@@ -0,0 +1,151 @@
...
+          .connect(root)
+          .manageGranter(
+            newAuthorizer.GENERAL_PERMISSION_SPECIFIER(),
+            transitionMigrator.address,
+            newAuthorizer.EVERYWHERE(),
+            true
+          );
+      });
+
+      it('migrates all roles properly', async () => {
```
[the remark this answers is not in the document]
jubeira: I tried this change but then I stepped back since we already have an outer `describe` for that. The current version, even though more redundant, is also more readable and follows the style of other tests (i.e. performing the operation in every `it` block).

jubeira:
Thanks for the review @EndymionJkb !

PR #1994, opened 2022-11-08
Deployment preparation: L2 Gauge Checkpointer

jubeira:
# Description

L2 Gauge checkpointer V2 deployment preparation.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Documentation or wording changes
- [x] Other

## Checklist:

- [ ] The diff is legible and has no extraneous changes
- [ ] Complex code has been commented, including external interfaces
- [ ] Tests are included for all code paths
- [ ] The base branch is either `master`, or there's a description of how to merge

## Issue Resolution

See #1990

PR #2154, opened 2023-01-06
L2 gauge checkpointer: granular checkpoint function

jubeira:
# Description

This PR adds a second checkpoint function to target a single gauge type. This can be useful e.g. if the total amount of gauges grows too big and gas becomes an issue at some point. This change is not strictly necessary for the next deployment, so we can just dismiss it if it doesn't add value.

Some thoughts:
- We could do a single function for the checkpoint that receives a type, and use 0 as a sentinel value to checkpoint all the types. Not 100% clean, but we'd reuse the lines before and after the inner checkpoint.
- Tests will be included in the fork tests for the next deployment (see #1994)

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:

- [ ] Tests are included for all code paths --> test pending (will be part of fork test)

## Issue Resolution

N/A


pkg/liquidity-mining/contracts/gauges/L2GaugeCheckpointer.sol
```diff
@@ -148,9 +135,28 @@ contract L2GaugeCheckpointer is IL2GaugeCheckpointer, ReentrancyGuard {
...
+        external
+        payable
+        override
+        withSupportedGaugeType(gaugeType)
+        nonReentrant
+    {
+        // solhint-disable-next-line not-rely-on-time
+        uint256 currentPeriod = _roundDownTimestamp(block.timestamp);
+
+        _checkpointGauges(gaugeType, minRelativeWeight, currentPeriod);
```
TomAFrench, line 149: This function makes the assumption that all gauges will have the same ETH cost. I imagine we have a mix of capped and uncapped arbitrum gauges so this assumption may break if we don't keep the factories in sync.

I think we can query this cost from each gauge similar to how we do it in `getTotalBridgeCost`.

pkg/liquidity-mining/contracts/gauges/L2GaugeCheckpointer.sol
```diff
@@ -117,9 +109,7 @@ contract L2GaugeCheckpointer is IL2GaugeCheckpointer, ReentrancyGuard {
         return _gauges[gaugeType].length();
     }
 
-    /**
-     * @dev See `IL2GaugeCheckpointer#getGaugeAt`.
-     */
+    /// @inheritdoc IL2GaugeCheckpointer
     function getGaugeAt(IGaugeAdder.GaugeType gaugeType, uint256 index)
```
TomAFrench, line 113:
```suggestion
    function getGaugeAtIndex(IGaugeAdder.GaugeType gaugeType, uint256 index)
```

This is a bit more standardised with similar functions imo.

pkg/liquidity-mining/contracts/gauges/L2GaugeCheckpointer.sol
TomAFrench, line 149: We can also simplify the zero length check in `_checkpointGauges` while saving gas.
 
```
EnumerableSet.AddressSet storage typeGauges = _gauges[gaugeType];

uint256 totalTypeGauges = typeGauges.length();
if (totalTypeGauges == 0) {
    // Return early if there's no work to be done.
    return;
}
```

TomAFrench (commented):
There's potential for leaking ETH on arbitrum gauges atm. Looks good otherwise.

pkg/liquidity-mining/contracts/gauges/L2GaugeCheckpointer.sol
jubeira: Hmm, nice catch!

jubeira:
Thanks for the comments!
Addressed in https://github.com/balancer-labs/balancer-v2-monorepo/pull/2154/commits/60811f5d7565cadf7e31330907e34497660d0e45; PTAL.

TomAFrench (approved):
LGTM

PR #2158, opened 2023-01-09
Remove concept of "canonical gauges" from gaugeadder

TomAFrench:
# Description

This PR removes the mapping from pool to gauge addresses in the gauge adder. This allows multiple gauges to be added for the same pool (useful for gauge migrations, etc.)

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/interfaces/contracts/liquidity-mining/IGaugeAdder.sol
+++ b/pkg/interfaces/contracts/liquidity-mining/IGaugeAdder.sol
@@ -36,15 +36,6 @@ interface IGaugeAdder is IAuthentication {
     function getGaugeController() external view returns (IGaugeController);
 
     /**
-     * @notice Returns the gauge corresponding to a Balancer pool `pool` on Ethereum mainnet.
-     * Only returns gauges which have been added to the Gauge Controller.
-     * @dev Gauge Factories also implement a `getPoolGauge` function which maps pools to gauges which it has deployed.
-     * This function provides global information by using which gauge has been added to the Gauge Controller
-     * to represent the canonical gauge for a given pool address.
-     */
-    function getPoolGauge(IERC20 pool) external view returns (ILiquidityGauge);
-
-    /**
      * @notice Returns the `index`'th factory for gauge type `gaugeType`
      */
     function getFactoryForGaugeType(GaugeType gaugeType, uint256 index) external view returns (address);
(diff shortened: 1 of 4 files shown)


pkg/interfaces/contracts/liquidity-mining/IGaugeAdder.sol
```diff
@@ -35,15 +35,6 @@ interface IGaugeAdder is IAuthentication {
...
     function getGaugeController() external view returns (IGaugeController);
 
-    /**
-     * @notice Returns the gauge corresponding to a Balancer pool `pool` on Ethereum mainnet.
-     * Only returns gauges which have been added to the Gauge Controller.
-     * @dev Gauge Factories also implement a `getPoolGauge` function which maps pools to gauges which it has deployed.
-     * This function provides global information by using which gauge has been added to the Gauge Controller
-     * to represent the canonical gauge for a given pool address.
-     */
-    function getPoolGauge(IERC20 pool) external view returns (ILiquidityGauge);
```
jubeira, line 45: I think we should update the changelog to reflect this, right?

pkg/liquidity-mining/contracts/admin/GaugeAdder.sol
```diff
@@ -64,23 +57,6 @@ contract GaugeAdder is IGaugeAdder, SingletonAuthentication, ReentrancyGuard {
...
     }
 
-    /**
-     * @notice Returns the gauge corresponding to a Balancer pool `pool` on Ethereum mainnet.
-     * Only returns gauges which have been added to the Gauge Controller.
-     * @dev Gauge Factories also implement a `getPoolGauge` function which maps pools to gauges which it has deployed.
-     * This function provides global information by using which gauge has been added to the Gauge Controller
-     * to represent the canonical gauge for a given pool address.
-     */
-    function getPoolGauge(IERC20 pool) public view override returns (ILiquidityGauge) {
```
jubeira, line 74: Keeping this would be pretty annoying with potentially more than one gauge per pool, and the simplicity this change brings is welcome.

Just to double check: are we completely sure none will need this information easily accessible on chain?

jubeira (approved):
LGTM!

pkg/liquidity-mining/contracts/admin/GaugeAdder.sol
TomAFrench: This information is inaccurate as it is so we don't lose anything by removing it. If we needed an accurate registry in future we'd have to curate it manually.

pkg/interfaces/contracts/liquidity-mining/IGaugeAdder.sol
TomAFrench: Let's add this to #1996

PR #2159, opened 2023-01-09
Gas optimization

swarna1101:


EndymionJkb:
Confirmed through the grants committee that this was indeed intended for another repo. Closing again, but not before taking a screenshot of the diff :)

PR #2174, opened 2023-01-12
Initial version of VotingEscrowRemapper for cross chain boosts

TomAFrench:
# Description

This PR has the initial proposed version of the `VotingEscrowRemapper` contract.

> This contract allows veBAL holders on Ethereum to designate an address on each L2 to receive its balance.
This is intended for smart contract systems where they cannot deploy to the same address on all networks, EOA's are
expected to utilise delegation mechanisms on the L2s themselves.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


nventuro (commented):
Only now got to reviewing this, it looks great! I don't have any suggestions or concerns.

My only question is about the 'potential griefing attacks' that are supposedly enabled by letting two contracts map their L1 veBAL balances to the same L2 address. Why would this be problematic?

nventuro:
@TomAFrench if you could reply to the comment above that'd be very useful. Thanks!

jubeira:
> My only question is about the 'potential griefing attacks' that are supposedly enabled by letting two contracts map their L1 veBAL balances to the same L2 address. Why would this be problematic?

Reading at the code, I understand that the mapping does not add up. 
Suppose a contract `A` with 100 veBAL and a contract `B` with 0.
Contract `A` creates a mapping to address `A*` on the remote chain. If contract `B` could then create a mapping to address `A*`, it could deny `A`'s veBAL balance in the remote chain.

By making the mapping unique, once an address in the local chain (e.g. `A`) maps to another one in the remote chain (`A*`), the address in the remote chain cannot be used by another contract in the local chain (`B`).

In this case, as the comments in the code mention, `B` could frontrun `A` and reserve `A*` in the remote chain before `A` does.

nventuro:
But I see no reason why we couldn't add those up. The same thing could happen without duplicate remapping: contract `A` could map to address `B`, which is an EOA that already holds veBAL. The remapper can't prevent this.

nventuro:
Ah I see, the issue is we intend to have 1-1 mapping of local and remote.

jubeira:
A few notes about the latest changes:
- In the end the remappings are going to be queried on-chain, so reverting when the mapping does not exist makes things more cumbersome.
- Defaulting the remaps to the same address inside the contract creates two mappings from remote to local. E.g.:
```
remap(local, remote, chainId);
getRemote(local, chainId);  --> returns remote
getLocal(remote, chainId);  --> returns local
getLocal(local, chainId);   --> returns local as well
```
- That being said, it is better to just return `address(0)` for non-remapped accounts, which also makes the code simpler. If a mapping does not exist:
```
getRemote(local, chainId);  --> returns `address(0)`; it can be assumed that the address in `chainID` is also `local`
getLocal(remote, chainId);  --> returns `address(0)`, but it cannot be assumed that `remote` in mainnet is not remapped.
```

On the other hand, an external function that forwards the call to `locked__end` in the voting escrow was also added for convenience.

pkg/liquidity-mining/contracts/VotingEscrowRemapper.sol
```diff
@@ -0,0 +1,201 @@
...
+     * @param chainId - The chain id of the remote chain on which `remoteUser` resides.
+     */
+    function setNetworkRemapping(
+        address localUser,
+        address remoteUser,
+        uint256 chainId
+    ) external {
+        _require(msg.sender == localUser || msg.sender == _localRemappingManager[localUser], Errors.SENDER_NOT_ALLOWED);
+        require(_isAllowedContract(localUser), "Only contracts which can hold veBAL can set up a mapping");
+        require(remoteUser != address(0), "Zero address cannot be used as remote user");
```
nventuro, line 135: How can a remapping be cleared then? It can be changed, but not cleared unless the contract stops passing the smart wallet check.

jubeira: The 'clear' feature is to deter griefing attacks.

If an account was misbehaving and mapping to somebody else's address, it can be deny-listed in the wallet checker, and then anyone can remove the mapping.

If the user wants to clear its own rightful mapping they can always do so.

pkg/liquidity-mining/contracts/VotingEscrowRemapper.sol
```diff
@@ -0,0 +1,202 @@
...
+
+import "@balancer-labs/v2-interfaces/contracts/solidity-utils/openzeppelin/IERC20.sol";
+import "@balancer-labs/v2-interfaces/contracts/liquidity-mining/IVotingEscrow.sol";
+
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/SingletonAuthentication.sol";
+
+/**
+ * @notice This contract allows veBAL holders on Ethereum to designate an address on each L2 to receive its balance.
+ * This is intended for smart contract systems where they cannot deploy to the same address on all networks, EOA's are
+ * expected to utilise delegation mechanisms on the L2s themselves.
```
EndymionJkb, line 26:
```suggestion
 * @notice This contract allows veBAL holders on Ethereum to assign their balance to designated addresses on each L2.
 * This is intended for smart contracts that are not deployed to the same address on all networks. EOA's are
 * expected to either use the same address, or manage delegation on L2 networks themselves.
```

pkg/liquidity-mining/contracts/VotingEscrowRemapper.sol
```diff
@@ -0,0 +1,202 @@
...
+ *
+ * @dev For each network (chainId) we maintain a mapping between Ethereum (local) and L2 (remote) addresses.
+ * Queries for a balance of an address on a remote network are then remapped to the linked local address.
+ * A local user can specify their own mapping or a delegate can be set up for them for the situation where the user
+ * cannot call this contract.
```
EndymionJkb, line 31:
```suggestion
 * @dev For each network (chainId), we maintain a mapping between local (Ethereum) and remote (L2) addresses.
 * This contract remaps balance queries on remote network addresses to their corresponding local addresses.
 * Users able to call this contract can set their own mappings, or delegate this function to another account if they cannot.
```

jubeira:
Moving back to draft because of the bridge-when-remap feature.

pkg/liquidity-mining/contracts/VotingEscrowRemapper.sol
```diff
@@ -0,0 +1,286 @@
...
+        require(
+            _remoteToLocalAddressMap[chainId][remoteUser] == address(0),
+            "Cannot overwrite an existing mapping by another user"
+        );
+
+        // TODO: confirm this.
+        // require(
+        //     _localToRemoteAddressMap[chainId][remoteUser] == address(0),
+        //     "Cannot to remap to an address that is in use locally"
+        // );
```
jubeira, line 171: @nventuro I think we should add this restriction.

Whenever a mapping takes place, 
- You can't step somebody else's remote address
- The remote address in the local chain shouldn't be remapped, since it 'belongs' to the local user.

On the other hand, we'll also be adding a check for the veBAL balance for the remote address.
Something like this:
```
require(veBAL.balanceOf(remoteUser) == 0)
```

pkg/liquidity-mining/contracts/VotingEscrowRemapper.sol
```diff
@@ -0,0 +1,286 @@
...
+        // Clear out the old remote user to avoid orphaned entries.
+        address oldRemoteUser = _localToRemoteAddressMap[chainId][localUser];
+        _remoteToLocalAddressMap[chainId][oldRemoteUser] = address(0);
+
+        // Set up new remapping.
+        _remoteToLocalAddressMap[chainId][remoteUser] = localUser;
+        _localToRemoteAddressMap[chainId][localUser] = remoteUser;
+
+        // Note: it is important to perform the bridge calls _after_ the mappings are settled, since the
+        // omni voting escrow will rely on the correct mappings to bridge the balances.
```
jubeira, line 182: This is not implemented yet, but I understand the bridge will query the remapper back to know _where_ to send the balances.

If my reasoning is correct, that should be fine:
- This function is now non-reentrant, so we can't alter the state (i.e. the mappings) again in a weird way.
- Reading the mappings after they are set (i.e. below these lines) should be correct, so the balance bridges below should target the correct addresses.

pkg/liquidity-mining/contracts/VotingEscrowRemapper.sol
```diff
@@ -0,0 +1,306 @@
...
+        if (oldRemoteUser != address(0)) {
+            // If there was an old mapping, send balance from (local) oldRemoteUser --> (remote) oldRemoteUser
+            // This should clean up the existing bridged balance from localUser --> oldRemoteUser.
+            omniVotingEscrow.sendUserBalance{ value: nativeFee }(
+                oldRemoteUser,
+                uint16(chainId),
+                payable(msg.sender),
+                address(0),
+                ""
+            );
```
jubeira, line 215: Note for the reviewer: to my best understanding, this is the right way to call the bridge, but let's double check.

pkg/liquidity-mining/contracts/VotingEscrowRemapper.sol
```diff
@@ -0,0 +1,306 @@
...
+        // Bridge balance for new mapping localUser --> remoteUser.
+        omniVotingEscrow.sendUserBalance{ value: nativeFee }(
+            localUser,
+            uint16(chainId),
+            payable(msg.sender),
+            address(0),
+            ""
+        );
+
+        emit AddressMappingUpdated(localUser, remoteUser, uint16(chainId));
```
nventuro, line 227: Move this before the bridge calls

pkg/liquidity-mining/contracts/VotingEscrowRemapper.sol
```diff
@@ -0,0 +1,306 @@
...
+        // - The remote address must not already be in use by another local user (C --> B is forbidden).
+        // - The remote address must not be a local address that has already been remapped (C --> A is forbidden).
+        // - The local address must not be the target remote address for another local user (B --> C is forbidden).
+        //
+        // Note that this means that it is possible to frontrun this call to grief a user by taking up their
+        // selected remote address before they do so. This is mitigated somewhat by restricting potential attackers to
+        // the set of contracts that are allowlisted to hold veBAL (and their remapping managers). Should
+        // one of them grief, then Balancer governance can remove them from these allowlists.
+
+        // Assuming A --> B, then C cannot be remapped to B.
```
nventuro, line 170:
```suggestion
        // B cannot be remapped to (i.e. be a remote) if a prior A --> B mapping exists.
```

pkg/liquidity-mining/contracts/VotingEscrowRemapper.sol
```diff
@@ -0,0 +1,306 @@
...
+        // To prevent it, we verify that the reverse mapping of B does not exist.
+        require(
+            _remoteToLocalAddressMap[chainId][remoteUser] == address(0),
+            "Cannot overwrite an existing mapping by another user"
+        );
+
+        // Assuming A --> B, then C cannot be remapped to A.
+        // To prevent it, we verify that the mapping of A does not exist.
```
nventuro, line 178:
```suggestion
        // A cannot be remapped to (i.e. be a remote) if a prior A --> B mapping exists.
        // To prevent it, we verify that the mapping of our remote does not exist.
```

pkg/liquidity-mining/contracts/VotingEscrowRemapper.sol
```diff
@@ -0,0 +1,306 @@
...
+        require(
+            _localToRemoteAddressMap[chainId][remoteUser] == address(0),
+            "Cannot remap to an address that is in use locally"
+        );
+
+        // Assuming A --> B, then B cannot be remapped to C.
+        // To prevent it, we verify that the reverse mapping of B does not exist.
```
nventuro, line 185:
```suggestion
        // B cannot be mapped from (i.e. be a local) if a prior A --> B mapping exists.
        // To prevent it, we verify that the reverse mapping of our local does not exist.
```

pkg/liquidity-mining/contracts/VotingEscrowRemapper.sol
nventuro: I found the original phrasing very confusing (spent over 10 minutes re-reading these statements). I think this new way of stating them is simpler.

pkg/liquidity-mining/contracts/VotingEscrowRemapper.sol
```diff
@@ -0,0 +1,306 @@
...
+        // - The remote address must not be a local address that has already been remapped (C --> A is forbidden).
+        // - The local address must not be the target remote address for another local user (B --> C is forbidden).
+        //
+        // Note that this means that it is possible to frontrun this call to grief a user by taking up their
+        // selected remote address before they do so. This is mitigated somewhat by restricting potential attackers to
+        // the set of contracts that are allowlisted to hold veBAL (and their remapping managers). Should
+        // one of them grief, then Balancer governance can remove them from these allowlists.
+
+        // Assuming A --> B, then C cannot be remapped to B.
+        // To prevent it, we verify that the reverse mapping of B does not exist.
```
nventuro, line 171:
```suggestion
        // To prevent it, we verify that the reverse mapping of our remote does not exist.
```

pkg/liquidity-mining/contracts/VotingEscrowRemapper.sol
```diff
@@ -0,0 +1,306 @@
...
+ * expected to either use the same address, or manage delegation on L2 networks themselves.
+ *
+ * @dev For each network (chainId), we maintain a mapping between local (Ethereum) and remote (L2) addresses.
+ * This contract remaps balance queries on remote network addresses to their corresponding local addresses.
+ * Users able to call this contract can set their own mappings, or delegate this function to another account if they
+ * cannot.
+ */
+contract VotingEscrowRemapper is SingletonAuthentication, ReentrancyGuard {
+    IVotingEscrow private immutable _votingEscrow;
+    IOmniVotingEscrow private _omniVotingEscrow;
```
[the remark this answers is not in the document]
jubeira: I don't think it's a bad idea, actually.
https://github.com/balancer/balancer-v2-monorepo/issues/2402

pkg/liquidity-mining/contracts/VotingEscrowRemapper.sol
```diff
@@ -0,0 +1,306 @@
...
+        view
+        returns (IVotingEscrow.Point memory)
+    {
+        address localRemap = getLocalUser(remoteUser, chainId);
+        address localUser = localRemap == address(0) ? remoteUser : localRemap;
+
+        IVotingEscrow votingEscrow = getVotingEscrow();
+        uint256 userEpoch = votingEscrow.user_point_epoch(localUser);
+        return votingEscrow.user_point_history(localUser, userEpoch);
+    }
```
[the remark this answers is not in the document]
jubeira: I'd say remove. This was also making some noise in my head.

pkg/liquidity-mining/contracts/VotingEscrowRemapper.sol
```diff
@@ -0,0 +1,306 @@
...
+    /**
+     * @notice Clears a local user's mapping for a particular network.
+     * @dev This is intended to discourage and also allow recovery from griefing attacks.
+     * If griefing occurs then the griefer can be removed from Smart Wallet Checker and have their remappings erased.
+     * The local user can always clear its own mapping, regardless the state of the Smart Wallet Checker.
+     * @param localUser - The address of the local user to erase.
+     * @param chainId - The chain id of the network to erase.
+     */
+    function clearNetworkRemapping(address localUser, uint256 chainId) external payable nonReentrant {
+        require(!_isAllowedContract(localUser) || localUser == msg.sender, "localUser is still in good standing");
```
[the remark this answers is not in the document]
jubeira: How about this? https://github.com/balancer/balancer-v2-monorepo/pull/2174/commits/ec210abac0278da1d5195cfbb8dc895eb9d930fc

Leaving the manager should be harmless, since nor the user nor the manager can set a mapping again if the address is not allow listed in the smart wallet checker.

The manager cleaning 'feature' was not even documented, and it doesn't make much sense in this context as you say.

pkg/liquidity-mining/test/VotingEscrowRemapper.test.ts
```diff
@@ -0,0 +1,560 @@
...
+      });
+
+      it('emits an AddressMappingUpdated event', async () => {
+        const tx = await doClearMap();
+
+        expectEvent.inReceipt(await tx.wait(), 'AddressMappingUpdated', {
+          localUser: local.address,
+          remoteUser: ZERO_ADDRESS,
+          chainId,
+        });
```
nventuro, line 356: I wonder if we should also emit an event indicating that `remote` no longer has a local

pkg/liquidity-mining/test/VotingEscrowRemapper.test.ts
```diff
@@ -0,0 +1,560 @@
...
+      });
+
+      it('emits an AddressDelegateUpdated event', async () => {
+        const tx = await doClearMap();
+
+        expectEvent.inReceipt(await tx.wait(), 'AddressDelegateUpdated', {
+          localUser: local.address,
+          delegate: ZERO_ADDRESS,
+        });
+      });
```
nventuro, line 366:
```suggestion
```

This shouldn't be emitted

pkg/liquidity-mining/test/VotingEscrowRemapper.test.ts
```diff
@@ -0,0 +1,560 @@
...
+        sharedBeforeEach(() => {
+          caller = other;
+        });
+
+        it('reverts', async () => {
+          await expect(doClearMap()).to.be.revertedWith('localUser is still in good standing');
+        });
+      });
+
+      context('when caller it local user', () => {
```
nventuro, line 463:
```suggestion
      context('when caller is local user', () => {
```

nventuro (approved):

pkg/liquidity-mining/test/VotingEscrowRemapper.test.ts
jubeira: Hmm if you overwrite an existing remapping, there's a remote that no longer maps to a local too.
We could do it, but maybe for clarity we should use two different events. WDTY?

jubeira: Added in https://github.com/balancer/balancer-v2-monorepo/pull/2174/commits/36326229a89ca1248d1838df02c493be0cda42bc.

jubeira:
Thanks for the comments! I'll merge after the CI finishes.

PR #2173, opened 2023-01-11
Deployment preparation: `TimelockAuthorizerTransitionMigrator`

jubeira:
# Deployment preparation: `TimelockAuthorizerTransitionMigrator`

Build-info: master @https://github.com/balancer-labs/balancer-v2-monorepo/commit/43e6405e189a8c69c24b396a88cef8388ea97e2e.
Supersedes #2151 (same, but rebased and squashed for convenience).

## Deployment preparation tasks checklist:

- [x] Deployments `README` is updated with the new task name and link
- [x] Task `README` has a description and links to the artifacts and expected output folders 
- [x] `build-info` is updated with artifacts compiled at the specified commit
- [x] Artifacts are generated from `build-info`
- [x] New task has a working fork test

## Code checklist:

- [x] The diff is legible and has no extraneous changes 
- [x] Complex code has been commented in tests and scripts
- [x] Fork test block number and test cases are appropriate for this deployment
- [x] The base branch is either `master`, or there's a description of how to merge

## Issue Resolution

See #1991

PR #2184, opened 2023-01-20
Handle delayed permissions in `TimelockAuthorizerTransitionMigrator`

### LINKED ISSUE #2177: Handle delayed permissions in `TimelockAuthorizerTransitionMigrator`
Part of #2143 - see https://github.com/balancer-labs/balancer-v2-monorepo/pull/2173#issuecomment-1384498940.

jubeira:
# Description

This PR adds the capability of handling delayed permissions to `TimelockAuthorizerMigrator`.
The migrator will check whether permissions to be granted have delays set in the authorizer, scheduling the grant permission for later if that's the case and storing the schedule ID.
Then, a second added method executes all the scheduled permissions at once, skipping those that may fail for any reason. Anyone can execute the delayed actions; the `TimelockAuthorizer` should only revert if the delay is not due, if the permission was canceled, or if it was already executed.

As of today, the only expected scheduled permission is `add_gauge` in `GaugeController` (see https://github.com/balancer-labs/balancer-v2-monorepo/pull/2173#issuecomment-1397500695).

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #2177.

### DIFF
--- a/pkg/governance-scripts/contracts/TimelockAuthorizerTransitionMigrator.sol
+++ b/pkg/governance-scripts/contracts/TimelockAuthorizerTransitionMigrator.sol
@@ -26,7 +26,7 @@ contract TimelockAuthorizerTransitionMigrator {
     using Address for address;
 
     event PermissionSkipped(bytes32 indexed role, address indexed grantee, address indexed target);
-    event ScheduledExecutionSkipped(uint256 indexed permissionId);
+    event ScheduledExecutionSkipped(uint256 indexed scheduledExecutionId);
 
     IBasicAuthorizer public immutable oldAuthorizer;
     TimelockAuthorizer public immutable timelockAuthorizer;
--- a/pkg/governance-scripts/contracts/TimelockAuthorizerTransitionMigrator.sol
+++ b/pkg/governance-scripts/contracts/TimelockAuthorizerTransitionMigrator.sol
@@ -115,7 +115,7 @@ contract TimelockAuthorizerTransitionMigrator {
     /**
      * @notice Executes permissions scheduled during migration, all at once.
      * @dev `migratePermissions` must be called successfully first.
-     * Emits `ScheduledPermissionSkipped` event for permission IDs that cannot be executed at the time for any reason
+     * Emits `ScheduledPermissionSkipped` event for execution IDs that cannot be executed at this time for any reason
      * (delay not yet due, action was canceled, or action already executed).
      */
     function executeScheduledPermissions() external {
--- a/pkg/governance-scripts/contracts/TimelockAuthorizerTransitionMigrator.sol
+++ b/pkg/governance-scripts/contracts/TimelockAuthorizerTransitionMigrator.sol
@@ -26,7 +26,6 @@ contract TimelockAuthorizerTransitionMigrator {
     using Address for address;
 
     event PermissionSkipped(bytes32 indexed role, address indexed grantee, address indexed target);
-    event ScheduledExecutionSkipped(uint256 indexed scheduledExecutionId);
 
     IBasicAuthorizer public immutable oldAuthorizer;
     TimelockAuthorizer public immutable timelockAuthorizer;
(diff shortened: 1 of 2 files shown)


pkg/governance-scripts/contracts/TimelockAuthorizerTransitionMigrator.sol
```diff
@@ -97,6 +112,26 @@ contract TimelockAuthorizerTransitionMigrator {
         );
     }
 
+    /**
+     * @notice Executes permissions scheduled during migration, all at once.
+     * @dev `migratePermissions` must be called successfully first.
+     * Emits `ScheduledPermissionSkipped` event for permission IDs that cannot be executed at the time for any reason
```
EndymionJkb, line 118:
```suggestion
     * Emits `ScheduledPermissionSkipped` event for execution IDs that cannot be executed at this time for any reason
```

EndymionJkb: In the code they're called scheduledExecutionIds, so I'd call them execution IDs here, vs "permission ID"

pkg/governance-scripts/contracts/TimelockAuthorizerTransitionMigrator.sol
```diff
@@ -97,6 +112,26 @@ contract TimelockAuthorizerTransitionMigrator {
...
+     * (delay not yet due, action was canceled, or action already executed).
+     */
```
EndymionJkb, line 120: Maybe add here why it's ok to call this multiple times (more explicitly). As I understand it, if you had 3 things with a week delay, and 3 with a 2-week delay, and you called this after a week, it would grant permissions on the 3 1-week ids, and skip the 2-week ones.

Then if you waited another week and called it again, it would skip the first three - canExecute would return false because it "executed" would have been set to true - and grant the last 3.

So I think you don't need that "ScheduledPermissionSkipped" event. Events usually indicate either a successful action, or an error. But here it would be emitted as part of normal operation (unless all permissions happened to be the same length, and you only called it once). You could call this forever after, and it would just keep emitting skipped events.

pkg/governance-scripts/test/TimelockAuthorizerTransitionMigrator.test.ts
```diff
@@ -151,6 +153,107 @@ describe('TimelockAuthorizerTransitionMigrator', () => {
...
+                scheduledExecutionId,
+              },
+              newAuthorizer.address,
+              1
+            );
+
+            await expect(transitionMigrator.scheduledPermissionIds(1)).to.be.reverted;
+          });
+
+          it('can grant a scheduled permission after the delay passes', async () => {
```
EndymionJkb, line 221: Maybe have 2 permissions with different delays? And test that it grants the earliest and keeps the second one pending until the longer delay expires?

pkg/governance-scripts/contracts/TimelockAuthorizerTransitionMigrator.sol
jubeira: +1, I'll update the docs.

Regarding the event, your reasoning is correct.

The intention behind it was to be able to reason about pending permissions by quickly looking at the transaction events. Without the event, if a permission or two is skipped on the first call, it would be a bit more difficult to tell whether the migration is complete or not, since the timelock authorizer will only emit an event when granting a permission in this case. That being said, do you still think we should remove it?

pkg/governance-scripts/test/TimelockAuthorizerTransitionMigrator.test.ts
jubeira: Done in https://github.com/balancer-labs/balancer-v2-monorepo/pull/2184/commits/af8a8ba369d0fe67ea729c7cf359dc4d025f7353.

pkg/governance-scripts/contracts/TimelockAuthorizerTransitionMigrator.sol
EndymionJkb: It's not the end of the world, but it seems tracking the "positive" events should be sufficient to know whether permissions have been granted. I don't particularly like having a function anybody can call forever that would emit a bunch of events. A bot could flood the subgraph with spurious events.

jubeira: Sounds good then.

Removed in https://github.com/balancer-labs/balancer-v2-monorepo/pull/2184/commits/fe318378668813bd162da10b5aa94c1422530fa2.

TomAFrench (approved):
LGTM

PR #2185, opened 2023-01-24
BalancerSorQueries - Helper contract for optimizing SOR initialization

danielmkm:
# Description

Helper contract `BalancerSorQueries` that aims to optimize SOR initialization. Helper functions for bulk loading of onchain data with an attempt to reduce both network-in and network-out of the multicall request.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/standalone-utils/contracts/BalancerSorQueries.sol
```diff
@@ -0,0 +1,293 @@
...
+        returns (uint256[][] memory)
+    {
+        uint256[] memory balances;
+        uint256 lastChangeBlock;
+        uint256[][] memory allBalances = new uint256[][](poolIds.length);
+
+        for (uint256 i = 0; i < poolIds.length; i++) {
+            (, balances, lastChangeBlock) = vault.getPoolTokens(poolIds[i]);
+
+            if (lastChangeBlock > blockNumber) {
```
mikemcdonald, line 178: I'm not sure this logic makes sense and can't think of scenarios where this is true. Since the `getPoolTokens` call happens no matter what, why not always return all balances since that is assumed to be the latest state

mikemcdonald, line 178: Is the idea here to just limit the rpc calldata size? Since the `getPoolTokens` call is made regardless it seems better to always update that data. The onchain node state is the source of truth. An example of this is right now on Polygon the subgraph is up to the latest block but has slightly bad/inaccurate data. This means the onchain data doesn't overwrite any existing bad subgraph data

pkg/standalone-utils/contracts/BalancerSorQueries.sol
```diff
@@ -0,0 +1,293 @@
...
+        uint256[] memory amps = new uint256[](poolAddresses.length);
+
+        for (uint256 i = 0; i < poolAddresses.length; i++) {
+            (amps[i], , ) = IPoolWithAmp(poolAddresses[i]).getAmplificationParameter();
+        }
+
+        return amps;
+    }
+
+    function getSwapFeePercentageForPools(address[] memory poolAddresses, SwapFeeType[] memory swapFeeTypes)
```
nventuro, line 206: I'd consider checking that these arrays have the same length, to help detect user error

pkg/standalone-utils/contracts/BalancerSorQueries.sol
```diff
@@ -0,0 +1,293 @@
...
+        }
+    }
+
+    function _getPoolActualSupply(address poolAddress) internal view returns (uint256) {
+        try IPoolWithActualSupply(poolAddress).getActualSupply() returns (uint256 actualSupply) {
+            return actualSupply;
+        } catch {
+            return 0;
+        }
+    }
```
nventuro, line 292: Why handle these silently like this? Wouldn't it be better to revert?

nventuro (commented):
Thanks for opening this @danielmkm!

I left a couple questions and small comments, and like @mikemcdonald am curious about `getPoolTokenBalancesWithUpdatesAfterBlock`.

Also, it seems like `pkg/deployments/tasks/20220721-balancer-queries/build-info/BalancerQueries.json` was accidentally deleted?

pkg/standalone-utils/contracts/BalancerSorQueries.sol
danielmkm: Hey sorry for the delay here, missed the notifications.

The overall goal of this contract was to limit both data in and data out, and the token balances represent the largest amount of data out for the query. The initial assumption was that the subgraph balances should be correct (until the last block processed), so only loading changes after that block would significantly reduce the data out. This could also extend to polling queries (ie: if you're fetching every 30-60 seconds, you really dont need all balances on every fetch).

You can always pass a `blockNumber` of 0 and bypass the check entirely, I added this to the config in the SOR, it previously was not there. Definitely not stuck on keeping it though, but it does produce meaningful reduction in the query time simply by reducing the amount of data being transferred across the network.

pkg/standalone-utils/contracts/BalancerSorQueries.sol
danielmkm: This is actually a solution for one of the bigger problems we have currently. The SOR does a big multicall fetching all data needed, but if any operation in the multicall reverts, the entire request fails without any information on what operation in the multicall failed.

pkg/standalone-utils/contracts/BalancerSorQueries.sol
danielmkm: totally fair, will add that!

pkg/standalone-utils/contracts/BalancerSorQueries.sol
nventuro: So you'd be using a supply of zero as a sentinel value to indicate one of these requests failed? If so I'd at least write comments here explaning this (though you could potentially better achieve this by e.g. returning a tuple or otherwise more clearly emulating a 'may'be value).

danielmkm: Agreed, the approach here was mainly for preventing the request from reverting. I had a chat with john today and he'd also like a way to identify pools with failed requests so they can be filtered out. I'll rework this part.

danielmkm: I did a pass here, ended up leaving the 0 values and adding some comments. The point is that if any of the functions above return 0, the pool should get filtered as well, so we catch both situations.

I extended `getPoolData` to return `ignoreIdxs`, which are any pools that returned 0 values where 0 is not an acceptable value, this should give the SDK/frontend the ability to more easily filter out pools from the list that shouldn't be used in routing

mikemcdonald:
I haven't fully tested this in the SDK yet, but all of the changes LGTM

pkg/standalone-utils/contracts/BalancerPoolDataQueries.sol
```diff
@@ -0,0 +1,415 @@
...
+
+        uint256[] memory errorIdxs = new uint256[](numErrors);
+        uint256 idx = 0;
+
+        for (i = 0; i < errors.length; i++) {
+            if (errors[i] == true) {
+                errorIdxs[idx] = i;
+                idx++;
+            }
+        }
```
nventuro, line 411: Nice

nventuro (commented):
I verified the build info corresponds to commit ef80e7268060ef85428486bc1c2a9dd7647bb2d5 of the [beethovenxfi fork of this repo](https://github.com/beethovenxfi/balancer-v2-monorepo).

Can you add this to the main deployments readme?

Also, it seems `pkg/deployments/tasks/20220721-balancer-queries/build-info/BalancerQueries.json` was accidentally deleted. Please restore that file.

nventuro:
Online CI checks are failing and will continue to fail as @danielmkm is not part of this org and therefore does not have access to the respository secrets (which include the RPC API keys).

That said, it seems like the task artifacts are wrong. Can you run `yarn hardhat extract-artifacts --id 20230124-balancer-pool-data-queries`?

danielmkm:
Restored `pkg/deployments/tasks/20220721-balancer-queries/build-info/BalancerQueries.json`, sorry about that. 

I ran `extract-artifacts` but the output is the same as what's already committed, there is no diff 🤷.

Screenshot 2023-04-11 at 10 22 29

pkg/deployments/tasks/20230124-balancer-pool-data-queries/readme.md
```diff
@@ -0,0 +1,4 @@
+# 2023-01-24 - Balancer Pool DataQueries
+
+Deployment of the `BalancerPoolDataQueriesDeployment` contract. It provides a way to perform bulk queries that is useful for optimizing SOR initialization of on chain data.
+
```
nventuro, line 4:
```suggestion

- [`BalancerPoolDataQueries` artifact](./artifact/BalancerPoolDataQueries.json)
```

pkg/deployments/tasks/20230124-balancer-pool-data-queries/readme.md
```diff
@@ -0,0 +1,4 @@
+# 2023-01-24 - Balancer Pool DataQueries
+
+Deployment of the `BalancerPoolDataQueriesDeployment` contract. It provides a way to perform bulk queries that is useful for optimizing SOR initialization of on chain data.
```
nventuro, line 3:
```suggestion
Deployment of the `BalancerPoolDataQueries` contract. It provides a way to perform bulk queries that is useful for optimizing SOR initialization of on chain data.
```

nventuro (approved):
Sorry, yes - it was the readme link that was missing. I added two suggestions since I can't write to the base repo, we can merge once those are in!

danielmkm:
done!

jubeira:
Superseded by #2522 and https://github.com/balancer/balancer-deployments/pull/53.

PR #1991, opened 2022-11-08
Deployment preparation: adaptor entrypoint

jubeira:
# Description

This PR adds the deployment task for the adaptor entrypont (scripts, smoke test, docs).
We need the new timelock authorizer to be able to actually test the entrypoint, so the test is rather skinny.

Artifacts synced to master @https://github.com/balancer-labs/balancer-v2-monorepo/commit/6dd6892f81443f16e3f62ff7f70c078ebdae40f0

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Documentation or wording changes
- [x] Other

## Checklist:

- [x] The diff is legible and has no extraneous changes
- [x] Complex code has been commented, including external interfaces
- [ ] Tests are included for all code paths --> `performAction` pending; it should be added in the authorizer task
- [x] The base branch is either `master`, or there's a description of how to merge

## Issue Resolution

See #1990

PR #2188, opened 2023-01-25
`TimelockAuthorizerTransitionMigrator`: fork tests improvement

jubeira:
# Description

Follow-up on #2173 (see https://github.com/balancer-labs/balancer-v2-monorepo/pull/2173#discussion_r1086629495).
Just fork tests improvements; no changes in contract features nor artifacts.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other

## Checklist:

## Issue Resolution

See #1991


EndymionJkb (approved):
LGTM: checks that it doesn't migrate after half the delay, and does after the full delay. You're checking both that the event was emitted, and the state of the permission.

The only other thing I can think of would be to run it again after the delay expires and it's already granted (should not revert, and just do nothing). But I don't think that's important

PR #2189, opened 2023-01-26
Deprecation cleanup: broken links and references

EndymionJkb:
# Description

Noticed the weighted pool readme pointed to stable... so looked over the whole directory: 1) links should all work; 2) original reasons for deprecation should be preserved; 3) it shouldn't say use the new version if it's also been deprecated.

We will want to update all these again to point to the current ones, once deployed. 

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira (approved):
LGTM!
I'm not super happy about the fact that references inside deprecated packages also break when we release a new version, but I guess that's just the way it is.

PR #2205, opened 2023-02-07
[Reentrancy patch] Weighted Pool V3

jubeira:
# Description

This PR protects weighted pool functions from read-only reentrancy, and improves docs.
Based on `weighted-pool-v3` branch. Note that the base commit of this branch should generate the build info for `weighted-pool-v2`: https://github.com/balancer-labs/balancer-v2-monorepo/commit/72f3682b133335b75195124f44c9f5eaa54860b0.

Not to be merged to master. This patch should be enough to generate the build info for `weighted-pool-v3`.

See https://forum.balancer.fi/t/reentrancy-vulnerability-scope-expanded/4345/1 for reference.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

- N/A Tests are included for all code paths

## Issue Resolution

N/A

### DIFF
--- a/pkg/pool-utils/contracts/RecoveryMode.sol
+++ b/pkg/pool-utils/contracts/RecoveryMode.sol
@@ -74,7 +74,7 @@ abstract contract RecoveryMode is IRecoveryMode, BasePoolAuthorization {
      * This function will revert when called within a Vault context (i.e. in the middle of a join or an exit).
      *
      * This function depends on the invariant value, which may be calculated incorrectly in the middle of a join or
-     * an exit, because the state of the pool could be out of sync with the state of the vault.
+     * an exit, because the state of the pool could be out of sync with the state of the Vault.
      * `_onDisableRecoveryMode` will revert when called from such a context for weighted pools, effectively
      * protecting this function.
      *
--- a/pkg/pool-utils/contracts/RecoveryMode.sol
+++ b/pkg/pool-utils/contracts/RecoveryMode.sol
@@ -70,6 +70,13 @@ abstract contract RecoveryMode is IRecoveryMode, BasePoolAuthorization {
      * @notice Disable recovery mode, which disables the special safe exit path for LPs.
      * @dev Protocol fees are not paid while in Recovery Mode, so it should only remain active for as long as strictly
      * necessary.
+     *
+     * This function will revert when called within a Vault context (i.e. in the middle of a join or an exit).
+     *
+     * This function depends on the invariant value, which may be calculated incorrectly in the middle of a join or
+     * an exit, because the state of the pool could be out of sync with the state of the vault.
+     * `_onDisableRecoveryMode` will revert when called from such a context for weighted pools, effectively
+     * protecting this function.
      */
     function disableRecoveryMode() external override authenticate {
         _setRecoveryMode(false);
--- a/pkg/pool-utils/contracts/RecoveryMode.sol
+++ b/pkg/pool-utils/contracts/RecoveryMode.sol
@@ -77,6 +77,8 @@ abstract contract RecoveryMode is IRecoveryMode, BasePoolAuthorization {
      * an exit, because the state of the pool could be out of sync with the state of the vault.
      * `_onDisableRecoveryMode` will revert when called from such a context for weighted pools, effectively
      * protecting this function.
+     *
+     * See https://forum.balancer.fi/t/reentrancy-vulnerability-scope-expanded/4345 for reference.
      */
     function disableRecoveryMode() external override authenticate {
         _setRecoveryMode(false);
--- a/pkg/pool-weighted/contracts/BaseWeightedPool.sol
+++ b/pkg/pool-weighted/contracts/BaseWeightedPool.sol
@@ -89,6 +89,7 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool {
      * Calculating the invariant requires the state of the pool to be in sync with the state of the vault.
      * That condition may not be true in the middle of a join or an exit, which is why the value returned by this
      * function under that circumstance could be incorrect.
+     * See https://forum.balancer.fi/t/reentrancy-vulnerability-scope-expanded/4345 for reference.
      */
     function getInvariant() public view returns (uint256) {
         (, uint256[] memory balances, ) = getVault().getPoolTokens(getPoolId());
--- a/pkg/pool-weighted/contracts/BaseWeightedPool.sol
+++ b/pkg/pool-weighted/contracts/BaseWeightedPool.sol
@@ -84,11 +84,15 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool {
      * @dev Returns the current value of the invariant.
      *
      * **IMPORTANT NOTE**: calling this function within a Vault context (i.e. in the middle of a join or an exit) is
-     * potentially unsafe, since the returned value may be incorrect. It is up to the caller to protect itself.
+     * potentially unsafe, since the returned value is manipulable. It is up to the caller to ensure safety.
      *
      * Calculating the invariant requires the state of the pool to be in sync with the state of the vault.
-     * That condition may not be true in the middle of a join or an exit, which is why the value returned by this
-     * function under that circumstance could be incorrect.
+     * That condition may not be true in the middle of a join or an exit.
+     *
+     * To call this function safely, attempt to trigger the reentrancy guard in the vault by calling a non-reentrant
+     * function before calling `getInvariant`. That will make the transaction revert in an unsafe context.
+     * See `whenNotInVaultContext` in `WeightedPool` for reference.
+     *
      * See https://forum.balancer.fi/t/reentrancy-vulnerability-scope-expanded/4345 for reference.
      */
     function getInvariant() public view returns (uint256) {
--- a/pkg/pool-weighted/contracts/BaseWeightedPool.sol
+++ b/pkg/pool-weighted/contracts/BaseWeightedPool.sol
@@ -86,10 +86,10 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool {
      * **IMPORTANT NOTE**: calling this function within a Vault context (i.e. in the middle of a join or an exit) is
      * potentially unsafe, since the returned value is manipulable. It is up to the caller to ensure safety.
      *
-     * Calculating the invariant requires the state of the pool to be in sync with the state of the vault.
+     * Calculating the invariant requires the state of the pool to be in sync with the state of the Vault.
      * That condition may not be true in the middle of a join or an exit.
      *
-     * To call this function safely, attempt to trigger the reentrancy guard in the vault by calling a non-reentrant
+     * To call this function safely, attempt to trigger the reentrancy guard in the Vault by calling a non-reentrant
      * function before calling `getInvariant`. That will make the transaction revert in an unsafe context.
      * (See `whenNotInVaultContext` in `WeightedPool`).
      *
--- a/pkg/pool-weighted/contracts/BaseWeightedPool.sol
+++ b/pkg/pool-weighted/contracts/BaseWeightedPool.sol
@@ -91,7 +91,7 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool {
      *
      * To call this function safely, attempt to trigger the reentrancy guard in the vault by calling a non-reentrant
      * function before calling `getInvariant`. That will make the transaction revert in an unsafe context.
-     * See `whenNotInVaultContext` in `WeightedPool` for reference.
+     * (See `whenNotInVaultContext` in `WeightedPool`).
      *
      * See https://forum.balancer.fi/t/reentrancy-vulnerability-scope-expanded/4345 for reference.
      */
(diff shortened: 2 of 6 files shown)


pkg/pool-weighted/contracts/BaseWeightedPool.sol
```diff
@@ -82,6 +82,14 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool {
 
     /**
      * @dev Returns the current value of the invariant.
+     *
+     * **IMPORTANT NOTE**: calling this function within a Vault context (i.e. in the middle of a join or an exit) is
+     * potentially unsafe, since the returned value may be incorrect. It is up to the caller to protect itself.
```
EndymionJkb, line 87:
```suggestion
     * potentially unsafe, since the returned value is manipulable. It is up to the caller to ensure safety.
```

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -301,7 +327,15 @@ contract WeightedPool is BaseWeightedPool, WeightedPoolProtocolFees {
         WeightedPoolProtocolFees._updatePostJoinExit(postJoinExitInvariant);
     }
 
-    function _beforeProtocolFeeCacheUpdate() internal override {
+    /**
+     * @dev This function will revert when called within a Vault context (i.e. in the middle of a join or an exit).
+     *
+     * This function depends on the invariant value, which may be calculated incorrectly in the middle of a join or
+     * an exit because the state of the pool could be out of sync with the state of the vault. This makes the function
+     * unsafe to call in such contexts, and hence it is protected.
```
EndymionJkb, line 335:
```suggestion
     * an exit, because the state of the pool could be out of sync with the state of the vault.
```
I think this is enough

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -342,8 +376,16 @@ contract WeightedPool is BaseWeightedPool, WeightedPoolProtocolFees {
      * even if they don't yet exist, they will effectively be included in any Pool operation that involves BPT.
      *
      * In the vast majority of cases, this function should be used instead of `totalSupply()`.
+     *
+     * **IMPORTANT NOTE**: calling this function within a Vault context (i.e. in the middle of a join or an exit) is
+     * potentially unsafe, since the returned value may be incorrect. It is up to the caller to protect itself.
```
EndymionJkb, line 381:
```suggestion
     * potentially unsafe, since the returned value is manipulable. It is up to the caller to ensure safety.
```
Same as above

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -342,8 +376,16 @@ contract WeightedPool is BaseWeightedPool, WeightedPoolProtocolFees {
...
+     *
+     * This is because this function calculates the invariant, which requires the state of the pool to be in sync
+     * with the state of the vault. That condition may not be true in the middle of a join or an exit, which is why
+     * the value returned by this function under that circumstance could be incorrect.
```
EndymionJkb, line 385:
```suggestion
     * with the state of the vault. That condition may not be true in the middle of a join or an exit.
```

pkg/pool-weighted/contracts/BaseWeightedPool.sol
```diff
@@ -82,6 +82,14 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool {
...
+     *
+     * Calculating the invariant requires the state of the pool to be in sync with the state of the vault.
+     * That condition may not be true in the middle of a join or an exit, which is why the value returned by this
+     * function under that circumstance could be incorrect.
```
EndymionJkb, line 91:
```suggestion
     * That condition may not be true in the middle of a join or an exit.
```

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -355,7 +397,15 @@ contract WeightedPool is BaseWeightedPool, WeightedPoolProtocolFees {
         return supply.add(protocolFeesToBeMinted);
     }
 
-    function _onDisableRecoveryMode() internal override {
+    /**
+     * @dev This function will revert when called within a Vault context (i.e. in the middle of a join or an exit).
+     *
+     * This function depends on the invariant value, which may be calculated incorrectly in the middle of a join or
+     * an exit because the state of the pool could be out of sync with the state of the vault. This makes the function
+     * unsafe to call in such contexts, and hence it is protected.
```
EndymionJkb, line 405:
```suggestion
     * an exit, because the state of the pool could be out of sync with the state of the vault.
```

EndymionJkb (commented):
Just shortened some of the comments for clarity.

pkg/pool-weighted/contracts/BaseWeightedPool.sol
```diff
@@ -82,6 +82,14 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool {
 
     /**
      * @dev Returns the current value of the invariant.
+     *
+     * **IMPORTANT NOTE**: calling this function within a Vault context (i.e. in the middle of a join or an exit) is
```
nventuro, line 86: We should also tell users _how_ to protect themselves, otherwise we're giving a warning but no actionable advice.

EndymionJkb:
You can also add the `manageUserBalance` to the MockVault, if that's easier

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -140,6 +140,32 @@ contract WeightedPool is BaseWeightedPool, WeightedPoolProtocolFees {
         _normalizedWeight7 = numTokens > 7 ? params.normalizedWeights[7] : 0;
     }
 
+    /**
+     * @dev Ensure we are not in a Vault context when this function is called, by attempting a zero-value internal
+     * balance operation. If we are already in a Vault transaction (e.g., a swap, join, or exit), the Vault's
```
nventuro, line 145: Note that we no longer use the zero-value approach

pkg/pool-weighted/contracts/WeightedPool.sol
nventuro: I don't agree with this suggestion, we should explain that we're adding the modifier and why. Otherwise we'd just say 'this is dangerous', instead of 'this would be dangerous, *but* we fixt it this way'.

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -301,7 +327,15 @@ contract WeightedPool is BaseWeightedPool, WeightedPoolProtocolFees {
...
-    function _beforeProtocolFeeCacheUpdate() internal override {
+    /**
+     * @dev This function will revert when called within a Vault context (i.e. in the middle of a join or an exit).
+     *
+     * This function depends on the invariant value, which may be calculated incorrectly in the middle of a join or
+     * an exit because the state of the pool could be out of sync with the state of the vault. This makes the function
+     * unsafe to call in such contexts, and hence it is protected.
+     * See https://forum.balancer.fi/t/reentrancy-vulnerability-scope-expanded/4345 for reference.
+     */
+    function _beforeProtocolFeeCacheUpdate() internal override whenNotInVaultContext {
```
nventuro, line 338: I'm not sure I agree with this patch. According to our mental model, the problematic function is `updateProtocolFeePercentageCache`, not this one. This function is dangerous and the caller must take care, but the caller here is the external function, and hence that should receive the fix.

Note that it doesn't matter much here since this is the only occurance, but still.

pkg/pool-weighted/test/WeightedPool.test.ts
```diff
@@ -42,17 +42,22 @@ describe('WeightedPool', function () {
 
     const initialBalances = range(1, 3).map(fp);
 
-    sharedBeforeEach('deploy pool', async () => {
+    sharedBeforeEach('deploy pool and real vault', async () => {
+      const vault = await Vault.create();
+
       tokens = allTokens.subset(2);
+      await tokens.mint({ to: lp.address, amount: max(initialBalances) });
+      await tokens.approve({ to: vault.address, from: lp, amount: max(initialBalances) });
```
nventuro, line 50: Is this so that tests don't revert when calling `manageUserBalance`?

jubeira: I would add it to the mock vault in master. For the patch, this should do just to turn the CI green, since only this test seems to be affected.

pkg/pool-weighted/contracts/WeightedPool.sol
jubeira: This is just for the patch.
The problem with the external function is that it `ProtocolFeeCache` does not know about the vault, so it'll need a reference to it. And we wanted to keep the patch as concise as possible.

The plan is to add that reference in master for a more definitive fix and for better maintainability in future versions.

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -140,6 +140,32 @@ contract WeightedPool is BaseWeightedPool, WeightedPoolProtocolFees {
         _normalizedWeight7 = numTokens > 7 ? params.normalizedWeights[7] : 0;
     }
 
+    /**
+     * @dev Ensure we are not in a Vault context when this function is called, by attempting a zero-value internal
```
EndymionJkb, line 144:
```suggestion
     * @dev Ensure we are not in a Vault context when this function is called, by attempting an internal
```
Good catch! Could just say this.

pkg/pool-weighted/contracts/WeightedPool.sol
jubeira: I've added some extra notes in the external functions in https://github.com/balancer-labs/balancer-v2-monorepo/pull/2205/commits/9f56af4c2319ca773bcade4da640681a6978faa8; PTAL.

Note again that these are only applicable to this patch in particular, since they actually refer to weighted pools even though the utils may be used in master for other pool types. But I think it makes sense for this particular deployment.

pkg/pool-weighted/contracts/BaseWeightedPool.sol
jubeira: how about now? https://github.com/balancer-labs/balancer-v2-monorepo/pull/2205/commits/ab7702eae6f4b17e50cdb89f1e52923e9d6d674e

pkg/pool-weighted/contracts/WeightedPool.sol
jubeira: how about now? https://github.com/balancer-labs/balancer-v2-monorepo/commit/ab7702eae6f4b17e50cdb89f1e52923e9d6d674e

nventuro (approved):

pkg/pool-utils/contracts/RecoveryMode.sol
```diff
@@ -70,6 +70,13 @@ abstract contract RecoveryMode is IRecoveryMode, BasePoolAuthorization {
      * @notice Disable recovery mode, which disables the special safe exit path for LPs.
      * @dev Protocol fees are not paid while in Recovery Mode, so it should only remain active for as long as strictly
      * necessary.
+     *
+     * This function will revert when called within a Vault context (i.e. in the middle of a join or an exit).
+     *
+     * This function depends on the invariant value, which may be calculated incorrectly in the middle of a join or
+     * an exit, because the state of the pool could be out of sync with the state of the vault.
+     * `_onDisableRecoveryMode` will revert when called from such a context for weighted pools, effectively
+     * protecting this function.
```
EndymionJkb, line 79: Do we want the forum reference here too, as in ProtocolFeeCache?

pkg/pool-weighted/contracts/BaseWeightedPool.sol
```diff
@@ -82,6 +82,18 @@ abstract contract BaseWeightedPool is BaseMinimalSwapInfoPool {
...
+     * potentially unsafe, since the returned value is manipulable. It is up to the caller to ensure safety.
+     *
+     * Calculating the invariant requires the state of the pool to be in sync with the state of the vault.
+     * That condition may not be true in the middle of a join or an exit.
+     *
+     * To call this function safely, attempt to trigger the reentrancy guard in the vault by calling a non-reentrant
+     * function before calling `getInvariant`. That will make the transaction revert in an unsafe context.
+     * See `whenNotInVaultContext` in `WeightedPool` for reference.
```
EndymionJkb, line 94:
```suggestion
     * (See `whenNotInVaultContext` in `WeightedPool`.)
```
Nit, but we already have a "for reference" below

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -342,8 +378,20 @@ contract WeightedPool is BaseWeightedPool, WeightedPoolProtocolFees {
...
+     *
+     * **IMPORTANT NOTE**: calling this function within a Vault context (i.e. in the middle of a join or an exit) is
+     * potentially unsafe, since the returned value is manipulable. It is up to the caller to ensure safety.
+     *
+     * This is because this function calculates the invariant, which requires the state of the pool to be in sync
+     * with the state of the vault. That condition may not be true in the middle of a join or an exit.
+     *
+     * To call this function safely, attempt to trigger the reentrancy guard in the vault by calling a non-reentrant
+     * function before calling `getActualSupply`. That will make the transaction revert in an unsafe context.
+     * See `whenNotInVaultContext` in `WeightedPool` for reference.
```
EndymionJkb, line 390:
```suggestion
     * (See `whenNotInVaultContext` in `WeightedPool`.)
```

EndymionJkb (approved):
Just nits at this point

PR #2207, opened 2023-02-07
Reentrancy Preparation: Introduce VaultReentrancyLib

EndymionJkb:
# Description

As the first step to addressing the read-only reentrancy issue going forward (see this [forum post](https://forum.balancer.fi/t/reentrancy-vulnerability-scope-expanded/4345) for context), introduce a library with a function that can be called to ensure the caller is not already within a Vault context (e.g., swap/join/exit).

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/pool-utils/contracts/test/MockVault.sol
```diff
@@ -222,4 +222,9 @@ contract MockVault is IPoolSwapStructs {
     function setAuthorizer(IAuthorizer newAuthorizer) external {
         _authorizer = newAuthorizer;
     }
+
+    // Support VaultReentrancyLib
+    function manageUserBalance(IVault.UserBalanceOp[] memory ops) external payable {
+        // solhint-disable-previous-line no-empty-blocks
+    }
```
nventuro, line 229: Shouldn't we also have the reentrancy guard here? Tests that use the new lib will fail with this mock if not

nventuro: (also I don't think we need this change in this PR at all?)

pkg/pool-utils/contracts/lib/VaultReentrancyLib.sol
```diff
@@ -0,0 +1,36 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+
+import "@balancer-labs/v2-interfaces/contracts/vault/IVault.sol";
+
+library VaultReentrancyLib {
+    /**
+     * @dev Ensure we are not in a Vault context when this function is called, by attempting an internal
```
jubeira, line 21:
```suggestion
     * @dev Ensure we are not in a Vault context when this function is called, by attempting a no-op internal
```

pkg/pool-utils/contracts/test/MockReentrancyPool.sol
```diff
@@ -0,0 +1,176 @@
...
+
+pragma solidity ^0.7.0;
+pragma experimental ABIEncoderV2;
+
+import "@balancer-labs/v2-interfaces/contracts/pool-weighted/WeightedPoolUserData.sol";
+
+import "../BaseGeneralPool.sol";
+import "../lib/VaultReentrancyLib.sol";
+
+contract MockReentrancyPool is BaseGeneralPool {
```
jubeira, line 23: Maybe add a small @dev block here?

pkg/pool-utils/contracts/test/MockReentrancyPool.sol
```diff
@@ -0,0 +1,176 @@
...
+            pauseWindowDuration,
+            bufferPeriodDuration,
+            owner
+        )
+    {
+        _totalTokens = tokens.length;
+    }
+
+    // Define a public function protected by the library. Works when called externally; reverts if called
+    // during a Vault operation (join/swap/exit).
```
jubeira, line 60:
```suggestion
    /**
     * @dev Public function protected by the library. Works when called externally; reverts if called
     * during a Vault operation (join/swap/exit).
     */
```

pkg/pool-utils/contracts/test/MockVault.sol
EndymionJkb: It's only here because it will be needed in multiple downstream PRs, which I wanted to keep independent.

pkg/pool-utils/contracts/test/MockReentrancyPool.sol
```diff
@@ -0,0 +1,176 @@
...
+        VaultReentrancyLib.ensureNotInVaultContext(getVault());
+
+        emit ProtectedFunctionCalled();
+    }
+  
+    // Vault hooks
+
+    // Testing all the hooks is technically overkill: if it works in one Vault context, it works in all.
+    // We mock up an entire pool with all the hooks defined mainly for documentation purposes: in a real pool,
+    // these are all the places you need to protect.
```
jubeira, line 71: I find this comment a bit confusing, since it's not the hooks that need to be protected (protecting them will cause basically any valid operation to revert, as the test shows).

I'd propose rephrasing it a bit, saying that these are all the hooks that are called within a vault context, callers should be protected against calling unsafe functions within that context.

jubeira (approved):
A few minor comments, otherwise looks good!

PR #2208, opened 2023-02-07
Reentrancy Preparation: store Vault reference

EndymionJkb:
# Description

The complete fix for the read-only reentrancy issue going forward (see this [forum post](https://forum.balancer.fi/t/reentrancy-vulnerability-scope-expanded/4345) for context), we need to protect external functions in base contracts (specifically, `RecoveryMode` and  `ProtocolFeeCache`) that do not currently reference the Vault: yet the Vault is required for the library call.

Since `ProtocolFeeCache` is derived from `RecoveryMode`, it is sufficient to store a Vault reference in RecoveryMode, and make it available internally for calls to the VaultReentrancyLib.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/pool-utils/test/ProtocolFeeCache.test.ts
+++ b/pkg/pool-utils/test/ProtocolFeeCache.test.ts
@@ -75,7 +75,11 @@ describe('ProtocolFeeCache', () => {
     it('reverts during deployment', async () => {
       await expect(
         deploy('MockProtocolFeeCache', {
-          args: [vault.protocolFeesProvider.address, { swap: 137, yield: ProtocolFee.SWAP, aum: ProtocolFee.YIELD }],
+          args: [
+            vault.address,
+            vault.protocolFeesProvider.address,
+            { swap: 137, yield: ProtocolFee.SWAP, aum: ProtocolFee.YIELD },
+          ],
           from: admin,
         })
       ).to.be.revertedWith('Non-existent fee type');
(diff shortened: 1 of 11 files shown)


pkg/pool-utils/test/ProtocolFeeCache.test.ts
```diff
@@ -75,7 +75,11 @@ describe('ProtocolFeeCache', () => {
     it('reverts during deployment', async () => {
       await expect(
         deploy('MockProtocolFeeCache', {
-          args: [vault.protocolFeesProvider.address, { swap: 137, yield: ProtocolFee.SWAP, aum: ProtocolFee.YIELD }],
+          args: [
+            vault.address,
+            vault.protocolFeesProvider.address,
+            { swap: 137, yield: ProtocolFee.SWAP, aum: ProtocolFee.YIELD },
+          ],
```
jubeira, line 82: I would add one small getter test here and in `RecoveryMode.test.ts` besides the ones you've added in base pools.

jubeira (commented):
One minor comment, otherwise LGTM!

pkg/pool-utils/test/ProtocolFeeCache.test.ts
EndymionJkb: There isn't really an independent RecoveryMode test, and I don't think the contortions to make one are worth it. But I can add it to ProtocolFeeCache at least.

jubeira (approved):

PR #2211, opened 2023-02-07
Pre-deployment preparation adjustments: interfaces & CSP fork test

jubeira:
# Description

Minor changes:
- Use USDC instead of USDT in Composable Stable Pool (CSP) fork test. The reentrancy attacker contract needs to call `approve`, and using USDT means we'd need to add a dependency to the safe ERC20 library (or copy the code). We can just use USDC instead; it even has the same decimals and just works for the fork test.
- Add methods that we'll be testing to interfaces, so that we only need to add a dependency between `deployments` and `interfaces`.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A


EndymionJkb (commented):
Does @inheritdoc work inside a /** */ block like that? Elsewhere we have it as:
/// @inheritdoc

nventuro (approved):

PR #2214, opened 2023-02-07
Reentrancy Preparation: Linear Pool

EndymionJkb:
# Description

For the complete fix going forward (see this [forum post](https://forum.balancer.fi/t/reentrancy-vulnerability-scope-expanded/4345) for context), we need to protect certain external functions that rely on balances (specifically, `setTargets` and `setSwapFeePercentage`).

I believe testing will be covered by fork tests as part of the deployments.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/pool-linear/contracts/LinearPool.sol
```diff
@@ -124,6 +125,18 @@ abstract contract LinearPool is ILinearPool, IGeneralPool, IRateProvider, NewBas
     event SwapFeePercentageChanged(uint256 swapFeePercentage);
     event TargetsSet(IERC20 indexed token, uint256 lowerTarget, uint256 upperTarget);
 
+    modifier whenNotInVaultContext() {
```
jubeira, line 128: Maybe add docs to the modifier too?

This is what we wrote in the patch:

    /**
     * @dev Ensure we are not in a Vault context when this function is called, by attempting a no-op internal
     * balance operation. If we are already in a Vault transaction (e.g., a swap, join, or exit), the Vault's
     * reentrancy protection will cause this function to revert.
     *
     * The exact function call doesn't really matter: we're just trying to trigger the Vault reentrancy check
     * (and not hurt anything in case it works). An empty operation array with no specific operation at all works
     * for that purpose, and is also the least expensive in terms of gas and bytecode size.
     *
     * Use this modifier with any function that can cause a state change in a pool and is either public itself,
     * or called by a public function *outside* a Vault operation (e.g., join, exit, or swap).
     * See https://forum.balancer.fi/t/reentrancy-vulnerability-scope-expanded/4345 for reference.
     */

jubeira (commented):
I've reviewed our plan checking all the calls once again and I agree with these exact functions being protected.

I'd consider improving the docs on `getRate` and `getVirtualSupply` adding larger warning signs as we did in the patches, considering those two cannot be protected.

For `rebalance`, `rebalanceWithExtraMain` and `initialize` I think it's fine not to add anything, since those alter the state and are already protected.

jubeira (approved):
Looking good!

pkg/interfaces/contracts/pool-linear/ILinearPool.sol
```diff
@@ -50,4 +50,35 @@ interface ILinearPool is IBasePool {
...
+     * @notice Set the lower and upper bounds of the zero-fee trading range for the main token balance.
+     * @dev For a new target range to be valid:
+     *      - the current balance must be between the current targets (meaning no fees are currently pending)
+     *      - the current balance must be between the new targets (meaning setting them does not create pending fees)
+     *
+     * The first requirement could be relaxed, as the LPs actually benefit from the pending fees not being paid out,
+     * but being stricter makes analysis easier at little expense.
+     *
+     * This is a permissioned function, and will revert when called within a Vault context (i.e. in the middle
+     * of a join or an exit).
```
nventuro, line 64:
```suggestion
     * This is a permissioned function, and will revert when called within a Vault context (i.e. in the middle
     * of a join or an exit).
```

I'd split these two concepts, as they are unrelated. Same thing in the other function

nventuro (approved):
Looks good! This should be all we need to publish

PR #2220, opened 2023-02-08
Do weighted and stable patch deploy

nventuro:
# Deployment: WeightedPool and ComposableStablePool

- [x] Contract(s) deployed to all target networks / testnets
- [x] TX IDs are properly generated
- [x] Action IDs are generated after deployment (if applicable) 
- [x] Contracts are verified in every network 
- [x] Deployments `CHANGELOG` is updated
- [x] There are no code changes of any kind


jubeira (approved):
Looks good!
OP verification for CSP is pending; we should retry in a while.

All the rest seems to be just fine.

PR #2244, opened 2023-02-10
Extract executor role

### LINKED ISSUE #2240: Improve 'executor' role
As per https://github.com/balancer-labs/balancer-v2-monorepo/issues/2239, this permission needs to be isolated from the regular permissions data structure.

Executors can only execute a single scheduled execution (there's no global executors), and the set of executors is determined at schedule time by the schedule - the set is immutable afterwards (unlike cancellers, see https://github.com/balancer-labs/balancer-v2-monorepo/issues/2238).

nventuro:
# Description

This extracts executor permissions into their own mapping, which means that it is now clear that the list of executors for a given scheduled execution id is immutable, and only set when scheduling. I also added an event to clearly identify these.

The tests could be improved (e.g. by testing that all executors work), but I didn't want to introduce too many changes there.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [x] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #2240

### DIFF
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -416,6 +416,13 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
     }
 
     /**
+     * @notice Returns true if `account` is an executor for `scheduledExecutionId`.
+     */
+    function isExecutor(uint256 scheduledExecutionId, address account) public view returns (bool) {
+        return _isExecutor[scheduledExecutionId][account];
+    }
+
+    /**
      * @notice Returns true if execution `scheduledExecutionId` can be executed.
      * Only true if it is not already executed or cancelled, and if the execution delay has passed.
      */
@@ -570,7 +577,7 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
         if (scheduledExecution.protected) {
             // Protected scheduled executions can only be executed by a set of accounts designated by the original
             // scheduler.
-            _require(_isExecutor[scheduledExecutionId][msg.sender], Errors.SENDER_NOT_ALLOWED);
+            _require(isExecutor(scheduledExecutionId, msg.sender), Errors.SENDER_NOT_ALLOWED);
         }
 
         scheduledExecution.executed = true;
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -97,7 +97,6 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
     // solhint-disable var-name-mixedcase
     bytes32 public immutable GRANT_ACTION_ID;
     bytes32 public immutable REVOKE_ACTION_ID;
-    bytes32 public immutable EXECUTE_ACTION_ID;
     bytes32 public immutable SCHEDULE_DELAY_ACTION_ID;
 
     // These action ids do not need to be used by external actors as the action ids above do.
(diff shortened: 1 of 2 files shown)


nventuro:
We could also extend this by storing an enumerable set for executors of each action (to make discoverabilty easier and not have to rely on events so much), but this is one of the simpler roles, and we may not be able to do this for all roles, so I wanted to wait until we see those before making that change.

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -111,17 +110,27 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
     IAuthorizerAdaptor private immutable _authorizerAdaptor;
     uint256 private immutable _rootTransferDelay;
 
+    // Authorizer permissions
     address private _root;
     address private _pendingRoot;
-    ScheduledExecution[] private _scheduledExecutions;
+    mapping(uint256 => mapping(address => bool)) private _isExecutor;
```
jubeira, line 116: Maybe add a comment describing what the outer / inner mappings represent?
For simple mappings the variable name is usually enough, but in this case a comment would help.

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -111,17 +110,27 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
...
+
     /**
      * @notice Emitted when a new execution `scheduledExecutionId` is scheduled.
      */
     event ExecutionScheduled(bytes32 indexed actionId, uint256 indexed scheduledExecutionId);
 
+    /**
+     * @notice Emitted when an executor is created for a scheduled execution `scheduledExecutionId`.
+     */
+    event ExecutorCreated(uint256 indexed scheduledExecutionId, address executor);
```
jubeira, line 132: do we want to index the executor as well?

pkg/vault/test/authorizer/TimelockAuthorizer.test.ts
```diff
@@ -1845,7 +1845,7 @@ describe('TimelockAuthorizer', () => {
 
               context('when an executor is specified', () => {
                 sharedBeforeEach('set executors', async () => {
-                  executors = [root];
+                  executors = [other];
```
jubeira, line 1848: can we are more than one for the test?

jubeira (commented):
This looks pretty good and clean.
I've done a first pass and left a few really minor comments, mostly about docs.

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
EndymionJkb: Something like:
// scheduledExecutionId => (executorAddress => true/false)

EndymionJkb (commented):
Seems like the executor should be indexed; it will likely be a known address (multisig, etc.)

nventuro:
Thanks! Addressed the comments.

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -562,15 +565,21 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
...
-        require(block.timestamp >= scheduledExecution.executableAt, "ACTION_NOT_EXECUTABLE");
+        require(block.timestamp >= scheduledExecution.executableAt, "ACTION_NOT_YET_EXECUTABLE");
+
         if (scheduledExecution.protected) {
-            bytes32 executeScheduledActionId = getExecuteExecutionActionId(scheduledExecutionId);
-            bool isAllowed = hasPermission(executeScheduledActionId, msg.sender, address(this));
-            _require(isAllowed, Errors.SENDER_NOT_ALLOWED);
+            // Protected scheduled executions can only be executed by a set of accounts designated by the original
+            // scheduler.
+            _require(_isExecutor[scheduledExecutionId][msg.sender], Errors.SENDER_NOT_ALLOWED);
```
jubeira, line 573: Would it make sense to have a public getter for `isExecutor(id, account)`? It might make things a bit easier to read.

jubeira (approved):
One last question to think about, otherwise looks good!

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
nventuro: Good call, yes - otherwise this information would only be available via events.

PR #2245, opened 2023-02-11
Artifact links

EndymionJkb:
# Description

The recent flurry of Linear Pool deployments, and associated excruciating analysis of all the files and verification processes that go into that, uncovered an inordinate number of broken links in readme files. This fixes those.

The next PR introduces a check to ensure we didn't miss any.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira (commented):
Changes look good; with this the readme files should reflect the actual artifacts present in deployments.

I have a side-question: should we have one artifact per output? This should not be strictly necessary since we haven't been doing it so far, and everything is generated from the existing build info files in any case. 

I've noticed that in some pools (e.g. weighted / composable v3) we have artifacts for pool and factory, whereas for linear pools we just have the factory. This is because a custom number of optimization runs generate multiple build-info files, which translates into multiple artifacts.

I don't think this has a practical impact, and maybe just fixing the links is the way forward.

EndymionJkb:
> Changes look good; with this the readme files should reflect the actual artifacts present in deployments.
> 
> I have a side-question: should we have one artifact per output? This should not be strictly necessary since we haven't been doing it so far, and everything is generated from the existing build info files in any case.
> 
> I've noticed that in some pools (e.g. weighted / composable v3) we have artifacts for pool and factory, whereas for linear pools we just have the factory. This is because a custom number of optimization runs generate multiple build-info files, which translates into multiple artifacts.
> 
> I don't think this has a practical impact, and maybe just fixing the links is the way forward.

This should just make sure that if there's an artifact, there's a link. I believe it does work for the case with multiple artifacts. (We can test after merging the PR that fixes up the existing links.)

jubeira (approved):
Alright, so this PR is correct because it fixes broken links and removes the ones that are not present.

If we want to have artifacts for every output (e.g. pool _and_ factory for every pool type) we can add the artifacts with their respective links later.

Issue #2197, opened 2023-02-06
Add broken MD links check to GH actions


jubeira:
See https://github.com/gaurav-nelson/github-action-markdown-link-check.

Should be particularly useful for deployments.

PR #2246, opened 2023-02-11
Verify links

EndymionJkb:
# Description

Building on #2245, this extends the artifact check to ensure we have a link to the artifact in the readme: and no links to artifacts that are not present.

It also adds a second, more general check on the readme, to ensure that any (local) links present are live, including ".."-relative directories, and the main README file.

This should fail until all link issues are corrected.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira:
This is pretty cool!

Btw have you seen #2197? I think there are pros and cons about having the code in-house (I'm personally not against it). I've done a first pass and code-wise it looks reasonable; if we decide to go this way I'll do a second pass.

EndymionJkb:
> This is pretty cool!
> 
> Btw have you seen #2197? I think there are pros and cons about having the code in-house (I'm personally not against it). I've done a first pass and code-wise it looks reasonable; if we decide to go this way I'll do a second pass.

I didn't! I'll link the issue. Rab noted they were off on the Aave deployment, so I checked the rest. This is a very targeted check, though, just on the artifact, so it's part of the existing artifact checking task: this is one more thing to verify.

The other one is a more general link check: but again, only for the readme file, and only for internal links: it doesn't try to verify external http links.

pkg/deployments/src/links.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    if (linkCnt > 0) {
+      logger.success(`Verified ${linkCnt} links in task '${task.id}'`);
+    }
+  } else {
+    const isTopLevelFolder = ['deprecated', 'scripts'].includes(path.basename(task.dir()));
+    if (!isTopLevelFolder) {
+      throw Error(`Missing readme.md for task '${task.id}'`);
+    }
```
nventuro, line 45: When would this not be the case?

nventuro (commented):
This is a great idea, thanks!

The logic is a bit complex, and I'm not a big fan of having to parse files looking for links etc. I'd give https://github.com/balancer-labs/balancer-v2-monorepo/issues/2197 a go and see if we can use that to find broken links, and then keep the snippet here were we check that all artifacts do indeed have their link.

pkg/deployments/src/links.ts
EndymionJkb: When it recurses through the directory, it includes just "/deprecated", along with "/deprecated/<task>". Same with scripts. There is no /deprecated/readme.md, but all the /deprecated/<task> directories should have readme's

EndymionJkb:
Superseded by #2261

Issue #2242, opened 2023-02-10
Improve handling of granters and revokers


nventuro:
(part of #2239)

These should look fairly similar to what we currently have (instant, done by root), paired with grant/schedule grant functions (and revoke), except their storage should be separate (as per #2239), and I'm not entirely sold on the notion of general granters.

Global granters make sense, as they can e.g. grant permission to set the swap fee on any pool, but general granters seem too broad. The only reasonable use I can think of for them is that of migrator scripts, but I'm not sure that scenario is worth the extra complexity. We need to revise past governance actions, migrations, etc., and figure out if we actually need this.

nventuro:
Another notion of that of destroying granters/revokers. We currently allow granters and revokers to remove their peers (not create them), but I'm not entirely sure this is needed. It doesn't seem like a security issue, but it does make their scope broader in a bit of a strange way.

nventuro:
A potential use case for general revokers is to e.g. let the emergency subdao revoke any permission of certain accounts (such as those that operate some pieces of infrastructure).

nventuro:
Also, with granting and revoking becoming separate from normal permissions, we'll need to create special storage for their delays. Granting and revoking the same permission should have different delays.

nventuro:
As pointed out in https://github.com/balancer-labs/balancer-v2-monorepo/issues/2241, this blocks a few other tasks, but unfortunately is also one of the largest ones. It should likely be tackled first, but we may want to split it into smaller chunks.

PR #2250, opened 2023-02-13
Only let root make and destroy granters and revokers

nventuro:
# Description

This makes it so only root can manage granters and revokers. We previously allowed for them to destroy their peers, but I found no reason to support this.

This is its own PR mostly to keep changes to the tests sane (which as you can probably see are not very sane to begin with).

Part of https://github.com/balancer-labs/balancer-v2-monorepo/issues/2242

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [x] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


nventuro:
This requires https://github.com/balancer-labs/balancer-v2-monorepo/pull/2255 to be merged so we don't need to fix those tests.

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -614,14 +614,18 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
     }
 
     /**
-     * @notice Sets `account`'s granter status to `allowed` for action `actionId` in target `where`.
-     * @dev Note that granters can revoke the granter status of other granters, even removing the root.
-     * However the root can always rejoin, and then remove any malicious granters.
+     * @notice Grants or revokes granter status to `account`'s for action `actionId` in target `where`.
```
EndymionJkb, line 617:
```suggestion
     * @notice Grants or revokes granter status to `account` for action `actionId` in target `where`.
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -614,14 +614,18 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
...
      * Note that there are no delays associated with adding or removing granters. This is based on the assumption that
      * any action which a malicous user could exploit to damage the protocol will have a sufficiently long delay
      * associated with either granting permission for or exercising that permission such that the root will be able to
-     * reestablish control and cancel the action before it can be executed.
+     * reestablish control and cancel either the granting or associated action before it can be executed, and then
+     * remove the granter.
+     *
+     * A malicious granter may also attempt to use their granter status to grant permission to multiple accounts, but
+     * they cannot create new granters. Therefore, root can simply first revoke their granter status, and then revoke
+     * the permissions granted to said no longer increasing set of accounts.
```
EndymionJkb, line 628:
```suggestion
     * they cannot create new granters. Therefore, the danger posed by a malicious granter is limited and self-
     * contained. Root can mitigate the situation simply and completely by revoking first their granter status,
     * and then any permissions granted by that account, knowing there cannot be any more.
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -674,25 +676,23 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
     }
 
     /**
-     * @notice Sets `account`'s revoker status to `allowed` for action `actionId` in target `where`.
-     * @dev Note that revokers can revoke the revoker status of other revokers, even banning the root.
-     * However the root can always rejoin, and then remove any malicious revokers.
+     * @notice Grants or revokes revoker status to `account`'s for action `actionId` in target `where`.
```
EndymionJkb, line 679:
```suggestion
     * @notice Grants or revokes revoker status to `account` for action `actionId` in target `where`.
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -674,25 +676,23 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
...
+     * @dev Only the root can add and remove revokers.
      *
      * Note that there are no delays associated with adding or removing revokers. This is based on the assumption that
      * any permissions for which revocation from key addresses would be dangerous (e.g. preventing the BalancerMinter
      * from minting BAL) have sufficiently long delays associated with revoking them that the root will be able to
      * reestablish control and cancel the revocation before the scheduled revocation can be executed.
+     *
+     * A malicious cannot create new revokers, so root can simply revoke their status once.
```
EndymionJkb, line 687:
```suggestion
     * A malicious revoker cannot create new revokers, so root can simply revoke their status once.
```

pkg/vault/test/authorizer/TimelockAuthorizer.test.ts
```diff
@@ -375,72 +375,13 @@ describe('TimelockAuthorizer', () => {
           context('for a specific contract', () => {
             const where = WHERE_1;
 
-            context('when the sender has permission', () => {
-              beforeEach('grant permission', async () => {
-                await authorizer.addGranter(actionId, from, where, { from: root });
-              });
-
-              it('can grant permission for that action in that contract only', async () => {
```
jubeira, line 383: [Nothing to address here, just question] Could it be that these test messages were out of sync with the actual tests?

pkg/vault/test/authorizer/TimelockAuthorizerRoot.test.ts
```diff
@@ -128,45 +128,5 @@ describe('TimelockAuthorizer', () => {
...
-      await authorizer.removeGranter(GENERAL_PERMISSION_SPECIFIER, root, EVERYWHERE, { from: grantee });
-      expect(await authorizer.canGrant(GENERAL_PERMISSION_SPECIFIER, root, WHERE_1)).to.be.false;
-      expect(await authorizer.canGrant(GENERAL_PERMISSION_SPECIFIER, root, EVERYWHERE)).to.be.false;
-
-      await authorizer.addGranter(GENERAL_PERMISSION_SPECIFIER, root, EVERYWHERE, { from: root });
-      expect(await authorizer.canGrant(GENERAL_PERMISSION_SPECIFIER, root, WHERE_1)).to.be.true;
-      expect(await authorizer.canGrant(GENERAL_PERMISSION_SPECIFIER, root, EVERYWHERE)).to.be.true;
-    });
-
-    it('cannot have their global grant permissions revoked by an authorized address for a specific contract', async () => {
```
jubeira, line 144: Same as above; why not keep this one? In this case it would be the same right?

jubeira (commented):
First pass done.
Solidity code changes look good.

About the test: I'd suggest keeping the ones in the `TimelockAuthorizerRoot` tests, adjusting them if necessary. If there are things that we now don't want to allow with these changes, we should probably keep tests that verify that behavior instead of removing them.

About the changes in `TimelockAuthorizer` itself: overall they look good, although I'm a bit confused by the test titles.

pkg/vault/test/authorizer/TimelockAuthorizer.test.ts
nventuro: Yes, the original tests were a bit of a mess

pkg/vault/test/authorizer/TimelockAuthorizerRoot.test.ts
```diff
@@ -128,45 +128,5 @@ describe('TimelockAuthorizer', () => {
       expect(await authorizer.canRevoke(GENERAL_PERMISSION_SPECIFIER, grantee, WHERE_1)).to.be.false;
       expect(await authorizer.canRevoke(GENERAL_PERMISSION_SPECIFIER, grantee, EVERYWHERE)).to.be.false;
     });
-
-    it('can have their global grant permissions revoked by an authorized address for any contract', async () => {
```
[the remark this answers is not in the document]
nventuro: Because there's nothing special about granters that'd let them do it for us to test this :shrug: If you hadn't seen the prior version of the code, you wouldn't think to write this test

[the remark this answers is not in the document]
jubeira: correct, but I've seen it :grimacing: 

My point is that we are coming from that version, and we could've made a mistake while doing so. But yes, if we went straight to the current state we'd never have written that test so it's also fine not to have it.

jubeira (approved):

PR #2255, opened 2023-02-14
Delete transition migrator

nventuro:
This deletes the timelock transition migrator, which we won't use moving forward (and will be incompatible with some aspects of the new design, so it's easier to get rid of it).


jubeira (approved):

jubeira:
Long live transition migrator.

PR #2261, opened 2023-02-15
Try third-party markdown checker

### LINKED ISSUE #2197: Add broken MD links check to GH actions

EndymionJkb:
# Description

Trying the third-party tool, retaining only the artifact check. Variation of #2246.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

- [X] The diff is legible and has no extraneous changes - (well, only one small fix. Rest fixed in previous PRs)

## Issue Resolution

Closes #2197


pkg/deployments/src/artifact.ts
```diff
@@ -58,6 +74,28 @@ export function checkArtifact(task: Task): void {
...
+
+  for (const line of readmeLines) {
+    // Tolerate differences in case (e.g., veBoost vs VeBoost).
+    if (line.toLowerCase() == expectedContent.toLowerCase()) {
+      found = true;
+      break;
+    }
+  }
+
+  if (found) {
```
jubeira, line 92:
```suggestion
  const linkFound = readmeLines.find((line) => line.toLowerCase() == expectedContent.toLowerCase()) !== undefined;

  if (linkFound) {
```

Slightly more compact version

pkg/deployments/src/artifact.ts
```diff
@@ -37,6 +38,15 @@ function _extractArtifact(task: Task, file: string, contract?: string) {
...
  */
 export function checkArtifact(task: Task): void {
+  let readmeLines: string[] = [];
+
+  const filePath = path.join(task.dir(), `readme.md`);
+  const readmeFileExists = fs.existsSync(filePath) && fs.statSync(filePath).isFile();
+  if (readmeFileExists) {
+    const readmeContents = fs.readFileSync(filePath).toString();
+    readmeLines = readmeContents.split('\n');
+  }
```
jubeira, line 48: I think you could also do this inside `checkReadmeArtifactLink`, since you only need the `readmeLines` inside there, and you are already doing `if (readmeFileExists)` below. 

Instead of sending the lines, you'd send the file path, which is also available.

jubeira (commented):
I think this makes sense and it'll help a lot.

I've left two minor improvement suggestions; otherwise looks good.

Some more thoughts:
- I think we should also modify the templates for the deployment preparation / deployment. If we add the expected output links in the preparation stage, won't the CI fail with this? 
- [Not for this PR] Eventually it'd be good to have an automated check that verifies which artifacts should be present (e.g. for a pool deployment, there should be one for the pool and one for the factory at the very least, which is something we don't enforce today).

pkg/deployments/src/artifact.ts
EndymionJkb: That's an optimization so that I don't have to read the file (a slow operation) more than once. Usually there's only one artifact, so it doesn't make much difference, but if there are three or four (happens sometimes), it would read the file multiple times.

EndymionJkb:
> Some more thoughts:
> 
> * I think we should also modify the templates for the deployment preparation / deployment. If we add the expected output links in the preparation stage, won't the CI fail with this?

Yes; we're hitting exactly this case with the recent Yearn deployment lol. Thanks, good catch!

jubeira (approved):
:rocket:

PR #2265, opened 2023-02-16
Deployment: Gearbox Linear Pool Factory

joaobrunoah:
# Deployment: `GearboxLinearPoolFactory`

# Deployment tasks checklist:

- [x] Contract(s) deployed to all target networks / testnets
- [x] TX IDs are properly generated
- [x] Action IDs are generated after deployment (if applicable)
- [x] Contracts are verified in every network
- [x] Deployments CHANGELOG is updated
- [x] There are no code changes of any kind - Just a minor adjustment in the deployment index.ts to correctly build the mocked token

# Issue Resolution

N/A


jubeira (approved):
This looks good! I've executed the failing tests locally and they pass (I see no reason for them not to work with these changes); I believe they don't work very well from forks?

Contract verification went well too.

nventuro:
Yes, they won't work from forks because they require access to an archive node, and the API keys for that are stored as repository secrets. PRs created from outside won't have access to those, and will therefore be unable to make the archive node queries.

PR #2267, opened 2023-02-16
Add FeeDistributor recipient option for SingleRecipientGauge

nventuro:
# Description

This adds a `feeDistributorRecipient` flag to the constructor of the gauge. If set to true, it expects for the recipient to implement the `depositTokens` function, which is in turn expected to transfer tokens from the gauge into itself. This will be highly useful for projects that intend to directly distribute tokens using a `FeeDistributor`-like contract.

I didn't bother taking this function out of the `FeeDistributor` interface and have it be its own thing as it seemed way overkill. We may at some point, but I don't think we need to do it _now_. The dependency is extremely small anyway.

I also had to get rid of the `IBaseGaugeFactory`, which assumed all gauges would feature the same `create` function, which proves to be an incorrect assumption. I don't think we'd have any programmatic gauge creation anyway, so this is not a big loss.

We don't have unit tests for this gauge (though it's getting to the point where maybe we should), so I extended the fork tests instead. These are likely best reviewed hiding whitespace changes.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [x] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/deployments/tasks/20230215-single-recipient-gauge-factory-v2/readme.md
```diff
@@ -3,6 +3,8 @@
 Deployment of the `SingleRecipientGaugeFactory`, for liquidity gauges which forward their entire BAL allowance onto a single address.
 This version adds a weight cap to single recipient gauges, and supersedes [single recipient gauge factory V1](../deprecated/20220325-single-recipient-gauge-factory/).
 
+Additionally, this version also supports the recipient implementing the `depositToken` function (from the `FeeDistributor` interface), directly deposoting the tokens in such a distributor instead of requiring a manual step after minting.
```
jubeira, line 6:
```suggestion
Additionally, this version also supports the recipient implementing the `depositToken` function (from the `FeeDistributor` interface), directly depositing the tokens in such a distributor instead of requiring a manual step after minting.
```

pkg/deployments/tasks/20230215-single-recipient-gauge-factory-v2/test/task.fork.ts
```diff
@@ -123,168 +124,207 @@ describeForkTest('SingleRecipientGaugeFactory V2', 'mainnet', 16627100, function
...
-    // relative weight in the controller.
-    const calldata = gaugeController.interface.encodeFunctionData('add_gauge(address,int128)', [gauge.address, 2]);
-    await authorizerAdaptor.connect(admin).performAction(gaugeController.address, calldata);
-
-    expect(await gaugeController.gauge_exists(gauge.address)).to.be.true;
-  });
+  function itWorksLikeACappedSingleRecipientGauge(mode: RecipientMode) {
+    // We're going to create gauges, vote for them, have time pass, etc. Because of that, we take a snapshot before we
+    // do any of that, and restore it at the end.
+    let snapshot: SnapshotRestorer;
```
jubeira, line 143: uhmm why not just use an empty `sharedBeforeEach`?

pkg/interfaces/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Breaking changes
```
jubeira, line 5: :sadge: :upside_down_face:

pkg/liquidity-mining/test/GaugeRelativeWeightCap.test.ts
```diff
@@ -70,21 +69,49 @@ describe('GaugeRelativeWeightCap', () => {
...
 
-  function testRelativeWeightCapForGauge(contractName: string) {
+  function testRelativeWeightCapForGauge(kind: GaugeKind) {
+    let factory: Contract;
+
+    sharedBeforeEach(async () => {
+      factory = kind == GaugeKind.LiquidityGauge ? liquidityGaugeFactory : stakelessGaugeFactory;
+    });
+
+    async function sendGaugeDeployTx(
```
jubeira, line 92: hmm how about `createGaugeTx` ?
I find it a bit hard to wrap my head around `send` and the return type of this helper without reading it hard.

nventuro:
CI fails because of https://github.com/balancer-labs/balancer-v2-monorepo/pull/2268

jubeira (approved):
Looks pretty good!
I've left a few minor comments / suggestion, otherwise I'd say it's good to go.

I can think some ways to keep the interface without adding the extra parameter in the factory, but I'm 93% sure you won't like it so I won't even suggest it :)

pkg/deployments/tasks/20230215-single-recipient-gauge-factory-v2/test/task.fork.ts
```diff
@@ -123,168 +124,207 @@ describeForkTest('SingleRecipientGaugeFactory V2', 'mainnet', 16627100, function
...
+
+      // And then forwarded to the recipient
+      expectTransferEvent(
+        await mintTx.wait(),
+        {
+          from: gauge.address,
+          to: recipient,
+          value: actualEmissions,
+        },
+        BAL
```
jubeira, line 266: shouldn't you also test the last step in case mode is `FeeDistributorRecipient`?

pkg/deployments/tasks/20230215-single-recipient-gauge-factory-v2/test/task.fork.ts
```diff
@@ -123,168 +124,207 @@ describeForkTest('SingleRecipientGaugeFactory V2', 'mainnet', 16627100, function
...
+      // And then forwarded to the recipient
+      expectTransferEvent(
+        await tx.wait(),
+        {
+          from: gauge.address,
+          to: recipient,
+          value: gaugeTransferEvent.args.value,
+        },
+        BAL
+      );
```
jubeira, line 327: Same as above, maybe check the last step if mode is `FeeDistributorRecipient`?

jubeira:
With the working fork test, let's just remove the build info and artifacts, and add `.skip` to it.

Then, we can keep the code changes separate from the deployment preparation, and put together a PR just with the new build info.

pkg/liquidity-mining/test/GaugeRelativeWeightCap.test.ts
nventuro: What do you find hard to understand about it?

pkg/deployments/tasks/20230215-single-recipient-gauge-factory-v2/test/task.fork.ts
nventuro: Because that'd cause for each `it` to revert to the setup snapshot, which we don't want (since each `it` performs steps that we build on top of).

nventuro:
I'm not changing the build info since this won't be the final version - we'll add one small change to the gauge before deployment.

PR #2269, opened 2023-02-16
Merkle orchard: claim fixes

jubeira:
# Description

Fix for the Merkle Orchard contract, removing optimizations when claiming distributions. We don't expect to use this contract regularly; we just need it to work properly for one final distribution, so removing optimizations is fine.
Issue described in [this post](https://medium.com/immunefi/balancer-logic-error-bugfix-review-74f5edca8b1a).

Based on `merkle-orchard-v2` branch, which is branched from https://github.com/balancer-labs/balancer-v2-monorepo/commit/a8e5f93c555cc842af586f74022f48dbcd58bf67 (where the build info for Orchard V1 was generated). This is to be verified by the reviewer.

This should not be merged back into master since the contract was removed from the codebase long ago, and this is a patch release.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

See #2218.

### DIFF
--- a/pkg/distributors/contracts/MerkleOrchard.sol
+++ b/pkg/distributors/contracts/MerkleOrchard.sol
@@ -223,8 +223,6 @@ contract MerkleOrchard {
 
             (uint256 distributionWordIndex, uint256 distributionBitIndex) = _getIndices(claim.distributionId);
 
-            // Start a new claims set
-            // Since channel ids are a hash, the initial zero id can be safely considered invalid
             bytes32 currentChannelId = _getChannelId(tokens[claim.tokenIndex], claim.distributor);
             uint256 currentBits = 1 << distributionBitIndex;
             _setClaimedBits(currentChannelId, claimer, distributionWordIndex, currentBits);
--- a/pkg/distributors/contracts/MerkleOrchard.sol
+++ b/pkg/distributors/contracts/MerkleOrchard.sol
@@ -210,14 +210,8 @@ contract MerkleOrchard {
         bool asInternalBalance
     ) internal {
         uint256[] memory amounts = new uint256[](tokens.length);
-
-        // To save gas when setting claimed statuses in storage, we group claims for each channel and word index
-        // (referred to as a 'claims set'), aggregating the claim bits to set and total claimed amount, only committing
-        // to storage when changing claims sets (or when processing the last claim).
-        // This means that callers should sort claims by grouping distribution channels and distributions with the same
-        // word index in order to achieve reduced gas costs.
-
         Claim memory claim;
+
         for (uint256 i = 0; i < claims.length; i++) {
             claim = claims[i];
 
--- a/pkg/distributors/contracts/MerkleOrchard.sol
+++ b/pkg/distributors/contracts/MerkleOrchard.sol
@@ -221,7 +221,6 @@ contract MerkleOrchard {
         for (uint256 i = 0; i < claims.length; i++) {
             claim = claims[i];
 
-            // New scope to avoid stack-too-deep issues
             (uint256 distributionWordIndex, uint256 distributionBitIndex) = _getIndices(claim.distributionId);
 
             // Start a new claims set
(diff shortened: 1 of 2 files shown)


ylv-io (approved):
Excellent PR. Love removing the code and simplifying it. Great work @jubeira 🚀

jubeira:
@nventuro if you look at the commits:
- https://github.com/balancer-labs/balancer-v2-monorepo/pull/2269/commits/dd6fcc1c7fe6ded6f21c4d11bb0f6094a04179d7 adds the test that should not pass, but passes with Orchard v1
- https://github.com/balancer-labs/balancer-v2-monorepo/pull/2269/commits/521ef9632a049cdf4eae17e4b4bebc7f045117d0 fixes the contract, effectively breaking the test
- https://github.com/balancer-labs/balancer-v2-monorepo/pull/2269/commits/1dd9b5e6f56c2820e4f1861a03530d28787aab93 inverts the logic of the test so that it verifies the fix now

pkg/distributors/contracts/MerkleOrchard.sol
```diff
@@ -217,63 +217,19 @@ contract MerkleOrchard {
...
-        uint256 currentWordIndex;
-
-        uint256 currentBits; // The accumulated claimed bits to set in storage
-        uint256 currentClaimAmount; // The accumulated tokens to be claimed from the current channel (not claims set!)
-
         Claim memory claim;
         for (uint256 i = 0; i < claims.length; i++) {
             claim = claims[i];
 
             // New scope to avoid stack-too-deep issues
```
nventuro, line 224:
```suggestion
```

pkg/distributors/contracts/MerkleOrchard.sol
```diff
@@ -217,63 +217,19 @@ contract MerkleOrchard {
...
-            // Since a claims set is only committed if the next one is not part of the same set, the last claims set
-            // must be manually committed always.
-            if (i == claims.length - 1) {
-                _setClaimedBits(currentChannelId, claimer, currentWordIndex, currentBits);
-                _deductClaimedBalance(currentChannelId, currentClaimAmount);
-            }
+            (uint256 distributionWordIndex, uint256 distributionBitIndex) = _getIndices(claim.distributionId);
+
+            // Start a new claims set
+            // Since channel ids are a hash, the initial zero id can be safely considered invalid
```
nventuro, line 228:
```suggestion
```

pkg/distributors/contracts/MerkleOrchard.sol
```diff
@@ -217,63 +217,19 @@ contract MerkleOrchard {
         // This means that callers should sort claims by grouping distribution channels and distributions with the same
         // word index in order to achieve reduced gas costs.
```
nventuro, line 218: This comment should also be removed

pkg/distributors/test/MerkleOrchard.test.ts
```diff
@@ -301,6 +307,46 @@ describe('MerkleOrchard', () => {
...
+        ];
+        merkleTree = new MerkleTree(elements);
+        const root = merkleTree.getHexRoot();
+
+        await merkleOrchard
+          .connect(distributor)
+          .createDistribution(token1.address, root, claimableBalance.mul(2), distributionId);
+        merkleProof = merkleTree.getHexProof(elements[0]);
+
+        claims = [
```
nventuro, line 326: While not an issue here, this is not a good pattern. You share the `merkleTree`, `claims`, etc variables accross multiple scopes, but here set them to a different value. If you don't have a similar `beforeEach` block that sets them in other scopes, they'll be affected by these changes.

I find it better to restrict the scope of the variables to where they're set/used to avoid these things (even if in this case that'd mean every block having their own version of `claims` etc)

nventuro (approved):
Looks great, thanks a lot both of you! :raised_hands: 

I also checked that compiling at commit `a8e5f93c555cc842af586f74022f48dbcd58bf67` produces the build info of the `20211012-merkle-orchard` task.

ylv-io:
I confirm that building `distributors` at commit `a8e5f93c555cc842af586f74022f48dbcd58bf67` produces the build info of the `20211012-merkle-orchard` task.

image

nventuro (approved):

PR #2270, opened 2023-02-16
Fix: Readme link CI check

jubeira:
CI fix.


EndymionJkb:
Could this be intermittent? It's a valid link, and I'd thought we had other etherscan links (apparently not - they're just in comments, it turns out).

Maybe try this for these "too many requests" issues?

        with:
          use-quiet-mode: 'yes'
          retryon429: 'yes'
          retry-after: '30s'

jubeira:
> Maybe try this for these "too many requests" issues?

Yes, it's valid indeed, but the error in master was actually 403 (forbidden).
It looks like the 'retry' feature is only for 429s, and today I've re-run CI and it didn't have this issue.

Let's close for now, and monitor how often this happens.

PR #2271, opened 2023-02-17
Improve Timelock revert reasons

nventuro:
# Description

This adds dedicated revert reasons for each failure type. On top of improving UX, it also increases confidence in the test suite as each failure mode now gets a different error message. This helped me realize that some of the test removals in https://github.com/balancer-labs/balancer-v2-monorepo/pull/2250 might have not been correct (i.e. we deleted too much).

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -574,7 +574,7 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
         require(where != address(_executor), "ATTEMPTING_EXECUTOR_REENTRANCY");
 
         bytes32 actionId = IAuthentication(where).getActionId(_decodeSelector(data));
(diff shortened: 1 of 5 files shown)


EndymionJkb (approved):
Looks a lot more specific. Do we have guidelines for when to use Errors vs. when to have contract-specific hard-coded text? I guess when there are a large number that are too specific?

jubeira (approved):
Looks sensible and consistent.

Two questions:
- Is there a reason not to add these to the standard balancer errors?
- Do you plan to restore some of the deleted tests in a separate PR?

nventuro:
>Do we have guidelines for when to use Errors vs. when to have contract-specific hard-coded text? I guess when there are a large number that are too specific?

Strings are better, they just take up more contract space. This is doubly true for user-facing contracts (like the Authorizer, where good UX is critical). We have plently of space here.

>Do you plan to restore some of the deleted tests in a separate PR?

Yes

PR #2273, opened 2023-02-17
Avax deployment

ylv-io:
# Deployment: full system to the Avalanche network.

Deployments package and CI infrastructure also updated to accommodate the new network.

`ComposableStablePool` and `LinearPool` factories could not be deployed, since they exceeded the fix block gas limit

- [X] Contracts are verified in every network 


nventuro:
Composable v3 and LinearPools cannot be deployed due to the Avax block size being 8M. They'll upgrade to 15M in March, at which point we'll be able to complete the deployment.

pkg/deployments/action-ids/avalanche/action-ids.json
```diff
@@ -0,0 +1,151 @@
...
+    "PoolRecoveryHelper": {
+      "useAdaptor": false,
+      "actionIds": {
+        "addPoolFactory(address)": "0x90ce8036a67bbe2451d4f50ab7353343339db901728ae2de09928006819ee686",
+        "enableRecoveryMode(address)": "0xd803b6aadd1ec41e42e680b246ebc3f68425ee6cfba1883b53b8fd01e7f37c6b",
+        "removePoolFactory(address)": "0x25cbdd96392463c2c0b1e5a2db364a18327ef561b6ceeb4a444b694dbf2c7890"
+      }
+    }
+  },
+  "20221124-authorizer-adaptor-entrypoint": {
```
EndymionJkb, line 121: Looks like we're missing the action-ids for the regular authorizer-adaptor

nventuro: Actually neither is permissioned - this entry shouldn't be here

EndymionJkb: Should they be removed from all the other networks too, then? (They just have "performAction".) All of them have entries for both.

nventuro: It doesn't hurt to have extra stuff here, but sure, go for it. We actually have lots of entries in these jsons that don't do anything (e.g. for functions that are not permissioned).

You could also add it to `ignored-functions.json`.

nventuro:
Merging since we can begin using some of this infra.

nventuro (approved):

PR #2280, opened 2023-02-23
Deployment: GnosisRootGaugeFactory

EndymionJkb:
# Deployment: GnosisRootGaugeFactory, and supporting contracts on Gnosis

- [~] Contracts are verified in every network 


pkg/deployments/tasks/20220413-child-chain-gauge-factory/output/gnosis.json
```diff
@@ -0,0 +1,5 @@
+{
+  "RewardsOnlyGauge": "0x45fFd460cC6642B8D8Fb12373DFd77Ceb0f4932B",
+  "ChildChainStreamer": "0x230a59F4d9ADc147480f03B0D3fFfeCd56c3289a",
```
jubeira, line 3: Is there a reason not to verify these two?
Maybe we should change `deploy` with `deployAndVerify` in the deployment script and execute them again.

nventuro: Verification doesn't work for vyper contracts, it has to be done manually. Thankfully it's easy to do since Vyper is single file

nventuro (approved):

nventuro (approved):

nventuro:
CI seems to be failing, but I checked both action ids and addresses locally on gnosis. Merging to see if that solves the issue?

Issue #2238, opened 2023-02-10
Create 'canceller' role


nventuro:
Currently only executors can cancel scheduled actions. We want to be able to create a canceler role, which would let accounts with it cancel actions scheduled by other parties (to be wielded by e.g. the emergency subDAO).

Tentative specs:
 - only root can create and destroy cancelers
 - destruction and creation are immediate
 - cancelling is also immediate
 - granular: can cancel either a specific execution id, or can globally cancel anything. the scheduler could be granted permission to cancel the action it is creating.

nventuro:
See https://github.com/balancer-labs/balancer-v2-monorepo/issues/2239 for context as to how this should be done.

nventuro:
Closed via #2297

PR #2297, opened 2023-02-24
Auth canceler

ylv-io:
# Description

Currently only executors can cancel scheduled actions. We want to be able to create a canceler role, which would let accounts with it cancel actions scheduled by other parties (to be wielded by e.g. the emergency subDAO).

Specs:
* only root can create and destroy cancelers
* destruction and creation are immediate
* cancelling is also immediate
* granular: can cancel either a specific execution id, or can globally cancel anything. 
* the scheduler is granted permission to cancel the action it is creating.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes issue https://github.com/balancer-labs/balancer-v2-monorepo/issues/2238

### DIFF
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -79,7 +79,9 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
     address public constant EVERYWHERE = address(-1);
 
     /**
-     * @notice A constant value for `scheduledExecutionId` that will match any execution Id
+     * @notice A constant value for `scheduledExecutionId` that will match any execution Id. Cancelers assigned to this Id
+     * will be able to cancel *any* scheduled action, which is very useful for e.g. emergency response dedicated
+     * teams that analyze these.
      */
     uint256 public constant GLOBAL_CANCELER_SCHEDULED_EXECUTION_ID = type(uint256).max;
 
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -470,7 +470,7 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
         require(isRoot(msg.sender), "SENDER_IS_NOT_ROOT");
         bytes32 actionId = getActionId(this.setPendingRoot.selector);
         bytes memory data = abi.encodeWithSelector(this.setPendingRoot.selector, newRoot);
-        return _scheduleWithDelay(actionId, address(this), data, getRootTransferDelay(), executors, msg.sender);
+        return _scheduleWithDelay(actionId, address(this), data, getRootTransferDelay(), executors);
     }
 
     /**
@@ -557,7 +557,7 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
 
         bytes32 scheduleDelayActionId = getScheduleDelayActionId(actionId);
         bytes memory data = abi.encodeWithSelector(this.setDelay.selector, actionId, newDelay);
-        return _scheduleWithDelay(scheduleDelayActionId, address(this), data, executionDelay, executors, msg.sender);
+        return _scheduleWithDelay(scheduleDelayActionId, address(this), data, executionDelay, executors);
     }
 
     /**
@@ -590,7 +590,15 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
 
         bytes32 actionId = IAuthentication(where).getActionId(_decodeSelector(data));
         require(hasPermission(actionId, msg.sender, where), "SENDER_DOES_NOT_HAVE_PERMISSION");
-        return _schedule(actionId, where, data, executors);
+
+        uint256 id = _schedule(actionId, where, data, executors);
+        // accounts that schedule actions are automatically made cancelers for them, so that they can manage their
+        // action. we check that they are not already a canceler since e.g. root may schedule actions (and root is
+        // always a global canceler).
+        if (!isCanceler(id, msg.sender)) {
+            _addCanceler(id, msg.sender);
+        }
+        return id;
     }
 
     /**
@@ -791,7 +799,14 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
         bytes memory data = abi.encodeWithSelector(this.grantPermissions.selector, _ar(actionId), account, _ar(where));
 
         // TODO: fix actionId for event (maybe overhaul _scheduleWithDelay?)
-        return _scheduleWithDelay(0x0, address(this), data, delay, executors);
+        uint256 id =_scheduleWithDelay(0x0, address(this), data, delay, executors);
+        // accounts that schedule actions are automatically made cancelers for them, so that they can manage their
+        // action. we check that they are not already a canceler since e.g. root may schedule actions (and root is
+        // always a global canceler).
+        if (!isCanceler(id, msg.sender)) {
+            _addCanceler(id, msg.sender);
+        }
+        return id;
     }
 
     /**
@@ -867,7 +882,14 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
         require(isRevoker(actionId, msg.sender, where), "SENDER_IS_NOT_REVOKER");
         bytes memory data = abi.encodeWithSelector(this.revokePermissions.selector, _ar(actionId), account, _ar(where));
         bytes32 revokePermissionId = getRevokePermissionActionId(actionId);
-        return _schedule(revokePermissionId, address(this), data, executors);
+        uint256 id = _schedule(revokePermissionId, address(this), data, executors);
+        // accounts that schedule actions are automatically made cancelers for them, so that they can manage their
+        // action. we check that they are not already a canceler since e.g. root may schedule actions (and root is
+        // always a global canceler).
+        if (!isCanceler(id, msg.sender)) {
+            _addCanceler(id, msg.sender);
+        }
+        return id;
     }
 
     /**
@@ -914,7 +936,7 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
     ) private returns (uint256 scheduledExecutionId) {
         uint256 delay = _delaysPerActionId[actionId];
         require(delay > 0, "CANNOT_SCHEDULE_ACTION");
-        return _scheduleWithDelay(actionId, where, data, delay, executors, msg.sender);
+        return _scheduleWithDelay(actionId, where, data, delay, executors);
     }
 
     function _scheduleWithDelay(
@@ -922,8 +944,7 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
         address where,
         bytes memory data,
         uint256 delay,
-        address[] memory executors,
-        address sender
+        address[] memory executors
     ) private returns (uint256 scheduledExecutionId) {
         scheduledExecutionId = _scheduledExecutions.length;
         emit ExecutionScheduled(actionId, scheduledExecutionId);
@@ -947,11 +968,6 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
             // Note that we allow for repeated executors - this is not an issue
             _isExecutor[scheduledExecutionId][executors[i]] = true;
             emit ExecutorAdded(scheduledExecutionId, executors[i]);
-        }
-
-        // if the scheduler doesn't have a permission to cancel the scheduled action then grant it
-        if (!isCanceler(scheduledExecutionId, sender)) {
-            _addCanceler(scheduledExecutionId, sender);
         }
     }
 
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -887,8 +887,8 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
         bytes memory data = abi.encodeWithSelector(this.revokePermissions.selector, _ar(actionId), account, _ar(where));
         bytes32 revokePermissionId = getRevokePermissionActionId(actionId);
         uint256 id = _schedule(revokePermissionId, address(this), data, executors);
-        // accounts that schedule actions are automatically made cancelers for them, so that they can manage their
-        // action. we check that they are not already a canceler since e.g. root may schedule actions (and root is
+        // Revokers that schedule actions are automatically made cancelers for them, so that they can manage their
+        // action. We check that they are not already a canceler since e.g. root may schedule revokes (and root is
         // always a global canceler).
         if (!isCanceler(id, msg.sender)) {
             _addCanceler(id, msg.sender);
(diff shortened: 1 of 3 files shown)


EndymionJkb:
Looks good! As I've noted before in related PRs, I think it should be add/remove vs create/destroy, for the canceler terminology.

It should also test that a canceler can actually cancel things (maybe that's coming, since this is still a draft).

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -129,9 +137,19 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
     event ExecutionScheduled(bytes32 indexed actionId, uint256 indexed scheduledExecutionId);
 
     /**
-     * @notice Emitted when an executor is created for a scheduled execution `scheduledExecutionId`.
+     * @notice Emitted when an executor is added for a scheduled execution `scheduledExecutionId`.
      */
-    event ExecutorCreated(uint256 indexed scheduledExecutionId, address indexed executor);
+    event ExecutorAdded(uint256 indexed scheduledExecutionId, address indexed executor);
```
nventuro, line 142: This being a small PR this is not an issue, but I would've included this rename in a separate PR.

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -830,7 +885,12 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
         for (uint256 i = 0; i < executors.length; i++) {
             // Note that we allow for repeated executors - this is not an issue
             _isExecutor[scheduledExecutionId][executors[i]] = true;
-            emit ExecutorCreated(scheduledExecutionId, executors[i]);
+            emit ExecutorAdded(scheduledExecutionId, executors[i]);
+        }
+
+        // if the scheduler doesn't have a permission to cancel the scheduled action then grant it
+        if (!isCanceler(scheduledExecutionId, sender)) {
+            _addCanceler(scheduledExecutionId, sender);
```
nventuro, line 893: I find this snippet a bit confusing. Functionality wise it is correct, but it makes the semantics of calling this function sort of strange (in some cases the account is made a canceler, in some it isn't).

What if instead we didn't have this, and created the canceler at the callsite? I think this will make things extra clear, since in many scenarios we actually won't create a canceler (only in `schedule`, `scheduleGrant` and `scheduleRevoke`). For example:

```
fn setDelay() {
  require(msg.sender == root)
  return _scheduleWithDelay
}

fn schedule() {
  require hasPermission(msg.sender)
  id = _schedule

  // accounts that schedule actions are automatically made cancelers for them, so that they can manage their
  // action. we check that they are not already a canceler since e.g. root may schedule actions (and root is
 // always a global canceler).
  if (!isCanceler(id, msg.sender))
    addCanceler(id, msg.sender)

  return id
}

File: pkg/vault/test/authorizer/TimelockAuthorizer.test.ts [hunk 3]
```diff
@@ -731,6 +734,192 @@ describe('TimelockAuthorizer', () => {
     });
   });
 
+  describe('addCanceler', () => {
+    context('when the sender is the root', () => {
+      let scheduledId = bn(0);
+      beforeEach('set sender', async () => {
+        from = root;
```
nventuro, line 741: This is very much a personal preference thing, but in this case I'm not a fan of this style (which is the style the rest of the test suite follows). We don't have parametrized tests that receive a `from` argument, we simply call from `root` (which is the only valid sender). Given that, I'd instead not have this block, and simply do `{ from: root }` in each call in each test.

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -78,6 +78,11 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
      */
     address public constant EVERYWHERE = address(-1);
 
+    /**
+     * @notice A constant value for `scheduledExecutionId` that will match any execution Id
```
nventuro, line 82:
```suggestion
     * @notice A constant value for `scheduledExecutionId` that will match any execution Id. Cancelers assigned to this Id
     * will be able to cancel *any* scheduled action, which is very useful for e.g. emergency response dedicated
     * teams that analyze these.
```

pkg/vault/test/authorizer/TimelockAuthorizer.test.ts
```diff
@@ -731,6 +734,192 @@ describe('TimelockAuthorizer', () => {
...
+      });
+
+      context('when creating canceler', () => {
```
EndymionJkb, line 744:
```suggestion
      context('when adding canceler', () => {
```

pkg/vault/test/authorizer/TimelockAuthorizer.test.ts
```diff
@@ -731,6 +734,192 @@ describe('TimelockAuthorizer', () => {
...
+            expect(await authorizer.isCanceler(scheduledId, other)).to.be.true;
+          });
+
+          it('emits an event', async () => {
+            const receipt = await authorizer.addCanceler(scheduledId, other, { from });
+
+            expectEvent.inReceipt(await receipt.wait(), 'CancelerAdded', { scheduledExecutionId: scheduledId });
+          });
+
+          it('cannot be add twice', async () => {
```
EndymionJkb, line 760:
```suggestion
          it('cannot be added twice', async () => {
```

pkg/vault/test/authorizer/TimelockAuthorizer.test.ts
```diff
@@ -731,6 +734,192 @@ describe('TimelockAuthorizer', () => {
...
+          });
+
+          it('cannot remove global canceler for a specific execution id', async () => {
+            await authorizer.addCanceler(GLOBAL_CANCELER_SCHEDULED_EXECUTION_ID, other, { from });
+            await expect(authorizer.removeCanceler(scheduledId, other, { from })).to.be.revertedWith(
+              'ACCOUNT_IS_GLOBAL_CANCELER'
+            );
+          });
+
+          it('cannot be remove twice', async () => {
```
EndymionJkb, line 854:
```suggestion
          it('cannot be removed twice', async () => {
```

pkg/vault/test/authorizer/TimelockAuthorizer.test.ts
```diff
@@ -731,6 +734,192 @@ describe('TimelockAuthorizer', () => {
...
+            await authorizer.addCanceler(scheduledId, other, { from });
+            const receipt = await authorizer.removeCanceler(scheduledId, other, { from });
+
+            expectEvent.inReceipt(await receipt.wait(), 'CancelerRemoved', {
+              scheduledExecutionId: scheduledId,
+              canceler: other.address,
+            });
+          });
+
+          it('cannot Removed', async () => {
```
EndymionJkb, line 887:
```suggestion
          it('cannot be removed', async () => {
```

pkg/vault/test/authorizer/TimelockAuthorizer.test.ts
```diff
@@ -731,6 +734,192 @@ describe('TimelockAuthorizer', () => {
...
+            await expect(authorizer.removeCanceler(scheduledId, other, { from })).to.be.revertedWith(
+              'ACCOUNT_IS_NOT_CANCELER'
+            );
+          });
+
+          it('cannot Removed', async () => {
```
EndymionJkb, line 893:
```suggestion
          it('cannot remove', async () => {
```

pkg/vault/test/authorizer/TimelockAuthorizer.test.ts
```diff
@@ -731,6 +734,192 @@ describe('TimelockAuthorizer', () => {
...
+      });
+
+      context('when creating canceler', () => {
+        context('for a specific scheduled execution id', () => {
+          it('can add canceler for a specific execution id', async () => {
+            expect(await authorizer.isCanceler(scheduledId, other)).to.be.false;
+
+            await authorizer.addCanceler(scheduledId, other, { from });
+
+            expect(await authorizer.isCanceler(scheduledId, other)).to.be.true;
```
nventuro, line 751: We should also test that they are not a canceler for a second, different `scheduledId`.

pkg/vault/test/authorizer/TimelockAuthorizer.test.ts
```diff
@@ -731,6 +734,192 @@ describe('TimelockAuthorizer', () => {
...
+    context('when the sender is the root', () => {
+      let scheduledId = bn(0);
+      beforeEach('set sender', async () => {
+        from = root;
+      });
+
+      context('when creating canceler', () => {
+        context('for a specific scheduled execution id', () => {
+          it('can add canceler for a specific execution id', async () => {
+            expect(await authorizer.isCanceler(scheduledId, other)).to.be.false;
```
nventuro, line 747: `other` typically means 'an unrelated account that is not special in any way' (e.g. `expect(isCanceler(other)).to.be.false`). I'd create a `canceler` signer and use that.

pkg/vault/test/authorizer/TimelockAuthorizer.test.ts
nventuro, line 760: Let's also have tests where:
 - we check root is already a canceler
 - we check root cannot be added as a canceler

pkg/vault/test/authorizer/TimelockAuthorizer.test.ts
```diff
@@ -731,6 +734,192 @@ describe('TimelockAuthorizer', () => {
...
+            await expect(authorizer.addCanceler(scheduledId, other, { from })).to.be.revertedWith(
+              'ACCOUNT_IS_ALREADY_CANCELER'
+            );
+          });
+        });
+
+        context('for any scheduled execution id', () => {
+          beforeEach('set schedule id', async () => {
+            scheduledId = GLOBAL_CANCELER_SCHEDULED_EXECUTION_ID;
+          });
```
nventuro, line 772: I would not have this, and instead inline the value in order to be more explicit (similar to my suggestion above to not use `from = root`).

pkg/vault/test/authorizer/TimelockAuthorizer.test.ts
```diff
@@ -1585,6 +1774,18 @@ describe('TimelockAuthorizer', () => {
...
+                it('receives canceler status', async () => {
+                  const id = await schedule();
+
+                  expect(await authorizer.isCanceler(id, grantee)).to.be.true;
+                });
+
+                it('can cancel the action immediately', async () => {
+                  const id = await schedule();
+                  // should not revert
+                  await expect(authorizer.cancel(id, { from: grantee })).not.to.reverted;
```
nventuro, line 1787: You can test for the `ExecutionCancelled` event here, which is slightly more robust

nventuro (approved):
Welcome to the monorepo :grin: great work, thank you very much! 

Left a suggestion on the automatic addition of cancelers when scheduling, and some style recommendations in the test, plus missing test cases.

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -573,7 +592,15 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
 
         bytes32 actionId = IAuthentication(where).getActionId(_decodeSelector(data));
         require(hasPermission(actionId, msg.sender, where), "SENDER_DOES_NOT_HAVE_PERMISSION");
-        return _schedule(actionId, where, data, executors);
+
+        uint256 id = _schedule(actionId, where, data, executors);
+        // accounts that schedule actions are automatically made cancelers for them, so that they can manage their
+        // action. we check that they are not already a canceler since e.g. root may schedule actions (and root is
+        // always a global canceler).
```
nventuro, line 599:
```suggestion
        // Accounts that schedule actions are automatically made cancelers for them, so that they can manage their
        // action. We check that they are not already a canceler since e.g. root may schedule actions (and root is
        // always a global canceler).
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -737,7 +801,14 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
         bytes memory data = abi.encodeWithSelector(this.grantPermissions.selector, _ar(actionId), account, _ar(where));
 
         // TODO: fix actionId for event (maybe overhaul _scheduleWithDelay?)
-        return _scheduleWithDelay(0x0, address(this), data, delay, executors);
+        uint256 id = _scheduleWithDelay(0x0, address(this), data, delay, executors);
+        // accounts that schedule actions are automatically made cancelers for them, so that they can manage their
```
nventuro, line 805:
```suggestion
        // Granters that schedule actions are automatically made cancelers for them, so that they can manage their
        // action. We check that they are not already a canceler since e.g. root may schedule grants (and root is
        // always a global canceler).
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -811,7 +884,14 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
         require(isRevoker(actionId, msg.sender, where), "SENDER_IS_NOT_REVOKER");
         bytes memory data = abi.encodeWithSelector(this.revokePermissions.selector, _ar(actionId), account, _ar(where));
         bytes32 revokePermissionId = getRevokePermissionActionId(actionId);
-        return _schedule(revokePermissionId, address(this), data, executors);
+        uint256 id = _schedule(revokePermissionId, address(this), data, executors);
+        // accounts that schedule actions are automatically made cancelers for them, so that they can manage their
+        // action. we check that they are not already a canceler since e.g. root may schedule actions (and root is
+        // always a global canceler).
```
nventuro, line 890:
```suggestion
        // Revokers that schedule actions are automatically made cancelers for them, so that they can manage their
        // action. We check that they are not already a canceler since e.g. root may schedule revokes (and root is
        // always a global canceler).
```

pkg/vault/test/authorizer/TimelockAuthorizer.test.ts
```diff
@@ -675,37 +663,247 @@ describe('TimelockAuthorizer', () => {
...
+              authorizer.addCanceler(GLOBAL_CANCELER_SCHEDULED_EXECUTION_ID, canceler, { from: root })
+            ).to.be.revertedWith('ACCOUNT_IS_ALREADY_CANCELER');
+          });
+        });
       });
+    });
+
+    context('when the sender is not the root', () => {
+      it('reverts', async () => {
+        await expect(authorizer.addCanceler(0, canceler, { from: canceler })).to.be.revertedWith('SENDER_IS_NOT_ROOT');
```
nventuro, line 753:
```suggestion
        await expect(authorizer.addCanceler(0, canceler, { from: other })).to.be.revertedWith('SENDER_IS_NOT_ROOT');
```

pkg/vault/test/authorizer/TimelockAuthorizer.test.ts
```diff
@@ -675,37 +663,247 @@ describe('TimelockAuthorizer', () => {
...
+            });
+          });
+
+          it('cannot remove', async () => {
+            await expect(
+              authorizer.removeCanceler(GLOBAL_CANCELER_SCHEDULED_EXECUTION_ID, canceler, { from: root })
+            ).to.be.revertedWith('ACCOUNT_IS_NOT_CANCELER');
+          });
+
+          it('cannot remove', async () => {
```
nventuro, line 835: It seems these descriptions are incomplete?

PR #2298, opened 2023-02-24
Managed Pool: use cash balance only for recovery mode exit

### LINKED ISSUE #2042: Make RecoveryMode work with cash amounts instead of total

EndymionJkb:
# Description

Managed Pool Controllers are asset managers, and some add/remove strategies involve asset managers. In any case, we anticipate that Managed Pools will have asset managers, and might therefore have managed balances.

Ensure that if recovery mode is enabled in this situation, that it uses the cash balances only. (Otherwise LPs exiting early would get a disproportionate share of any tokens with managed balances, potentially preventing stragglers from exiting at all.)

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #2042


jubeira (commented):
The implementation indeed fixes the issue; I've verified that the new test fails if I undo the changes and go back to the total balances.

But if I'm not mistaken, the idea of the recovery mode helper contract was to provide an external implementation for this, effectively reducing the pool's bytecode size. See the `IRecoveryModeHelper` [description](https://github.com/balancer/balancer-v2-monorepo/blob/4570a029a50591b06e583d07133bf6941fc1d208/pkg/interfaces/contracts/pool-utils/IRecoveryModeHelper.sol#L19-L23). In other words, maybe we should do an external call to the recovery mode helper to compute the amounts instead of computing them in the pool?

Deferring approval to @nventuro who has more context.

EndymionJkb:
> But if I'm not mistaken, the idea of the recovery mode helper contract was to provide an external implementation for this, effectively reducing the pool's bytecode size. See the `IRecoveryModeHelper` [description](https://github.com/balancer/balancer-v2-monorepo/blob/4570a029a50591b06e583d07133bf6941fc1d208/pkg/interfaces/contracts/pool-utils/IRecoveryModeHelper.sol#L19-L23). In other words, maybe we should do an external call to the recovery mode helper to compute the amounts instead of computing them in the pool?

Yes; I don't think we ever deployed that, though. All I see is the PoolRecoveryHelper, which is the thing that puts pools with failing RateProviders into RecoveryMode. Was the intent to deploy that, then call it from ManagedPool?

nventuro:
Very clean :clap:

Issue #2289, opened 2023-02-24
Remove factory interface and `update_voting_escrow` mechanism


jubeira:
The address of the voting escrow will be a fixed proxy that points to the actual implementation, so there's no need to update the escrow in the `ChildChainGauge`s.

jubeira:
The current Vyper factory has functionality that we won't be using:
- Everything related to special treatment to BAL
- Voting escrow update functionality
- Gauge deployment via proxies

The proposed changes are the following:
- Replace the factory with a simple one, similar to the ones used for liquidity / stakeless gauges.
- Use `VotingEscrowDelegationProxy` as the fixed voting escrow, since it can already change its implementation
- Remove all BAL special treatment / checkpoint functionality in the gauge

jubeira:
The current Vyper factory has functionality that we won't be using:
- BAL pseudo minter functionality
- Voting escrow update functionality
- Gauge deployment via proxies

The proposed changes are the following:
- Replace the factory with a simple one, similar to the ones used for liquidity / stakeless gauges.
- Use `VotingEscrowDelegationProxy` as the fixed voting escrow, since it can already change its implementation
- Create a third contract, analogous to the balancer minter in the L1, with pseudo mint functionality currently in the factory. Then, replace the references in the gauge with the minter.

jubeira:
With #2313 this would be complete: 
- the new factory is simpler and has only one purpose
- the pseudo mint functionality was moved to the pseudo minter contract (which is a singleton)
- the voting escrow is now fixed for the gauges

#2322 completes some tech debt for the base layer that the pseudo minter uses.

PR #2306, opened 2023-02-28
`ChildChainGauge` feature cleanup

jubeira:
# Description

The factory's mint functionality shall be replaced by a third singleton contract, analogous to the BAL minter in the L1. In short, it'll receive the BAL from the tokens during the checkpoint and keep track of the tokens already distributed, so that the right amount can be sent to users. This will decouple that logic from the factory, which doesn't need to be a singleton anymore.

This PR refactors the references to the factory in the gauge, replacing them with the minter (which will be added in a subsequent PR).
It also ~removes the kill / unkill functionality which is not used~ restores the kill / unkill functionality, which was not used before.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [x] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Part 2, #2289


pkg/liquidity-mining/contracts/gauges/ChildChainGauge.vy
```diff
@@ -110,17 +110,16 @@ reward_integral_for: public(HashMap[address, HashMap[address, uint256]])
 # user -> token -> [uint128 claimable amount][uint128 claimed amount]
 claim_data: HashMap[address, HashMap[address, uint256]]
 
-is_killed: public(bool)
```
jubeira, line 113: I'd like to give this one last thought.

If it's really unnecessary then these changes are OK, but are we sure we don't want to implement a `kill` functionality?
I guess you can always do that with the root gauges though right?

jubeira: Right; the point is that if we don't have the kill behavior in L2s, the only thing that we'd be missing is getting the already bridged BAL out of circulation. Besides that, it's not that it continues to be a problem after the L1 counterpart is killed.

If you think it's still worth it, let's go for it.

jubeira: It seems it's a bit more simple than in the liquidity gauge, since in this case the loop just accumulates inflation in a variable and we can just skip it if killed.
See https://github.com/balancer-labs/balancer-v2-monorepo/pull/2306/commits/cf0dac798ab58c53d615d3cde2f7bffd4f7cf4d0.

nventuro (approved):

PR #2308, opened 2023-03-01
Batch Relayer: explicit support for Stable Pools

### LINKED ISSUE #2305: Explicitly support stable joins and exits in BatchRelayer
`VaultActions` performs `userData` value replacements by inspecting the join/exit kind enum in the encoded data. Officially, only WeightedPool is supported, but both Stable and Weighted used to have fully compatible join and exit kinds, which means this could be used for both types. Unfortunately, proportional exits were not included in `ComposableStablePool`, and when they were later added back (#1912) in v2 the encoding was incompatible.

We should now extend `VaultActions` so that both Weighted and Stable are officially supported, using their own encode/decode libraries, and specifically make sure proportional exits work in ComposableStablePool.

EndymionJkb:
# Description

VaultActions performs userData value replacements by inspecting the join/exit kind enum in the encoded data. Weighted and Stable pools used to be identical, but over multiple releases of Stable Pools, support for proportional joins/exits was dropped and then restored, and in the process the enum values (and names) diverged.

Ensure the Batch Relayer fully supports Stable Pools, including proportional joins and exits.

Fully testing old pools will need fork tests.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #2305


pkg/standalone-utils/contracts/relayer/VaultActions.sol
```diff
@@ -160,20 +167,52 @@ abstract contract VaultActions is IBaseRelayerLibrary {
...
+
+        // Save gas by only re-encoding the data if we actually performed a replacement
+        return
+            _replacedAmounts(amountsIn)
+                ? abi.encode(StablePoolUserData.JoinKind.EXACT_TOKENS_IN_FOR_BPT_OUT, amountsIn, minBPTAmountOut)
+                : userData;
+    }
+
+    // Mutates amountsIn, and returns true if any replacements were made
+    function _replacedAmounts(uint256[] memory amountsIn) private returns (bool replacedAmounts) {
```
nventuro, line 208: I'd more explicitly have an anonymous return value and explicitly return:

```
fn _foo() returns (bool) {
  replaced = false
  for ...
    if x 
      replaced = true

  return replaced
```

pkg/standalone-utils/contracts/relayer/VaultActions.sol
```diff
@@ -160,20 +167,52 @@ abstract contract VaultActions is IBaseRelayerLibrary {
...
+                : userData;
+    }
+
+    /**
+     * @dev All versions of stable pool have the same enum and processing for EXACT_TOKENS_IN_FOR_BPT_OUT, so we can
+     * just use the current one. If proportional join is supported, it's a "given out" (parameter is a BPT amount), so
+     * there would be no replacements, and it can be passed through directly to the pool.
+     *
+     * All versions are identical for joins, except the latest ComposableStable appends a proportional join. If you
+     * tried to do a proportional join on any legacy pool, the pool would revert.
```
nventuro, line 183: I'd move this comment to the body of the function since it explains how it works, not what it does.

nventuro: Also I found the wording a bit awkward. Maybe try something like:

>The only 'given in' join (in which the parameters are the amounts in) is EXACT_TOKENS_IN_FOR_BPT_OUT, so that is the only one in which we do replacements. Luckily all versions of Stable Pool share the same enum value for it, so we can treat them all at once.
>
>Note that ComposableStablePool v3 supports a proportional join kind which some previous versions did not: while it is not rejected here, if passed to the Pool it will revert.

pkg/standalone-utils/contracts/relayer/VaultActions.sol
```diff
@@ -239,7 +278,18 @@ abstract contract VaultActions is IBaseRelayerLibrary {
         if (kind == PoolKind.WEIGHTED) {
             return _doWeightedExitChainedReferenceReplacements(userData);
         } else {
-            _revert(Errors.UNHANDLED_EXIT_KIND);
+            // exitKind just looks at the first value, same for all, so we can use the current library function
+            uint8 exitKind = uint8(StablePoolUserData.exitKind(userData));
```
nventuro, line 282: I find this odd here, all we gain is to not have to make the call further down. I'd move this inside each of the `_doReplacements` functions to clarity, also making the structure of this `if-else` chain clearer.

pkg/standalone-utils/contracts/relayer/VaultActions.sol
```diff
@@ -140,6 +141,12 @@ abstract contract VaultActions is IBaseRelayerLibrary {
...
         if (kind == PoolKind.WEIGHTED) {
             return _doWeightedJoinChainedReferenceReplacements(userData);
+        } else if (
+            kind == PoolKind.LEGACY_STABLE ||
+            kind == PoolKind.LEGACY_COMPOSABLE_STABLE ||
+            kind == PoolKind.COMPOSABLE_STABLE
+        ) {
+            return _doStableJoinChainedReferenceReplacements(userData);
         } else {
             _revert(Errors.UNHANDLED_JOIN_KIND);
```
nventuro, line 151: This error should be changed to 'unhandled pool kind'

pkg/standalone-utils/contracts/relayer/VaultActions.sol
```diff
@@ -239,7 +278,18 @@ abstract contract VaultActions is IBaseRelayerLibrary {
...
+            uint8 exitKind = uint8(StablePoolUserData.exitKind(userData));
+
+            if (kind == PoolKind.LEGACY_STABLE) {
+                return _doLegacyStableExitChainedReferenceReplacements(userData, exitKind);
+            } else if (kind == PoolKind.LEGACY_COMPOSABLE_STABLE) {
+                return _doLegacyComposableStableExitChainedReferenceReplacements(userData, exitKind);
+            } else if (kind == PoolKind.COMPOSABLE_STABLE) {
+                return _doComposableStableExitChainedReferenceReplacements(userData, exitKind);
+            } else {
+                _revert(Errors.UNHANDLED_EXIT_KIND);
```
nventuro, line 291: This error should aslo be changed to 'unhandled pool kind'

pkg/standalone-utils/contracts/relayer/VaultActions.sol
```diff
@@ -280,4 +330,108 @@ abstract contract VaultActions is IBaseRelayerLibrary {
...
+            // so we don't do replacements for those.
+            return userData;
+        }
+    }
+
+    // Shared Stable Exit utility functions
+
+    /**
+     * @dev Can use the standard library for parsing out the arguments, as this is common to all types, but we need
+     * to pass in the exitKind itself, as this could be different.
```
nventuro, line 382: I think we need to have a note somewhere (perhaps when choosing stable pool kinds) that explains that they don't all support the same exit kinds, and that in some cases the encoding is the same but the exit kind id has different values, so we handle them this way.

It's otherwise a bit confusing to follow, and close to impossible if you don't already know that.

pkg/standalone-utils/test/VaultActions.test.ts
```diff
@@ -1196,4 +1206,678 @@ describe('VaultActions', function () {
       });
     });
   });
+
+  describe('stable pools', () => {
+    let poolIdStable: string;
```
nventuro, line 1211: Given that this is so large, I'd consider moving it to its own file (`VaultActionsStableJoinExit.test.ts` or w/e)

nventuro (commented):
The changes look good (I like the `_replacedAmounts` function!), but I'm not sure about the tests - it seems we're only testing the COMPOSABLE_STABLE pool kind, and none of the other two?

pkg/standalone-utils/contracts/relayer/VaultActions.sol
```diff
@@ -102,7 +103,7 @@ abstract contract VaultActions is IBaseRelayerLibrary {
         getVault().manageUserBalance{ value: value }(ops);
     }
 
-    enum PoolKind { WEIGHTED }
+    enum PoolKind { WEIGHTED, LEGACY_STABLE, LEGACY_COMPOSABLE_STABLE, COMPOSABLE_STABLE }
```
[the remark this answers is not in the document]
EndymionJkb: Yes; I had it like that originally, and thought this way would be more consistent, but now that we have actual onchain versions, it makes sense to use them

pkg/standalone-utils/contracts/relayer/VaultActions.sol
```diff
@@ -151,29 +161,64 @@ abstract contract VaultActions is IBaseRelayerLibrary {
...
+                ? abi.encode(WeightedPoolUserData.JoinKind.EXACT_TOKENS_IN_FOR_BPT_OUT, amountsIn, minBPTAmountOut)
+                : userData;
+    }
+
+    function _doStableJoinChainedReferenceReplacements(bytes memory userData) private returns (bytes memory) {
+        // The only 'given in' join (in which the parameters are the amounts in) is EXACT_TOKENS_IN_FOR_BPT_OUT,
+        // so that is the only one where we do replacements. Luckily all versions of Stable Pool share the same
+        // enum value for it, so we can treat them all the same, and just use the latest version.
+
+        // Note that ComposableStablePool v3 supports a proportional join kind which some previous versions did not.
```
nventuro, line 185:
```suggestion
        // Note that ComposableStablePool v2 and up support a proportional join kind which some previous versions did not.
```

pkg/standalone-utils/contracts/relayer/VaultActions.sol
```diff
@@ -232,14 +277,38 @@ abstract contract VaultActions is IBaseRelayerLibrary {
...
+     * they do NOT all support the same exit kinds. Also, though the encoding of the data associated with the exit
+     * is uniform across pool kinds for the same exit method, the ExitKind ID itself may have a different value.
+     *
+     * For instance, BPT_IN_FOR_EXACT_TOKENS_OUT is 2 in legacy Stable Pools, but 1 in Composable Stable Pools.
+     * (See the reference comment and libraries below.)
+     *
+     * To accommodate this, we delegate first to custom ChainedReferenceReplacement functions for each pool type.
+     * These functions determine the general exit method (e.g., single or multi-token), and then - since the encoding
+     * of the arguments is the same for each method across pool kinds - forward the specific ExitKind ID appropriate
+     * to that pool kind to the shared functions that perform the actual recoding.
```
nventuro, line 294: I'd move this inside the body of the function

nventuro: Also I'm not sure we need to explain all of this here? I think we could not have this comment and simply explain in `_doStableExactBptInForOneTokenOutReplacements` and `_doStableExactBptInForTokensOutReplacements` that those kinds have different enum values depending on the pool version, which is why they receive it as an argument and use to to re-encode, instead of hardcoding a constant.

pkg/standalone-utils/test/VaultActionsRelayer.setup.ts
```diff
@@ -0,0 +1,167 @@
...
+  relayerLibrary: Contract,
+  tokens: TokenList,
+  params: {
+    poolId: string;
+    userData: string;
+    toInternalBalance: boolean;
+    outputReferences?: Dictionary<BigNumberish>;
+    sender: Account;
+    recipient: Account;
+    poolKind?: number;
```
nventuro, line 94: I'd not make this argument optional

pkg/standalone-utils/test/VaultActionsStableJoinExit.test.ts
```diff
@@ -0,0 +1,757 @@
...
+  setupRelayerEnvironment,
+  encodeJoinPool,
+  encodeExitPool,
+  PoolKind,
+  approveVaultForRelayer,
+  getJoinExitAmounts,
+  encodeSwap,
+} from './VaultActionsRelayer.setup';
+
+describe('stable pools', () => {
```
nventuro, line 32:
```suggestion
describe('Vault Actions - Stable Pools', () => {
```

pkg/standalone-utils/test/VaultActionsStableJoinExit.test.ts
```diff
@@ -0,0 +1,757 @@
...
+  let bptIndex2: number;
+
+  before('setup environment', async () => {
+    ({ user, other, vault, relayer, relayerLibrary } = await setupRelayerEnvironment());
+  });
+
+  before('setup common recipient', () => {
+    // All the tests use the same recipient; this is a simple abstraction to improve readability.
+    recipient = user;
+  });
```
nventuro, line 50: Make this a separate account, e.g. using `randomAddress()` if it doesn't send any transactions - it can be confusing and misleading if we're using the same account under two names.

pkg/standalone-utils/test/VaultActionsStableJoinExit.test.ts
```diff
@@ -0,0 +1,757 @@
...
+            }),
+          ])
+        ).to.be.revertedWith('Incorrect sender');
+      });
+    });
+
+    context('when caller is authorized', () => {
+      context('sender = user', () => {
+        beforeEach(async () => {
+          sender = user;
```
nventuro, line 103:
```suggestion
      context('sender = user', () => {
        let sender: SignerWithAddress;
      
        beforeEach(async () => {
          sender = user;
```

You'll need to repeat this in other tests, but it'll make to so we're not sharing the `sender` variable accross blocks, which can lead to user error and hard to debug scenarios (if e.g. we forget to initialize it in one block, since some tests would then depend on unrelated tests, based on execution order)

nventuro (approved):

pkg/standalone-utils/test/VaultActionsStableJoinExit.test.ts
EndymionJkb: In the old VaultActions too?

nventuro: Yes!

pkg/standalone-utils/contracts/relayer/VaultActions.sol
EndymionJkb: Ok; I've distributed the content to lower levels, instead of concentrating it all here. This just has the top sentence, saying what the overall process is doing.

Then I have a section for "Stable Pool version-dependent recoding dispatch functions," which extract the exitKind and have version-dependent logic. That explains the middle part (why there have to be separate functions for each pool version).

Then the "Shared Stable Exit recoding functions" second has an abbreviated explanation of why ExitKind has to be passed in. See how you like it now.

pkg/standalone-utils/test/VaultActionsRelayer.setup.ts
EndymionJkb: Yes; we did just get burned by that :) Originally it was to make minimal changes (i.e., not change VaultActions.test.ts; just add the new file). But it required a lot of changes anyway, so might as well add one more.

PR #2294, opened 2023-02-24
Begin granter overhaul

nventuro:
# Description

This is a first attempt at #2242 - likely ~30% of the work required for that. #2238 will likely look very similar to this, as will a future revoker overhaul.

It changes `addGranter` so that granter permissions are stored in a separate data structure, and are not regular permissions (i.e. they don't use the 'base permission' and 'specifier' scheme). This makes analyzing the rules for granting and revoking permissions much simpler. Similarly, granting now also has dedicated delays.

The API is the following:
 - addGranter
 - removeGranter
 - isGranter
 - canGrant (the name is terrible, but I don't want to change it yet. This is basically isGranter + delay semantics)
 - scheduleGrantDelayChange (not yet implemented)
 - getGrantDelay

I introduced two changes to the behavior of the code. First, root is always a granter of everything everywhere (and this cannot be undone), so I removed the calls where we set this manually, migrate permissions from current root to pending root, etc. Second, since root can already grant everything, I removed the notion of being a 'general granter' (a granter of any permission), since I find no reasonable use case for this. We can still have global granters of a permission (e.g. can grant permission to set swap fees on any pool), but not of *all* permissions.

----

I tried to keep the changes to a minimum to make reviewing this PR possible. In particular:
 - I didn't greatly modify existing tests, only performed small adjustments. The tests can be *greatly* improved (in fact I already have that ready), but it'd be too messy to do that here, all at once.
 - I skipped the tests that are related to delays in granting, since I have not yet implemented that functionality. The reasoning is the same: I don't want to bloat this PR with tons of changes, even if it means that this contract will temporarily lose functionality.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [x] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Documentation or wording changes
- [ ] Other

## Checklist:

- [x] The diff is legible and has no extraneous changes
- [x] Complex code has been commented, including external interfaces
- [x] Tests are included for all code paths
- [x] The base branch is either `master`, or there's a description of how to merge

Issue #2241, opened 2023-02-10
Improve self-scheduling


nventuro:
(related to #2239)

Some Authorizer native actions use delays themselves (e.g. root transfers, changing delays, etc.), and these use the same scheduling mechanism as the rest of the system. Schedule-only functions typically have the `onlyExecutor` modifier, meaning they can only be called via a call to `execute`.

We should rework `_schedule` and `_scheduleWithDelay` slightly to more clearly stablish a separation between external scheduled actions (where the target *must not* be the Authorizer), and internal ones (where the target is always the Authorizer).

nventuro:
This should likely be tackled after https://github.com/balancer-labs/balancer-v2-monorepo/issues/2242, since grant/revoke delays are involved here (and part of the source of confusion).

nventuro:
We could now remove the event from `_scheduleWithDelay` and have dedicated events for `schedule`, `scheduleDelayChange`, `scheduleRootChange`, etc., and inline `_schedule` inside `schedule`.

PR #2310, opened 2023-03-02
Add grant delays

nventuro:
This adds back the grant delays removed in #2294, by introducing `scheduleGrantDelayChange`. I started to adjust some small parts of the code and tests so that also adding revoke delays will be simple.

This is a work in progress, mostly as the current tests are not great, and I'm trying to expand on them. However, only after a while did I notice that I was actually working mostly on `scheduleGrantPermissions` and not `scheduleGrantDelayChange` :sweat_smile: I might roll those changes back and include them in a follow-up PR.


pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -180,6 +180,11 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
      */
     event ActionDelaySet(bytes32 indexed actionId, uint256 delay);
 
+    /**
+     * @notice Emitted when a new `delay` is set in order to grant permission to execute action action `actionId`.
```
jubeira, line 184:
```suggestion
     * @notice Emitted when a new `delay` is set in order to grant permission to execute action `actionId`.
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -513,25 +514,72 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
...
+
+        _grantDelays[actionId] = delay;
+        emit GrantDelaySet(actionId, delay);
+    }
+
+    function _isDelayShorterThanSetAuthorizer(uint256 delay) private view returns (bool) {
+        // No delay can be greater than the current delay for changing the Authorizer itself (`IVault.setAuthorizer`).
+        // Otherwise, it'd be possible to execute the action with a shorter delay by simply replacing the
+        // TimelockAuthorizer with a different contract that didn't enforce these delays.
+        // Note that it is still possible for an action to end up with a delay shorter than `setAuthorizer` if e.g.
```
jubeira, line 538:
```suggestion
        // Note that it is still possible for an action to end up with a delay larger than `setAuthorizer` if e.g.
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -513,25 +514,72 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
...
+
+        uint256 executionDelay = _getDelayChangeExecutionDelay(_grantDelays[actionId], newDelay);
+
+        bytes memory data = abi.encodeWithSelector(this.setGrantDelay.selector, actionId, newDelay);
+        // TODO: fix actionId for event (maybe overhaul _scheduleWithDelay?)
+        return _scheduleWithDelay(0x0, address(this), data, executionDelay, executors);
+    }
+
+    function _getDelayChangeExecutionDelay(uint256 currentDelay, uint256 newDelay) private pure returns (uint256) {
+        require(newDelay <= MAX_DELAY, "DELAY_TOO_LARGE");
```
jubeira, line 581: We are doing both this check and `isRoot` for `scheduleDelayChange` and `scheduleGrantDelayChange`.

How about grouping the the two of them in a separate helper and calling it at the very beginning of the function?
I find it a bit weird that an internal `_get...` starts with a `require` when this could be solved at the beginning of the external method.

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -776,10 +820,15 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
...
-            // Note that `canGrant` will return true for the executor if the permission has a delay.
-            require(canGrant(actionIds[i], msg.sender, where[i]), "SENDER_IS_NOT_GRANTER");
+            if (_grantDelays[actionIds[i]] == 0) {
+                require(isGranter(actionIds[i], msg.sender, where[i]), "SENDER_IS_NOT_GRANTER");
+            } else {
+                // Some actions may have delays associated with granting them - these permissions cannot be granted
+                // directly, even if the caller is a granter, and must instead be scheduled for future execution via
+                // `scheduleGrantPermission`.
+                require(msg.sender == address(_executor), "GRANT_MUST_BE_SCHEDULED");
+            }
```
jubeira, line 830: I found `canGrant` a bit confused because of the existence of `isGranter`, but the helper did make this for loop more readable.

How about keeping the helper, maybe thinking about a new name?
On the other hand, I thought it was a useful helper to expose.

jubeira (commented):
First pass done!
Overall looks good; left a handful of questions and minor observations.

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
nventuro: You mean `canGrant`? I think it's a terrible concept :/ We are forced to have `canPerform`, but I'd not have it unless we needed to. How did putting this behind a function make it more readable?

jubeira: I think the name was not great, but it was convenient at least in this context. The structure was straightforward: iterate action IDs, can they be granted? if yes, go, otherwise revert.
But yeah, doing it like this exposes two different error messages which is good.

Feel free to disregard the comment; perhaps I'd suggest moving the comment inside the `else` block above the `if`.
How about something like

```
  // If the current action does not have an associated delay, the permission can be granted right away provided that the caller is a granter.
  // On the other hand, some actions may have delays associated with granting them - these permissions cannot be granted
  // directly, even if the caller is a granter, and must instead be scheduled for future execution via
  // `scheduleGrantPermission`.
```

nventuro: The thing is 'can they be granted' really means 'can they be granted in the context of a non-scheduled action', which is a complex and weird concept - I find the code just being there makes everything much more obvious.

nventuro:
I've extracted the `scheduleDelayChange` tests into a separate file, extended them to be more comprehensive, and then copied them to support `scheduleGrantDelayChange` as well.

pkg/vault/test/authorizer/TimelockAuthorizerDelays.test.ts
```diff
@@ -0,0 +1,372 @@
...
+
+      it('execution can be unprotected', async () => {
+        const id = await authorizer.scheduleDelayChange(ACTION_1, ACTION_DELAY, [], { from: root });
+        const execution = await authorizer.getScheduledExecution(id);
+
+        expect(execution.protected).to.be.false;
+      });
+
+      it('execution can be protected', async () => {
+        const executors = range(4).map(() => ethers.Wallet.createRandom().address);
```
jubeira, line 73: We have a wrapper for `ethers.Wallet.createRandom` in `constants.ts`.
Although I think that wrapper should use `.address` instead of `getAddress()` :sweat_smile:

pkg/vault/test/authorizer/TimelockAuthorizerDelays.test.ts
```diff
@@ -0,0 +1,372 @@
...
+        const id = await authorizer.scheduleDelayChange(ACTION_1, ACTION_DELAY, executors, { from: root });
+        const execution = await authorizer.getScheduledExecution(id);
+
+        expect(execution.protected).to.be.true;
+        await Promise.all(
+          executors.map(async (executor) => expect(await authorizer.isExecutor(id, executor)).to.be.true)
+        );
+      });
+
+      it('root can cancel the execution', async () => {
```
jubeira, line 83: What about cancelers? Is that tested elsewhere?

pkg/vault/test/authorizer/TimelockAuthorizerDelays.test.ts
```diff
@@ -0,0 +1,372 @@
...
+      it('reverts if setDelay is called directly', async () => {
+        await expect(authorizer.instance.setDelay(ACTION_1, DAY)).to.be.revertedWith('CAN_ONLY_BE_SCHEDULED');
+      });
+
+      it('reverts if the delay is greater than the setAuthorizer delay', async () => {
+        const setAuthorizerDelay = await authorizer.delay(await actionId(vault, 'setAuthorizer'));
+        const id = await authorizer.scheduleDelayChange(ACTION_1, setAuthorizerDelay.add(1), [], { from: root });
+
+        // This condition is only tested at the time the delay is actually set (in case e.g. there was a scheduled action
+        // to change setAuthorizer's delay), so we must attempt to execute the action to get the expected revert.
```
jubeira, line 202: This is interesting.
We could also catch this at `scheduleDelayChange` time right?

pkg/vault/test/authorizer/TimelockAuthorizerDelays.test.ts
```diff
@@ -0,0 +1,372 @@
...
+
+        const id = await authorizer.scheduleDelayChange(ACTION_1, ACTION_DELAY, [], { from: root });
+
+        await advanceTime(expectedExecutionDelay);
+        await authorizer.execute(id);
+
+        expect(await authorizer.delay(ACTION_2)).to.be.equal(previousAction2Delay);
+      });
+
+      it('does not set any the grant action delay when executed', async () => {
```
jubeira, line 119:
```suggestion
      it('does not set the grant action delay when executed', async () => {
```

pkg/vault/test/authorizer/TimelockAuthorizerDelays.test.ts
```diff
@@ -0,0 +1,372 @@
...
+        const previousGrantDelay = await authorizer.getActionIdGrantDelay(ACTION_1);
+
+        const id = await authorizer.scheduleDelayChange(ACTION_1, ACTION_DELAY, [], { from: root });
+
+        await advanceTime(expectedExecutionDelay);
+        await authorizer.execute(id);
+
+        expect(await authorizer.delay(ACTION_2)).to.be.equal(previousGrantDelay);
```
jubeira, line 127: hmm shouldn't this be `ACTION_1`? I think I'm not getting why `delay(action2)` is compared with `grantDelay(action1)`.

pkg/vault/test/authorizer/TimelockAuthorizerDelays.test.ts
```diff
@@ -0,0 +1,372 @@
...
+        const id = await authorizer.scheduleGrantDelayChange(ACTION_1, ACTION_GRANT_DELAY, executors, { from: root });
+        const execution = await authorizer.getScheduledExecution(id);
+
+        expect(execution.protected).to.be.true;
+        await Promise.all(
+          executors.map(async (executor) => expect(await authorizer.isExecutor(id, executor)).to.be.true)
+        );
+      });
+
+      it('root can cancel the execution', async () => {
```
jubeira, line 244: same as above; what about cancelers?

pkg/vault/test/authorizer/TimelockAuthorizerDelays.test.ts
jubeira: Looking at the test below, it should be the same action (1) and it should compare the current grant action delay with the old one.

jubeira (approved):
Wow, this is huge!
I've left a few questions and minor comments, but it's looking pretty good IMO.
I'd suggest getting one more pair of eyes on top of this given the amount of cases under test.

pkg/vault/test/authorizer/TimelockAuthorizerDelays.test.ts
```diff
@@ -0,0 +1,372 @@
...
+        const id = await authorizer.scheduleDelayChange(ACTION_1, ACTION_DELAY, [], { from: root });
+
+        await advanceTime(expectedExecutionDelay);
+        const receipt = await authorizer.execute(id);
+        expectEvent.inReceipt(await receipt.wait(), 'ActionDelaySet', { actionId: ACTION_1, delay: ACTION_DELAY });
+      });
+    }
+
+    context('when the delay is being increased', () => {
+      // When incrasing the delay, the execution delay should always be the MINIMUM_EXECUTION_DELAY.
```
ylv-io, line 140:
```suggestion
      // When increasing the delay, the execution delay should always be the MINIMUM_EXECUTION_DELAY.
```

pkg/vault/test/authorizer/TimelockAuthorizerDelays.test.ts
nventuro: You are right!

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -527,25 +528,72 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
...
+        address[] memory executors
+    ) external returns (uint256 scheduledExecutionId) {
         require(isRoot(msg.sender), "SENDER_IS_NOT_ROOT");
+        require(newDelay <= MAX_DELAY, "DELAY_TOO_LARGE");
+
+        uint256 executionDelay = _getDelayChangeExecutionDelay(_grantDelays[actionId], newDelay);
+
+        bytes memory data = abi.encodeWithSelector(this.setGrantDelay.selector, actionId, newDelay);
+        // TODO: fix actionId for event (maybe overhaul _scheduleWithDelay?)
+        return _scheduleWithDelay(0x0, address(this), data, executionDelay, executors);
```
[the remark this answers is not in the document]
nventuro: Because we're getting rid of extended action ids (see #2316 and #2241). The todo refers to #2241, which now has my current thoughts on the matter.

pkg/vault/test/authorizer/TimelockAuthorizerDelays.test.ts
nventuro: Testing cancelers is out of scope - that's part of `add/removeCanceler`, and then the `cancel` function itself. All we test here is that root can cancel this execution (which it itself created). I added a small comment about the fact that we assume this.

ylv-io (approved):
🚢 Love being explicit with auth calls.

pkg/vault/test/authorizer/TimelockAuthorizerDelays.test.ts
nventuro: Potentially, though it would have some side effects. E.g. perhaps the new delay is not valid at the time of scheduling, but it will be valid at execution. E.g. during system setup, you'll need to first schedule and execute a setAuthroizer delay change, and only then will you be able to execute delay changes for other stuff.

pkg/vault/test/authorizer/TimelockAuthorizerDelays.test.ts
nventuro: Replied above.

PR #2311, opened 2023-03-02
Liquidity Bootstrapping Pool: fee fix

EndymionJkb:
# Description

Fee calculation was inverted for swaps given in. Tests that would have caught this were in BaseWeightedPool.behavior.ts. However, the pool type defaulted to WeightedPool: and we didn't override it. So it was never actually running those tests against the LBP.

Now it does: and I removed the defaulted parameter so that this doesn't happen again (e.g., with Managed or a future derived kind of Weighted Pool).

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/pool-weighted/test/BaseWeightedPool.behavior.ts
+++ b/pkg/pool-weighted/test/BaseWeightedPool.behavior.ts
@@ -13,10 +13,7 @@ import { sharedBeforeEach } from '@balancer-labs/v2-common/sharedBeforeEach';
 import { expectBalanceChange } from '@balancer-labs/v2-helpers/src/test/tokenBalance';
 import Vault from '@balancer-labs/v2-helpers/src/models/vault/Vault';
 
-export function itBehavesAsWeightedPool(
-  numberOfTokens: number,
-  poolType: WeightedPoolType = WeightedPoolType.WEIGHTED_POOL
-): void {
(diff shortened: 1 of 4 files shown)


pkg/pool-weighted/test/BaseWeightedPool.behavior.ts
```diff
@@ -569,15 +575,15 @@ export function itBehavesAsWeightedPool(
...
+            pool.instance[MINIMAL_SWAP_INFO_ONSWAP](
               {
                 kind: SwapKind.GivenIn,
                 tokenIn: tokens.first.address,
                 tokenOut: tokens.second.address,
                 amount: 0,
                 poolId: await pool.getPoolId(),
                 lastChangeBlock: 0,
-                from: other.address,
+                from: lp.address,
```
jubeira, line 586: Why are we using `lp` here, but not below in `given out` context?

jubeira (approved):
Looks good! +1 to removing the default argument; the effects of omitting it are highly undesirable. 

Just left one small question.

pkg/pool-weighted/test/BaseWeightedPool.behavior.ts
EndymionJkb: Only because it didn't fail, I guess. I can change it there as well for consistency.

PR #2313, opened 2023-03-03
Add `L2BalancerPseudoMinter`

jubeira:
# Description

L2 pseudo minter implementation.
This contract is analogous to the `MainnetBalancerMinter`, but for L2s. It works together with child chain gauges to distribute BAL to users by transferring bridged tokens. It replaces the functionality present in Curve's `ChildChainGaugeFactory`.

Last but not least, it has a mechanism to validate that the gauges are correct (similar to the `GaugeAdder` in some sense) which is compatible with our permission system.
In short, at creation time the factory will mark the gauge as valid, and the gauge will store a reference to the factory. The pseudo minter can then retrieve the reference to the factory, and ask the factory whether the gauge was actually created the factory or not. A malicious gauge can pretend to be created from a valid factory, but a valid factory will reject the claim. Moreover, the external `mint` functions are protected against reentrancy in the base `BalancerMinter` layer, so it shouldn't be a problem to give the gauges the control of the execution.
Finally, valid factories need to be allowlisted in the pseudo minter by a permissioned account. So if the factory is valid, and the gauges come from a valid factory, gauges should be valid too.

I have some tests for this, but I wanted to split the work in smaller PRs to make the review easier.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

See #2289

### DIFF
--- a/pkg/liquidity-mining/contracts/L2BalancerPseudoMinter.sol
+++ b/pkg/liquidity-mining/contracts/L2BalancerPseudoMinter.sol
@@ -1,0 +1,134 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+
+import "@balancer-labs/v2-interfaces/contracts/liquidity-mining/IChildChainGauge.sol";
+import "@balancer-labs/v2-solidity-utils/contracts/helpers/SingletonAuthentication.sol";
+import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/SafeERC20.sol";
+import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/SafeMath.sol";
+
+import "./BalancerMinter.sol";
+
+/**
+ * @dev Distributes bridged BAL tokens in child chains, using the same interface as the mainnet Balancer minter.
+ *
+ * This contract is analogous to the mainnet minter: it has the same interface and interacts with (L2) gauges in a
+ * similar manner, keeping track of how many tokens were already distributed to each gauge and user.
+ *
+ * The difference with the mainnet minter is that this contract does not have a way of minting BAL directly: the tokens
+ * are only minted in mainnet, and then bridged to L2s. Then, this contract accumulates the emissions received by
+ * child chain gauges and distributes them to users.
+ *
+ * Every time the tokens are 'minted' (i.e. distributed) from a gauge to a given user, the pseudo minter will query the
+ * child chain gauge for the total amount of tokes that need to be distributed to that user. By keeping track of the
+ * amount that it has already distributed for that gauge / user, the pseudo minter can then transfer the difference
+ * to the user and update the total transferred amount.
+ */
+contract L2BalancerPseudoMinter is BalancerMinter, SingletonAuthentication {
+    event GaugeFactoryAdded(ILiquidityGaugeFactory indexed factory);
+    event GaugeFactoryRemoved(ILiquidityGaugeFactory indexed factory);
+
+    using SafeMath for uint256;
+    using SafeERC20 for IERC20;
+
+    mapping(ILiquidityGaugeFactory => bool) private _validFactories;
+
+    constructor(IVault vault, IERC20 balancerToken)
+        BalancerMinter(balancerToken, "Balancer Pseudo Minter", "1")
+        SingletonAuthentication(vault)
+    {
+        // solhint-disable-previous-line no-empty-blocks
+    }
+
+    /**
+     * @notice Adds a given child chain gauge factory to the allowlist.
+     * @dev This is a permissioned function.
+     * Reverts if the given factory was added beforehand; emits `GaugeFactoryAdded` event upon success.
+     */
+    function addGaugeFactory(ILiquidityGaugeFactory factory) external authenticate {
+        require(!_validFactories[factory], "FACTORY_ALREADY_ADDED");
+        _validFactories[factory] = true;
+        emit GaugeFactoryAdded(factory);
+    }
+
+    /**
+     * @notice Removes a given child chain gauge factory from the allowlist.
+     * @dev This is a permissioned function.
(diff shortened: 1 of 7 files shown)


pkg/liquidity-mining/contracts/L2BalancerPseudoMinter.sol
```diff
@@ -0,0 +1,134 @@
...
+    /**
+     * @notice Removes a given child chain gauge factory from the allowlist.
+     * @dev This is a permissioned function.
+     * Reverts if the given factory had not been added beforehand; emits `GaugeFactoryRemoved` event upon success.
+     */
+    function removeGaugeFactory(ILiquidityGaugeFactory factory) external authenticate {
+        require(_validFactories[factory], "FACTORY_NOT_ADDED");
+        _validFactories[factory] = false;
+        emit GaugeFactoryRemoved(factory);
+    }
```
jubeira, line 75: We might not want this, since removing a factory effectively kills all the gauges created by it (`_updateGauge` would revert).

We already have a `kill` permissioned function for individual gauges, so this would be just a shortcut. I've added it to the initial version because it's a low hanging fruit and for discussion purposes mostly.

nventuro: Would this affect the other reward tokens in any way?

jubeira: No, it should not.
This would make any calls to `mint` (mint for / mint many, etc) revert, but that's only used for BAL emissions.

One detail is that if you call `mintMany`, or `mintManyFor` with a gauge whose factory's been killed, the whole operation would revert, even if there are some valid gauges in the array. On the other hand, calling `mintMany` with killed gauges would not revert, it would just waste some gas.

But you can always choose what gauges to add to the array, so I don't see it as a huge issue anyways.

pkg/liquidity-mining/contracts/L2BalancerPseudoMinter.sol
```diff
@@ -0,0 +1,134 @@
...
+     * total amount of tokens as indicated by the gauge and the tokens that have already been transferred to the user.
+     */
+    function _updateGauge(address gauge, address user) internal returns (uint256 tokensToMint) {
+        // First, we retrieve the factory address registered from the gauge.
+        // If the factory address is allowlisted in this contract, we verify that the gauge was actually created by
+        // the factory (otherwise it could be just a malicious gauge that claims to be created by an allowed factory).
+        IChildChainGauge ccGauge = IChildChainGauge(gauge);
+        ILiquidityGaugeFactory factory = ccGauge.factory();
+        require(isValidGaugeFactory(factory), "INVALID_GAUGE_FACTORY");
+        require(factory.isGaugeFromFactory(gauge), "INVALID_GAUGE");
```
nventuro, line 115: I think handling permissions this way is fine.

An alternative way would be to have an 'addGauge' function here that is permissioned, and then grant permission to the factory to call it. The problem with that is we'd need to assign a delay to granting said permission, which would have parallels with the GaugeController's add_gauge, but there's no gauge controller here (also I find delay on grant is harder to understand).

I'd rather we keep it like this, same as the L1 (someone needs to register new factories).

nventuro (commented):
Looking good! Let's definitely add tests here though, I don't see a need to split those somewhere else.

pkg/liquidity-mining/contracts/L2BalancerPseudoMinter.sol
jubeira: +1; I also find it more intuitive.
The difference with the L1 is the back and forth between the gauge and the factory (i.e. the gauge needs to hold a reference to the factory so that the pseudo minter can then go and query the factory). But I think that is reasonable, and the overall system should be robust.

pkg/liquidity-mining/contracts/L2BalancerPseudoMinter.sol
nventuro: I think this is fine, yes. Thanks!

nventuro: Maybe document that though.

pkg/liquidity-mining/test/L2BalancerPseudoMinter.test.ts
```diff
@@ -0,0 +1,325 @@
...
+    it('reverts with valid gauge from invalid factory', async () => {
+      await gauges[0].setMockFactory(ANY_ADDRESS);
+      await expect(pseudoMinter.connect(user).mintMany(gaugeAddresses)).to.be.revertedWith('INVALID_GAUGE_FACTORY');
+    });
+
+    it('reverts with malicious gauge prentending to come from a valid factory', async () => {
+      const maliciousGauge = await deploy('MockChildChainGauge', { args: ['test'] });
+      await maliciousGauge.setMockFactory(gaugeFactory.address);
+      const gaugesWithMalicious = [...gaugeAddresses, maliciousGauge.address];
+      await expect(pseudoMinter.connect(user).mintMany(gaugesWithMalicious)).to.be.revertedWith('INVALID_GAUGE');
```
nventuro, line 240: Try also passing an EOA as the gauge (this will revert with no reason, since the EOA won't return any returndata), and a contract with no `factory` function (also revert with no reason, in this case because the callee reverted).

pkg/liquidity-mining/test/L2BalancerPseudoMinter.test.ts
```diff
@@ -0,0 +1,325 @@
...
+
+      it('updates total minted amount for the user for each gauge', async () => {
+        const totalsBefore = await Promise.all(gauges.map((gauge) => pseudoMinter.minted(user.address, gauge.address)));
+        await pseudoMinter.connect(user).mintMany(gaugeAddresses);
+        const totalsAfter = await Promise.all(gauges.map((gauge) => pseudoMinter.minted(user.address, gauge.address)));
+        expect(totalsAfter).to.be.deep.eq(totalsBefore.map((totalBefore) => totalBefore.add(mockCheckpointStep)));
+      });
+
+      it('emits an event per gauge', async () => {
+        await Promise.all(
```
nventuro, line 268: `inReceipt` is not an async function, so you don't need to do this.

pkg/liquidity-mining/test/L2BalancerPseudoMinter.test.ts
```diff
@@ -0,0 +1,325 @@
...
+              recipient: user.address,
+              gauge: gauge.address,
+              minted: mockCheckpointStep,
+            })
+          )
+        );
+      });
+
+      it('calls gauge user checkpoint for every gauge', async () => {
+        await Promise.all(
```
nventuro, line 280: nor here

nventuro (approved):

pkg/liquidity-mining/test/L2BalancerPseudoMinter.test.ts
jubeira: :sweat_smile: I got carried out.

jubeira:
Thanks for the review! I'll merge after CI.

PR #2317, opened 2023-03-03
Rename TimelockExecutor, add docs

nventuro:
# Description

This renames the `TimelockExecutor` to `TimelockExecutionHelper`, since we already have 'executors' (i.e. accounts that can execute a scheduled execution), to avoid confusion. It also does some light renaming internally (e.g. `onlyExecutor` became `onlyScheduled`), and adds documentation around why the ExecutionHelper exists, how `schedule`, `execute` and `canPerform` work, etc. There are not functionality changes, which is why tests remain the same except to adjust for the renaming.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -407,18 +428,21 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
...
      */
     function canPerform(
         bytes32 actionId,
         address account,
         address where
     ) public view override returns (bool) {
         if (msg.sender == address(_authorizerAdaptor)) {
-            // We special case the situation where the caller is the `AuthorizerAdaptor`, as it can be tricked into
-            // passing an incorrect `actionId` value, potentially resulting in escalation of privileges.
+            // We special case the situation where the caller is the `AuthorizerAdaptor`, as due to a bug it can be
```
ylv-io, line 456:
```suggestion
            // In the situation where the caller is the `AuthorizerAdaptor`, as due to a bug it can be
```

ylv-io (approved):
More clarity 👁️
All good except small typo in one comment.

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -224,8 +224,29 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
...
-        require(msg.sender == address(_executor), "CAN_ONLY_BE_SCHEDULED");
+    /**
+     * @dev Prevents a TimelockAuthorizer function from being called directly, making it only possible to call it by
+     * scheduling a delayed execution.
+     *
+     * Each function that has this modifier applied to it should have an associated function that performs proper
+     * permission validation and then schedules a call.
+     */
+    modifier onlyScheduled() {
+        // Checking that we're being called by the TimelockExecutionHelper is a suffient check, given that:
```
EndymionJkb, line 235:
```suggestion
        // Checking that we're being called by the TimelockExecutionHelper is a sufficient check, given that:
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -224,8 +224,29 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
...
+        //
+        //  1) The TimelockExecutionHelper can only make external calls (and cause for this modifier to not revert) if
```
EndymionJkb, line 237:
```suggestion
        //  1) The TimelockExecutionHelper can only make external calls (and cause this modifier to not revert) if
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -224,8 +224,29 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
...
+        //
+        //  2) The TimelockAuthorizer only makes external non-view calls in a single place: when the `execute` function
+        //    is called by an executor. This is the only possible time it could call the TimelockExecutionHelper.
+        //
+        //  3) `execute` can only be called after scheduling a delayed execution.
+        //
+        //  4) Scheduled delayed executions either target the TimelockAuthorizer directly (such as in
+        //    `scheduleRootChange` or `scheduleDelayChange`), in which this modifier will not revert (as intended, given
+        //    those functions check proper permissions), or explictly forbid targeting the TimelockAuthorizer (in the
+        //    `schedule` function), making it impossible for the TimelockExecutionHelper to call into it.
```
EndymionJkb, line 248:
```suggestion
        //    `scheduleRootChange` or `scheduleDelayChange`), in which case this modifier will not revert (as intended,
        //    given those functions check proper permissions), or explictly forbid targeting the TimelockAuthorizer
        //    (in the `schedule` function), making it impossible for the TimelockExecutionHelper to call into it.
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -442,8 +466,25 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
             return account == address(_authorizerAdaptorEntrypoint);
         }
 
+        // Actions with no delay can only be performed by accounts that have the associated permission.
+        // Actions with a non-zero delay however cannot be performed by permissioned accounts: they can only be made by
```
EndymionJkb, line 470:
```suggestion
        // However, actions with a non-zero delay cannot be performed by permissioned accounts: they can only be made by
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -442,8 +466,25 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
...
+        // the TimelockAuthorizerExecutionHelper, which works alongisde the TimelockAuthorizer itself to ensure that
+        // said executions have been properly scheduled in advance by an authorized party via the `schedule` function.
```
EndymionJkb, line 472:
```suggestion
        // executions have been properly scheduled in advance by an authorized party via the `schedule` function.
```
Not technically wrong, but using "said" like that is most common as "legalese" in court documents. I'd try to limit it. Can say "aforementioned" if necessary to mean the same thing

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -664,13 +705,28 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
     }
 
     /**
-     * @notice Schedules an arbitrary execution of `data` in target `where`.
+     * @notice Schedules an arbitrary execution of `data` in target `where`. Returns a scheduledExecutionId that can be
+     * used to call `execute`, `cancel`, and associated getters such as `getScheduledExecution`.
+     *
+     * If `executors` is an empty array, that any account in the network will be able to initiate the scheduled
```
EndymionJkb, line 711:
```suggestion
     * If `executors` is an empty array, then any account in the network will be able to initiate the scheduled
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -664,13 +705,28 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
...
+     * change this after scheduling: the list of executors is immutable, and cannot be changed by any account (including
+     * root).
+     *
+     * The caller of the `schedule` function is automatically made a canceler for the scheduled execution, meaning they
+     * can call the `cancel` function for it. Other accounts, such as root, may also have or be granted permission to
+     * cancel any scheduled execution.
+     *
+     * This is the only way to execute actions in external contracts that have a delay associated with them. Calling
+     * said functions directly will cause `canPerform` to return false, even if the caller has permission. An account
+     * that has permission over an action with a delay cannot call it directly, and must intead schedule a delayed
```
EndymionJkb, line 722:
```suggestion
     * that has permission over an action with a delay cannot call it directly, and must instead schedule a delayed
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -683,29 +739,45 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
         // these actions can only be scheduled through specialised functions.
         require(where != address(this), "CANNOT_SCHEDULE_AUTHORIZER_ACTIONS");
 
-        // We also disallow the TimelockExecutor from attempting to call into itself. Otherwise the above protection
-        // could be bypassed by wrapping a call to `setPendingRoot` inside of a call causing the TimelockExecutor to
-        // reenter itself, essentially hiding the fact that `where == address(this)` inside `data`.
+        // We also disallow the TimelockExecutionHelper from attempting to call into itself. Otherwise the above
+        // protection could be bypassed by wrapping a call to `setPendingRoot` inside of a call causing the
```
EndymionJkb, line 743:
```suggestion
        // protection could be bypassed by wrapping a call to `setPendingRoot` inside of a call, causing the
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -683,29 +739,45 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
...
+        // TimelockExecutionHelper to reenter itself, essentially hiding the fact that `where == address(this)` inside
+        // `data`.
         //
-        // Note: The TimelockExecutor only accepts calls from the TimelockAuthorizer (i.e. not from itself) so this
-        // scenario should be impossible but this check is cheap so we enforce it here as well anyway.
-        require(where != address(_executor), "ATTEMPTING_EXECUTOR_REENTRANCY");
+        // Note: The TimelockExecutionHelper only accepts calls from the TimelockAuthorizer (i.e. not from itself) so
+        // this scenario should be impossible but this check is cheap so we enforce it here as well anyway.
```
EndymionJkb, line 748:
```suggestion
        // this scenario should be impossible: but this check is cheap so we enforce it here as well anyway.
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -683,29 +739,45 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
...
+        return scheduledExecutionId;
     }
 
     /**
-     * @notice Executes a scheduled action `scheduledExecutionId`.
+     * @notice Executes a scheduled action `scheduledExecutionId`. This is used to execute all scheduled executions,
+     * both those that originate from `schedule` but also internal TimelockAuthorizer functions such as
+     * `scheduleRootChange` or `scheduleDelayChange`.
+     *
+     * If any executors were setup when scheduling, `execute` can only be called by them. If none were set, the
```
EndymionJkb, line 769:
```suggestion
     * If any executors were set up when scheduling, `execute` can only be called by them. If none were set, the
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -683,29 +739,45 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
...
+     * scheduled execution is said to be 'unprotected', and can be executed by anyone.
+     *
+     * Once executed, a scheduled execution cannot be executed again. It also cannot be executed if canceled.
+     *
+     * We mark this function as `nonReentrant` out of an abundance of caution, as in theory this and the Authorizer
+     * should be resilient to reentrant executions. The non-reentrancy check means that it is not possible to execute a
+     * scheduled action during the execution of another scheduled action - an unlikely and convoluted scenario that we
+     * knowlingly forbid.
```
EndymionJkb, line 777:
```suggestion
     * explicitly forbid.
```

EndymionJkb (approved):
Modulo a couple typos and suggestions, should be good to go

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
EndymionJkb: This isn't wrong even as written, since "special case" is commonly used as a verb like that, but the rewrite is even clearer.

nventuro:
I can't write to save my life :sob:

PR #2322, opened 2023-03-06
`BalancerMinter` tests

jubeira:
# Description

Add tests for the base `BalancerMinter` layer, which is now common to the mainnet minter (already deployed) and the L2 pseudo minter (not yet deployed).
No changes in functionality, just tests.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other

## Checklist:

## Issue Resolution

N/A

### DIFF
--- a/pkg/liquidity-mining/test/BalancerMinter.test.ts
+++ b/pkg/liquidity-mining/test/BalancerMinter.test.ts
@@ -1,25 +1,249 @@
 import { ethers } from 'hardhat';
-import { Contract } from 'ethers';
+import { BigNumber, Contract, ContractReceipt } from 'ethers';
 import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/dist/src/signer-with-address';
 
+import { sharedBeforeEach } from '@balancer-labs/v2-common/sharedBeforeEach';
 import { deploy } from '@balancer-labs/v2-helpers/src/contract';
 import { expect } from 'chai';
-import { ZERO_ADDRESS } from '@balancer-labs/v2-helpers/src/constants';
+import { ANY_ADDRESS, ZERO_ADDRESS } from '@balancer-labs/v2-helpers/src/constants';
 import { BalancerMinterAuthorization } from '@balancer-labs/balancer-js';
 import { currentTimestamp, HOUR } from '@balancer-labs/v2-helpers/src/time';
 import * as expectEvent from '@balancer-labs/v2-helpers/src/test/expectEvent';
+import { randomAddress } from '@balancer-labs/v2-helpers/src/constants';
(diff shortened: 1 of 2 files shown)


pkg/liquidity-mining/test/BalancerMinter.test.ts
```diff
@@ -1,25 +1,249 @@
...
+      itCallsVirtualHooks();
+    });
+
+    describe('mintManyFor', () => {
+      sharedBeforeEach(async () => {
+        mintManyFunction = async () =>
+          await (await minterContract.connect(minter).mintManyFor(gauges, user.address)).wait();
+        mintManyStaticFunction = async () =>
+          await minterContract.connect(minter).callStatic.mintManyFor(gauges, user.address);
+      });
```
nventuro, line 214: You could also test `user` calling these if you wanted

pkg/liquidity-mining/test/BalancerMinter.test.ts
```diff
@@ -1,25 +1,249 @@
...
+        mintManyStaticFunction = async () => await minterContract.connect(user).callStatic.mint_many(gauges8);
+        gauges8 = await Promise.all(range(8).map(randomAddress)); // mint_many accepts a fixed length array of 8.
+      });
+
+      it('calls internal _mintFor virtual hook for each gauge', async () => {
+        const receipt = await mintManyFunction();
+        for (const gauge of gauges8) {
+          expectEvent.inReceipt(receipt, 'MintFor', { gauge, user: user.address });
+        }
+      });
```
nventuro, line 245: Why are you not calling `itCallsVirtualHooks();` here?

nventuro (approved):
Looks great! Thanks for improving these :raised_hands:

pkg/liquidity-mining/test/BalancerMinter.test.ts
jubeira: because in this case `mint_many` ends up calling `_mintFor` repeateadly instead of `_mintForMany`, so it's not equivalent to `mintMany` :(

pkg/liquidity-mining/test/BalancerMinter.test.ts
jubeira: actually... it's the same case.
If user calls `mintFor` for themselves, it'll revert because by default it's not allowed to. In other words, to call `mintFor` you'd need to first call `setMinterApproval` for yourself.

We could correct that by changing the require to
```
        require(_allowedMinter[msg.sender][user] || msg.sender == user, "Caller not allowed to mint for user");
```

PR #2323, opened 2023-03-06
Remove unused actionIds and IAuthenthicate interface

nventuro:
# Description

This removes the `IAuthenticate` interface (since we don't use the `getActionId` function, and #2241 will introduce custom events anyway), as well as some leftover functions and constants that were not removed in previous PRs but are no longer used (`_canPerformSpecificallyOrGenerally`, general permission specifiers, comments about base action ids, etc.). 

I also ended up removing the `TimelockAuthorizerRoot` tests, since the way root works now has the permissions baked in as opposed to granted, so we don't need to test those in the complex way we were doing it before, and instead added simple tests to check that `isGranter` and `isRevoker` return true for root.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [x] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes https://github.com/balancer/balancer-v2-monorepo/issues/2316


ylv-io (approved):
This is extremely satisfying PR. I like how much more easy to understand Authorizer is now.

PR #2325, opened 2023-03-06
Individual checkpoints for `ChildChainGauge` rewards

### LINKED ISSUE #2286: Skip tokens when claiming rewards
`ChildChainGauge` only supports claiming every registered reward at once. This mechanism would break if one of the registered tokens goes bad and reverts during a `transfer`, effectively locking the other rewards as well in the gauge since there's no way to remove reward tokens.

To prevent this, the proposal is to add the capability to select which tokens to claim (or skip), so that if one token goes rogue LPs can still claim the other rewards.

jubeira:
# Description

This PR allows selecting which rewards to checkpoint, instead of performing a checkpoint for all of them.
This makes the system more robust in case one of the added rewards reverts when transferring tokens, which would deny claiming any of the sane rewards otherwise.

Some notes about the changes:

- Checkpointing all the rewards is potentially unsafe only when claiming. Without the `_claim` flag, no external calls are made inside `_checkpoint_rewards`.
- `_checkpoint_rewards` makes use of the user's balance of gauge tokens to do the math. This means that whenever the gauge token balance changes for a user (i.e. `deposit`, `withdraw` and `transfer`), a checkpoint for the rewards has to be performed. Exposing the capability to checkpoint arbitrary rewards in those cases could be problematic, because otherwise the checkpoint after a balance change would be incorrect for rewards that were skipped during the balance change.
- With this change all the rewards shall be checkpointed whenever there is a balance change just like before. But since the external transfer might be problematic, we won't be able to claim the rewards in the same function call (that's why `_claim_rewards` flag is disabled in those cases). As a rule of thumb: if we are not claiming, we can (and should) checkpoint all the rewards. If we are just claiming, it is fine to choose which ones to checkpoint and claim.

And a few comments for the reviewer:
- I'm assuming that disallowing external token transfers during `transfer`, `withdraw` and `deposit` is fine and doesn't affect usability.
- Each reward stores it's state separately, and neither checkpoints nor claims for a given reward affect any of the others. Therefore, skipping problematic ones when claiming should be fine.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #2286


pkg/liquidity-mining/contracts/gauges/ChildChainGauge.vy
```diff
@@ -227,8 +243,14 @@ def _checkpoint_rewards(_user: address, _total_supply: uint256, _claim: bool, _r
...
-    for i in range(MAX_REWARDS):
-        if i == reward_count:
+    reward_indexes: DynArray[uint256, MAX_REWARDS] = []
+    if len(_input_reward_indexes) == 0:
+        reward_indexes = self._all_indexes()
+    else:
+        reward_indexes = _input_reward_indexes
+
+    for i in reward_indexes:
+        if i >= reward_count:
```
jubeira, line 253: The input indexes can potentially be higher than the reward count now. In that case the loop will just break, but we could revert as well.

pkg/liquidity-mining/contracts/gauges/ChildChainGauge.vy
```diff
@@ -1,4 +1,4 @@
-# @version 0.3.1
+# @version 0.3.3
```
EndymionJkb, line 1: Incrementing the version by 2?

jubeira: This is the Vyper compiler version.
I need 0.3.3 in particular to be able to use 'dynamic' arrays. This compiler version is [already listed](https://github.com/balancer/balancer-v2-monorepo/blob/45bfdc266fbff3b3b79cc68af7f4193b6ebd2124/pkg/liquidity-mining/hardhat.config.ts#L21) in the compilers we use, so we don't need to modify the general configuration for this to work.

pkg/liquidity-mining/contracts/gauges/ChildChainGauge.vy
```diff
@@ -210,8 +210,24 @@ def _update_liquidity_limit(_user: address, _user_balance: uint256, _total_suppl
...
+    return indexes
+
+
+@internal
+def _checkpoint_rewards(
+    _user: address,
+    _total_supply: uint256,
+    _claim: bool,
+    _receiver: address,
+    _input_reward_indexes: DynArray[uint256, MAX_REWARDS]
```
jubeira, line 229: Note: I've used input indexes instead of tokens so as to introduce smaller changes, but I guess we could also implement this as a list of tokens to redeem.

The indexes can be retrieved by reading the public `reward_tokens` array.

pkg/liquidity-mining/contracts/gauges/ChildChainGauge.vy
jubeira: Note for the reviewer: I changed my mind when writing tests. There is no reason to allow invalid indexes here, so I changed this check for an assertion which I believe is safer and more clear. Tests were updated accordingly.

pkg/liquidity-mining/test/ChildChainGauge.test.ts
```diff
@@ -0,0 +1,414 @@
...
+    }
+
+    describe('checkpoint', () => {
+      sharedBeforeEach('deposit BPT to gauge and "bridge" BAL rewards', async () => {
+        await stakeBPTAndBridgeBAL(bptAmount, bptAmount);
+      });
+
+      it('transfers balance of BAL to the pseudo minter on the first checkpoint', async () => {
+        // The first checkpoint updates liquidity limit with the new stake after depositing.
+        const tx1 = await gauge.connect(user1).user_checkpoint(user1.address);
```
nventuro, line 129: You could also test that the inflation rate for the current week has increased.

pkg/liquidity-mining/test/ChildChainGauge.test.ts
```diff
@@ -0,0 +1,414 @@
...
+        });
+
+        context('two users, unequal BPT stake', () => {
+          sharedBeforeEach(async () => {
+            await stakeBPTAndBridgeBAL(bptAmount, bptAmount.mul(2));
+            await checkpointAndAdvanceWeek();
+          });
+
+          // User 2 has double the stake, so 1/3 of the rewards go to User 1, and 2/3 go to User 2.
+          itMintsRewardsForUsers(balAmountPerWeek.div(3), balAmountPerWeek.div(3).mul(2));
```
nventuro, line 204:
```suggestion
          itMintsRewardsForUsers(balAmountPerWeek.div(3), balAmountPerWeek.mul(2).div(3));
```

pkg/liquidity-mining/test/ChildChainGauge.test.ts
```diff
@@ -0,0 +1,414 @@
...
+          });
+
+          // Maximum boosted rewards is 2.5x larger than non-boosted rewards.
+          // Since User 1 is the only boosted user and the stake among the two users is the same, the adjusted working
+          // balance for User 1 will be 2.5x the adjusted working balance for User 2.
+          // Since the rewards are proportional to the adjusted working balance, then:
+          // (Rewards_User_1) + (Rewards_User_2) = Total_rewards
+          // 2.5 RU2 + RU2 = Total_rewards
+          // Solving for (RU1, RU2) --> User 1 gets 5/7 of the total, User 2 gets 2/7 of the total.
+          itMintsRewardsForUsers(balAmountPerWeek.mul(5).div(7), balAmountPerWeek.mul(2).div(7));
```
nventuro, line 231: Very nice

pkg/liquidity-mining/test/ChildChainGauge.test.ts
```diff
@@ -0,0 +1,414 @@
...
+    let selectedRewards: TokenList;
+    let claimer: SignerWithAddress;
+
+    function itTransfersRewardsToClaimer(claimFunction: () => Promise<unknown> = claimSelected) {
+      it("transfers rewards to claimer without affecting other users' rewards", async () => {
+        const claimerStake = await gauge.balanceOf(claimer.address);
+        const gaugeTotalSupply = await gauge.totalSupply();
+
+        // Claimer rewards are proportional to their BPT stake in the gauge given that staking time is constant for all
+        // users.
```
nventuro, line 300: For completeness you could give them boosts, which should not do anything

pkg/liquidity-mining/test/ChildChainGauge.test.ts
```diff
@@ -0,0 +1,414 @@
...
+
+    sharedBeforeEach('add reward', async () => {
+      await Promise.all(
+        rewards.addresses.map((reward) =>
+          vault.authorizerAdaptorEntrypoint
+            .connect(admin)
+            .performAction(gauge.address, gauge.interface.encodeFunctionData('add_reward', [reward, admin.address]))
+        )
+      );
+    });
```
nventuro, line 379: It'd be good to at least test the scenario where there are no reward tokens, and a scenario in which there are _some_ reward tokens, where otherwise valid indices (e.g. 5) then fail

pkg/liquidity-mining/contracts/gauges/ChildChainGauge.vy
```diff
@@ -227,9 +243,15 @@ def _checkpoint_rewards(_user: address, _total_supply: uint256, _claim: bool, _r
...
-        if i == reward_count:
-            break
+    reward_indexes: DynArray[uint256, MAX_REWARDS] = []
+    if len(_input_reward_indexes) == 0:
+        reward_indexes = self._all_indexes()
+    else:
+        reward_indexes = _input_reward_indexes
+
+    for i in reward_indexes:
+        assert i < reward_count, "INVALID_REWARD_INDEX"
```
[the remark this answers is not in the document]
jubeira: yes, it should be fine.
I'll add a test that does the same claim twice to verify.

jubeira:
Thanks for the review; the comments have been addressed. This should be ready for a second pass.

pkg/liquidity-mining/test/ChildChainGauge.test.ts
```diff
@@ -523,6 +523,88 @@ describe('ChildChainGauge', () => {
...
+      it('can still claim working rewards', async () => {
+        const tx = await gauge.connect(claimer)[claim](claimer.address, ZERO_ADDRESS, [1]);
+        const event = expectTransferEvent(
+          await tx.wait(),
+          { from: gauge.address, to: claimer.address },
+          functionalReward.address
+        );
+
+        expect(event.args.value).to.be.almostEqual(rewardAmount);
+      });
```
nventuro, line 582: Can you add a test where we claim the working rewards, then fix the broken rewards, then reclaim, and

a) the broken rewards are claimed correctly
b) the already claimed rewards are not re-claimed

(i.e. claims are fully independent)

jubeira: https://github.com/balancer/balancer-v2-monorepo/pull/2325/commits/15ada0d0912fef81f13f6c11475c3f09d1563dcb there we go.

nventuro (approved):

PR #2326, opened 2023-03-06
Better self scheduling

### LINKED ISSUE #2241: Improve self-scheduling

ylv-io:
# Description

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #2241

### DIFF
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -140,6 +140,36 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
     ScheduledExecution[] private _scheduledExecutions;
 
     /**
+     * @notice Emitted when a revoke permission is scheduled.
+     */
+    event RevokePermissionScheduled(bytes32 indexed actionId, uint256 indexed scheduledExecutionId);
+
+    /**
+     * @notice Emitted when a grant permission is scheduled.
+     */
+    event GrantPermissionScheduled(bytes32 indexed actionId, uint256 indexed scheduledExecutionId);
+
+    /**
+     * @notice Emitted when a revoke delay change is scheduled.
+     */
+    event RevokeDelayChangeScheduled(bytes32 indexed actionId, uint256 indexed scheduledExecutionId);
+
+    /**
+     * @notice Emitted when a grant delay change is scheduled.
+     */
+    event GrantDelayChangeScheduled(bytes32 indexed actionId, uint256 indexed scheduledExecutionId);
+
+    /**
+     * @notice Emitted when a delay change is scheduled.
+     */
+    event DelayChangeScheduled(bytes32 indexed actionId, uint256 indexed scheduledExecutionId);
+
+    /**
+     * @notice Emitted when a root change is scheduled.
+     */
+    event RootChangeScheduled(bytes32 indexed actionId, uint256 indexed scheduledExecutionId);
+
+    /**
      * @notice Emitted when a new execution `scheduledExecutionId` is scheduled.
      */
     event ExecutionScheduled(bytes32 indexed actionId, uint256 indexed scheduledExecutionId);
@@ -525,7 +555,9 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
 
         // Since this can only be called by root, which is always a canceler for all scheduled executions, we don't
         // bother creating any new cancelers.
-        return _scheduleWithDelay(actionId, address(this), data, getRootTransferDelay(), executors);
+        uint256 scheduledExecutionId = _scheduleWithDelay(actionId, address(this), data, getRootTransferDelay(), executors);
+        emit RootChangeScheduled(actionId, scheduledExecutionId);
+        return scheduledExecutionId;
     }
 
     /**
@@ -632,7 +664,9 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
 
         // Since this can only be called by root, which is always a canceler for all scheduled executions, we don't
         // bother creating any new cancelers.
-        return _scheduleWithDelay(scheduleDelayActionId, address(this), data, executionDelay, executors);
+        uint256 scheduledExecutionId = _scheduleWithDelay(scheduleDelayActionId, address(this), data, executionDelay, executors);
+        emit DelayChangeScheduled(actionId, scheduledExecutionId);
+        return scheduledExecutionId;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -525,7 +555,9 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
 
         // Since this can only be called by root, which is always a canceler for all scheduled executions, we don't
         // bother creating any new cancelers.
-        return _scheduleWithDelay(actionId, address(this), data, getRootTransferDelay(), executors);
+        uint256 scheduledExecutionId = _scheduleWithDelay(address(this), data, getRootTransferDelay(), executors);
+        emit RootChangeScheduled(actionId, scheduledExecutionId);
```
ylv-io, line 559: I know that `actionId` is always the same but I say we keep it for consistency but it is a pretty weak opinion of mine.

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -642,18 +666,19 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
         bytes32 actionId,
         uint256 newDelay,
         address[] memory executors
-    ) external returns (uint256 scheduledExecutionId) {
+    ) external returns (uint256) {
```
ylv-io, line 669: I removed these for consistency. Not sure what are our codesyte regarding return var names.

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -627,12 +650,13 @@ contract TimelockAuthorizer is IAuthorizer, IAuthentication, ReentrancyGuard {
...
         uint256 executionDelay = _getDelayChangeExecutionDelay(_delaysPerActionId[actionId], newDelay);
 
-        bytes32 scheduleDelayActionId = getScheduleDelayActionId(actionId);
         bytes memory data = abi.encodeWithSelector(this.setDelay.selector, actionId, newDelay);
 
         // Since this can only be called by root, which is always a canceler for all scheduled executions, we don't
         // bother creating any new cancelers.
-        return _scheduleWithDelay(scheduleDelayActionId, address(this), data, executionDelay, executors);
+        uint256 scheduledExecutionId = _scheduleWithDelay(address(this), data, executionDelay, executors);
+        emit DelayChangeScheduled(actionId, scheduledExecutionId);
```
ylv-io, line 658: I dropped `scheduleDelayActionId` in favor of `actionId`. I assume we don't need the specifier anymore because we now have specific event `DelayChangeScheduled`. But I might be wrong. Let me know.

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
EndymionJkb: Generally an explicit return is preferred, vs named variables, but there are some good reasons to use named return variables; e.g., when: 1) it is clearer (e.g., the return value is set inside complex logic); 2) we're trying to save bytecode; 3) it makes getting return values from tests a little easier

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -456,11 +486,11 @@ contract TimelockAuthorizer is IAuthorizer, ReentrancyGuard {
         require(isRoot(msg.sender), "SENDER_IS_NOT_ROOT");
         bytes memory data = abi.encodeWithSelector(this.setPendingRoot.selector, newRoot);
 
-        // TODO: fix event action id
-
         // Since this can only be called by root, which is always a canceler for all scheduled executions, we don't
         // bother creating any new cancelers.
-        return _scheduleWithDelay(0x0, address(this), data, getRootTransferDelay(), executors);
+        uint256 scheduledExecutionId = _scheduleWithDelay(address(this), data, getRootTransferDelay(), executors);
+        emit RootChangeScheduled(scheduledExecutionId);
```
nventuro, line 492:
```suggestion

        emit RootChangeScheduled(scheduledExecutionId);
```

Highly subjective, but I'd add whitespace here to separate the complex call from smaller bits like the event and return value :upside_down_face:

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
nventuro: Correct!

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -907,16 +946,15 @@ contract TimelockAuthorizer is IAuthorizer, ReentrancyGuard {
         bytes memory data = abi.encodeWithSelector(this.grantPermissions.selector, _ar(actionId), account, _ar(where));
 
         // TODO: fix actionId for event (maybe overhaul _scheduleWithDelay?)
```
nventuro, line 948:
```suggestion
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -679,7 +713,12 @@ contract TimelockAuthorizer is IAuthorizer, ReentrancyGuard {
         bytes32 actionId = IAuthentication(where).getActionId(_decodeSelector(data));
         require(hasPermission(actionId, msg.sender, where), "SENDER_DOES_NOT_HAVE_PERMISSION");
 
-        uint256 scheduledExecutionId = _schedule(actionId, where, data, executors);
+        uint256 delay = _delaysPerActionId[actionId];
+        require(delay > 0, "CANNOT_SCHEDULE_ACTION");
```
nventuro, line 717:
```suggestion
        require(delay > 0, "CANNOT_SCHEDULE_ACTION");

```

Here I'd similarly separate the input check from the state change

nventuro (commented):
Looks great! The only thing is the suggested extra event arguments, plus tests for those :)

nventuro (approved):

PR #2296, opened 2023-02-24
Reentrancy Preparation: Managed Pool

EndymionJkb:
# Description

Add reentrancy protection for Managed Pool. This is a new release, not a patch, so it is based off master. 

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Documentation or wording changes
- [ ] Other

## Checklist:

- [X] The diff is legible and has no extraneous changes
- [N/A] Complex code has been commented, including external interfaces
- [X] Tests are included for all code paths
- [X] The base branch is either `master`, or there's a description of how to merge

## Issue Resolution

Closes #2295

PR #2330, opened 2023-03-08
Weighted Pool: ongoing reentrancy protection

EndymionJkb:
# Description

WeightedPool (version 3) was patched for reentrancy; this applies the same (generalized) protections so that any future deployments are safe.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

- [N/A] Tests are included for all code paths (would be in the deployment fork test)


EndymionJkb:
Thanks, @jubeira !

jubeira:
Fixed compile issue with message `UnimplementedFeatureError: Encoding type "struct IVault.UserBalanceOp memory[] memory" not yet implemented.` (missing `pragma experimental ABIEncoderV2`) in https://github.com/balancer/balancer-v2-monorepo/pull/2330/commits/596aea82508a60545787919eeaa4be55d0538955.

nventuro:
This was the patch change https://github.com/balancer/balancer-v2-monorepo/pull/2205

nventuro (approved):
This seems to apply the same behavior as #2205, except that one also included changes to `updateProtocolFeePercentageCache` and this one doesn't (because that was added in https://github.com/balancer/balancer-v2-monorepo/pull/2296/files).

nventuro:
Incidentally #2296 did not add tests for the protocol fee cache update, but it should have.

pkg/pool-utils/contracts/RecoveryMode.sol
```diff
@@ -85,13 +86,24 @@ abstract contract RecoveryMode is IRecoveryMode, BasePoolAuthorization {
...
+     *
+     * See https://forum.balancer.fi/t/reentrancy-vulnerability-scope-expanded/4345 for reference.
      */
     function disableRecoveryMode() external override authenticate {
         // Some derived contracts respond to disabling recovery mode with state changes (e.g., related to protocol fees,
         // or otherwise ensuring that enabling and disabling recovery mode has no ill effects on LPs). When called
         // outside of recovery mode, these state changes might lead to unexpected behavior.
         _ensureInRecoveryMode();
 
+        VaultReentrancyLib.ensureNotInVaultContext(_getVault());
```
nventuro, line 105: This should be accompanied by a test.

nventuro (changes requested):

pkg/pool-utils/contracts/RecoveryMode.sol
jubeira: Side comment (maybe for a follow-up PR): we are doing this here and in `ProtocolFeeCache` (which is derived from `RecoveryMode`) without a modifier.

Wouldn't it be more consistent to define the modifier here, and use it in both contracts? That way the code would be the same both in pools and in these auxiliary contracts.

Qawtr (approved):

PR #2331, opened 2023-03-08
Stable Pool: ongoing reentrancy protection

EndymionJkb:
# Description

ComposableStablePool (version 3) was patched for reentrancy; this applies the same (generalized) protections so that any future deployments are safe.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

- [N/A] Tests are included for all code paths (would be fork tests in the deployment)


nventuro:
This was the patch change: https://github.com/balancer/balancer-v2-monorepo/pull/2206

pkg/pool-stable/contracts/ComposableStablePoolRates.sol
```diff
@@ -141,6 +142,8 @@ abstract contract ComposableStablePoolRates is IComposableStablePoolRates, Compo
 
     /// @inheritdoc IComposableStablePoolRates
     function setTokenRateCacheDuration(IERC20 token, uint256 duration) external override authenticate {
+        VaultReentrancyLib.ensureNotInVaultContext(_getVault());
```
jubeira, line 145: Why not define a modifier and use it here, like we did for the patch?

jubeira (commented):
First pass done.

Besides the inline comment,,
- don't we also need to protect `updateTokenRateCache` in `ComposableStablePoolRates`?
- what about `disableRecoveryMode`? It already has access to the vault, but I don't see the protection applied in this branch either.

nventuro (commented):
This seems to apply the same behavior as https://github.com/balancer/balancer-v2-monorepo/pull/2206, except that one also included changes to `updateProtocolFeePercentageCache` and `disableRecoveryMode` and this one doesnt (because those were added in https://github.com/balancer/balancer-v2-monorepo/pull/2296 and https://github.com/balancer/balancer-v2-monorepo/pull/2330 respectively).

EndymionJkb:
> * what about `disableRecoveryMode`? It already has access to the vault, but I don't see the protection applied in this branch either. EDIT: it's added in #2330

Yes; didn't want to potentially conflict.

jubeira (commented):
Patch looks good; all the functions that should be protected and documented are covered with this.

I've just left one super minor comment regarding a comment.

About the tests: we could include them in the fork tests as we did with the patches instead of doing unit tests. I don't have a strong opinion regarding what's the best way forward. WDYT @nventuro ?

PR #2336, opened 2023-03-08
Improve readme: suggested Node version

jubeira:
# Description

Small doc update; selecting the correct Node version unequivocally is handy.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

- N/A - Complex code has been commented, including external interfaces
- N/A - Tests are included for all code paths

## Issue Resolution

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,12 @@ Active development occurs in this repository, which means some contracts in it m
 
 The build & test instructions below should work out of the box with Node ^14.18.0. (Please note that it needs Node 14 specifically, and will NOT work with Node 16 or higher. Minor version should be at least 18).
 
+To use the suggested Node version, run:
+
+```bash
+$ nvm use
+```
+
 ## Clone
 
 This repository uses git submodules; use `--recurse-submodules` option when cloning. For example, using https:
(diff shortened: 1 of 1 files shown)


EndymionJkb (approved):
Should help! People might have different minor versions, so it's best not to give a specific version, like you have it.

README.md
```diff
@@ -33,6 +33,12 @@ Active development occurs in this repository, which means some contracts in it m
 
 The build & test instructions below should work out of the box with Node ^14.18.0. (Please note that it needs Node 14 specifically, and will NOT work with Node 16 or higher. Minor version should be at least 18).
 
+To use the suggested Node version, run:
+
+```bash
+$ nvm use
+```
+
```
nventuro, line 41: This implies that nvm is used? I don't use nvm myself.

jubeira: yes; this is just a shortcut if you are lazy to install a version that works manually (the right versions are specified above these lines). Do you suggest rewording this?

EndymionJkb: I suppose it could be generalized a bit, noting that using an incompatible Node version can cause issues, and that it's possible to install multiple Node versions using a manager such as nvm. If you use nvm, you can switch between them with `nvm use`. I think that's the most popular one, so reasonable to mention it like this.

PR #2338, opened 2023-03-08
Improve schedule delay change docs

### LINKED ISSUE #2243: Prevent nested delay changes
Delay changes are scheduled so that at any point in time the current delay is enforced, e.g. if an action's delay is decreased, we wait for the difference in delays before performing the change.

However, if multiple delay changes are scheduled for the same action, this property is lost. An action may have a delay, but have a scheduled change that will reduce it with little to no duration. We could consider having some sort of lock so that each delay can have one scheduled change at a time (maybe also exposing what the new delay will be).

Note that, as per #2238, some accounts would be able to cancel these, so not being able to create new scheduled executions would not be an issue. `cancel` would need to correctly update the lock though.

nventuro:
# Description

This simply improves documentation for the schedule delay change functions.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #2243 
Note that we're not actually preventing nested delays, as the issue instructed, since that would be too complicated and fragile. We instead advice to not do it, noting that the system becomes hard to analyze otherwise.

### DIFF
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -596,7 +596,26 @@ contract TimelockAuthorizer is IAuthorizer, ReentrancyGuard {
     }
 
     /**
-     * @notice Schedules an execution to set the delay for `actionId`' to `newDelay`.
+     * @notice Schedules an execution to set the delay for `actionId`' to `newDelay`. This makes it impossible to
+     * execute `actionId` without an immutable public on-chain committment for the execution at least `newDelay` seconds
+     * in advance.
+     *
+     * Critical actions that are expected to be performed by EOAs or multisigs are typically subject to said delays to
+     * allow for public scrutiny.
+     *
+     * How long it will take to make this change will depend on the current and new delays: if increasing by more than
+     * 5 days, then the time difference between the delays must pass. Otherwise, the minimum delay change execution
+     * delay of 5 days must pass instead.
+     *
+     * Only `executors` will be able to execute the scheduled action, unless `executors` is an empty array, in which
+     * case any account can execute it.
+     *
+     * Avoid scheduling multiple delay changes for the same action at the same time, as this makes it harder to reason
+     * about the state of the system. If there is already a scheduled delay change and there is a desire to change the
+     * future delay to some other value, cancel the first scheduled change and schedule a new one.
+     *
+     * Only root can call this function, but other accounts may be granted permission to cancel the scheduled execution
+     * (including global cancelers).
      */
     function scheduleDelayChange(
         bytes32 actionId,
@@ -619,7 +638,26 @@ contract TimelockAuthorizer is IAuthorizer, ReentrancyGuard {
     }
 
     /**
-     * @notice Schedules an execution to set the delay for granting permission over `actionId` to `newDelay`.
+     * @notice Schedules an execution to set the delay for granting permission over `actionId` to `newDelay`. This makes
+     * it impossible to grant permission to execute `actionId` without an immutable public on-chain committment for the
+     * granting at least `newDelay` seconds in advance.
+     *
+     * Critical actions that are expected to be performed by smart contracts are typically subject to said grant delays
+     * to allow for public scrutiny of new contracts that are granted the permission.
+     *
+     * How long it will take to make this change will depend on the current and new grant delays: if increasing by more
+     * than 5 days, then the time difference between the grant delays must pass. Otherwise, the minimum delay change
+     * execution delay of 5 days must pass instead.
+     *
+     * Only `executors` will be able to execute the scheduled action, unless `executors` is an empty array, in which
+     * case any account can execute it.
+     *
+     * Avoid scheduling multiple grant delay changes for the same action at the same time, as this makes it harder to
+     * reason about the state of the system. If there is already a scheduled grant delay change and there is a desire to
+     * change the future grant delay to some other value, cancel the first scheduled change and schedule a new one.
+     *
+     * Only root can call this function, but other accounts may be granted permission to cancel the scheduled execution
+     * (including global cancelers).
      */
     function scheduleGrantDelayChange(
         bytes32 actionId,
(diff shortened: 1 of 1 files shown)


pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -596,7 +596,26 @@ contract TimelockAuthorizer is IAuthorizer, ReentrancyGuard {
...
+     *
+     * Only `executors` will be able to execute the scheduled action, unless `executors` is an empty array, in which
+     * case any account can execute it.
+     *
+     * Avoid scheduling multiple delay changes for the same action at the same time, as this makes it harder to reason
+     * about the state of the system. If there is already a scheduled delay change and there is a desire to change the
+     * future delay to some other value, cancel the first scheduled change and schedule a new one.
+     *
+     * Only root can call this function, but other accounts may be granted permission to cancel the scheduled execution
+     * (including global cancelers).
```
ylv-io, line 618:
```suggestion
     * @notice Schedules an execution to set the delay for `actionId`' to `newDelay`. This makes it impossible to
     * execute `actionId` without an immutable public on-chain commitment for the execution at least `newDelay` seconds
     * in advance.
     *
     * Critical actions that are expected to be performed by EOAs or multisigs are typically subject to such delays to
     * allow for public scrutiny.
     *
     * How long it will take to make this change will depend on the current and new delays: if increasing by more than
     * 5 days, then the time difference between the delays must pass. Otherwise, the minimum delay change execution
     * delay of 5 days must pass instead.
     *
     * Only `executors` will be able to execute the scheduled action, unless `executors` is an empty array, in which
     * case any account can execute it.
     *
     * Avoid scheduling multiple delay changes for the same action at the same time, as this makes it harder to reason
     * about the state of the system. If there is already a scheduled delay change and there is a desire to change the
     * future delay to some other value, cancel the first scheduled change and schedule a new one.
     *
     * Only root can call this function, but other accounts may be granted permission to cancel the scheduled execution
     * (including global cancelers).
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -619,7 +638,26 @@ contract TimelockAuthorizer is IAuthorizer, ReentrancyGuard {
...
+     *
+     * Only `executors` will be able to execute the scheduled action, unless `executors` is an empty array, in which
+     * case any account can execute it.
+     *
+     * Avoid scheduling multiple grant delay changes for the same action at the same time, as this makes it harder to
+     * reason about the state of the system. If there is already a scheduled grant delay change and there is a desire to
+     * change the future grant delay to some other value, cancel the first scheduled change and schedule a new one.
+     *
+     * Only root can call this function, but other accounts may be granted permission to cancel the scheduled execution
+     * (including global cancelers).
```
ylv-io, line 660:
```suggestion
     * @notice Schedules an execution to set the delay for granting permission over `actionId` to `newDelay`. This makes
     * it impossible to grant permission to execute `actionId` without an immutable public on-chain commitment for the
     * granting at least `newDelay` seconds in advance.
     *
     * Critical actions that are expected to be performed by smart contracts are typically subject to such grant delays
     * to allow for public scrutiny of new contracts that are granted the permission.
     *
     * How long it will take to make this change will depend on the current and new grant delays: if increasing by more
     * than 5 days, then the time difference between the grant delays must pass. Otherwise, the minimum delay change
     * execution delay of 5 days must pass instead.
     *
     * Only `executors` will be able to execute the scheduled action, unless `executors` is an empty array, in which
     * case any account can execute it.
     *
     * Avoid scheduling multiple grant delay changes for the same action at the same time, as this makes it harder to
     * reason about the state of the system. If there is already a scheduled grant delay change and there is a desire to
     * change the future grant delay to some other value, cancel the first scheduled change and schedule a new one.
     *
     * Only root can call this function, but other accounts may be granted permission to cancel the scheduled execution
     * (including global cancelers).
```

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -641,7 +679,27 @@ contract TimelockAuthorizer is IAuthorizer, ReentrancyGuard {
...
+     *
+     * Only `executors` will be able to execute the scheduled action, unless `executors` is an empty array, in which
+     * case any account can execute it.
+     *
+     * Avoid scheduling multiple revoke delay changes for the same action at the same time, as this makes it harder to
+     * reason about the state of the system. If there is already a scheduled revoke delay change and there is a desire
+     * to change the future grant delay to some other value, cancel the first scheduled change and schedule a new one.
+     *
+     * Only root can call this function, but other accounts may be granted permission to cancel the scheduled execution
+     * (including global cancelers).
```
ylv-io, line 702:
```suggestion
     * @notice Schedules an execution to set the delay for revoking permission over `actionId` to `newDelay`. This makes
     * it impossible to revoke permission to execute `actionId` without an immutable public on-chain commitment for the
     * revoking at least `newDelay` seconds in advance.
     *
     * Critical actions that are performed by smart contracts and to which there is a long term commitment (e.g. minting
     * of BAL as part of the Liquidity Mining Program) are typically subject to such revoke delays, making it impossible
     * to disable the system without sufficient notice.
     *
     * How long it will take to make this change will depend on the current and new revoke delays: if increasing by more
     * than 5 days, then the time difference between the revoke delays must pass. Otherwise, the minimum delay change
     * execution delay of 5 days must pass instead.
     *
     * Only `executors` will be able to execute the scheduled action, unless `executors` is an empty array, in which
     * case any account can execute it.
     *
     * Avoid scheduling multiple revoke delay changes for the same action at the same time, as this makes it harder to
     * reason about the state of the system. If there is already a scheduled revoke delay change and there is a desire
     * to change the future grant delay to some other value, cancel the first scheduled change and schedule a new one.
     *
     * Only root can call this function, but other accounts may be granted permission to cancel the scheduled execution
     * (including global cancelers).
```

ylv-io (approved):
Great to have these explanations but I find them hard to understand. Saying that it is not easy to explain how the system operates in simple terms.

PR #2339, opened 2023-03-09
Batch Relayer: make deployable

EndymionJkb:
# Description

Turns out that adding 12 new protocols to the BatchRelayer took up a little bytecode: over 26k!

Luckily they are all very similar - and even already had a common base contract - so a lot of the code could be factored out.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/standalone-utils/contracts/relayer/IBaseRelayerLibrary.sol
+++ b/pkg/standalone-utils/contracts/relayer/IBaseRelayerLibrary.sol
@@ -61,13 +61,13 @@ abstract contract IBaseRelayerLibrary is AssetHelpers {
      */
     function _resolveAmountPullTokenAndApproveSpender(
         IERC20 token,
-        IERC20 wrappedToken,
+        address spender,
         uint256 amount,
         address sender
     ) internal returns (uint256 resolvedAmount) {
         resolvedAmount = _resolveAmountAndPullToken(token, amount, sender);
 
-        token.safeApprove(address(wrappedToken), resolvedAmount);
+        token.safeApprove(spender, resolvedAmount);
     }
 
     /**
--- a/pkg/standalone-utils/contracts/relayer/IBaseRelayerLibrary.sol
+++ b/pkg/standalone-utils/contracts/relayer/IBaseRelayerLibrary.sol
@@ -115,7 +115,7 @@ abstract contract IBaseRelayerLibrary is AssetHelpers {
     }
 
     /**
-     * Check for a chained output reference, and encode the given `amount` if necessary.
+     * @dev Check for a chained output reference, and encode the given `amount` if necessary.
      * This is internal, since some wrappers call it independently.
      */
     function _setChainedReference(uint256 outputReference, uint256 amount) internal {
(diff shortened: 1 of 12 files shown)


jubeira (approved):
Wow, nice improvement!

I've left a few minor comments, mostly about docs.

Side comment: at some point I was thinking whether it made sense to standarize how we name the variables. E.g. `mainToken` and `wrappedToken` could make sense, instead of using different names for each relayer. But this might be a better discussion for a future PR if we consider that it's worth it.

EndymionJkb:
> Side comment: at some point I was thinking whether it made sense to standarize how we name the variables. E.g. `mainToken` and `wrappedToken` could make sense, instead of using different names for each relayer. But this might be a better discussion for a future PR if we consider that it's worth it.

Yes; I wanted to make minimal changes, and not change any real code in the wrappers. Would be nice to at least have a convention going forward.

pkg/standalone-utils/contracts/relayer/IBaseRelayerLibrary.sol
```diff
@@ -50,4 +53,53 @@ abstract contract IBaseRelayerLibrary is AssetHelpers {
     function _setChainedReferenceValue(uint256 ref, uint256 value) internal virtual;
 
     function _getChainedReferenceValue(uint256 ref) internal virtual returns (uint256);
+
+    function _resolveAmountPullAndApproveToken(
```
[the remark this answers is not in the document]
rabmarut: Can we call it something like `_resolveAmountPullTokenAndApproveSpender`? The approved entity is not always the `wrappedToken` itself - and often, it's not even a token at all.

pkg/standalone-utils/contracts/relayer/IBaseRelayerLibrary.sol
```diff
@@ -50,4 +53,74 @@ abstract contract IBaseRelayerLibrary is AssetHelpers {
...
     function _getChainedReferenceValue(uint256 ref) internal virtual returns (uint256);
+
+    /**
+     * @dev This reuses `_resolveAmountAndPullToken` to adjust the `amount` if it is a chained reference, and
+     * pulls that amount of the `mainToken` to the relayer. Additionally, approve the `wrappedToken` to enable
+     * wrapping operations.
+     */
+    function _resolveAmountPullAndApproveToken(
+        IERC20 mainToken,
+        IERC20 wrappedToken,
```
rabmarut, line 64: Would suggest making this an `address` since that's all it really needs to be in order to make this approval, and in some cases (like Euler and Gearbox) we are writing things like this which don't make any sense because these aren't actually tokens:

```
IERC20(address(eulerProtocol))
```

pkg/standalone-utils/contracts/relayer/IBaseRelayerLibrary.sol
```diff
@@ -50,4 +53,74 @@ abstract contract IBaseRelayerLibrary is AssetHelpers {
...
+    ) internal {
+        if (recipient != address(this)) {
+            token.safeTransfer(recipient, amount);
+        }
+
+        _setChainedReference(outputReference, amount);
+    }
+
+    /**
+     * Check for a chained output reference, and encode the given `amount` if necessary.
```
rabmarut, line 118:
```suggestion
     * @dev Check for a chained output reference, and encode the given `amount` if necessary.
```

rabmarut (approved):
This is a really great change. I left a few small suggestions but I also approve unconditionally.

PR #2343, opened 2023-03-09
add shared before each

ylv-io:
# Description

No more!
image

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


nventuro:
@jubeira will like this

EndymionJkb:
> @jubeira will like this

Me too: I was never comfortable with this voodoo where it still finds things somehow even when they're not imported anywhere. There might be more things like this

jubeira (approved):
Yes!

I realized this was wrong only a few months ago, and I was slowly adding it to the files that I was modifying for other purposes one at the time (it never affected the functionality of the tests apparently).
+1 to solve those red lines once and for all.

EndymionJkb:
We just got used to all these little things over time.

We were boiled like frogs.

Thanks to the new frog for noticing the water temperature! :)

PR #2347, opened 2023-03-10
Fix Linear import

nventuro:
LinearPool doesn't use a price rate cache, so it doesn't need to import this library. It was getting `WordCodec` indirectly from it though, so we now explicitly.


nventuro:
Looks like we have a flaky test here. Looking into it.

nventuro:
The test failure was due to https://github.com/balancer/balancer-v2-monorepo/pull/2348.

EndymionJkb (approved):

PR #2357, opened 2023-03-14
Do not allow repeated executors

### LINKED ISSUE #2355: Forbid repeated Timelock scheduled execution executors
We currently allow for this, but there's no good reason to do so. We could easily check that the mapping entry is not yet set and revert if so.

ylv-io:
# Description

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Fixes #2355

### DIFF
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -1184,7 +1184,7 @@ contract TimelockAuthorizer is IAuthorizer, ReentrancyGuard {
         );
 
         for (uint256 i = 0; i < executors.length; i++) {
-            require(!_isExecutor[scheduledExecutionId][executors[i]], "ALREADY_AN_EXECUTOR");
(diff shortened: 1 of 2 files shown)


pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -1184,7 +1184,7 @@ contract TimelockAuthorizer is IAuthorizer, ReentrancyGuard {
         );
 
         for (uint256 i = 0; i < executors.length; i++) {
-            // Note that we allow for repeated executors - this is not an issue
+            require(!_isExecutor[scheduledExecutionId][executors[i]], "ALREADY_AN_EXECUTOR");
```
nventuro, line 1187:
```suggestion
            require(!_isExecutor[scheduledExecutionId][executors[i]], "DUPLICATE_EXECUTORS");
```

I think this is slightly more explicit - they weren't 'already' an executor (i.e. before scheduling), but rather they sent the same one multiple times

nventuro (approved):

PR #2358, opened 2023-03-14
Deployment: Batch Relayer V5

EndymionJkb:
# Deployment: Batch Relayer V5

- [X] Contracts are verified in every network 


pkg/deployments/action-ids/arbitrum/action-ids.json
```diff
@@ -590,5 +590,50 @@
         "renameProtocolId(uint256,string)": "0xdc8681b640b547e98ee6f4d708df559b807526d976644c3aca24bc12c4f839c3"
       }
     }
+  },
+  "20230314-batch-relayer-v5": {
+    "BatchRelayerLibrary": {
+      "useAdaptor": true,
+      "actionIds": {
+        "approveVault(address,uint256)": "0x0a5a412edef3a6979567862645360015b536a00bc877d64660ee50428fca701d",
```
nventuro, line 598: BatchRelayer is actually not a permissioned contract, so you don't need to add any action ids here. I'd remove these (which will also remove the collisions etc)

nventuro (commented):
CI is failing as we keep having issues with our gnosis RPC provider, but I checked addresses locally and they are indeed correct.

nventuro (approved):

PR #2365, opened 2023-03-16
Split and refactor TimelockAuthorizer tests

ylv-io:
# Description

This PR splits TimelockAuthorizer tests into the following files for the sake of clarity:
* TimelockAuthorizerActors
* TimelockAuthorizerExecute
* TimelockAuthorizerPermissions
* TimelockAuthorizerRoot

Tests are refactored for better readability at code and output level.
image
Two new tests added: `scheduleGrantPermission` and `scheduleRevokePermission`.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
```diff
@@ -0,0 +1,431 @@
...
+
+        const receipt = await authorizer.execute(id, { from: account });
+        expectEvent.inIndirectReceipt(
+          await receipt.wait(),
+          authenticatedContract.interface,
+          'ProtectedFunctionCalled',
+          {
+            data: functionData,
+          }
+        );
```
nventuro, line 312: If you really want to, you can use snapshots to undo the execute and then re-do it from the other account. But tbh I think that's overkill.

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
```diff
@@ -0,0 +1,431 @@
...
+      await expect(schedule(where)).to.be.revertedWith('CANNOT_SCHEDULE_AUTHORIZER_ACTIONS');
+    });
+
+    it('reverts the target is the execution helper', async () => {
+      const where = await authorizer.instance.getTimelockExecutionHelper();
+      await expect(schedule(where)).to.be.revertedWith('ATTEMPTING_EXECUTION_HELPER_REENTRANCY');
+    });
+  });
+
+  describe('execute', () => {
```
nventuro, line 258: If you're feeling spicy, we could test a scheduled execution where the scheduled action attempts to execute another (or even the same) scheduled action. This will fail because of the reentrancy guard.

nventuro, line 258: We should also look at `execute`'s return value, which will be whatever was returned by the target contract. You can query this by doing `authorizer.callStatic.execute(...)`

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
```diff
@@ -0,0 +1,431 @@
...
+      return id;
+    };
+
+    it('cancel the action', async () => {
+      const id = await schedule();
+      await authorizer.cancel(id, { from: canceler });
+
+      const scheduledExecution = await authorizer.getScheduledExecution(id);
+      expect(scheduledExecution.cancelled).to.be.true;
+    });
```
nventuro, line 397: I'd do this with both a specific and a global canceler, since we otherwise never have a global canceler use its permission (we only check that `isCanceler` returns true).  We could also add a third test with `root`, which will be able to do it even if it never received the permission.

pkg/vault/test/authorizer/TimelockAuthorizerRoot.test.ts
```diff
@@ -0,0 +1,149 @@
...
+            await expect(authorizer.execute(id)).to.be.revertedWith('ACTION_NOT_YET_EXECUTABLE');
+
+            await advanceTime(ROOT_CHANGE_DELAY);
+            await authorizer.execute(id);
+
+            expect(await authorizer.isRoot(root)).to.be.true;
+            expect(await authorizer.isPendingRoot(newPendingRoot)).to.be.true;
+          });
+
+          it('emits an event', async () => {
```
nventuro, line 75: I thought this test would check that `RootChangeScheduled` is correctly emitted.

The contents of this test are correct, but they'd fit better under e.g. 'sets the pending root when executed'

pkg/vault/test/authorizer/TimelockAuthorizerRoot.test.ts
```diff
@@ -0,0 +1,149 @@
...
+        ROOT_CHANGE_DELAY = await authorizer.instance.getRootTransferDelay();
+      });
+
+      sharedBeforeEach('initiate a root transfer', async () => {
+        const id = await authorizer.scheduleRootChange(user, [], { from: root });
+        await advanceTime(ROOT_CHANGE_DELAY);
+        await authorizer.execute(id);
+      });
+
+      context('when the sender is the pending root', async () => {
```
nventuro, line 123: I'd remove this `context`, it adds unnecessary nesting

pkg/vault/test/authorizer/TimelockAuthorizerRoot.test.ts
```diff
@@ -0,0 +1,149 @@
...
+          await authorizer.claimRoot({ from: user });
+          expect(await authorizer.isRoot(root)).to.be.false;
+          expect(await authorizer.isRoot(user)).to.be.true;
+        });
+
+        it('resets the pending root address to the zero address', async () => {
+          await authorizer.claimRoot({ from: user });
+          expect(await authorizer.isPendingRoot(root)).to.be.false;
+          expect(await authorizer.isPendingRoot(user)).to.be.false;
+          expect(await authorizer.isPendingRoot(ZERO_ADDRESS)).to.be.true;
```
nventuro, line 134: We also have `getPendingRoot` and `getRoot` :sweat_smile:

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
```diff
@@ -0,0 +1,802 @@
...
+
+        // should not be able to execute before delay
+        await expect(authorizer.execute(id, { from: root })).to.be.revertedWith('ACTION_NOT_YET_EXECUTABLE');
+
+        await advanceTime(delay);
+        await authorizer.execute(id, { from: root });
+
+        expect(await authorizer.canPerform(ACTION_1, user, WHERE_1)).to.be.true;
+        expect(await authorizer.canPerform(ACTION_2, user, WHERE_2)).to.be.false;
+      });
```
nventuro, line 76: This is a `scheduleGrantPermission` test, i'd not have it here

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
```diff
@@ -0,0 +1,802 @@
...
+    });
+
+    context('when there is a no delay set to grant permissions', () => {
+      it('reverts if the sender is not the granter', async () => {
+        await expect(authorizer.grantPermission(ACTION_1, user, WHERE_1, { from: granter })).to.be.revertedWith(
+          'SENDER_IS_NOT_GRANTER'
+        );
+      });
```
nventuro, line 84: This is a bit odd because you're actually using `granter` here, but it's not yet been made a granter. I'd remove this test and check granter status in other places

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
```diff
@@ -0,0 +1,802 @@
...
+      context('when the sender is root', () => {
+        itGrantsPermissionCorrectly(() => root);
+      });
+
+      context('when the sender is granter', () => {
+        sharedBeforeEach('makes a granter', async () => {
+          await authorizer.addGranter(ACTION_1, granter, EVERYWHERE, { from: root });
+          await authorizer.addGranter(ACTION_2, granter, EVERYWHERE, { from: root });
+        });
+        itGrantsPermissionCorrectly(() => granter);
```
nventuro, line 277: We could add two `it`s here where `granter` is unable to grant permission over a third action. 

It would also be good to have tests for a non global granter, who is unable to grant permissions over _any_ address. We could maybe make `granter` a granter over `WHERE_1` and `WHERE_2`, so that we don't need to change `itGrantsPermissionCorrectly`, and then simply add a few `it`s.

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
```diff
@@ -0,0 +1,802 @@
...
+        await advanceTime(delay);
+        const receipt = await authorizer.execute(id);
+        expectEvent.inReceipt(await receipt.wait(), 'ExecutionExecuted', { scheduledExecutionId: id });
+      });
+
+      it('grants the permission when executed', async () => {
+        const id = await authorizer.scheduleGrantPermission(ACTION_1, user, WHERE_1, [], { from: getSender() });
+
+        await advanceTime(delay);
+        await authorizer.execute(id);
```
nventuro, line 354: You could also test the `PermissionGranted` event here, for completeness

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
```diff
@@ -0,0 +1,802 @@
...
+    context('when the sender is root', () => {
+      itScheduleGrantPermissionCorrectly(() => root);
+    });
+
+    context('when the sender is granter', () => {
+      sharedBeforeEach('makes a granter', async () => {
+        await authorizer.addGranter(ACTION_1, granter, EVERYWHERE, { from: root });
+        await authorizer.addGranter(ACTION_2, granter, EVERYWHERE, { from: root });
+      });
+      itScheduleGrantPermissionCorrectly(() => granter);
```
nventuro, line 391: Similar to my `grantPermission` suggestion, I'd try to improve handling of scenarios where the action fails because the caller does not have the right permissions.

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
```diff
@@ -0,0 +1,802 @@
...
+        await authorizer.execute(id);
+
+        expect(await authorizer.hasPermission(ACTION_1, user, WHERE_1)).to.be.equal(true);
+      });
+
+      it('does not grant any other permissions when executed', async () => {
+        const id = await authorizer.scheduleGrantPermission(ACTION_1, user, WHERE_1, [], { from: getSender() });
+
+        await advanceTime(delay);
+        await authorizer.execute(id);
```
nventuro, line 363: I really like these simple tests for the `cancel` and `execute` flow, without going crazy (e.g. not testing what happens if you execute a cancelled action, since we already test that in `execute` tests), but showing a reasonable happy path for all features.

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
```diff
@@ -0,0 +1,802 @@
...
+
+        // should not be able to execute before delay
+        await expect(authorizer.execute(id, { from: root })).to.be.revertedWith('ACTION_NOT_YET_EXECUTABLE');
+
+        await advanceTime(delay);
+        await authorizer.execute(id, { from: root });
+
+        expect(await authorizer.canPerform(ACTION_1, user, WHERE_1)).to.be.false;
+        expect(await authorizer.canPerform(ACTION_2, user, WHERE_2)).to.be.false;
+      });
```
nventuro, line 422: Similar to `grant`, this is a `scheduleRevoke` test and I'd remove it.

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
```diff
@@ -0,0 +1,802 @@
...
+          'SENDER_IS_NOT_REVOKER'
+        );
+      });
+
+      function itRevokesPermissionCorrectly(getSender: () => SignerWithAddress) {
+        context('when the user does not have the permission', () => {
+          it('ignores the request and cannot perform the requested action everywhere', async () => {
+            await expect(authorizer.revokePermission(ACTION_1, user, WHERE_1, { from: getSender() })).not.to.be
+              .reverted;
+
```
nventuro, line 437: How do you feel about this? Should we revert instead?

nventuro: I don't think it adds any value, can make things confusing, enables silent user error, and makes tests more complex.

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
```diff
@@ -0,0 +1,802 @@
...
+        itRevokesPermissionCorrectly(() => root);
+      });
+
+      context('when the sender is revoker', () => {
+        sharedBeforeEach('makes a revoker', async () => {
+          await authorizer.addRevoker(revoker, EVERYWHERE, { from: root });
+        });
+        itRevokesPermissionCorrectly(() => revoker);
+      });
+    });
```
nventuro, line 574: Same suggestions regarding permission checks for this and scheduleRevoke as I submitted for grant and scheduleGrant (since they're all similar)

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
```diff
@@ -0,0 +1,802 @@
...
+  });
+
+  describe('renouncePermission', () => {
+    context('when the sender does not have the permission', () => {
+      context('when renouncing the permission for a specific contract', () => {
+        it('ignores the request and still cannot perform the requested action everywhere', async () => {
+          await expect(authorizer.renouncePermission(ACTION_1, WHERE_1, { from: user })).not.to.be.reverted;
+
+          expect(await authorizer.canPerform(ACTION_1, user, EVERYWHERE)).to.be.false;
+        });
```
nventuro, line 701: I'd also consider reverting here

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
```diff
@@ -0,0 +1,802 @@
...
+        it('ignores the request and still cannot perform the requested action in any specific contract', async () => {
+          await expect(authorizer.renouncePermissionGlobally(ACTION_1, { from: user })).not.to.be.reverted;
+
+          expect(await authorizer.canPerform(ACTION_1, user, NOT_WHERE)).to.be.false;
+          expect(await authorizer.canPerform(ACTION_1, user, WHERE_2)).to.be.false;
+        });
+      });
+    });
+
+    context('when the user has the permission for a specific contract', () => {
```
nventuro, line 727: I'd add a test where `revoke` has a delay (so we can only do `scheduleRevoke`), but `renounce` still works immediately, with no delay.

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
```diff
@@ -0,0 +1,802 @@
...
+    });
+
+    context('when there is a no delay set to grant permissions', () => {
+      it('reverts if the sender is not the granter', async () => {
+        await expect(authorizer.grantPermission(ACTION_1, user, WHERE_1, { from: granter })).to.be.revertedWith(
+          'SENDER_IS_NOT_GRANTER'
+        );
+      });
+
+      function itGrantsPermissionCorrectly(getSender: () => SignerWithAddress) {
```
nventuro, line 86: This `getSender` thing is a very nice way to handle the async parameter issue we face elsewhere :ok_hand:

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
```diff
@@ -0,0 +1,431 @@
...
+      });
+    });
+
+    it('cannot be executed twice', async () => {
+      const id = await schedule(authenticatedContract);
+      await advanceTime(delay);
+
+      await authorizer.execute(id);
+      await expect(authorizer.execute(id)).to.be.revertedWith('ACTION_ALREADY_EXECUTED');
+    });
```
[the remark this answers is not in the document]
ylv-io: They certainly are but we tend to have overlap in tests suites of close functions, for example, schedule/execute, scheduleGrantPermission/grantPermission. I am neutral on this. If you want to remove this let me know.

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
```diff
@@ -0,0 +1,431 @@
...
+      });
+    });
+
+    it('cannot be executed twice', async () => {
+      const id = await schedule(authenticatedContract, [executor]);
+      await advanceTime(delay);
+
+      await authorizer.execute(id, { from: executor });
+      await expect(authorizer.execute(id, { from: executor })).to.be.revertedWith('ACTION_ALREADY_EXECUTED');
+    });
```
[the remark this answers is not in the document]
ylv-io: Same as above. Do you want to remove these tests?

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
ylv-io: This would require another contract which calls authorizer, correct?

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
ylv-io: I agree but let's make it a separate PR. This one is big enough.

ylv-io:
Thank you for the comments! I've addressed most comments. Going from commit to commit should be a nice experience. I would implement reverts for `revokePermission/renouncePermission` as a separate PR. I also need to do address extra tests for grantPermission, revokePermission, scheduleGrantPermission, and scheduleRevokePermission.

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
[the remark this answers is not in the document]
nventuro: I'd remove these because a) we have the same exact test for `execute`, so they don't really add anything, and b) reading these caused me some confusion as I started wondering what we should test in each `describe` block, thinking about other missing `execute` tests, going to that section and double-checking coverage, etc., which made it harder for me to be confident about correctness of the test suite.

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
```diff
@@ -253,6 +253,12 @@ describe('TimelockAuthorizer execute', () => {
       const where = await authorizer.instance.getTimelockExecutionHelper();
       await expect(schedule(where)).to.be.revertedWith('ATTEMPTING_EXECUTION_HELPER_REENTRANCY');
     });
+
+    it('reverts if data is less than 4 bytes', async () => {
+      await expect(authorizer.schedule(authenticatedContract.address, '0x00', [], { from: user })).to.be.revertedWith(
+        'SENDER_DOES_NOT_HAVE_PERMISSION'
```
nventuro, line 274: Hm not a fan of this error message. Let's change `_decodeSelector` so that it reverts if it cannot get a selector.

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
```diff
@@ -165,6 +165,13 @@ describe('TimelockAuthorizer execute', () => {
       }
     });
 
+    it('emits ExecutionScheduled event', async () => {
+      const receipt = await authorizer.instance.connect(user).schedule(authenticatedContract.address, data, []);
+
+      // _scheduledExecutions doesn't have a getter so we can't fetch it before the call
+      expectEvent.inReceipt(await receipt.wait(), 'ExecutionScheduled', { scheduledExecutionId: 2 });
```
nventuro, line 174: I don't understand this comment, what do you meant?

Also, let's test the `actionId` param of the event.

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
```diff
@@ -0,0 +1,431 @@
...
+
+      const scheduledExecution = await authorizer.getScheduledExecution(id);
+      expect(scheduledExecution.executed).to.be.false;
+      expect(scheduledExecution.data).to.be.equal(data);
+      expect(scheduledExecution.where).to.be.equal(authenticatedContract.address);
+      expect(scheduledExecution.protected).to.be.true;
+      expect(scheduledExecution.executableAt).to.be.at.almostEqual((await currentTimestamp()).add(delay));
+    });
+
+    it('emits ExecutorAdded events', async () => {
```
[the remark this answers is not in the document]
nventuro: Agreed, even if the helper does it today indirectly, let's have an explicit test for it.

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
```diff
@@ -370,6 +370,17 @@ describe('TimelockAuthorizer execute', () => {
...
 
+    it('cannot execute another scheculed action within a scheduled action', async () => {
+      const id = await schedule();
+
+      const data = authorizer.interface.encodeFunctionData('execute', [id]);
+      const nextId = await authorizer.schedule(authorizer, data, [], { from: user });
+
+      await advanceTime(delay);
+
+      await expect(authorizer.execute(nextId, { from: executor })).to.be.revertedWith('CANNOT_SCHEDULE_AUTHORIZER_ACTIONS');
```
nventuro: Note that this revert is not what you're trying to test: you're attempting to schedule an authorizer call directly (which is not allowed). We actually already have a test for this.

But I think we should maybe remove the reentrancy guard check altogether.

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
nventuro: Yes, it would. But now I think maybe we should remove the guard.

pkg/vault/test/authorizer/TimelockAuthorizerRoot.test.ts
```diff
@@ -74,11 +74,9 @@ describe('TimelockAuthorizer root', () => {
 
           it('emits an event', async () => {
             const newPendingRoot: SignerWithAddress = getNewPendingRoot();
-            const id = await authorizer.scheduleRootChange(newPendingRoot, [], { from: root });
+            const receipt = await authorizer.instance.connect(root).scheduleRootChange(newPendingRoot.address, []);
 
-            await advanceTime(ROOT_CHANGE_DELAY);
-            const receipt = await authorizer.execute(id);
-            expectEvent.inReceipt(await receipt.wait(), 'PendingRootSet', { pendingRoot: newPendingRoot.address });
+            expectEvent.inReceipt(await receipt.wait(), 'RootChangeScheduled', { newRoot: newPendingRoot.address });
```
nventuro: This new test is correct, but let's also keep the old test (i.e. that `PendingRootSet` is emitted correctly when executing). Sorry if my previous message was not clear.

nventuro (commented):
Thank you for carefully addressing each comment in a separate commit! It made reviewing the changes very simple and pleasant :)

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
nventuro: Created https://github.com/balancer/balancer-v2-monorepo/issues/2379

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
nventuro: Created https://github.com/balancer/balancer-v2-monorepo/issues/2379

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
nventuro: Created https://github.com/balancer/balancer-v2-monorepo/issues/2380

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
```diff
@@ -312,6 +312,17 @@ describe('TimelockAuthorizer execute', () => {
...
     });
 
+    it('execute returns a correct result', async () => {
+      const id = await schedule();
+      await advanceTime(delay);
+      const ret = await authorizer.instance.connect(executor).callStatic.execute(id);
+
+      // we have to slice first 4 selector bytes from the input data to get the return data
+      expect(ret).to.be.eq(
+        '0x' + authenticatedContract.interface.encodeFunctionData('protectedFunction', [functionData]).slice(10)
```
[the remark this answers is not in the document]
ylv-io: I don't think it is the case. ABI encoding bytes seems to add zeros padding.

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
```diff
@@ -254,6 +254,12 @@ describe('TimelockAuthorizer execute', () => {
       await expect(schedule(where)).to.be.revertedWith('ATTEMPTING_EXECUTION_HELPER_REENTRANCY');
     });
 
+    it('reverts if schedule for EOA', async () => {
+      // we do not specify reason here because call to an EOA results in the following error:
+      // Transaction reverted without a reason
+      await expect(authorizer.schedule(other.address, '0x00', [], { from: user })).to.be.reverted;
```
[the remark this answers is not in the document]
nventuro: Related to https://github.com/balancer/balancer-v2-monorepo/issues/2380

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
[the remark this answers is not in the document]
nventuro: Well, yes, the abi-encoded bytes :stuck_out_tongue:. I think you can do `defaultAbiCoder.encode([bytes], [functionData])`

nventuro (approved):
The last round of changes is brilliant :) Thanks a lot for your work!

PR #2382, opened 2023-03-23
Add new expected collisions to .gitignore

nventuro:
Follow-up to https://github.com/balancer/balancer-v2-monorepo/pull/2359. Keeping the old ignore rule in case someone still has these files in their filesystem.


EndymionJkb (approved):
Thanks! Didn't think of that

PR #2385, opened 2023-03-23
Remove reentrancy guard on TimelockAuthorizer

### LINKED ISSUE #2381: Remove the reentrancy guard checks on the timelock auth
* Note that the TimelockAuthorizer doesn't make use of reentrancy guards on the majority of external functions.
* The only function which makes an external non-view call (and so could initate a reentrancy attack) is execute
* Which executes a scheduled execution and so this is the only protected function.

Related https://github.com/balancer/balancer-v2-monorepo/pull/2365#discussion_r1145169851

ylv-io:
# Description

This PR removes ReentrancyGuard from the TimelockAuthorizer's `execute` function because it is secured by check-effects-interactions pattern. Other functions of TimelockAuthrorizer do not make external calls.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:

~~- [ ] Tests are included for all code paths (Does not apply)~~

## Issue Resolution

Closes #2381

### DIFF
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -22,7 +22,6 @@ import "@balancer-labs/v2-interfaces/contracts/vault/IAuthorizer.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/helpers/InputHelpers.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/math/Math.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/Address.sol";
-import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/ReentrancyGuard.sol";
 import "./TimelockExecutionHelper.sol";
 import "./TimelockAuthorizerManagement.sol";
 
@@ -47,9 +46,9 @@ import "./TimelockAuthorizerManagement.sol";
  *   target contract (where). This identifier is called `permissionId` and is computed as
  *   `keccak256(actionId, account, where)`.
  *
- * Note that the TimelockAuthorizer doesn't make use of reentrancy guards on the majority of external functions.
+ * Note that the TimelockAuthorizer doesn't use reentrancy guard on its external functions.
  * The only function which makes an external non-view call (and so could initate a reentrancy attack) is `execute`
- * which executes a scheduled execution and so this is the only protected function.
+ * which executes a scheduled execution and protected by Checks-Effects-Interactions pattern.
  * In fact a number of the TimelockAuthorizer's functions may only be called through a scheduled execution so reentrancy
  * is necessary in order to be able to call these.
  */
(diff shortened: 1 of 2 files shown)


pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -47,9 +46,9 @@ import "./TimelockAuthorizerManagement.sol";
  *   target contract (where). This identifier is called `permissionId` and is computed as
  *   `keccak256(actionId, account, where)`.
  *
- * Note that the TimelockAuthorizer doesn't make use of reentrancy guards on the majority of external functions.
+ * Note that the TimelockAuthorizer doesn't use reentrancy guard on its external functions.
  * The only function which makes an external non-view call (and so could initate a reentrancy attack) is `execute`
- * which executes a scheduled execution and so this is the only protected function.
+ * which executes a scheduled execution and protected by Checks-Effects-Interactions pattern.
```
EndymionJkb, line 51:
```suggestion
 * which executes a scheduled execution, protected by Checks-Effects-Interactions pattern.
```

EndymionJkb (commented):
ITimelockAuthorizer docs (which would also have to change if we did this) say that during the execution of an action, we explicitly forbid executing any further actions: just because that sounds like trouble, and we can't anticipate every possible action.

*This* action is protected by the pattern (technically, we could also emit the event before making the call) - you can't reenter and execute it again, for instance - but I believe the issue is with executing *other* actions.

Maybe the first action changes the state of something *else* (maybe something we don't control, like a gauge), and executing a second action in the middle could do something nefarious. I don't think it hurts anything to leave this protection in place.

pkg/vault/contracts/authorizer/TimelockAuthorizerManagement.sol
```diff
@@ -290,7 +289,7 @@ abstract contract TimelockAuthorizerManagement is ITimelockAuthorizer, Reentranc
 
         // Note that this is the only place in the entire contract we perform a non-view call to an external contract,
         // i.e. this is the only context in which this contract can be re-entered, and by this point we've already
-        // completed all state transitions.
+        // completed all state transitions. Which satisfies the requirements of the Checks-Effects-Interactions pattern.
```
nventuro, line 292:
```suggestion
        // completed all state transitions (in other words, we follow the Checks-Effects-Interactions pattern).
```

nventuro (approved):

nventuro:
>ITimelockAuthorizer docs (which would also have to change if we did this) say that during the execution of an action, we explicitly forbid executing any further actions: just because that sounds like trouble, and we can't anticipate every possible action.
>
>This action is protected by the pattern (technically, we could also emit the event before making the call) - you can't reenter and execute it again, for instance - but I believe the issue is with executing other actions.

That comment refers to scheduled executions, and it remains true: you can't call `execute` while it is being executed (because of a restriction placed by the `TimelockExecutionHelper`, which was redundant with this check).

PR #2389, opened 2023-03-24
Revert on bad schedule data

### LINKED ISSUE #2380: [timelock] Revert on bad schedule data
Currently `_decodeSelector` returns `bytes(0)` if the data contains no selector. This causes two issues:
 - we get a bad error message ('not allowed') since we query for a bad action id
 - if any function where to end up having the `0x0000` selector, then someone with that permission could also schedule calls with fewer than 3 bytes, since we'd wrongly check for that selector

We should revert when failing to get a selector.

ylv-io:
# Description

This PR adds additional validation for the `data` parameter of the `schedule` method.  If `data` is less than 4 bytes we revert with 'DATA_TOO_SHORT' too short error.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #2380


nventuro:
Fantastic, thanks!

PR #2394, opened 2023-03-30
Deployment Preparation: TimelockAuthorizer

ylv-io:
# Deployment preparation: TimelockAuthorizer

## Deployment preparation tasks checklist:

- [x] Deployments `README` is updated with the new task name and link
- [x] Task `README` has a description and links to the artifacts
- [x] `build-info` is updated with artifacts compiled at the specified commit
- [x] Artifacts are generated from `build-info`
- [x] New task has a working fork test 

## Deprecating old tasks checklist: 

- [x] Old task was moved to `deprecated/` directory
- [x] Deployments `README` is updated: old task is moved to the 'deprecated' section, and links are updated with the `deprecated/` prefix
- [x] Old task imports in `index.ts`, `input.ts` and fork tests are updated
- [x] Old task `README` is updated with a warning sign, a link to the replacement task and a short description 

## Code checklist:

- [x] Complex code has been commented in tests and scripts
- [x] Fork test block number and test cases are appropriate for this deployment


nventuro (commented):
There's multiple build info files uploaded here. We get multiple ones because hardhat sometimes splits compilation jobs depending on a number of factors, including compiler settings. In this case, at least one of these originates from the fact that the Vault has dedicated settings (with a lower `runs` value).

You should only upload a single build info, corresponding to the contract you mean to deploy. In this case, that's the `TimelockAuthorizerMigrator`, You can find which build info corresponds to it by looking at the `.dbg` file:

```
pkg/governance-scripts$ cat artifacts/contracts/TimelockAuthorizerMigrator.sol/TimelockAuthorizerMigrator.dbg.json 
{
  "_format": "hh-sol-dbg-1",
  "buildInfo": "../../build-info/60244e7354f59d8bf4ae80e91fa51f18.json"
}
```

nventuro:
As discussed offline, we want to have basic fork tests (checking that roles are granted etc) done for each network, so that we know the config is correct (i.e. being tested).

For now let's only commit the goerli config and tests so we can deploy there. We'll do other networks when the time comes.

pkg/deployments/tasks/20230403-timelock-authorizer/readme.md
```diff
@@ -0,0 +1,11 @@
+# 2023-04-03 - Timelock Authorizer
+
+Deployment of the `TimelockAuthorizer` in order to replace the basic Authorizer deployed with the Vault.
+This Authorizer implementation allows defining a delay per action identifier. Users can only execute functions directly when there is no delay. Otherwise, they're granted permission to schedule an action, which can then be executed by the Authorizer after the delay. It also introduces Granters and Revokers, to allow controlled delegation of permission handling to third parties, which eases the burden on governance and allows the system to scale. For instance, a smaller multisig might be designated as a Granter for certain veBAL-related functions for new pools.
+
+## Useful Files
+
+- [Ethereum mainnet addresses](./output/mainnet.json)
```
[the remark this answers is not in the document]
EndymionJkb: Also, it won't pass the link check. Per the docs, the output addresses get added in the Deployment PR

ylv-io:
There are two fork tests Mainnet and Goerli. I've also deprecated the old task.

pkg/deployments/tasks/20230403-timelock-authorizer/test/task.fork.ts
```diff
@@ -0,0 +1,153 @@
...
+    });
+
+    it('allows minting after the migration', async () => {
+      const balancerMinterTask = new Task('20220325-gauge-controller', TaskMode.READ_ONLY, getForkedNetwork(hre));
+      const balancerMinter = await balancerMinterTask.deployedInstance('BalancerMinter');
+
+      const balancerTokenAdminTask = new Task(
+        '20220325-balancer-token-admin',
+        TaskMode.READ_ONLY,
+        getForkedNetwork(hre)
```
nventuro, line 108: Does this work on goerli? I didn't remember we had set it up.

At any rate, it'll definitely fail on the L2s (and also is irrelevant). Can we extract it out of the generic tests?

pkg/deployments/tasks/20230403-timelock-authorizer/input/goerli.ts
```diff
@@ -0,0 +1,619 @@
...
+const SmartWalletChecker = new Task('20220420-smart-wallet-checker', TaskMode.READ_ONLY, 'goerli');
+
+const DAO_MULTISIG = '0x171C0fF5943CE5f133130436A29bF61E26516003';
+
+// Start: block that contains the transaction that deployed the `TimelockAuthorizer`.
+// etherscan.io/tx/0x20eb23f4393fd592240ec788f44fb9658cc6ef487b88398e9b76c910294c4eae
+// End: close to the current block at the time the `TimelockAuthorizerMigrator` is deployed.
+// It is expected that no roles were granted to the old authorizer after it.
+export const TRANSITION_START_BLOCK = 8745337;
+export const TRANSITION_END_BLOCK = 8745437;
```
nventuro, line 22: I'm actually not sure we need this now? We used to require the transition fix because the migration took ~1 month, but now that the migration only takes a couple days we don't need to have a transition period. We can just use the permissions hardcoded here and that's it.

pkg/deployments/tasks/20230403-timelock-authorizer/input.ts
```diff
@@ -0,0 +1,79 @@
...
+  Revokers: RoleData[];
+  ExecuteDelays: DelayData[];
+  GrantDelays: DelayData[];
+  TRANSITION_START_BLOCK: number;
+  TRANSITION_END_BLOCK: number;
+};
+
+export async function getOnChainRoles(roles: RoleData[], start: number, end: number): Promise<RoleData[]> {
+  const OldAuthorizerTask = new Task('20210418-authorizer', TaskMode.READ_ONLY);
+  const oldAuthorizerAddress = OldAuthorizerTask.output({ network: getForkedNetwork(hre) }).Authorizer;
```
nventuro, line 26: While this is correct for fork tests (because we want the forked network, e.g. mainnet, and not the real network, e.g. hardhat), it will fail in a real deployment (because we're not forking).

pkg/deployments/tasks/20230403-timelock-authorizer/input/goerli.ts
```diff
@@ -0,0 +1,619 @@
...
+      id: '0x7b8a1d293670124924a0f532213753b89db10bde737249d4540e9a03657d1aff',
+    },
+    txHash: '0x79102fe6eb2ddddbd19801b99e1438635d14196806420150115ecf6f4e9ccb22',
+  },
+];
+
+export const GrantDelays: DelayData[] = [
+  {
+    actionId: BalancerTokenAdmin.actionId('BalancerTokenAdmin', 'mint(address,uint256)'),
+    newDelay: 30 * DAY,
```
nventuro, line 554: Let's have shorter delays for goerli, since there's nothing at risk here. We still want to have delays for testing, but they don't need to be as long. The 30/14/7 day tiers could be reduced to 2/1/0.25 days

pkg/deployments/tasks/20230403-timelock-authorizer/input/goerli.ts
```diff
@@ -0,0 +1,619 @@
...
+export const TRANSITION_START_BLOCK = 8745337;
+export const TRANSITION_END_BLOCK = 8745437;
+
+export const Root = DAO_MULTISIG;
+
+// Permission fetched from TheGraph
+// thegraph.com/hosted-service/subgraph/balancer-labs/balancer-authorizer-goerli
+const theGraphRoles = [
+  {
+    id: '0x014d6b43e6c60cd9f1183053447be80d6c6ca5d245f2fd594f23289f1a847769',
```
nventuro, line 30: What is this id?

pkg/deployments/tasks/20230403-timelock-authorizer/input/goerli.ts
```diff
@@ -0,0 +1,619 @@
...
+    account: '0x7b9b6f094dc2bd1c12024b0d9cc63d6993be1888',
+    action: {
+      id: '0xc149e88b59429ded7f601ab52ecd62331cac006ae07c16543439ed138dcb8d34',
```
nventuro, line 33: It'd be great if we could have some reference for what these are, at least a task and function name. We'll defnitely want to do such a thing for the L2 ones.

It'd also be good to set a `target` for these, which would be the contract itself in most cases (for singeltons), and everywhere for pools.

While this is not strictly required for goerli, we'll want to stablish a good foundation to do this in the other networks.

nventuro (commented):
Looking great! :raised_hands: 

I've not yet reviewed the mainnet permissions since we're not yet doing that. Perhaps we should actually remove that input file for now until the time comes to do the migration?

pkg/deployments/tasks/20230403-timelock-authorizer/input/goerli.ts
ylv-io: It is the unique ids TheGraph generates for each permission. Nothing to do with our permissions.

nventuro: Ah I see, it's the unique id we assign to each entry. I'd consider renaming these to e.g. `entryId`. They're set here: https://github.com/balancer/authorizer-subgraph/blob/b1215c75e51a6db120276a5020ad797e91378c30/src/utils/actions.ts#L5

pkg/deployments/tasks/20230403-timelock-authorizer/input/goerli.ts
ylv-io: We don't. You are right. That was part of the design a copied.

nventuro (commented):
Love the type alias! Could you use `get-action-id-info` to populate the input file for goerli with the task contract and function names? Once that's ready we can merge and deploy

ylv-io:
I've written a Hardhat task which fetches permission from TheGraph API matches them against actions-ids.json file and builds the output roles json file to use for tests and migration.
```
hh get-action-ids-info --network goerli > ./tasks/20230403-timelock-authorizer/input/goerli.json
```
It can be used for any network given the correct TheGraphAPI url.

pkg/deployments/README.md
```diff
@@ -113,7 +112,12 @@ Returns an object with all contracts from a deployment and their addresses.
 | L2 Voting Escrow Delegation Proxy                      | [`20230316-l2-ve-delegation-proxy`](./tasks/20230316-l2-ve-delegation-proxy)                         |
 | Weighted Pool V4                                       | [`20230320-weighted-pool-v4`](./tasks/20230320-weighted-pool-v4)                                     |
 | Composable Stable Pools V4                             | [`20230320-composable-stable-pool-v4`](./tasks/20230320-composable-stable-pool-v4)                   |
+<<<<<<< HEAD
+| Composable Stable Pools V4                             | [`20230320-composable-stable-pool-v4`](./tasks/20230320-composable-stable-pool-v4)                   |
+| Timelock Authorizer, governance contract               | [`20230403-timelock-authorizer`](./tasks/20230403-timelock-authorizer)                               |
+=======
 | Voting Escrow Remapper                                 | [`20230403-vebal-remapper`](./tasks/20230403-vebal-remapper)                                         |
+>>>>>>> origin
```
EndymionJkb, line 120:
```suggestion
| Timelock Authorizer, governance contract               | [`20230403-timelock-authorizer`](./tasks/20230403-timelock-authorizer)                               |
| Voting Escrow Remapper                                 | [`20230403-vebal-remapper`](./tasks/20230403-vebal-remapper)                                         |
```

EndymionJkb (commented):
Can't add much to Nico; just noted the merge conflict in the README

pkg/deployments/README.md
ylv-io: Wow. No idea how I missed that. Thanks!

PR #2395, opened 2023-03-30
base goerli testnet deployment

mikemcdonald:
# Deployment: Base Goerli Testnet

Contracts aren't verified since Base Goerli isn't a recognized network within hardhat. We would need to rework the verification scripts to support or upgrade to a hardhat version once it's added.

Let me know if there's a preference on direction, just wanted to submit this so the deployment addresses are available somewhere.


nventuro:
Do you mean etherscan-like verification, or our own deployment checks? I believe those should work, since we don't rely on anything native to hardhat for them. We'll need to edit the CI scripts and add the RPC endpoint as a repository secret.

mikemcdonald:
Right, etherscan verification won't work with the existing hardhat version / verification setup.

jubeira:
@mikemcdonald would you like to get this ported to [`balancer-deployments`](https://github.com/balancer/balancer-deployments)? I can get this in shape if needed, including the contract verification.

EndymionJkb:
Base mainnet deployed.

PR #2403, opened 2023-03-31
Add fork test with gauge liq limit update

nventuro:
This is a proof of concept of the second mechanism described in https://github.com/balancer/balancer-v2-monorepo/issues/2399, where we use `manageUserBalance` to force a liquidity limit update in batch for user gauges without requiring any token approvals or prior setup other than relayer approval, using BatchRelayerV5.


pkg/deployments/tasks/20230314-batch-relayer-v5/test/test.fork.ts
```diff
@@ -145,4 +152,39 @@ describeForkTest('BatchRelayerLibrary', 'mainnet', 15485000, function () {
...
+    const tx = await relayer.connect(sender).multicall([
+      library.interface.encodeFunctionData('manageUserBalance', [
+        [RETH_STABLE_GAUGE, ETH_STETH_GAUGE].map((gauge) => ({
+          kind: TRANSFER_EXTERNAL_USER_BALANCE_OP_KIND,
+          asset: gauge,
+          amount: 1, // The amount must be non-zero to trigger the token trasnfer and liquidity limit update
+          sender: sender.address,
+          recipient: sender.address,
+        })),
+        0,
```
jubeira, line 166:
```suggestion
        0, // value: 0 (no need to transfer ETH)
```

pkg/deployments/tasks/20230314-batch-relayer-v5/test/test.fork.ts
```diff
@@ -145,4 +152,39 @@ describeForkTest('BatchRelayerLibrary', 'mainnet', 15485000, function () {
...
+    );
+
+    expectEvent.inIndirectReceipt(
+      await tx.wait(),
+      gaugeInterface,
+      'UpdateLiquidityLimit',
+      { user: sender.address },
+      ETH_STETH_GAUGE
+    );
+  });
```
jubeira, line 189: Fair enough. I thought about checking the balance and / or the working supply before the update, but on second thought  it doesn't really add much value in this context.

jubeira (approved):
This makes sense, pretty smart :+1:

PR #2405, opened 2023-04-03
Gauge Working Balance Helper

EndymionJkb:
# Description

Get the current and projected `working_balance` (effective veBAL boosted balance) for a user on a gauge, then expose a method that computes the ratio of the working or projected balances to the supply (a measure of the actual gauge share, and how it would change if checkpointed).

The `working_balance` can range between 40% and 100% of the nominal user balance on a gauge, depending on the veBAL "voting power" of the user. This value generally decays over time, but will increase with additional veBAL locking.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/liquidity-mining/contracts/GaugeWorkingBalanceHelper.sol
```diff
@@ -0,0 +1,153 @@
...
+        } else {
+            uint256 projectedWorkingSupply = currentWorkingSupply.add(
+                projectedWorkingBalance.sub(currentWorkingBalance)
+            );
+
+            return (
+                currentWorkingBalance.divDown(currentWorkingSupply),
+                projectedWorkingBalance.divDown(projectedWorkingSupply)
+            );
+        }
```
nventuro, line 151:
```suggestion
        uint256 projectedWorkingSupply = currentWorkingSupply.add(
            projectedWorkingBalance.sub(currentWorkingBalance)
        );

        return (
            currentWorkingBalance.divDown(currentWorkingSupply),
            projectedWorkingBalance.divDown(projectedWorkingSupply)
        );
```

The otpimization is not incorrect, but we hardly care about this small gas saving, specially considering this will be queried offchain. I'd rather keep the simpler impl that always works.

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
```diff
@@ -0,0 +1,300 @@
...
+  function itStoresParameters(readTotalSupplyFromVE: boolean) {
+    it('stores the veDelegationProxy', async () => {
+      expect(await workingBalanceHelper.getVotingEscrowDelegationProxy()).to.equal(veDelegationProxy.address);
+    });
+
+    it('stores the votingEscrow', async () => {
+      expect(await workingBalanceHelper.getVotingEscrow()).to.equal(votingEscrow.address);
+    });
+
+    it('indicates the network', async () => {
```
nventuro, line 80:
```suggestion
    it('indicates where to read supply from', async () => {
```

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
```diff
@@ -0,0 +1,300 @@
...
+    function projectedIsMaxedOut() {
+      it(`projected ratio should be greater than current by the maximum ratio (${MAX_BALANCE_RATIO})`, async () => {
+        const [currentWorkingBalance, projectedWorkingBalance] = await workingBalanceHelper.getWorkingBalances(
+          gauge.address,
+          user.address
+        );
+
+        // Ensure we have equal balances (that are non-zero)
+        expect(fromFp(projectedWorkingBalance) / fromFp(currentWorkingBalance)).to.eq(MAX_BALANCE_RATIO);
+        expect(projectedWorkingBalance).to.be.gt(0);
```
nventuro, line 144:
```suggestion
        expect(fromFp(projectedWorkingBalance) / fromFp(currentWorkingBalance)).to.eq(MAX_BALANCE_RATIO);

```

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
```diff
@@ -0,0 +1,300 @@
...
+      it('projected balance should be close to or less than current', async () => {
+        const [currentWorkingBalance, projectedWorkingBalance] = await workingBalanceHelper.getWorkingBalances(
+          gauge.address,
+          user.address
+        );
+
+        // Ensure we have equal balances (that are non-zero)
+        expect(projectedWorkingBalance).to.be.almostEqual(currentWorkingBalance);
+        expect(projectedWorkingBalance).to.be.gt(0);
+        expect(projectedWorkingBalance).to.be.lte(currentWorkingBalance);
```
nventuro, line 173:
```suggestion
        expect(projectedWorkingBalance).to.be.almostEqual(currentWorkingBalance);
        expect(projectedWorkingBalance).to.be.gt(0);
        expect(projectedWorkingBalance).to.be.lte(currentWorkingBalance);
```

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
```diff
@@ -0,0 +1,300 @@
...
+      it('current and projected ratios should be 1', async () => {
+        const [current, projected] = await workingBalanceHelper.getWorkingBalanceToSupplyRatios(
+          gauge.address,
+          user.address
+        );
+
+        // Ensure we have equal ratios (that equal 1)
+        expect(projected).to.eq(current);
+        expect(projected).to.eq(FP_ONE);
+      });
```
nventuro, line 132: This is the only ratio test we have - it'd be great to also test how the ratios change in a couple scenarios.

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
```diff
@@ -0,0 +1,300 @@
...
+    describe('with 1% veBAL', () => {
+      sharedBeforeEach('deposit', async () => {
+        // Another person stakes 100x as much
+        await depositIntoGauge(user, stakeAmount);
+        await depositIntoGauge(other, stakeAmount);
+
+        await createLockForUser(user, bptAmount, 365 * DAY);
+        await createLockForUser(other, bptAmount.mul(100), 365 * DAY);
+      });
+
```
nventuro, line 225: These are very good, I like them a lot. I'd add one small thing: in both of these cases, have a `context` in which we do a checkpoint, advance time by some amount (say half of the lock period), and then check that the projected values (both balance and ratio) are stricly lower than the current ones (because of the veBAL decay).

nventuro (commented):
Fantastic work @EndymionJkb, I like how this turned out quite a bit, and the tests also look very good.

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
```diff
@@ -0,0 +1,312 @@
...
+        );
+
+        // Ensure we have equal ratios (that equal 1)
+        expect(projected).to.eq(current);
+        expect(projected).to.eq(FP_ONE);
+      });
+    }
+
+    function projectedRatioIncreases() {
+      it('current and projected ratios should be 1', async () => {
```
nventuro, line 136:
```suggestion
      it('projected ratio should be larger than current', async () => {
```

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
```diff
@@ -0,0 +1,339 @@
...
+      deployment: 'test-deployment',
+    });
+
+    let pseudoMinter: Contract;
+
+    sharedBeforeEach('deploy helper', async () => {
+      await deployHelper(false);
+    });
+
+    itStoresParameters(false);
```
jubeira, line 314:
```suggestion
    const version = JSON.stringify({
      name: 'ChildChainGauge',
      version: '0',
      deployment: 'test-deployment',
    });
    const readTotalSupplyFromVe = false;

    let pseudoMinter: Contract;

    sharedBeforeEach('deploy helper', async () => {
      await deployHelper(readTotalSupplyFromVe);
    });

    itStoresParameters(readTotalSupplyFromVe);
```

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    function projectedRatioIsMaxedOut() {
+      it(`projected ratio should be greater than current by the maximum ratio (${MAX_BALANCE_RATIO})`, async () => {
+        const [currentWorkingBalance, projectedWorkingBalance] = await workingBalanceHelper.getWorkingBalances(
+          gauge.address,
+          user.address
+        );
+
+        expect(fromFp(projectedWorkingBalance) / fromFp(currentWorkingBalance)).to.eq(MAX_BALANCE_RATIO);
```
jubeira, line 153: `fromFp` returns a `Decimal`; can we use `/` just like that?

Shouldn't it be something like
```
fromFp(projectedWorkingBalance).div(fromFp(currentWorkingBalance))
```

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
```diff
@@ -0,0 +1,339 @@
...
+    function projectedBalanceIsIncreased(maxPctIncrease: number) {
+      it(`projected balance should be slightly more than current, by less than ${maxPctIncrease}%`, async () => {
+        const [currentWorkingBalance, projectedWorkingBalance] = await workingBalanceHelper.getWorkingBalances(
+          gauge.address,
+          user.address
+        );
+
+        expect(projectedWorkingBalance).to.be.gt(currentWorkingBalance);
+
+        const ratio = fromFp(projectedWorkingBalance) / fromFp(currentWorkingBalance);
```
jubeira, line 166: Same as above

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
```diff
@@ -0,0 +1,339 @@
...
+          await gauge.connect(user).user_checkpoint(user.address);
+        });
+
+        currentShouldAlmostEqualProjected();
+      });
+    }
+
+    function veBALDecaysOverTime() {
+      it('veBAL decays over time', async () => {
+        const [, projecteBalanceBefore] = await workingBalanceHelper.getWorkingBalances(gauge.address, user.address);
```
jubeira, line 197:
```suggestion
        const [, projectedBalanceBefore] = await workingBalanceHelper.getWorkingBalances(gauge.address, user.address);
```

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
```diff
@@ -0,0 +1,339 @@
...
+        await advanceTime(LOCK_PERIOD / 2);
+
+        const [, projectedBalanceAfter] = await workingBalanceHelper.getWorkingBalances(gauge.address, user.address);
+        const [, projectedRatioAfter] = await workingBalanceHelper.getWorkingBalanceToSupplyRatios(
+          gauge.address,
+          user.address
+        );
+
+        // Projections should be uniformly lower
+        expect(projectedBalanceAfter).to.be.lt(projecteBalanceBefore);
```
jubeira, line 215:
```suggestion
        expect(projectedBalanceAfter).to.be.lt(projectedBalanceBefore);
```

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
```diff
@@ -0,0 +1,339 @@
...
+        await createLockForUser(another, bptAmount.mul(100), LOCK_PERIOD);
+        await advanceTime(LOCK_PERIOD / 2);
+
+        const [, projectedBalanceAfter] = await workingBalanceHelper.getWorkingBalances(gauge.address, user.address);
+        const [, projectedRatioAfter] = await workingBalanceHelper.getWorkingBalanceToSupplyRatios(
+          gauge.address,
+          user.address
+        );
+
+        // Projections should be uniformly lower
```
jubeira, line 214: What about projected before vs current after? Shouldn't they be equal after the checkpoint?

That could also be a good indication that the test is correct.

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
```diff
@@ -0,0 +1,339 @@
...
+      expect(await workingBalanceHelper.readsTotalSupplyFromVE()).to.equal(readTotalSupplyFromVE);
+    });
+  }
+
+  function itComputesWorkingBalances() {
+    const TOKENLESS_PRODUCTION = 0.4;
+    const MAX_BALANCE_RATIO = 1 / TOKENLESS_PRODUCTION;
+    const LOCK_PERIOD = 365 * DAY;
+    const stakeAmount = fp(5);
+    const bptAmount = fp(10);
```
jubeira, line 90: I was thinking whether it made sense to made some of these parameters and run the test with different values, but it might be an overkill. WDYT?

jubeira (commented):
Great work! Code and tests are pretty clean, I think this is headed in the right direction.

I've left a bunch of minor comments and opinionated observations, PTAL.

pkg/liquidity-mining/contracts/GaugeWorkingBalanceHelper.sol
```diff
@@ -0,0 +1,144 @@
...
+    function balanceOf(address user) external view returns (uint256);
+
+    function totalSupply() external view returns (uint256);
+
+    // solhint-disable func-name-mixedcase
+    function working_balances(address user) external view returns (uint256);
+
+    function working_supply() external view returns (uint256);
+    // solhint-enable func-name-mixedcase
+}
```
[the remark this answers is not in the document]
EndymionJkb: It's a one-off interface, only used in this helper. Since it can't really be shared or reused, and this is a helper function, I thought it was clearer to have it self-contained.

We do this elsewhere (e.g., in the vendor-specific Linear Pools, when we need to call functions on external non-Balancer contracts)

pkg/liquidity-mining/contracts/GaugeWorkingBalanceHelper.sol
```diff
@@ -0,0 +1,144 @@
...
+        _veDelegationProxy = veDelegationProxy;
+        _veBAL = veDelegationProxy.getVotingEscrow();
+
+        _readTotalSupplyFromVE = readTotalSupplyFromVE;
+    }
+
+    /**
+     * @dev Returns the VotingEscrowDelegationProxy.
+     */
+    function getVotingEscrowDelegationProxy() external view returns (address) {
```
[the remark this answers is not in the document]
EndymionJkb: I could have it return an IveDelegation, I suppose, which is in the exported interface. What do you think? I'd say either leave it as an address, or make it IveDelegation. I wouldn't make it return the one-off one (IVeDelegationProxy)

pkg/liquidity-mining/contracts/GaugeWorkingBalanceHelper.sol
```diff
@@ -0,0 +1,144 @@
...
+        return address(_veDelegationProxy);
+    }
+
+    /**
+     * @dev Returns the VotingEscrow contract associated with the proxy.
+     */
+    function getVotingEscrow() external view returns (address) {
+        return address(_veBAL);
```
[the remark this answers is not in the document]
EndymionJkb: This is a weird case, since it's defined as an ERC-20, but is actually a Curve contract (VotingEscrow). Some of our contracts (e.g., FeeDistributor) return it as IVotingEscrow (from us, not Curve). Others (e.g., VotingEscrowDelegationProxy) return it as an IERC20. I used that pattern.

So I suppose I could make it return an IERC20.

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
```diff
@@ -0,0 +1,339 @@
...
+  }
+
+  context('on L1', () => {
+    const relativeWeightCap = fp(0.1);
+
+    sharedBeforeEach('deploy helper', async () => {
+      await deployHelper(true);
+    });
+
+    itStoresParameters(true);
```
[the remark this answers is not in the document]
EndymionJkb: Yes; also have readTotalSupplyFromProxy = false;

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
EndymionJkb: Well, I had it as readTotalSupplyFromProxy = false;

Slightly confusing either way.

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
EndymionJkb: Maybe there's another way to do it, but div like that seems to truncate it, and the test fails.

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
EndymionJkb: 🦅 👁️

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
EndymionJkb: Added

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
EndymionJkb: I can randomize the lock period (has to be significant, but can't be more than 1 year). The stakeAmounts got flaky (1 wei rounding issues) when I varied them. Might be clearer to leave it as is, than introduce equalWithError?

jubeira: I think introducing `equalWithError` would be fine; we could even make the error explicitly small. I guess the 1 wei error is more than reasonable considering how we are doing the math here (mulDown).

I'd leave it up to you. Either way doesn't sound bad.

pkg/liquidity-mining/contracts/GaugeWorkingBalanceHelper.sol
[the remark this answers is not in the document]
jubeira: I would return the interface; it's a bit more specific than the address and it gives the user some more information about the returned address.

pkg/liquidity-mining/contracts/GaugeWorkingBalanceHelper.sol
```diff
@@ -0,0 +1,144 @@
...
+     * the gauge), and adding a "boost" proportional to that user's share of the total voting power.
+     *
+     * @param gauge - address of a gauge (L1 or L2).
+     * @param user - address of a user.
+     * @return current and projected balances.
+     */
+    function getWorkingBalances(IGauge gauge, address user) public view returns (uint256, uint256) {
+        uint256 gaugeUserBalance = gauge.balanceOf(user);
+        uint256 projectedWorkingBalance = gaugeUserBalance.mulDown(_TOKENLESS_PRODUCTION);
+        IVeDelegationProxy proxy = _veDelegationProxy;
```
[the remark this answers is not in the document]
jubeira: Alright, so the proxy is used in the constructor and it would fail if it was null.

pkg/liquidity-mining/test/GaugeWorkingBalanceHelper.test.ts
jubeira:
```     expect(fromFp(projectedWorkingBalance).div(fromFp(currentWorkingBalance)).toNumber()).to.eq(MAX_BALANCE_RATIO);
```

works for me. It's a bit more verbose, but I get no warnings in the IDE.

jubeira (approved):
Looks great!

I think there's some pending suggestion with Nico, and I've replied to some of the open threads.
Other than that looks good to go :)

PR #2413, opened 2023-04-06
TimelockAuthorizer extension to allow for auditability of scheduled tasks onchain

danielmkm:
# Description

First of all, love this contract! This would be a huge step forward for the authorizer.

The one thing I see missing is the ability to provide better insights into scheduled tasks without having to rely on offchain tooling. With some simple extensions to the `ScheduledExecution` struct and a view function, we could make it possible to spin up a pretty robust UI leveraging only RPC calls.

The implementation is presented as a PoC. **IF** you guys think it's worthwhile, will properly add it to the interface with docs and build out tests.

- Extend the ScheduledExecution struct with fields for better auditability
- Add getNumScheduledExecutions to return the length of the array
- Add getScheduledExecutions to allow for fetching part of the array

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizerManagement.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizerManagement.sol
@@ -201,6 +201,32 @@ abstract contract TimelockAuthorizerManagement is ITimelockAuthorizer {
         return _scheduledExecutions[scheduledExecutionId];
     }
 
+    function getNumScheduledExecutions() external view returns (uint256) {
+        return _scheduledExecutions.length;
+    }
+
+    function getScheduledExecutions(uint256 offset, uint256 limit, bool reverseOrder)
+        external
+        view
+        returns (ITimelockAuthorizer.ScheduledExecution[] memory items)
+    {
+        require(offset < _scheduledExecutions.length, "INVALID_OFFSET");
+      
+        uint256 rest = _scheduledExecutions.length - offset;
+        uint256 size = rest > limit ? limit : rest;
+        items = new ITimelockAuthorizer.ScheduledExecution[](size);
+
+        if (reverseOrder) {
+            for (uint256 i = rest - 1; i >= rest - size; i--) {
+                items[i] = _scheduledExecutions[i];
+            }
+        } else {
+            for (uint256 i = offset; i < offset + size; i++) {
+                items[i] = _scheduledExecutions[i];
+            }
+        }
+    }
+
     /**
      * @inheritdoc ITimelockAuthorizer
      */
@@ -286,6 +312,9 @@ abstract contract TimelockAuthorizerManagement is ITimelockAuthorizer {
         }
 
         scheduledExecution.executed = true;
+        scheduledExecution.executedBy = msg.sender;
+        // solhint-disable-next-line not-rely-on-time
+        scheduledExecution.executedAt = block.timestamp;
 
         // Note that this is the only place in the entire contract we perform a non-view call to an external contract,
         // i.e. this is the only context in which this contract can be re-entered, and by this point we've already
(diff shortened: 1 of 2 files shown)


danielmkm:
It is theoretically true that you can leverage `eth_getLogs` to fetch the events and reconcile the data, but in practice this is quite a bit harder than you'd expect. Most RPCs have limitations on the block range you can fetch in a single query, so finding the events for past actions can be challenging. Alongside of that, reconciling the events is error prone and introduces another opportunity for showing incorrect data.

imo, it is important that scheduled actions are as easy as possible to audit, as this is the safeguard that the timelock authorizer is introducing. With the additions in this PR, building the interface below requires 2 `eth_calls`. One call to `getNumScheduledExecutions` and one call to `getScheduledExecutions`.

Screenshot 2023-04-13 at 08 32 30

pkg/vault/contracts/authorizer/TimelockAuthorizerManagement.sol
```diff
@@ -201,6 +201,30 @@ abstract contract TimelockAuthorizerManagement is ITimelockAuthorizer {
         return _scheduledExecutions[scheduledExecutionId];
     }
 
+    function getNumScheduledExecutions() external view returns (uint256) {
```
nventuro, line 204:
```suggestion
    function getScheduledExecutionsCount() external view returns (uint256) {
```

pkg/vault/contracts/authorizer/TimelockAuthorizerManagement.sol
```diff
@@ -201,6 +201,30 @@ abstract contract TimelockAuthorizerManagement is ITimelockAuthorizer {
...
+
+    function getScheduledExecutions(uint256 offset, uint256 limit, bool reverseOrder)
+        external
+        view
+        returns (ITimelockAuthorizer.ScheduledExecution[] memory items)
+    {
+        require(offset < _scheduledExecutions.length, "INVALID_OFFSET");
+      
+        uint256 rest = _scheduledExecutions.length - offset;
+        uint256 size = rest > limit ? limit : rest;
```
nventuro, line 216:
```suggestion
        uint256 size = Math.min(rest, limit);
```

I find this makes reasoning about `size` easier (i.e. that `size` is never greater than `rest`, and that `limit` may reduce `rest`).

pkg/vault/contracts/authorizer/TimelockAuthorizerManagement.sol
```diff
@@ -286,6 +310,9 @@ abstract contract TimelockAuthorizerManagement is ITimelockAuthorizer {
         }
 
         scheduledExecution.executed = true;
+        scheduledExecution.executedBy = msg.sender;
+        // solhint-disable-next-line not-rely-on-time
```
nventuro, line 314: I'd move this to the top of the file, disable it for the entire file, and add a short comment mentioning that timelocks are time based, and expected to be in the order of days, so any time changes or manipulation in the order of seconds or minutes are irrelevant.

nventuro (commented):
Been thinking about this and I like it. Other data structures are too complex to iterate over, but this one specifically does make a lot of sense I think. Thanks for the suggestion!

@ylv-io curious to hear your thoughts on this matter

ylv-io (approved):
It is a great value addition. In fact, we didn't even consider this, which I now feel ashamed of. Given that TimelockAuthorizer is not a user-facing contract, we do not care that much about gas costs, but the value added from being able to fetch data on-chain is significant.

I would like to second @danielmkm's experience regarding fetching eth_logs. It is neither reliable nor easy to do on a frontend.

nventuro:
Superceded by https://github.com/balancer/balancer-v2-monorepo/pull/2485, thanks a lot @danielmkm!

PR #2418, opened 2023-04-08
Make VaultReentrancy check a view function

EndymionJkb:
# Description

Midas implemented the VaultReentrancyLib this way, which allows the check to be a view function.

An external lib version of this is already in #2412, but in case we don't go that direction in handling Managed Pool bytecode, this should make the library less restrictive, and easier for partners to use directly.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/pool-utils/contracts/lib/VaultReentrancyLib.sol
```diff
@@ -33,8 +34,24 @@ library VaultReentrancyLib {
...
+        //
+        // IVault.UserBalanceOp[] memory noop = new IVault.UserBalanceOp[](0);
+        // _vault.manageUserBalance(noop);
+
+        // solhint-disable-next-line var-name-mixedcase
+        bytes32 REENTRANCY_ERROR_HASH = keccak256(abi.encodeWithSignature("Error(string)", "BAL#400"));
+
+        // read-only re-entrancy protection - we are making a static call on a non-view function. It will either
+        // "succeed" (return with no transaction), or revert. if it reverts, we need to make sure it didn't fail
+        // due to a re-entrancy attack.
```
nventuro, line 50: This explanation seems wrong. This call will always revert: the reentrancy guard check is non-view. These are the first lines that a call to `manageUserBalance` executes:

```
function _enterNonReentrant() private {
    _require(_status != _ENTERED, Errors.REENTRANCY);
    _status = _ENTERED;
}
```

So the transaction will either revert due to the require, or revert with no reason due to the forbidden `SSTORE`.

pkg/pool-utils/contracts/lib/VaultReentrancyLib.sol
```diff
@@ -33,8 +34,24 @@ library VaultReentrancyLib {
...
+        (, bytes memory revertData) = address(vault).staticcall(
+            abi.encodeWithSelector(vault.manageUserBalance.selector, new address[](0))
+        );
+
+        _require(keccak256(revertData) != REENTRANCY_ERROR_HASH, Errors.REENTRANCY);
```
nventuro, line 55: Bytecode size has been a concern, and this could likely have its bytecode size reduced greatly by using assembly, which would let us hardcode the selector, expected revert reason, calldata, etc. It might be a good idea to strengthen the tests in that case.

Alternatively we could move this to a separate contract and call there. Or we could also not care about this.

nventuro (commented):
Thanks for adding this! Indeed, having this be `view` would make things much easier for third party developers.

pkg/pool-utils/contracts/lib/VaultReentrancyLib.sol
ylv-io: This code is dangerous because the  `static_call` might revert for other reasons before hitting reentrancy error. We should make sure it never happens. 
The previous version is different because it requires a successful `manageUserBalance` call which is different compare to only failing with the reentrancy error. In fact would `manageUserBalance` call succeed if it doesn't change the state?

jubeira (commented):
The reasoning looks correct and the comments are pretty clear. Well thought!

Before approving: is there any test that should be added to this? Apparently the two outcomes are pretty well bounded, but I guess it doesn't hurt to be extra careful here.

EndymionJkb:
> Before approving: is there any test that should be added to this?
I'd be happy to add a test, but I'm not sure what it would be. Is there a way to make it fail with some "other" exception, and show that it's still protected?

Given that it fails in the modifier, before even calling manageUserBalance, not really. Though I suppose we could use a Mock Vault with an upstream modifier that fails in a different way when that function is called.

pkg/pool-utils/test/VaultReentrancyLib.test.ts
```diff
@@ -189,4 +190,17 @@ describe('VaultReentrancyLib', function () {
       });
     });
   });
+
+  describe('in MockVault context (test protecting view function', () => {
```
jubeira, line 194:
```suggestion
  describe('in MockVault context test protecting view function', () => {
```

pkg/pool-utils/contracts/test/MockVault.sol
```diff
@@ -224,8 +224,22 @@ contract MockVault is IPoolSwapStructs {
...
-    // Note that it isn't itself protected by the Vault context, like the real Vault would be.
-    function manageUserBalance(IVault.UserBalanceOp[] memory ops) external payable {
+    function manageUserBalance(IVault.UserBalanceOp[] memory) external payable nonReentrant {
         // solhint-disable-previous-line no-empty-blocks
     }
+
+    // The real Vault doesn't have any hooks that are view functions, so add one to this Vault
+    // specifically to test read-only reentrancy protection on view functions.
+    function functionWithHook(address pool) external nonReentrant {
+        (bool success, bytes memory returnData) = pool.call(abi.encodeWithSignature("viewHook()"));
```
jubeira, line 234: We could also call `protectedViewFunction` directly right? Or am I missing something?

pkg/pool-utils/test/VaultReentrancyLib.test.ts
```diff
@@ -189,4 +190,17 @@ describe('VaultReentrancyLib', function () {
...
+    let vault: Contract, pool: Contract;
+
+    sharedBeforeEach('deploy mock vault', async () => {
+      ({ instance: vault } = await Vault.create({ admin, mocked: true }));
+      pool = await deployPool({ vault: vault.address, tokens, owner: poolOwner });
+    });
+
+    it('reverts when calling a protected view function', async () => {
+      await expect(vault.connect(poolOwner).functionWithHook(pool.address)).to.be.revertedWith('REENTRANCY');
```
jubeira, line 203: :100:

jubeira (approved):
Looking good! I think the test is a good addition, thanks for taking care of it!

pkg/pool-utils/contracts/test/MockVault.sol
EndymionJkb: I'm just making it match the pattern of the others (i.e., be more realistic).

PR #2431, opened 2023-04-13
Add ChildChainGaugeRegistry

ylv-io:
# Description

* We need to have an enumerable list of gauges on the L2s.
* Store this list and provide functionality for querying it (e.g., return the total number of elements and a function to return a slice of the array from element N to element M).
* Allow someone to update the list by adding new entries.
* The second function should be permissioned.
* Additionally, we want to check that the gauges are valid.
  * This validation is similar to the check from L2BalancerPseudoMinter:
  * require(isValidGaugeFactory(factory), "INVALID_GAUGE_FACTORY");
  * require(factory.isGaugeFromFactory(gauge), "INVALID_GAUGE");
   * (You'd want to call l2BalancerPseudoMiner.isValidGaugeFactory().)
* With the list of gauges, we can then mass-update them when the veBAL balance gets bridged.
* Duplication is not allowed
* Gagues can be removed by authorized actor

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/liquidity-mining/contracts/gauges/ChildChainGaugeRegistry.sol
+++ b/pkg/liquidity-mining/contracts/gauges/ChildChainGaugeRegistry.sol
@@ -54,7 +54,7 @@ contract ChildChainGaugeRegistry is SingletonAuthentication, ReentrancyGuard {
 
     /**
      * @notice Add a gauge to the registry after validating its legitimacy.
-     * This function checks that the gauge's factory is registered with the L2BalancerPseudoMinter,
+     * @dev This function checks that the gauge's factory is registered with the L2BalancerPseudoMinter,
      * and that the gauge has been deployed from the registered factory. If these conditions are met,
      * the gauge is added to the registry, and a GaugeAdded event is emitted.
      * @param gauge The gauge to add to the registry.
--- a/pkg/liquidity-mining/contracts/gauges/ChildChainGaugeRegistry.sol
+++ b/pkg/liquidity-mining/contracts/gauges/ChildChainGaugeRegistry.sol
@@ -74,7 +74,7 @@ contract ChildChainGaugeRegistry is SingletonAuthentication, ReentrancyGuard {
 
     /**
      * @notice Remove a registered gauge from the registry and emit a GaugeRemoved event.
-     * If the gauge is not registered, the function reverts with a "GAUGE_NOT_REGISTERED" error.
+     * @dev If the gauge is not registered, the function reverts with a "GAUGE_NOT_REGISTERED" error.
      * Remove a gauge might affect the order of the remaining gauges.
      * @param gauge The gauge to remove from the registry.
      */
--- a/pkg/liquidity-mining/contracts/gauges/ChildChainGaugeRegistry.sol
+++ b/pkg/liquidity-mining/contracts/gauges/ChildChainGaugeRegistry.sol
@@ -19,10 +19,17 @@ import "@balancer-labs/v2-interfaces/contracts/liquidity-mining/IChildChainGauge
 
 import "@balancer-labs/v2-solidity-utils/contracts/helpers/SingletonAuthentication.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/EnumerableSet.sol";
+import "@balancer-labs/v2-solidity-utils/contracts/openzeppelin/ReentrancyGuard.sol";
 
 import "../L2BalancerPseudoMinter.sol";
 
-contract ChildChainGaugeRegistry is SingletonAuthentication {
+/**
+ * @title ChildChainGaugeRegistry
+ * @notice Registry for all child chain gauges.
+ * This contract enables the addition and removal of child chain gauges to the registry.
+ * Duplication is not permitted. Gauges are verified to be valid.
+ */
+contract ChildChainGaugeRegistry is SingletonAuthentication, ReentrancyGuard {
     using EnumerableSet for EnumerableSet.AddressSet;
 
     L2BalancerPseudoMinter private immutable _l2BalancerPseudoMinter;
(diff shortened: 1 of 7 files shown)


ylv-io:
I am not certain if the tests are sufficient. Any feedback would be greatly appreciated.

pkg/liquidity-mining/contracts/gauges/ChildChainGaugeRegistry.sol
```diff
@@ -0,0 +1,81 @@
...
+        require(_gauges.remove(address(gauge)), "GAUGE_NOT_REGISTERED");
+
+        emit GaugeRemoved(gauge);
+    }
+
+    function totalGauges() external view returns (uint256) {
+        return _gauges.length();
+    }
+
+    function getGauges(uint256 startIndex, uint256 endIndex) external view returns (IChildChainGauge[] memory) {
```
jubeira, line 68: Maybe worth explaining in docs that this gets gauges from `startIndex` (included) to `endIndex` (excluded).

pkg/liquidity-mining/test/ChildChainGaugeCheckpointer.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    await registry.connect(admin).addGauge(gauge1.address);
+    await registry.connect(admin).addGauge(gauge2.address);
+
+    checkpointer = await deploy('ChildChainGaugeCheckpointer', {
+      args: [registry.address],
+    });
+  });
+
+  describe('onVeBalBridged', () => {
```
jubeira, line 54: Since there are no `require`s nor `if-else` chains, the only thing that comes to my mind that could be tested here is adding / removing some gauges after testing the base case, and checking that it only checkpoints the ones that are currently in the registry.

pkg/liquidity-mining/test/ChildChainGaugeCheckpointer.test.ts
```diff
@@ -0,0 +1,65 @@
...
+  let gauge2: Contract;
+  let checkpointer: Contract;
+  let admin: SignerWithAddress, user1: SignerWithAddress, other: SignerWithAddress;
+  let registry: Contract;
+
+  before('setup signers', async () => {
+    [, admin, user1, other] = await ethers.getSigners();
+  });
+
+  sharedBeforeEach('deploy gauge ChildChainGaugeCheckpointer', async () => {
```
jubeira, line 23: (not strong opinion) maybe consider splitting this block in two smaller ones: one that just deploys the base contracts up to the checkpointer, and one that sets up the permissions, adds the factory to the minter, deploys the gauges and adds them to the registry.

pkg/liquidity-mining/test/ChildChainGaugeRegistry.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    name: 'ChildChainGauge',
+    version: '0',
+    deployment: 'test-deployment',
+  });
+
+  before('setup signers', async () => {
+    [, admin, other] = await ethers.getSigners();
+  });
+
+  sharedBeforeEach('deploy gauge registry', async () => {
```
jubeira, line 37: Same as above, consider splitting the setup in smaller blocks if that makes sense to you.

pkg/liquidity-mining/test/ChildChainGaugeRegistry.test.ts
```diff
@@ -0,0 +1,135 @@
...
+  });
+
+  describe('addGauge', () => {
+    it('can add a valid gauge to the registry', async () => {
+      const tx = await registry.connect(admin).addGauge(gauge.address);
+
+      expectEvent.inReceipt(await tx.wait(), 'GaugeAdded', { gauge: gauge.address });
+      expect(await registry.totalGauges()).to.eq(1);
+      expect((await registry.getGauges(0, 1))[0]).to.eq(gauge.address);
+    });
```
jubeira, line 88: You could split this `it` block in 3 small tests:
- `it('emits an event')` --> checks only that the event is emitted in the tx.
- `it('increments total gauges') --> checks total gauges before and after calling `addGauge` (after = before + 1, you don't need to hardcode the size).
- `it('adds the gauge at the end') --> checks that the gauge can be retrieved with `getGauges` using the right indexes. You can call `totalGauges` before to know what index to use as start so that you don't need to hard code it.

Then, you can add all those 3 `it` blocks inside a function called `itAddsGaugeCorrectly` and use it below.

pkg/liquidity-mining/test/ChildChainGaugeRegistry.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      const maliciousGauge = await deploy('MockChildChainGauge', { args: ['test'] });
+      await maliciousGauge.setMockFactory(gaugeFactory.address);
+      await expect(registry.connect(admin).addGauge(maliciousGauge.address)).to.be.revertedWith(
+        'GAUGE_NOT_FROM_FACTORY'
+      );
+    });
+
+    it('reverts if gauge factory is invalid', async () => {
+      await expect(registry.connect(admin).addGauge(otherGauge.address)).to.be.revertedWith('INVALID_GAUGE_FACTORY');
+    });
```
jubeira, line 109: These look good :+1: 

Following the style of other unit tests, you could also use contexts to describe what's happening, and rename the `it` blocks to just `it('reverts')`.

E.g.:

```typescript
context('when the caller is not authorized', () => {
  it('reverts', async () => {
    await expect(registry.connect(other).addGauge(gauge.address)).to.be.revertedWith('SENDER_NOT_ALLOWED');
  });
});
```

And so on and so forth for the rest of the invalid cases.

For the valid one:

```typescript
context('when the gauge, factory and caller are valid', () => {
  itAddsGaugeCorrectly();
});
```

pkg/liquidity-mining/test/ChildChainGaugeRegistry.test.ts
```diff
@@ -0,0 +1,135 @@
...
+  });
+
+  describe('removeGauge', () => {
+    sharedBeforeEach('add gauge', async () => {
+      await registry.connect(admin).addGauge(gauge.address);
+    });
+
+    it('can remove a valid gauge from the registry', async () => {
```
jubeira, line 117: Same as above; I'd split this in smaller super simple blocks.

pkg/liquidity-mining/test/ChildChainGaugeRegistry.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      const tx = await registry.connect(admin).removeGauge(gauge.address);
+
+      expectEvent.inReceipt(await tx.wait(), 'GaugeRemoved', { gauge: gauge.address });
+      expect(await registry.totalGauges()).to.eq(0);
+      await expect(registry.getGauges(0, 1)).to.be.revertedWith('END_INDEX_OUT_OF_BOUNDS');
+    });
+
+    it('reverts if gauge is not present', async () => {
```
jubeira, line 125: You can consider using `context` here as well, although it might be a bit of an overkill given that this case is a bit simpler than `add`.

jubeira (commented):
This is great work @ylv-io ! I think the code and the tests are really well oriented.

I've left a bunch of minor comments, mostly about docs and some opinionated feedback about the tests. But overall I think this looks correct and mostly complete, great job!

pkg/liquidity-mining/contracts/gauges/ChildChainGaugeRegistry.sol
```diff
@@ -0,0 +1,81 @@
...
+        ILiquidityGaugeFactory factory = gauge.factory();
+        require(_l2BalancerPseudoMinter.isValidGaugeFactory(factory), "INVALID_GAUGE_FACTORY");
+        require(factory.isGaugeFromFactory(address(gauge)), "GAUGE_NOT_FROM_FACTORY`");
+
+        require(_gauges.add(address(gauge)), "GAUGE_ALREADY_REGISTERED");
+
+        emit GaugeAdded(gauge);
+    }
+
+    function removeGauge(IChildChainGauge gauge) external authenticate {
```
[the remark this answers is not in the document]
ylv-io: Is order important?

pkg/liquidity-mining/test/ChildChainGaugeCheckpointer.test.ts
ylv-io: Splitting one `sharedBeforeEach` into multiple would require declaring many function vars at the test scope level which would pollute the list of vars used in the test. Also the whole code being inside one function makes it more obvious that it is executed sequentially.

pkg/liquidity-mining/test/ChildChainGaugeRegistry.test.ts
ylv-io: Hardcoding values can be beneficial in that it eliminates reliance on potentially faulty functions. In fact, during testing, it is essential to ensure that specific inputs yield predetermined outputs. If your inputs are not fixed, you may be susceptible to elusive bugs.

pkg/liquidity-mining/test/ChildChainGaugeRegistry.test.ts
ylv-io:
```javascript
context('when the gauge, factory and caller are valid', () => {
  itAddsGaugeCorrectly();
});
```
What is the rationale behind not inlining this? Because if you have to jump to the itAddsGaugeCorrectly implementation, you can't see the context description anymore. I understand it is useful when you have to call itAddsGaugeCorrectly in multiple contexts, but it should be inlined in the case it is used only once. Right?

jubeira: I agree that it's a tradeoff, and for this test in particular it's a gray area.

> This modification introduces two extra lines to the code and adds an additional nesting level to the CLI output. While it may be reasonable to implement this when handling multiple reverts within a single context, the rationale for applying it to a single revert is unclear. I am interested in hearing any arguments supporting this approach.

In general we use `context` for more complex tests, and I agree that it's not necessarily justified for these tests. I think the 'pros' of nesting is that it follows the style of other tests, it is more maintainable in some sense, and personally I think the outputs look slightly better. Using contexts also helps me reason about covering all the cases that can go wrong, but again that's a personal appreciation.
None of these reasons are very strong for these tests in particular; as you say the readability gets worse by nesting, and just inlining is more straightforward.

> What is the rationale behind not inlining this? Because if you have to jump to the itAddsGaugeCorrectly implementation, you can't see the context description anymore. I understand it is useful when you have to call itAddsGaugeCorrectly in multiple contexts, but it should be inlined in the case it is used only once. Right?

Same as above; maybe I'm too used to reading other tests that follow that style. Again for this case which is relatively simple maybe it doesn't add much.

All in all, I think the initial implementation is also just fine.

pkg/liquidity-mining/contracts/gauges/ChildChainGaugeRegistry.sol
```diff
@@ -33,38 +40,64 @@ contract ChildChainGaugeRegistry is SingletonAuthentication {
...
+        require(factory.isGaugeFromFactory(address(gauge)), "GAUGE_NOT_FROM_FACTORY");
 
         require(_gauges.add(address(gauge)), "GAUGE_ALREADY_REGISTERED");
 
         emit GaugeAdded(gauge);
     }
 
+    /**
+     * @notice Remove a registered gauge from the registry and emit a GaugeRemoved event.
+     * If the gauge is not registered, the function reverts with a "GAUGE_NOT_REGISTERED" error.
```
jubeira, line 77:
```suggestion
     * @dev If the gauge is not registered, the function reverts with a "GAUGE_NOT_REGISTERED" error.
```

pkg/liquidity-mining/test/ChildChainGaugeCheckpointer.test.ts
```diff
@@ -56,10 +60,53 @@ describe('ChildChainGaugeCheckpointer', () => {
...
+        await tx.wait(),
+        gauge3.interface,
+        'UserCheckpoint',
+        {
+          user: user1.address,
+        },
+        gauge3.address
+      );
+
+      expect(expectEvent.arrayFromIndirectReceipt(await tx.wait(), gauge1.interface, 'UserCheckpoint').length).to.eq(1);
```
jubeira, line 109: Same as above.

jubeira (approved):
I think this is looking good! Great job!

I've left a few minor comments, but otherwise looks good to me!

PR #2451, opened 2023-04-27
Deployment Preparation: Gauge Working Balance Helper

EndymionJkb:
# Deployment preparation: GaugeWorkingBalanceHelper

Build-info: master@https://github.com/balancer/balancer-v2-monorepo/commit/72f869c0faea8360b9e60d1eef3c69a50ea4669e

## Deployment preparation tasks checklist:

- [X] Deployments `README` is updated with the new task name and link
- [X] Task `README` has a description and links to the artifacts
- [X] `build-info` is updated with artifacts compiled at the specified commit
- [X] Artifacts are generated from `build-info`
- [X] New task has a working fork test 

## Code checklist:

- [X] Complex code has been commented in tests and scripts
- [X] Fork test block number and test cases are appropriate for this deployment


pkg/deployments/tasks/20230427-gauge-working-balance-helper/readme.md
```diff
@@ -0,0 +1,9 @@
+# 2023-04-27 - Gauge Working Balance Helper
+
+Deployment of the `GaugeWorkingBalanceHelper`, which reads the current and projected `working_balance` (effective veBAL boosted balance) for a user on a gauge, as well as the ratio of the working_balance to the supply. This can be used (e.g., by front-end processes) to determine whether checkpointing a gauge would be advantageous.
+
+## Useful Files
+
+- [`GaugeWorkingBalanceHelper` artifact](./artifact/GaugeWorkingBalanceHelper.json)
+- [`LiquidityGaugeV5` artifact](./artifact/LiquidityGaugeV5.json)
+- [`ChildChainGauge` artifact](./artifact/ChildChainGauge.json)
```
nventuro, line 9: Why are these included? I'd remove them entirely

pkg/deployments/tasks/20230427-gauge-working-balance-helper/test/L1GaugeHelper.fork.ts
```diff
@@ -0,0 +1,171 @@
...
+      expect(projectedWorkingBalance).to.gt(0);
+    });
+  });
+
+  context('with veBAL', () => {
+    const TOKENLESS_PRODUCTION = 0.4;
+    const MAX_BALANCE_RATIO = 1 / TOKENLESS_PRODUCTION;
+
+    before('create veBAL whale', async () => {
+      bal80weth20Pool = await deployedAt('v2-pool-weighted/WeightedPool', VEBAL_POOL);
```
nventuro, line 113: You should use the corresponding task for this, not the artifact from the live package.

pkg/deployments/tasks/20230427-gauge-working-balance-helper/test/L1GaugeHelper.fork.ts
```diff
@@ -0,0 +1,171 @@
...
+          gauge.address,
+          veBALHolder.address
+        );
+
+        expect(projectedWorkingBalance).to.be.almostEqual(currentWorkingBalance);
+        expect(projectedWorkingBalance).to.be.gt(0);
+        expect(projectedWorkingBalance).to.be.lte(currentWorkingBalance);
+      });
+
+      it('current and projected ratios should now be equal', async () => {
```
nventuro, line 161: This is conceptually wrong, in that there's no reason why the rates would be equal. The rate could also be slightly lower. In this case, because we checkpoint and then immediately query it turns out both the balances and ratios are the same. A better test would be to check for them to be very close or equal here, then move forward e.g. one week, and then check that they are close but not the same.

pkg/deployments/tasks/20230427-gauge-working-balance-helper/test/L2GaugeHelper.fork.ts
```diff
@@ -0,0 +1,155 @@
...
+
+      veDelegationProxy = await deploy('v2-liquidity-mining/VotingEscrowDelegationProxy', {
+        args: [vault.address, votingEscrow.address, ZERO_ADDRESS],
+      });
+
+      workingBalanceHelper = await deploy('v2-liquidity-mining/GaugeWorkingBalanceHelper', {
+        args: [veDelegationProxy.address, false],
+      });
+    });
+
```
nventuro, line 107: I don't get this, why are you redeploying these contracts, and then directly from the package?

pkg/deployments/tasks/20230427-gauge-working-balance-helper/readme.md
EndymionJkb: The fork tests don't work without them

pkg/deployments/tasks/20230427-gauge-working-balance-helper/test/L1GaugeHelper.fork.ts
EndymionJkb: Well, I took this from the Gnosis Root Gauge Factory deployment. I need that exact pool, not just any pool.

I just call `getPoolId` and `approve` on it though; I don't need it "as" a WeightedPool. So what I can do is hard-code the poolId instead of the address. Then I can just get it as an IERC20 to call approve.

const [poolAddress] = await vault.getPool(VEBAL_POOL_ID);
const bal80weth20Pool = await deployedAt('IERC20', poolAddress);

pkg/deployments/tasks/20230427-gauge-working-balance-helper/test/L2GaugeHelper.fork.ts
EndymionJkb: Maybe there's a better way to do it, but I'm redeploying because the regular task version uses the real VotingEscrow, while I want to use my `MockL2VotingEscrow`, so that I can directly set a veBAL balance (the "veBAL whale" step is then a one-liner).

EndymionJkb:
Moving to Deployments repo.

PR #2457, opened 2023-05-02
Add missing output links

EndymionJkb:
# Description

Noticed there were a few missing output links in /deployments. Also some obsolete mentions of kovan, which were replaced with goerli. Finally, made the working/order consistent.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/deployments/tasks/20220707-distribution-scheduler/readme.md
```diff
@@ -5,5 +5,6 @@ Deployment of the `DistributionScheduler`, which can be used to schedule token d
 ## Useful Files
 
 - [Ethereum mainnet addresses](./output/mainnet.json)
+- [Avalanche mainnet addresses](./output/avalanche.json)
```
jubeira, line 8: Hmm the entry here is correct, but the fact that it is not deployed in other L2s makes me think that this is probably not required in Avalanche, and it shouldn't be considered a 'canonical' deployment.

jubeira (approved):
Just one side comment; otherwise LGTM.

Isn't there a way to add this to CI? (i.e. check that all outputs have a link in the readme)

pkg/deployments/tasks/20220707-distribution-scheduler/readme.md
EndymionJkb: Yes; I'm not sure why that was deployed there. (It wasn't one of the recent ones I just did.) Anyway not hurting anything.

EndymionJkb:
> Isn't there a way to add this to CI? (i.e. check that all outputs have a link in the readme)

Well, the problem is our two-stage deployment. The artifacts always have to be there: and that is checked. But the outputs will only be there in the actual Deployment PR, not the Deployment Preparation PR, so we'd have to know somehow which one we were in to do the check, and I don't think we should have "logic" like that in CI.

It's on the checklist; we just have to be more vigilant. (Most of these are from before we had the checklist in place.)

PR #2463, opened 2023-05-02
Add `OmniVotingEscrowAdaptor`

### LINKED ISSUE #2402: Call `OmniVotingEscrow` using an adaptor in the remapper
See https://github.com/balancer/balancer-v2-monorepo/pull/2174#discussion_r1154753328 for reference.

The adaptor would add the extra data that might be required by the omni voting escrow, providing the remapper a stable and simple interface (e.g. user, chain ID, maybe refund address).

jubeira:
# Description

This PR adds the `OmniVotingEscrowAdaptor`, an intermediate contract between the `VotingEscrowRemapper` and the `OmniVotingEscrow` that provides a stable interface to the remapper, while allowing to configure certain parameters to be used when the omni voting escrow is called.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #2402


pkg/liquidity-mining/contracts/VotingEscrowRemapper.sol
```diff
@@ -205,8 +199,8 @@ contract VotingEscrowRemapper is IVotingEscrowRemapper, SingletonAuthentication,
     function clearNetworkRemapping(address localUser, uint16 chainId) external payable override nonReentrant {
         require(localUser != address(0), "localUser cannot be zero address");
         require(!_isAllowedContract(localUser) || localUser == msg.sender, "localUser is still in good standing");
-        IOmniVotingEscrow omniVotingEscrow = getOmniVotingEscrow();
-        require(address(omniVotingEscrow) != address(0), "Omni voting escrow not set");
+        IOmniVotingEscrowAdaptor omniVotingEscrowAdaptor = getOmniVotingEscrowAdaptor();
+        require(address(omniVotingEscrowAdaptor.getOmniVotingEscrow()) != address(0), "Omni voting escrow not set");
```
jubeira, line 203: With the separation of concerns in https://github.com/balancer/balancer-v2-monorepo/pull/2463/commits/e70e32dabd25890f48eb8b93e702207373c14d9e, `getOmniVotingEscrow` is out of the settings just because of this.

This is just a small gas optimization in the end. I could move the require to `sendUserBalance` and `estimateSendUserBalance` inside the adaptor, but that would mean that we would be checking if the voting escrow is set more than once in clear / set network remapping.

WDYT?

pkg/interfaces/contracts/liquidity-mining/IVotingEscrowRemapper.sol
```diff
@@ -16,6 +16,7 @@ pragma solidity ^0.7.0;
 pragma experimental ABIEncoderV2;
 
 import "./IOmniVotingEscrow.sol";
```
nventuro, line 18: I don't think we need to import this here?

nventuro (approved):

PR #2474, opened 2023-05-05
Prepare zkEVM deployment

nventuro:
I did the Authorizer, Vault and Adaptor (which we need for action ids). This was a bit more involved than other networks because it is not supported by the verification plugin, so I had to add this as a custom network and pass it to the verifier.


pkg/deployments/addresses/zkevm.json
```diff
@@ -0,0 +1,110 @@
...
+  },
+  "0x8df317a729fcaA260306d7de28888932cb579b88": {
+    "task": "20230414-authorizer-wrapper",
+    "name": "AuthorizerWithAdaptorValidation"
+  },
+  "0x3B1eb8EB7b43882b385aB30533D9A2BeF9052a98": {
+    "task": "20211202-no-protocol-fee-lbp",
+    "name": "NoProtocolFeeLiquidityBootstrappingPoolFactory"
+  },
+  "0x3c87FF3e9307dbEbfAe720E04C6439e49f79BF9B": {
```
jubeira, line 78: The block explorer is not showing any code here: https://zkevm.polygonscan.com/address/0x3c87ff3e9307dbebfae720e04c6439e49f79bf9b

jubeira (commented):
Looking good!

There is one contract that was not verified (I've tried and failed as well):
- `ProtocolFeesCollector`: `0xce88686553686DA562CE7Cea497CE749DA109f9F`

And I've left one more comment. It sounds like the block explorer has some trouble with contracts deployed by contracts; another example was `BalancerRelayer` which wasn't verified, but in that case I could verify it myself.

EndymionJkb:
Yes; looks like this is one of those chains that can't verify contract deployments :(

jubeira:
> Yes; looks like this is one of those chains that has trouble with / can't verify contract deployments :(

Sounds good. Note that it doesn't _always_ happen: the `BalancerRelayer` is deployed by the library, and I could verify it.
On the other hand, the fact that it wasn't verified when I reviewed the PR means that the verification failed the first time, so there's probably something flaky in the verification process.

pkg/deployments/addresses/zkevm.json
nventuro: This address does have code. Checked with the zkEVM team, they say it's an issue on Etherscan's end and they're working on it.

EndymionJkb (approved):

pkg/deployments/package.json
```diff
@@ -25,11 +25,11 @@
...
+    "build-address-lookup": "hardhat build-address-lookup --network mainnet && hardhat build-address-lookup --network polygon && hardhat build-address-lookup --network arbitrum && hardhat build-address-lookup --network optimism && hardhat build-address-lookup --network goerli && hardhat build-address-lookup --network gnosis && hardhat build-address-lookup --network bsc && hardhat build-address-lookup --network avalanche && hardhat build-address-lookup --network sepolia && hardhat build-address-lookup --network zkevm",
+    "check-address-lookup": "hardhat build-address-lookup --network mainnet && hardhat build-address-lookup --network polygon && hardhat build-address-lookup --network arbitrum && hardhat build-address-lookup --network optimism && hardhat build-address-lookup --network goerli && hardhat build-address-lookup --network gnosis && hardhat build-address-lookup --network bsc && hardhat build-address-lookup --network avalanche && hardhat build-address-lookup --network sepolia && hardhat build-address-lookup --network zkevm",
```
jubeira, line 32: :grimacing: 
I'll fix this when porting these deployments

PR #2476, opened 2023-05-05
Update deployment addresses

### LINKED ISSUE #2445: Revive deployments address registry
We have files with deployment addresses that is not completely up to date: https://github.com/balancer/balancer-v2-monorepo/tree/master/pkg/deployments/addresses.

These are convenient to generate documentation.
Some notes:
- Update missing releases.
- Handle deprecated tasks.
- Add step to deployments process to keep it up to date.

jubeira:
# Description

Update deployment addresses and add step to deployment process to keep it up to date.
The changes in the PR were generated with `yarn build-address-lookup`.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

- N/A Complex code has been commented, including external interfaces
- N/A Tests are included for all code paths

## Issue Resolution

Closes #2445.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/deployment-template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/deployment-template.md
@@ -6,6 +6,7 @@
 - [ ] TX IDs are properly generated
 - [ ] Action IDs are generated after deployment (if applicable) <!-- This includes contracts deployed by contracts (e.g. mock pools deployed by factories) -->
 - [ ] Outputs for all target networks are generated, and linked in the task `readme`
+- [ ] Addresses files are up to date with the new deployment <!-- yarn build-address-lookup -->
 - [ ] Contracts are verified in every network <!-- This includes contracts deployed by contracts (e.g. mock pools deployed by factories) -->
 - [ ] Deployments `CHANGELOG` is updated
 - [ ] There are no code changes of any kind
--- a/pkg/deployments/addresses/arbitrum.json
+++ b/pkg/deployments/addresses/arbitrum.json
@@ -226,5 +226,185 @@
   "0x5cF4928a3205728bd12830E1840F7DB85c62a4B9": {
     "task": "20230223-protocol-id-registry",
     "name": "ProtocolIdRegistry"
+  },
+  "0xf23b4DB826DbA14c0e857029dfF076b1c0264843": {
+    "task": "20230206-aave-rebalanced-linear-pool-v4",
+    "name": "AaveLinearPoolFactory"
+  },
+  "0x76578ecf9a141296Ec657847fb45B0585bCDa3a6": {
+    "task": "20230206-aave-rebalanced-linear-pool-v4",
+    "name": "MockAaveLendingPool"
+  },
+  "0xb9F8AB3ED3F3aCBa64Bc6cd2DcA74B7F38fD7B88": {
+    "task": "20230206-aave-rebalanced-linear-pool-v4",
+    "name": "MockStaticAToken"
+  },
+  "0x11884dA90FB4221b3aa288a7741C51eC4Fc43B2f": {
+    "task": "20230206-aave-rebalanced-linear-pool-v4",
+    "name": "MockAaveLinearPool"
+  },
+  "0x1c99324EDC771c82A0DCCB780CC7DDA0045E50e7": {
+    "task": "20230206-composable-stable-pool-v3",
+    "name": "ComposableStablePoolFactory"
+  },
+  "0x231B05F3a92d578EFf772f2Ddf6DacFFB3609749": {
+    "task": "20230206-composable-stable-pool-v3",
+    "name": "MockComposableStablePool"
+  },
+  "0xa3B9515A9c557455BC53F7a535A85219b59e8B2E": {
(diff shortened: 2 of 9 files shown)


EndymionJkb:
So this already worked? We just need to run that existing task after deployments?

jubeira:
> So this already worked? We just need to run that existing task after deployments?

Yes, that should be pretty much it. This doesn't consider deprecated contracts (although as per off-thread chat it should be fine right?)

EndymionJkb:
Maybe also have it print a "green check" message when it succeeds, like the others? It works, but doesn't print anything.

nventuro:
We don't have any CI checks that look at the correctness of these files, do we?

EndymionJkb:
> No. Maybe we could add a task that generates them, and compares the generated files against the existing ones. That task will fail if these files are not updated during a new deployment. WDYT?

The other directories work that way, and the reason for resurrecting this was integrations/maxis wanted to rely on it to produce documentation, so we should have some guarantee that it's accurate (and complete).

jubeira:
Alright, I've modified the scripts a bit so that the address lookup file is generated all at once, and always respecting the same order. Before these changes, the file's entries could be ordered differently according to the order of the tasks added to it, which is not ideal.

I've also added the missing entries, and a CI task that checks whether the committed file is up to date as discussed. PTAL.

pkg/deployments/hardhat.config.ts
```diff
@@ -256,24 +256,48 @@ task('get-action-ids-info', `Reconstructs all the permissions from TheGraph AP a
...
+
+  // Create Task objects, excluding tokens tasks.
+  const tasks = Task.getAllTaskIds()
+    .filter((taskId) => !taskId.startsWith('00000000-'))
+    .map((taskId) => new Task(taskId, TaskMode.READ_ONLY, hre.network.name));
+
+  const addressLookupFileOk = checkContractDeploymentAddresses(tasks, hre.network.name);
+  if (!addressLookupFileOk) {
+    throw new Error(
+      `Address lookup file is incorrect for network ${hre.network.name}. Please re-generate file by running 'build-address-lookup' task`
```
EndymionJkb, line 295:
```suggestion
      `Address lookup file is incorrect for network ${hre.network.name}. Please run 'build-address-lookup' to regenerate it`
```

pkg/deployments/src/network.ts
```diff
@@ -42,25 +42,53 @@ export function getContractDeploymentTransactionHash(deployedAddress: string, ne
   return txHash;
 }
 
-export function saveContractDeploymentAddresses(task: Task): void {
-  if (task.network === 'hardhat') return;
+/**
+ * Saves file with canonical deployment addresses for all the given tasks in the given network,
```
EndymionJkb, line 46:
```suggestion
 * Saves a file with the canonical deployment addresses for all tasks in a given network.
```

pkg/deployments/src/network.ts
```diff
@@ -42,25 +42,53 @@ export function getContractDeploymentTransactionHash(deployedAddress: string, ne
...
+  if (network === 'hardhat') return;
+
+  const allTaskEntries = buildContractDeploymentAddressesEntries(tasks);
+  const filePath = path.join(CONTRACT_ADDRESSES_DIRECTORY, `${network}.json`);
+
+  fs.writeFileSync(filePath, _stringifyEntries(allTaskEntries));
+}
+
+/**
+ * Builds object that maps deployment addresses to {task ID, contract name} for all the given tasks.
```
EndymionJkb, line 58:
```suggestion
 * Builds an object that maps deployment addresses to {task ID, contract name} for all given tasks.
```

pkg/deployments/src/network.ts
```diff
@@ -42,25 +42,53 @@ export function getContractDeploymentTransactionHash(deployedAddress: string, ne
...
+
+  return allTaskEntries;
+}
 
-  const newEntries = Object.fromEntries(
-    Object.entries(task.output({ ensure: false })).map(([name, address]) => [address, { task: task.id, name }])
-  );
+/**
+ * Returns true if the existing deployment addresses file stored in `CONTRACT_ADDRESSES_DIRECTORY` matches the
+ * canonical one for the given set of tasks in the given network; false otherwise.
```
EndymionJkb, line 78:
```suggestion
 * Returns true if the existing deployment addresses file stored in `CONTRACT_ADDRESSES_DIRECTORY` matches the
 * canonical one for the given network; false otherwise.
```

EndymionJkb (approved):
One possible refinement. Macs tend to have lots of .DS_Store files strewn about from the Finder, and this causes all the tasks without --id flags to break, since they get included in the /tasks directory listing. (You have to know to delete the `tasks/.DS_Store` file.)

Consider filtering out hidden files along with the task file, to be Mac-friendly?
.filter((taskId) => !taskId.startsWith('00000000-') && !(taskId[0] == '.'))

Maybe even factor out that filtering code somehow, as it's in multiple places, though that might be hard given the structure of these files.

jubeira:
The address registry is happy, CI is happy, and Tim Cook is happy :+1:

PR #2485, opened 2023-05-12
Timelock authorizer scheduled tasks auditability

nventuro:
Supercedes https://github.com/balancer/balancer-v2-monorepo/pull/2413, adding tests and interface improvements.

### DIFF
--- a/pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
+++ b/pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
@@ -413,7 +413,7 @@ describe('TimelockAuthorizer execute', () => {
       return id;
     };
 
-    it('stores executor information', async () => {
+    it('stores canceler information', async () => {
       const id = await schedule();
       await advanceTime(delay);
       await authorizer.cancel(id, { from: canceler });
--- a/pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
+++ b/pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
@@ -10,7 +10,7 @@ import { actionId } from '@balancer-labs/v2-helpers/src/models/misc/actions';
 import { advanceTime, currentTimestamp, DAY } from '@balancer-labs/v2-helpers/src/time';
 import Vault from '@balancer-labs/v2-helpers/src/models/vault/Vault';
 import { sharedBeforeEach } from '@balancer-labs/v2-common/sharedBeforeEach';
-import { MAX_UINT256 } from '@balancer-labs/v2-helpers/src/constants';
+import { MAX_UINT256, ZERO_ADDRESS } from '@balancer-labs/v2-helpers/src/constants';
 
 describe('TimelockAuthorizer execute', () => {
   let authorizer: TimelockAuthorizer, vault: Contract, authenticatedContract: Contract;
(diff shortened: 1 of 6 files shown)


pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
```diff
@@ -78,6 +79,32 @@ describe('TimelockAuthorizer execute', () => {
...
+      const countAfter = await authorizer.instance.getScheduledExecutionsCount();
+
+      expect(countAfter).to.equal(countBefore.add(1));
+    });
+
+    it('stores scheduler information', async () => {
+      const id = await schedule(authenticatedContract);
+
+      const scheduledExecution = await authorizer.getScheduledExecution(id);
+      expect(scheduledExecution.scheduledBy).to.equal(user.address);
```
jubeira, line 94: minor detail: the fact that `user` is scheduling the execution is not obvious just from reading this diff. This is not the case in the other tests added in this PR.

Do you think it would make sense to make it a bit more explicit?

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
```diff
@@ -426,4 +495,175 @@ describe('TimelockAuthorizer execute', () => {
...
+
+    it('reverts if the maxSize value is zero', async () => {
+      await expect(authorizer.instance.getScheduledExecutions(0, 0, false)).to.be.revertedWith('ZERO_MAX_SIZE_VALUE');
+    });
+
+    context('in chronological order', () => {
+      const reverseOrder = false;
+
+      it('returns entries in chronological order', async () => {
+        const entries = await authorizer.instance.getScheduledExecutions(0, 10, reverseOrder);
```
jubeira, line 555: Here and below: maybe make 0 and 10 (or at least 10) a named constant, and expect `MAX_SIZE` > `TOTAL_ENTRIES` to document which use case we are in.

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
```diff
@@ -426,4 +495,175 @@ describe('TimelockAuthorizer execute', () => {
...
+        expect(entries[0].where).to.equal(authorizer.address);
+        expect(entries[0].data).to.equal(
+          authorizer.interface.encodeFunctionData('setDelay', [await actionId(vault, 'setAuthorizer'), 2 * delay])
+        );
+
+        // The last entry is the one that we scheduled
+        expect(entries[entries.length - 1].where).to.equal(authenticatedContract.address);
+        expect(entries[entries.length - 1].data).to.equal(
+          authenticatedContract.interface.encodeFunctionData('protectedFunction', [`0x04`])
+        );
```
jubeira, line 569: Would it make sense to validate the ones in the middle with (or without) a loop?

jubeira (approved):
Looks pretty clean and correct! This is a nice addition, great work.

I've left just a few comments, mostly typos and some suggestions to make the tests slightly more explicit.

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
nventuro: I'm reusing the `schedule` function, which is defined 10 lines above, and which clearly sends the tx from `user`. All of the other tests in this file make that clear. I think this is fine.

pkg/vault/test/authorizer/TimelockAuthorizerExecute.test.ts
nventuro: Eh I dunno, it'd be a bit annoying, and by looking at the code I think it's clear that if the edges are correct then everything else is as well :shrug:

Issue #2482, opened 2023-05-09
Add Polygon ZK EVM root gauge


jubeira:
Stakeless gauge, for Polygon ZK EVM.

PR #2492, opened 2023-05-15
Polygon ZkEVM Root Gauge

EndymionJkb:
# Description

The Root Gauge and Factory contracts for the recently deployed Polygon zkEVM network. It seems to be a fairly simple bridge, and I've tested it through Etherscan, successfully bridging BAL. (Along with using the UI many times, and in both directions.)

Next step is the deployment prep + fork test.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Starts #2482 .

### DIFF
--- a/pkg/liquidity-mining/contracts/gauges/zkevm/PolygonZkEVMRootGauge.sol
+++ b/pkg/liquidity-mining/contracts/gauges/zkevm/PolygonZkEVMRootGauge.sol
@@ -1,0 +1,80 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.7.0;
+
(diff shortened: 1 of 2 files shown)


jubeira:
Nice! Looks rather simple indeed.

Could you post the bridge tx so that we can compare it against this code to double check?

And one question: there's no need to send any ETH in this case then?

EndymionJkb:
> Nice! Looks rather simple indeed.
> 
> Could you post the bridge tx so that we can compare it against this code to double check?
> 
> And one question: there's no need to send any ETH in this case then?
Mainnet: 0x3351455ea444ba06caa8ad3aea8369652f4a5baaa9e66d09b0db0e649c0510ab
zkEVM: 0x3a6d992fd9b1ec6367e1a2c7cf3492c5e9a86f47966184917397867dc21e9d2d

I did not send any ETH (on Etherscan, it prompted for an amount, and it was 0), if that's what you mean.

jubeira:
> I did not send any ETH (on Etherscan, it prompted for an amount, and it was 0), if that's what you mean.

Reading at the docs through Etherscan, the function you've called in the transaction is indeed payable, and there's a parameter for the token address:

```
Token address, 0 address is reserved for ether
```

If the function is payable just to bridge ETH, then we are fine.
I'm asking because other bridges (namely Arbitrum) require sending a bit of ETH as value to pay for bridge costs, and we need to know about that in our `_postMintAction`. Do their docs say anything about bridge costs anywhere?

EndymionJkb:
> If the function is payable just to bridge ETH, then we are fine. I'm asking because other bridges (namely Arbitrum) require sending a bit of ETH as value to pay for bridge costs, and we need to know about that in our `_postMintAction`. Do their docs say anything about bridge costs anywhere?

Not that I could see. The code path we're going through even seems to require that it be zero.

nventuro (approved):

PR #2499, opened 2023-05-17
Make Timelock grant and revoke consistent

nventuro:
# Description

Grant and revoke can be made globally by passing the EVERYWHERE sentinel value. This is the same as granters and cancelers. However, we were not considering the special case when mixing specific and global permissions. In particular, we shouldn't allow specific granting if the user is already global, and we should disawllow specific revoking for global users.

This PR brings the behavior of grant and revoke in line with that of granters, revokers and cancelers, fixes existing tests, and adds a couple new ones. The test style of grant and revoke is not ideal (and inferior to that of granter/revoker/canceler), but I followed it nonetheless to avoid causing too much disruption.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [x] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -315,7 +315,7 @@ contract TimelockAuthorizer is IAuthorizer, TimelockAuthorizerManagement {
             require(msg.sender == getTimelockExecutionHelper(), "GRANT_MUST_BE_SCHEDULED");
         }
 
-        require(!_isPermissionGranted[actionId][account][where], "PERMISSION_ALREADY_GRANTED");
+        require(!hasPermission(actionId, account, where), "PERMISSION_ALREADY_GRANTED");
 
         _isPermissionGranted[actionId][account][where] = true;
         emit PermissionGranted(actionId, account, where);
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -412,6 +412,8 @@ contract TimelockAuthorizer is IAuthorizer, TimelockAuthorizerManagement {
         address account,
         address where
     ) private {
+        require(hasPermission(actionId, account, where), "PERMISSION_NOT_GRANTED");
+
         if (_isPermissionGranted[actionId][account][EVERYWHERE()]) {
             // If an account has global permission, then it must explicitly lose this global privilege. This prevents
             // scenarios where an account has their permission revoked over a specific contract, but they can still
@@ -420,9 +422,6 @@ contract TimelockAuthorizer is IAuthorizer, TimelockAuthorizerManagement {
             // permission over some contracts after losing global privilege. This is considered an unlikely scenario,
             // and would require manual removal of the specific permissions even after removal of the global one.
             require(where == EVERYWHERE(), "ACCOUNT_HAS_GLOBAL_PERMISSION");
-        } else {
-            // Alternatively, they must currently have the permission in order to have it revoked.
-            require(_isPermissionGranted[actionId][account][where], "PERMISSION_NOT_GRANTED");
         }
 
         _isPermissionGranted[actionId][account][where] = false;
--- a/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
+++ b/pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
@@ -412,7 +412,18 @@ contract TimelockAuthorizer is IAuthorizer, TimelockAuthorizerManagement {
         address account,
         address where
     ) private {
-        require(_isPermissionGranted[actionId][account][where], "PERMISSION_NOT_GRANTED");
+        if (_isPermissionGranted[actionId][account][EVERYWHERE()]) {
+            // If an account has global permission, then it must explicitly lose this global privilege. This prevents
+            // scenarios where an account has their permission revoked over a specific contract, but they can still
+            // use it (including in that contract!) because they have global permission.
+            // There's an edge case in which an account could have both specific and global permission, and still have
+            // permission over some contracts after losing global privilege. This is considered an unlikely scenario,
+            // and would require manual removal of the specific permissions even after removal of the global one.
+            require(where == EVERYWHERE(), "ACCOUNT_HAS_GLOBAL_PERMISSION");
+        } else {
+            // Alternatively, they must currently have the permission in order to have it revoked.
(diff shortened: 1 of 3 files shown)


ylv-io (approved):
🙏

pkg/vault/contracts/authorizer/TimelockAuthorizer.sol
```diff
@@ -315,7 +315,7 @@ contract TimelockAuthorizer is IAuthorizer, TimelockAuthorizerManagement {
             require(msg.sender == getTimelockExecutionHelper(), "GRANT_MUST_BE_SCHEDULED");
         }
 
-        require(!_isPermissionGranted[actionId][account][where], "PERMISSION_ALREADY_GRANTED");
+        require(!hasPermission(actionId, account, where), "PERMISSION_ALREADY_GRANTED");
```
EndymionJkb, line 318: Maybe add a comment that this is now also checking for global permission for this action, as well as specific permission at where?

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
```diff
@@ -665,6 +598,48 @@ describe('TimelockAuthorizer permissions', () => {
             });
           });
         });
+
+        context('when the user has the permission in a specific contract and everywhere', () => {
```
EndymionJkb, line 602:
```suggestion
        context('when the user has both global permission and permission in a specific contract', () => {
```

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
```diff
@@ -665,6 +598,48 @@ describe('TimelockAuthorizer permissions', () => {
...
+
+          context('when revoking the permission for a contract', () => {
+            it('cannot revoke the permission', async () => {
+              await expect(
+                authorizer.revokePermission(ACTION_1, user, WHERE_1, { from: getSender() })
+              ).to.be.revertedWith('ACCOUNT_HAS_GLOBAL_PERMISSION');
+            });
+          });
+
+          context('when revoking the permission for a everywhere', () => {
```
EndymionJkb, line 616:
```suggestion
          context('when revoking a global permission', () => {
```

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
```diff
@@ -665,6 +598,48 @@ describe('TimelockAuthorizer permissions', () => {
...
+              await authorizer.revokePermissionGlobally(ACTION_1, user, { from: getSender() });
+
+              expect(await authorizer.canPerform(ACTION_1, user, EVERYWHERE)).to.be.false;
+              expect(await authorizer.canPerform(ACTION_1, user, WHERE_2)).to.be.false;
+            });
+
+            it('can still perform the requested action in the specific contract', async () => {
+              await authorizer.revokePermissionGlobally(ACTION_1, user, { from: getSender() });
+
+              expect(await authorizer.canPerform(ACTION_1, user, WHERE_1)).to.be.true;
```
EndymionJkb, line 627: Could also check that they can't do WHERE_2:
              expect(await authorizer.canPerform(ACTION_1, user, WHERE_2)).to.be.false;

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
```diff
@@ -665,6 +598,48 @@ describe('TimelockAuthorizer permissions', () => {
             });
           });
         });
+
+        context('when the user has the permission in a specific contract and everywhere', () => {
+          sharedBeforeEach('grants the permissions', async () => {
+            await authorizer.grantPermission(ACTION_1, user, WHERE_1, { from: root });
+            await authorizer.grantPermissionGlobally(ACTION_1, user, { from: root });
```
EndymionJkb, line 605: For completion, could check that it can do both after the grant, contrasted with tests below:
expect(await authorizer.canPerform(ACTION_1, user, WHERE_1)).to.be.true;
expect(await authorizer.canPerform(ACTION_1, user, WHERE_2)).to.be.true;
expect(await authorizer.canPerform(ACTION_1, user, EVERYWHERE)).to.be.true;

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
```diff
@@ -1067,5 +1031,59 @@ describe('TimelockAuthorizer permissions', () => {
         });
       });
     });
+
+    context('when the user has the permission for a specific contract and everywhere', () => {
```
EndymionJkb, line 1035:
```suggestion
    context('when the user has both global permission and permission in a specific contract', () => {
```

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
```diff
@@ -1067,5 +1031,59 @@ describe('TimelockAuthorizer permissions', () => {
...
+          expect(await authorizer.canPerform(ACTION_1, user, WHERE_2)).to.be.true;
+        });
+
+        it('can perform the requested action everywhere', async () => {
+          expect(await authorizer.canPerform(ACTION_1, user, EVERYWHERE)).to.be.true;
+          expect(await authorizer.canPerform(ACTION_1, user, WHERE_2)).to.be.true;
+        });
+      });
+
+      context('when renouncing the permission for everywhere', () => {
```
EndymionJkb, line 1059:
```suggestion
      context('when renouncing the global permission', () => {
```

EndymionJkb (commented):
Has some odd test phrasing:
"...the permission for a everywhere" -> "...a global permission"

Also consider renaming NOT_WHERE; I get it, but it might be clearer to call it:
SOMEWHERE_ELSE, or
ANYWHERE (i.e., a contract at ANY_ADDRESS)

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
nventuro: While this is better, I'm keeping (here and elsewhere) the original phrasing as it matches the rest of the suite. I'd rather it be sort of ugly but consistent, inconsistency will only make this even harder to read.

pkg/vault/test/authorizer/TimelockAuthorizerPermissions.test.ts
nventuro: Note that this is already done in the test above.

PR #2509, opened 2023-05-29
Make `peekChainedReferenceValue` payable, non-view.

jubeira:
# Description

`peek` fails to be used inside `multicall` whenever there's a value transfer because view functions are not payable.
This PR fixes it, and adjusts the tests accordingly.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A

### DIFF
--- a/pkg/standalone-utils/contracts/relayer/BaseRelayerLibrary.sol
+++ b/pkg/standalone-utils/contracts/relayer/BaseRelayerLibrary.sol
@@ -89,8 +89,11 @@ contract BaseRelayerLibrary is IBaseRelayerLibrary {
     /**
      * @notice Returns the amount referenced by chained reference `ref`.
      * @dev It does not alter the reference (even if it's marked as temporary).
+     *
+     * This function does not alter the state in any way. It is not marked as view because it has to be `payable`
+     * in order to be used in a batch transaction.
      */
-    function peekChainedReferenceValue(uint256 ref) public view override returns (uint256 value) {
+    function peekChainedReferenceValue(uint256 ref) public payable override returns (uint256 value) {
         (, value) = _peekChainedReferenceValue(ref);
     }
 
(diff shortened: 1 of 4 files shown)


pkg/standalone-utils/contracts/relayer/BaseRelayerLibrary.sol
```diff
@@ -89,8 +89,11 @@ contract BaseRelayerLibrary is IBaseRelayerLibrary {
     /**
      * @notice Returns the amount referenced by chained reference `ref`.
      * @dev It does not alter the reference (even if it's marked as temporary).
+     *
+     * This function does not alter the state in any way. It is not marked as view because it has to be `payable`
+     * in order to be used in a batch transaction.
      */
-    function peekChainedReferenceValue(uint256 ref) public view override returns (uint256 value) {
+    function peekChainedReferenceValue(uint256 ref) public payable override returns (uint256 value) {
```
EndymionJkb, line 96:
```suggestion
    function peekChainedReferenceValue(uint256 ref) external payable override returns (uint256 value) {
```
Everything could be external except getVault, the only thing we call

EndymionJkb, line 96: Everything could be external (except `getVault`, which is the only thing we call)

EndymionJkb (approved):
Looks good!

Maybe a comment here at the top of the BaseRelayerLibrary (like there is in contracts included by the BalancerRelayer: VaultActions, etc.) 
 * All functions must be payable so they can be called from a multicall involving ETH
 
Could also add a comment in the tests for why you're doing an explicit callStatic, and maybe in the contract if it changes how users have to call it

PR #2510, opened 2023-06-01
Add checkpoint gauges capability to `GaugeActions` (relayer)

### LINKED ISSUE #2399: Add checkpointGauges to BatchRelayer
The L2 `ChildChainGauge` lets anyone call `user_checkpoint(addr)`, which both checkpoints the user and updates their liquidity limit. We want to be able to update the liquidity limit as that is when the veBAL balance read is triggered and the new working balance computed, which we need whenever there's veBAL balance changes (e.g. on new locks, state pushes to L2s, etc.). So for these we can simply call this function from any contract.

For the L1 `LiquidityGaugeV5` however, the `user_checkpoint(addr)` function is restricted so that only the minter and the user can call it, so we want to find a different way to tackle this issue. There's two different ways we can address this.

One path is to try to mint BAL. There's two issues with this. The first one is that we'll mint BAL if there's any to mint, which the user didn't request. The second is that the gauge needs to be allowed to mint for the user (though ths can be done via signatures).

The second path is to perform a gauge token transfer, which also updates the liquidity limit. We could have a contract hold gauge lp tokens and transfer them, though this would be annoying operationally as we'd need to do top ups, etc. Alternatively, we can transfer gauge tokens from the user to themselves. We can do this with a single token, though ultimately it doesn't matter since they're both sender and recipient. This requires gauge token allowance, _but_ the Vault already has infinite allowance for these, so we can use the `TRANSFER_EXTERNAL` user balance operation kind to perform these. Because batch relayer is already a relayer, there's no extra setup we need to do.

Based on this, I think the self transfer is the best method to achieve this. We can add a `checkpointGauges(address user, address[] gauges)` function (maybe requiring `user == msg.sender` for this initial version), and then do the above. We could receive a constructor flag that indicates if we need to use the L1 or L2 mechanism.

jubeira:
# Description

Stab at #2399 (see details in ticket).
Tests pending; this is an initial draft for early feedback.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #2399.

### DIFF
--- a/pkg/standalone-utils/contracts/relayer/GaugeActions.sol
+++ b/pkg/standalone-utils/contracts/relayer/GaugeActions.sol
@@ -30,23 +30,23 @@ abstract contract GaugeActions is IBaseRelayerLibrary {
     using SafeERC20 for IERC20;
 
     IBalancerMinter private immutable _balancerMinter;
-    bool private immutable _isL2Relayer;
+    bool private immutable _canCallUserCheckpoint;
 
     /**
      * @dev The zero address may be passed as balancerMinter to safely disable features
      *      which only exist on mainnet
      */
-    constructor(IBalancerMinter balancerMinter, bool isL2Relayer) {
+    constructor(IBalancerMinter balancerMinter, bool canCallUserCheckpoint) {
         _balancerMinter = balancerMinter;
-        _isL2Relayer = isL2Relayer;
+        _canCallUserCheckpoint = canCallUserCheckpoint;
     }
 
     /**
-     * @notice Returns true if the relayer is configured to checkpoint L2 gauges.
+     * @notice Returns true if the relayer is configured to checkpoint gauges directly via `user_checkpoint`.
      * @dev This method is not expected to be called inside `multicall` so it is not marked as `payable`.
      */
-    function isL2Relayer() external view returns (bool) {
-        return _isL2Relayer;
+    function canCallUserCheckpoint() external view returns (bool) {
+        return _canCallUserCheckpoint;
     }
 
     function gaugeDeposit(
@@ -112,14 +112,14 @@ abstract contract GaugeActions is IBaseRelayerLibrary {
      * @dev Both mainnet and child chain gauges are supported.
      */
     function gaugeCheckpoint(address user, IStakingLiquidityGauge[] calldata gauges) external payable {
-        if (_isL2Relayer) {
-            _checkpointL2Gauges(user, gauges);
+        if (_canCallUserCheckpoint) {
+            _checkpointGaugesViaUserCheckpoint(user, gauges);
         } else {
-            _checkpointL1Gauges(user, gauges);
+            _checkpointGaugesViaUserBalance(user, gauges);
         }
     }
 
-    function _checkpointL2Gauges(address user, IStakingLiquidityGauge[] calldata gauges) internal {
+    function _checkpointGaugesViaUserCheckpoint(address user, IStakingLiquidityGauge[] calldata gauges) internal {
         uint256 numGauges = gauges.length;
         // In L2s (child chain gauges), `user_checkpoint` is not permissioned, so we can just call it directly.
         for (uint256 i = 0; i < numGauges; ++i) {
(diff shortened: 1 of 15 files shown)


jubeira:
As pointed out in https://github.com/balancer/balancer-v2-monorepo/issues/2399#issuecomment-1492616941, this can already be done by the SDK by constructing the right call for `VaultActions`, and that should work for the L1 and the L2.

In other words, this doesn't add a new capability at the end of the day; it just makes it more convenient and intuitive to use.
If the direction is correct I can polish the rough edges and write the tests.

pkg/standalone-utils/contracts/relayer/GaugeActions.sol
```diff
@@ -96,4 +106,47 @@ abstract contract GaugeActions is IBaseRelayerLibrary {
             gauges[i].claim_rewards(msg.sender);
         }
     }
+
+    /**
+     * @notice Perform a user checkpoint for the given user at the given set of gauges.
```
EndymionJkb, line 111:
```suggestion
     * @notice Perform a user checkpoint for the given user on the given set of gauges.
```

pkg/standalone-utils/contracts/relayer/GaugeActions.sol
```diff
@@ -96,4 +106,47 @@ abstract contract GaugeActions is IBaseRelayerLibrary {
...
+        }
+    }
+
+    function _checkpointL1Gauges(address user, IStakingLiquidityGauge[] calldata gauges) internal {
+        uint256 numGauges = gauges.length;
+        IVault.UserBalanceOp[] memory ops = new IVault.UserBalanceOp[](numGauges);
+
+        // In mainnet, `user_checkpoint` is permissioned for liquidity gauges, so we cannot call it directly.
+        // The checkpoint in this case is triggered by a gauge token transfer from the user to itself.
+        // This is done using the Vault's allowance which is unlimited for gauge tokens.
```
EndymionJkb, line 136:
```suggestion
        // However, some non-permissioned actions cause the gauge to checkpoint a user as a side effect,
        // even if the operation itself is a no-op. The simplest of these is a gauge token transfer, which we
        // perform here. Since the Vault has an unlimited allowance for gauge tokens, and user balance
        // operations use the Vault allowance, no approvals are necessary.
```
Maybe make this "Curve quirk" a little clearer? (Also a Vault user balance trick.)

pkg/standalone-utils/contracts/relayer/GaugeActions.sol
```diff
@@ -30,13 +30,23 @@ abstract contract GaugeActions is IBaseRelayerLibrary {
     using SafeERC20 for IERC20;
 
     IBalancerMinter private immutable _balancerMinter;
+    bool private immutable _isL2Relayer;
```
EndymionJkb, line 33: We need to know this because L1 is permissioned (can't call checkpoint directly), while L2 is not (so you can). Could this ever change in future gauge releases, or even networks? (e.g., what will happen on zkSync? Could we ever have a permissioned L2?)

Consider renaming this semantically: maybe isCheckpointPermissioned, or something like that.

EndymionJkb (approved):
I had one substantive comment about potential renaming, but otherwise LGTM; just suggesting a little extra clarification, since we're using non-intuitive Curve quirks (and relatively seldom-used user balance functionality).

Unfortunate that we need to add another parameter, but no way around it. It's not the first gauge-related thing that has to distinguish between L1 and L2.

pkg/standalone-utils/contracts/relayer/GaugeActions.sol
jubeira: How about https://github.com/balancer/balancer-v2-monorepo/pull/2510/commits/738dc332992c5e61f4dc0918ccf7caa1160f4cc1?

L2s are not permissioned because of the version of the contract that we are using (i.e. the 'new' `ChildChainGauge`). The original contract is from Curve, but we've decided to remove that restriction -since there was always a way around it via transfers, so it was just better to just be honest about it and allow anyone to checkpoint-.

In other words, I don't think that will ever change. But I still agree that it might be better to just name the boolean according to the action.

PR #2515, opened 2023-06-06
Relayer: Chain internal balance operations

### LINKED ISSUE #2351: Support chaining internal balance ops in Batch Relayer
A recent security incident has highlighted one potential pitfall of recovery mode exits (already mentioned in the docs): exits to external addresses require the token transfers to work, and in some failure modes, they might not.

It is still possible to withdraw to internal balance, however, which doesn't do any transfers (yet another hidden use case of that feature). If the "outage is temporary, users would be able to withdraw from internal balance after the issue was resolved.

It would be therefore be useful to be able to combine recovery mode exits and internal balance withdrawals (or transfers), using chained references. Internal balance operations currently do not support chained references.

EndymionJkb:
# Description

The Batch Relayer has always supported internal balance operations - but not with chained references. (And, somehow, there were no tests for it at all!)

Extending it to support chaining would allow more efficient operation; e.g., you could deposit a bunch of tokens into internal balance, then do a big batch swap, all in the same transaction.

It also facilitates certain kinds of "rescue" operations. For example, if a given token got "bricked," such that all transfers failed. This would defeat even a Recovery Mode exit (to an external wallet). Now you could exit to internal balance and, e.g., internal transfer to some sort of helper contract empowered to assist in mitigation, or swap into another token that could be withdrawn, etc.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [X] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #2351

### DIFF
--- a/pkg/standalone-utils/contracts/relayer/VaultActions.sol
+++ b/pkg/standalone-utils/contracts/relayer/VaultActions.sol
@@ -95,11 +95,29 @@ abstract contract VaultActions is IBaseRelayerLibrary {
         }
     }
 
-    function manageUserBalance(IVault.UserBalanceOp[] calldata ops, uint256 value) external payable {
+    function manageUserBalance(
+        IVault.UserBalanceOp[] memory ops,
+        uint256 value,
+        OutputReference[] calldata outputReferences
+    ) external payable {
         for (uint256 i = 0; i < ops.length; i++) {
             require(ops[i].sender == msg.sender || ops[i].sender == address(this), "Incorrect sender");
+
+            uint256 amount = ops[i].amount;
+            if (_isChainedReference(amount)) {
+                ops[i].amount = _getChainedReferenceValue(amount);
+            }
         }
+
         getVault().manageUserBalance{ value: value }(ops);
+
+        // `manageUserBalance` does not return results, but there is no calculation of amounts as with swaps.
+        // We can just use the original amounts.
+        for (uint256 i = 0; i < outputReferences.length; ++i) {
+            require(_isChainedReference(outputReferences[i].key), "invalid chained reference");
+
+            _setChainedReferenceValue(outputReferences[i].key, ops[outputReferences[i].index].amount);
+        }
     }
 
     enum PoolKind { WEIGHTED, LEGACY_STABLE, COMPOSABLE_STABLE, COMPOSABLE_STABLE_V2 }
--- a/pkg/standalone-utils/test/VaultActions.test.ts
+++ b/pkg/standalone-utils/test/VaultActions.test.ts
@@ -3,7 +3,7 @@ import Vault from '@balancer-labs/v2-helpers/src/models/vault/Vault';
 import { BigNumberish, fp } from '@balancer-labs/v2-helpers/src/numbers';
 import TokenList from '@balancer-labs/v2-helpers/src/models/tokens/TokenList';
 import WeightedPool from '@balancer-labs/v2-helpers/src/models/pools/weighted/WeightedPool';
-import { getPoolAddress, SwapKind, WeightedPoolEncoder } from '@balancer-labs/balancer-js';
+import { getPoolAddress, SwapKind, UserBalanceOpKind, WeightedPoolEncoder } from '@balancer-labs/balancer-js';
 import { MAX_INT256, MAX_UINT256, randomAddress, ZERO_ADDRESS } from '@balancer-labs/v2-helpers/src/constants';
 import { expectBalanceChange } from '@balancer-labs/v2-helpers/src/test/tokenBalance';
 import * as expectEvent from '@balancer-labs/v2-helpers/src/test/expectEvent';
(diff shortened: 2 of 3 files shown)


pkg/standalone-utils/test/VaultActions.test.ts
```diff
@@ -1161,4 +1190,324 @@ describe('VaultActions', function () {
...
+
+        itTestsUserBalance();
+      });
+
+      function itTestsUserBalance() {
+        it('sends immediate amounts', async () => {
+          // Internal balance of sender doesn't change
+          // Tokens are transferred from sender to recipient's internal balance
+
+          await expectBalanceChange(
```
jubeira, line 1244:
```suggestion
          // Internal balance of sender doesn't change
          // Tokens are transferred from sender to recipient's internal balance
          // `expectBalanceChange` in this case will track internal balances, since we are passing the vault as an argument.
          await expectBalanceChange(
```

jubeira, line 1244: would it make sense to track the external balance changes..?
Just thinking out loud here, it might be an overkill. This is just a thought because without considering the tiny detail that by sending the vault we are tracking internal balances the changes are not intuitive (i.e. "why is the sender balance not changing?" was the first thing that came to my mind).

pkg/standalone-utils/test/VaultActions.test.ts
```diff
@@ -1161,4 +1190,324 @@ describe('VaultActions', function () {
...
+          expectEvent.inIndirectReceipt(receipt, vault.instance.interface, 'InternalBalanceChanged', {
+            user: TypesConverter.toAddress(recipient),
+            token: tokens.SNX.address,
+            delta: amountSNX,
+          });
+
+          await expectChainedReferenceContents(relayer, toChainedReference(0), amountDAI);
+
+          await expectChainedReferenceContents(relayer, toChainedReference(1), amountSNX);
+        });
```
jubeira, line 1306: This can probably be split in two tests:
- one that checks the events in the receipt
- one that checks the output references

pkg/standalone-utils/test/VaultActions.test.ts
```diff
@@ -1161,4 +1190,324 @@ describe('VaultActions', function () {
...
+
+          expectEvent.inIndirectReceipt(receipt, vault.instance.interface, 'InternalBalanceChanged', {
+            user: TypesConverter.toAddress(recipient),
+            token: tokens.DAI.address,
+            delta: amountDAI,
+          });
+        });
+
+        it('is chainable via multicall', async () => {
+          const receipt = await (
```
jubeira, line 1334:
```suggestion
          // Sender deposits DAI and SNX to relayer, and relayer deposits DAI back to sender, SNX to recipient.
          const receipt = await (
```

pkg/standalone-utils/test/VaultActions.test.ts
```diff
@@ -1161,4 +1190,324 @@ describe('VaultActions', function () {
...
+          const BPT = (await getBPT(poolIdA)).get(0).instance;
+          BPT.connect(user).transfer(TypesConverter.toAddress(sender), await BPT.balanceOf(user.address));
+
+          const amountInBPT = fp(1);
+
+          // Exit Pool A (DAI, MKR) to internal balance
+          // Pretend MKR is bricked (i.e., external transfers fail)
+          // Swap *internally* with Pool B MKR -> SNX
+          // Withdraw DAI and SNX back out to wallet
+          // So external token balances of DAI/MKR should be unchanged, and SNX should equal token out from swap
```
jubeira, line 1432: Are we withdrawing DAI though?
I understand DAI is not withdrawn, so DAI/MKR external token balances are unchanged indeed.

pkg/standalone-utils/test/VaultActions.test.ts
```diff
@@ -1161,4 +1190,324 @@ describe('VaultActions', function () {
...
+
+          const {
+            args: { amountOut: amountOutSNX },
+          } = expectEvent.inIndirectReceipt(receipt, vault.instance.interface, 'Swap', { poolId: poolIdB });
+
+          await expectChainedReferenceContents(relayer, toChainedReference(0), daiAmountOut);
+          expect(snxAmountWithdrawn).to.eq(amountOutSNX);
+
+          // SNX should be in recipient's account.
+          expect(await tokens.SNX.balanceOf(recipient)).to.eq(amountOutSNX);
```
jubeira, line 1508: maybe checking the external transfer event would be safer using `expectTransferEvent`?
We are not checking whether the recipient had SNX before the transfers -- by any means, this check should be correct anyways.

jubeira (approved):
This is great; pretty clean and complete.
I've left just a handful of minor comments and questions, but other than that it's looking good!

pkg/standalone-utils/contracts/relayer/VaultActions.sol
```diff
@@ -95,11 +95,29 @@ abstract contract VaultActions is IBaseRelayerLibrary {
...
         }
+
         getVault().manageUserBalance{ value: value }(ops);
+
+        // `manageUserBalance` does not return results, but there is no calculation of amounts as with swaps.
+        // We can just use the original amounts.
+        for (uint256 i = 0; i < outputReferences.length; ++i) {
+            require(_isChainedReference(outputReferences[i].key), "invalid chained reference");
+
+            _setChainedReferenceValue(outputReferences[i].key, ops[outputReferences[i].index].amount);
```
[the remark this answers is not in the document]
EndymionJkb: Looking at the rest of the code, it's not really documented in a single place, and that's as good a place as any to explain it. Added a giant doc block explaining what references are and how they work (i.e., what index and key mean in the struct). Might be overkill, but you can never have too much documentation :)

pkg/standalone-utils/test/VaultActions.test.ts
```diff
@@ -97,12 +98,17 @@ describe('VaultActions', function () {
     outputReferences?: Dictionary<BigNumberish>;
     sender: Account;
     recipient?: Account;
+    useInternalBalance?: boolean;
```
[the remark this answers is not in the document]
EndymionJkb: There was previously no way to exit to internal balance (in the tests). It was hard-coded to false. Since I needed to do this in one of the tests, I had to add it. It defaults to the previously hard-coded value, so that the existing tests don't change.

pkg/standalone-utils/test/VaultActions.test.ts
EndymionJkb: Yes; that's confusing/unintuitive behavior of `expectBalanceChange`. I wanted to keep the tests as similar as possible, so I didn't change anything with it, but I agree it could benefit from a comment here. I'll probably expand even more.

EndymionJkb: Well, I wrote it this way because the function under test specifically affects internal balances. I added a big documentation block explaining how it works, so hopefully that will address the concern.

PR #2516, opened 2023-06-06
Relayer: query swap/join/exit (V1)

### LINKED ISSUE #2209: Batch relayer: allow calling queries
The next batch relayer should be able to call query swap, join and exit.

EndymionJkb:
Description

This is a version of the Relayer with the same interface as the regular one: except it does queries instead of the actual operations. So the first question is whether we want this stripped down version that only handles Vault operations, or we want to add query methods (with slightly different signatures) to the regular Relayer.

See #2521 for an alternate version of the full Relayer that supports both multicall and queryMulticall operations, with the query version delegated to a second relayer (essentially this one).

It's a draft pending that, and also validation of the testing strategy (before filling out the rest of the test suite). There's just one example test there now, which calls BalancerQuery, then does the corresponding query through the relayer, and verifies the results are identical.

Putting it up for early feedback.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #2209


EndymionJkb:
Closing for now, in favor of the V2 approach (in #2521).

PR #2519, opened 2023-06-08
Improve expect event

jubeira:
# Description

Improve `expectEvent.inIndirectReceipt` when the amount of events are specified.

Without this change, as long as one event matched the given arguments it would pass, even if the expected amount was larger than one. With this change, if the amount is specified we expect all the matches to contain the given arguments.

The return type also changes: it returns a single event for the regular use-case (i.e. when no amount is specified), but returns all of the matches when the amount is specified.

This should be backwards compatible, equally intuitive to use in general when (at least) one event is expected, but more flexible and powerful when analyzing multiple logs in a single receipt.

## Type of change

- [x] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A

### DIFF
--- a/pvt/helpers/src/test/expectEvent.ts
+++ b/pvt/helpers/src/test/expectEvent.ts
@@ -61,7 +61,7 @@ export function inIndirectReceipt(
   }
 
   const exceptions: Array<string> = [];
-  const event = expectedEvents.find(function (e) {
+  const events = expectedEvents.filter(function (e) {
     for (const [k, v] of Object.entries(eventArgs)) {
       try {
         if (e.args == undefined) {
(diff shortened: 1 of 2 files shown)


pvt/helpers/src/test/expectEvent.ts
```diff
@@ -42,6 +42,19 @@ export function inReceipt(receipt: ContractReceipt, eventName: string, eventArgs
...
+ * Expecting a specific amount of events from a particular address is optional.
+ * @param receipt Receipt to analyze.
+ * @param emitter Interface of the contract emitting the event(s).
+ * @param eventName Name of the event(s).
+ * @param eventArgs Arguments of the event(s). This does not need to be a complete list; as long as the event contains
+ *  the specified ones, the function will not throw.
+ * @param address Contract address that emits the event(s). If undefined, the logs will not be filtered by address.
+ * @param amount Amount of expected events that match all the specified conditions. If not specified, at least one is
+ *  expected.
+ * @returns First matching event if the amount is not specified, all matching events otherwise.
```
EndymionJkb, line 56:
```suggestion
 * @returns First matching event if the amount is not specified; all matching events otherwise.
```

pvt/helpers/src/test/expectEvent.ts
```diff
@@ -77,13 +90,20 @@ export function inIndirectReceipt(
     return true;
   });
 
-  if (event === undefined) {
-    // Each event entry may have failed to match for different reasons,
-    // throw the first one
-    throw exceptions[0];
+  // Each event entry may have failed to match for different reasons; in case of failure we throw the first one.
+  if (amount === undefined) {
+    // If amount is undefined, we don't care about the amount of events. If no events were found, we throw.
```
EndymionJkb, line 95:
```suggestion
    // If amount is undefined, we don't care about the number of events. If no events were found, we throw.
```

EndymionJkb (approved):
I wondered whether we actually used this amount specifier anywhere, and we do: in the gauge checkpointer fork test. So it is being tested. (I was going to suggest adding some tests for the test, but I guess running the whole suite does verify that it's backwards compatible.)

I added some lines to VaultActions.test.ts to try it out:
          const events = expectEvent.inIndirectReceipt(receipt, vault.instance.interface, 'Swap', {}, vault.address, 4);
          expect(events.map((e) => e.signature)).to.deep.equal(Array(4).fill('Swap(bytes32,address,address,uint256,uint256)'))

And got an appropriate error message when I changed the expected amount to 5. So it does appear to be working:
      4 'Swap' events found; expected 5

Maybe we could still add some tests (maybe even my "test" tests) to verify the unused features, such as that it returns all the events with amount > 1? (Unless there are already examples I just couldn't find.)

PR #2528, opened 2023-06-27
deprecate security@balancer.finance email

markusbkoch:
balancer.finance is being phased out. We should direct everyone to the docs page where we reference the official communication channels (for now, Immunefi for smart contract issues; email for UI etc)


EndymionJkb (approved):
Good catch! I'd say it's already phased out; only admins have access now, I believe.

PR #2529, opened 2023-07-06
Relayer: support recovery mode exit

### LINKED ISSUE #2458: Add recovery mode exit support in balancer relayer
The relayer supports regular pool exits (given in / given out, etc) for various pool kinds.

However, if the exit kind is 'recovery mode', it reverts for all the existing pool kinds [at this point](https://github.com/balancer/balancer-v2-monorepo/blob/845aff0ae1a6d07dcecb0d033246cb5263da9ca7/pkg/standalone-utils/contracts/relayer/VaultActions.sol#L255).

This is because both [weighted](https://github.com/balancer/balancer-v2-monorepo/blob/master/pkg/interfaces/contracts/pool-weighted/WeightedPoolUserData.sol#L22) and [stable](https://github.com/balancer/balancer-v2-monorepo/blob/master/pkg/interfaces/contracts/pool-stable/StablePoolUserData.sol#L19) pool user data libraries attempt to decode the exit kind, and it's not supported in neither of them. In principle, we wouldn't need to modify the user data in this case.

EndymionJkb:
# Description

Recovery mode exits currently revert on both Weighted and Stable pools. This adds support for them.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #2458

### DIFF
--- a/pkg/standalone-utils/contracts/relayer/VaultActions.sol
+++ b/pkg/standalone-utils/contracts/relayer/VaultActions.sol
@@ -319,7 +319,7 @@ abstract contract VaultActions is IBaseRelayerLibrary {
         returns (bytes memory)
     {
         // Must check for the recovery mode ExitKind first, which is common to all pool types.
-        // If it is just a regular exit, pass it to the appropriate PoolKind handler for interpretation.  
+        // If it is just a regular exit, pass it to the appropriate PoolKind handler for interpretation.
         if (BasePoolUserData.isRecoveryModeExitKind(userData)) {
             return _doRecoveryExitReplacements(userData);
         } else if (kind == PoolKind.WEIGHTED) {
--- a/pkg/standalone-utils/contracts/relayer/VaultActions.sol
+++ b/pkg/standalone-utils/contracts/relayer/VaultActions.sol
@@ -18,6 +18,7 @@ pragma experimental ABIEncoderV2;
 import "@balancer-labs/v2-interfaces/contracts/vault/IVault.sol";
 import "@balancer-labs/v2-interfaces/contracts/pool-weighted/WeightedPoolUserData.sol";
 import "@balancer-labs/v2-interfaces/contracts/pool-stable/StablePoolUserData.sol";
+import "@balancer-labs/v2-interfaces/contracts/pool-utils/BasePoolUserData.sol";
 
 import "@balancer-labs/v2-solidity-utils/contracts/helpers/InputHelpers.sol";
 import "@balancer-labs/v2-solidity-utils/contracts/helpers/VaultHelpers.sol";
@@ -317,7 +318,9 @@ abstract contract VaultActions is IBaseRelayerLibrary {
         private
         returns (bytes memory)
     {
-        if (kind == PoolKind.WEIGHTED) {
+        if (BasePoolUserData.isRecoveryModeExitKind(userData)) {
+            return _doRecoveryExitReplacements(userData);
+        } else if (kind == PoolKind.WEIGHTED) {
             return _doWeightedExitChainedReferenceReplacements(userData);
         } else {
             if (kind == PoolKind.LEGACY_STABLE) {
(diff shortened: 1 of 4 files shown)


pkg/standalone-utils/contracts/relayer/VaultActions.sol
```diff
@@ -317,7 +318,9 @@ abstract contract VaultActions is IBaseRelayerLibrary {
         private
         returns (bytes memory)
     {
-        if (kind == PoolKind.WEIGHTED) {
+        if (BasePoolUserData.isRecoveryModeExitKind(userData)) {
```
jubeira, line 321: Can we add a small comment saying that this check should always go first? Otherwise the weighted / stable encoders will revert.

pkg/standalone-utils/test/VaultActionsStableJoinExit.test.ts
```diff
@@ -723,6 +724,237 @@ describe('Vault Actions - Stable Pools', () => {
...
+      sharedBeforeEach('enter recovery mode', async () => {
+        await stablePool.enableRecoveryMode(admin);
+
+        expect(await stablePool.inRecoveryMode()).to.be.true;
+      });
+
+      context('sender = user', () => {
+        beforeEach(async () => {
+          sender = user;
+          await tokens.approve({ to: vault, from: user });
```
jubeira, line 763: Question: This would be just for the chained swaps right?

pkg/standalone-utils/test/VaultActions.test.ts
```diff
@@ -1196,6 +1204,234 @@ describe('VaultActions', function () {
...
 
+  describe('exit pool in recovery mode', () => {
+    const amountInBPT = fp(1);
+
+    context('when caller is not authorized', () => {
+      it('reverts', async () => {
+        await expect(
+          relayer.connect(other).multicall([
+            await encodeExitPool(vault, relayerLibrary, tokens, {
+              poolKind: PoolKind.WEIGHTED,
```
jubeira, line 1215: We actually don't care about the pool kind in this case right?
How about using an invalid pool kind just to show that it works regardless of the kind?

pkg/standalone-utils/test/VaultActionsStableJoinExit.test.ts
```diff
@@ -723,6 +724,237 @@ describe('Vault Actions - Stable Pools', () => {
...
+    }
+
+    context('when caller is not authorized', () => {
+      it('reverts', async () => {
+        await expect(
+          relayer.connect(other).multicall([
+            await encodeExitPool(vault, relayerLibrary, tokens, {
+              poolKind: PoolKind.COMPOSABLE_STABLE_V2,
+              poolId: poolIdStable,
+              userData: BasePoolEncoder.recoveryModeExit(amountInBPT),
```
jubeira, line 741: (unrelated with this PR): the JS pool encoder uses `(uint256, uint256)`, whereas the code in Solidity uses `uint8` for the enum. Should we rectify that in a different PR?

jubeira (approved):
Better late than never; this looks pretty good and complete, great job!
I've only left a few minor questions and suggestions; otherwise looks good!

pkg/standalone-utils/test/VaultActionsStableJoinExit.test.ts
EndymionJkb: I think this is generally inconsistent all over the place; we've noted that before, and it might be a big change to a lot of files.

We may have even had this conversation before, about whether it's worth it, and (I believe) concluded it should be fine to have JS all be uint256, which should still resolve to the same value whether the underlying Solidity is 8, 16, 32, etc. (it can only be equal or smaller), even if the definition of enum changes at some point.

In fact, it might even be "safer" to leave JS at 256, so that we don't have to change it if enums go from uint8 to uint16 in a future version

pkg/standalone-utils/test/VaultActionsStableJoinExit.test.ts
EndymionJkb: Yes; if you comment it out, that specific test fails.

pkg/standalone-utils/test/VaultActions.test.ts
EndymionJkb: That doesn't seem to work; low level failure, I guess when mapping the enum. It doesn't care what the pool kind is, but it seems to have to be a valid pool kind.

jubeira: then would it make sense to make it random in the test to convey that it's the same? Or do you think it'd be too much?

EndymionJkb: Seems a little much :) Maybe if I just add a comment?

jubeira:
:ship: when CI finishes

jubeira (approved):

PR #2533, opened 2023-07-17
L2 gauge checkpointer fixes and refactor

### LINKED ISSUE #2524: L2GaugeCheckpointer: fix checkpoint timestamp
We are currently checkpointing the current week.

The problem is that for a gauge to mint, it needs to have votes over a week, _wait for the week to finish_, and then perform the checkpoint.

If the votes change from week to week (in particular, if they go down), the relative weight filter may filter out gauges that need to be checkpointed.

The current checkpointer still allows checkpointing individual gauges or lowering the threshold to 0, so it's not impossible to poke the gauges in those cases. But it'd be more efficient to just fix the filter to consider the previous period (or any custom period, now that we are at it).

jubeira:
# Description

The name `L2GaugeCheckpointer` is a bit confusing now that we actually have a checkpointer for child chain gauges. What this checkpointer does is automating the checkpoint of stakeless (root) gauges in mainnet, so a refactor (`L2GaugeCheckpointer` --> `StakelessGaugeCheckpointer`) seemed appropriate.

On top of that, this PR introduces two fixes to address corner cases:
- We now use the start of the _previous_ week to check the relative weight of a gauge in the gauge controller. This is to match the [actual behavior in stakeless gauges](https://github.com/balancer/balancer-v2-monorepo/blob/c4cc3d466eaa3c1e5fa62d303208c6c4a10db48a/pkg/liquidity-mining/contracts/gauges/StakelessGauge.sol#L147).
- We now checkpoint the gauges in the controller before filtering gauges by relative weight in case they are not updated (if they are not, the controller returns 0 and we might accidentally skip gauges with non-zero weight).

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:

- [ ] Tests are included for all code paths --> This will be done in a fork test

## Issue Resolution

Closes #2524.

### DIFF
--- a/pkg/interfaces/CHANGELOG.md
+++ b/pkg/interfaces/CHANGELOG.md
@@ -18,8 +18,9 @@
   - Added event `GaugeTypeAdded`, and changed `GaugeFactoryAdded` for `GaugeFactorySet`.
   - Added `getGaugeTypes` method.
 - Moved `GaugeType` from `IGaugeAdder` to `IL2GaugeCheckpointer`, and adjusted types to accept new networks.
-- Refactored `IL2GaugeCheckpointer`.
+- Refactored `IL2GaugeCheckpointer` to `StakelessGaugeCheckpointer`.
   - Removed `isSupportedGaugeType` from interface.
+  - Added `getRoundedDownBlockTimestamp` to interface.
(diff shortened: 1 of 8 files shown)


ylv-io (approved):
The code gods have smiled upon thee!

PR #2534, opened 2023-07-17
Stakeless gauge checkpointer: multi gauge checkpoint

jubeira:
# Description

This PR adds the capability to checkpoint multiple specific gauges in the same call.
The effect would be the same as calling `checkpointSingleGauge` multiple times, but it should be much cheaper to do so in one call because the storage reads are cheaper after the first time they are performed in the same tx.

Built on top of #2533; merge that one first.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A


ylv-io (approved):
🔥

PR #2542, opened 2023-08-11
Remove unnecessary test contract

EndymionJkb:
# Description

The old Avalanche bridge had a unit test and supporting Mock. The new one no longer needs this. We removed the test, but forgot the Mock contract.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira (approved):
So this is a weird case.

I'm actually using this contract [here](https://github.com/balancer/balancer-deployments/blob/master/tasks/20230811-avalanche-root-gauge-factory-v2/test/task.fork.ts#L340) as it was included in the `build-info` for that release.

I need it to test some specific conditions that happen when interacting with the real bridge while sending different amounts of BAL. This is because the bridge we're using truncates the amounts being sent, and under some conditions it could revert if we don't do things right. Testing those conditions with the mock is very convenient because we have control over the exact transfer amounts, which is required to verify that the gauge works properly.

On the other hand, we cannot just include this mock in the sources of the deployments repo; we'd need either need to copy the root gauge code (a deal breaker) or make `deployments` depend on `liquidity-mining` which is also kinda awkward.

We can remove this anyways, since the fork test will still work. If we do so, we might want to leave a reference in that PR or somewhere in the fork test so that the original code for the mock can be searched for easily. WDYT?

EndymionJkb:
Made a PR in development that moves the code there (commented out so it doesn't actually compile and introduce dependencies), and pointed to it from the fork test.

PR #2555, opened 2023-11-09
Support "recovery" exits from stable phantom pools

EndymionJkb:
# Description

We noted that the relayer doesn't support "recovery" exits from StablePhantom pools. I believe this was omitted because such exits are only supported when the pools are paused, and the original StablePhantom pools are past their pause window. So this would really only be needed if someone redeployed the factory (e.g., on a new L2). (Or, I suppose, if someone were fork testing something in an old block within the pause window.)

Mostly a dotting i's, crossing t's thing, since V5 was supposed to support *all* stable pool types but (understandably / deliberately) left out StablePhantoms: and included a test for them that didn't actually test the relayer.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

- [ ] Tests are included for all code paths (there is a working fork test in deployments, based off this build)


EndymionJkb:
Not supported

PR #2577, opened 2025-06-09
Veboost v2.1

EndymionJkb:
# Description

Allow Stake DAO and Tetu to use veBAL held in their respective lockers to be used for boosts. Redeploy veBoostV2 with no semantic changes, just a "preseeding" mechanism similar to the earlier "preseeded voting escrow delegation" mechanism, which migrates all current (non-expired) boosts and grants infinite allowance for Stake DAO and Tetu "operators" to create new boosts with using their locked veBAL.

It is version 2.1 instead of 3, since the contract code is unchanged. (V1 to V2 was a substantive change, as V1 used NFTs to represent boosts.) This just removes V1 and NFT migration references, adds the preseeding mechanism, modifies the internal `_boost` to accept "start times" in the past (necessary for simple preseeding), and checks for the Stake DAO/Tetu approval exceptions for new boosts.

There is an incompatibility between the Vyper ABI and what ethers expects (tuple vs. array), so transformed it on deployment to keep the hardhat tests passing.

See [this deployment PR](https://github.com/balancer/balancer-deployments/pull/270) for the current boosts and (unconfirmed) operator/delegator configuration. Note that two were expiring on 6/12, so the deployment target  was set to 6/13 to avoid them. There are two others expiring in June: 6/19 and 6/26.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/liquidity-mining/contracts/VeBoostV2.vy
+++ b/pkg/liquidity-mining/contracts/VeBoostV2.vy
@@ -1,6 +1,6 @@
 # @version 0.3.3
 """
-@title Boost Delegation V2
+@title Boost Delegation V2.1
 @author CurveFi
 """
 
@@ -47,7 +47,7 @@ struct Point:
 
 NAME: constant(String[32]) = "Vote-Escrowed Boost"
 SYMBOL: constant(String[8]) = "veBoost"
-VERSION: constant(String[8]) = "v2.0.0"
+VERSION: constant(String[8]) = "v2.1.0"
 
 EIP712_TYPEHASH: constant(bytes32) = keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)")
 PERMIT_TYPEHASH: constant(bytes32) = keccak256("Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)")
--- a/pkg/liquidity-mining/contracts/VeBoostV2.vy
+++ b/pkg/liquidity-mining/contracts/VeBoostV2.vy
@@ -22,9 +22,6 @@ event Boost:
     _slope: uint256
     _start: uint256
 
-event Migrate:
-    _token_id: indexed(uint256)
-
 
 interface BoostV1:
     def ownerOf(_token_id: uint256) -> address: view
@@ -72,9 +69,6 @@ delegated_slope_changes: public(HashMap[address, HashMap[uint256, uint256]])
 
 received: public(HashMap[address, Point])
 received_slope_changes: public(HashMap[address, HashMap[uint256, uint256]])
-
-migrated: public(HashMap[uint256, bool])
-
 
 @external
 def __init__(_boost_v1: address, _ve: address):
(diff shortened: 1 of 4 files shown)


pkg/liquidity-mining/contracts/VeBoostV2.vy
```diff
@@ -73,15 +88,42 @@ delegated_slope_changes: public(HashMap[address, HashMap[uint256, uint256]])
 received: public(HashMap[address, Point])
 received_slope_changes: public(HashMap[address, HashMap[uint256, uint256]])
 
-migrated: public(HashMap[uint256, bool])
+# Flag to ensure migration is only performed once
+migrated: public(bool)
+
+# Checked on calls to `boost`, enabling certain Stake DAO and Tetu accounts
+# to create boosts on behalf of their respective lockers
+isApprovedForAll: public(HashMap[address, HashMap[address, bool]])
```
jubeira, line 96: Is this really needed though? Just setting the approval to max should be enough without modifying the code, as the allowance is not reduced when used in that case.

pkg/liquidity-mining/contracts/VeBoostV2.vy
```diff
@@ -225,39 +271,15 @@ def _boost(_from: address, _to: address, _amount: uint256, _endtime: uint256):
 
 @external
 def boost(_to: address, _amount: uint256, _endtime: uint256, _from: address = msg.sender):
-    # reduce approval if necessary
-    if _from != msg.sender:
+    if _from != msg.sender and self.isApprovedForAll[_from][msg.sender] == False:
```
jubeira, line 274: If we just set the allowance from delegator to operator to `MAX_UINT256` we shouldn't need anything related to `isApprovedForAll`. Am I missing anything?

pkg/liquidity-mining/contracts/VeBoostV2.vy
```diff
@@ -287,11 +317,24 @@ def permit(_owner: address, _spender: address, _value: uint256, _deadline: uint2
     )
 
     if _owner.is_contract:
-        sig: Bytes[65] = concat(_abi_encode(_r, _s), slice(convert(_v, bytes32), 31, 1))
+        sig: Bytes[65] = concat(
```
jubeira, line 320: Are these changes coming from a linter?

jubeira (commented):
Good work @EndymionJkb, first pass done.

Overall looks fair. I'll do a second pass looking at the fork test to do a cross check. If I'm not mistaken we don't really need `isApprovedForAll`, other than that LGTM.

pkg/liquidity-mining/contracts/VeBoostV2.vy
EndymionJkb: No; as above, was just done for consistency with the previous preseed contract, and self-documentation. Otherwise there's no indication it was done (so I can just document it here instead).

pkg/liquidity-mining/contracts/VeBoostV2.vy
```diff
@@ -395,3 +437,169 @@ def DOMAIN_SEPARATOR() -> bytes32:
 @external
 def VE() -> address:
     return VE
+
+
+# Called once for each existing active boost. We're migrating from a previous
```
jubeira, line 442: :+1:

jubeira (approved):
Just one more minor comment, otherwise LGTM! Thanks for pushing and documenting this!

PR #2585, opened 2025-07-23
Docs: fix broken Boosted Pool link in Linear Pool README

Hopium21:
pkg/pool-linear/README.md
Old link:
https://docs.balancer.fi/concepts/pools/boosted.html#boosted-pools
New link:
https://docs.balancer.fi/concepts/explore-available-balancer-pools/boosted-pool.html#boosted-pools

The previous URL no longer resolves. This fix points to the correct, current location of Boosted Pool docs to avoid confusion for devs reading this README.


EndymionJkb (commented):
V2 should point to V2 docs when available.

EndymionJkb (approved):
Looks good now; thanks!

PR #2588, opened 2025-07-29
Fix broken docs link in Linear Pools README

Hopium21:
Found a broken link in the README pointing to the old Balancer docs.
Replaced it with the updated working link to the Boosted Pools section on v2.balancer.fi


EndymionJkb (approved):
LGTM! Thanks for pointing to V2.

PR #2590, opened 2025-08-01
 fix: correct typo in test report documentation

Hopium21:
Fixed typo "BTP" -> "BPT" (Balancer Pool Token) in the test report file. This is the standard abbreviation in the Balancer ecosystem that should be written as "BPT".


EndymionJkb (commented):
Looks good - but there are more. I'd also fix the test that generated the report (e.g., BaseWeightedPool.behavior.ts). It's also still in BasePool and NewBasePool (just search for BTP). Might as well get all instances of it.

Hopium21:
These were actually the last remaining instances I could find - the ones you mentioned (BasePool, NewBasePool, etc.) were already updated earlier. Just double-checked everything again to be sure.

Should be all good now

EndymionJkb (approved):
Right! I hadn't pulled the latest. Thanks again!

PR #2595, opened 2025-08-12
Fix typo in comment in FeeDistributor.sol

Hopium21:
Corrected a typo in the inline comment, changing "though" to "through" for clarity.


EndymionJkb (approved):
Thanks!

PR #2598, opened 2025-08-14
fix: Hardhat artifact imports and contract factory usage

defiberrys:
# Description

updated the code to use the public Hardhat API.

* replaced `import { Artifacts } from 'hardhat/internal/artifacts'` with `import { artifacts, Artifact } from 'hardhat'` and rewrote `getArtifact` using `artifacts.readArtifactSync`.
* switched from the non-existent `ethers.getContractFactoryFromArtifact` to `ethers.getContractFactory(artifact.abi, artifact.bytecode, signer)`.

everything now works with the official Hardhat API without relying on internal modules.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pvt/helpers/src/contract.ts
+++ b/pvt/helpers/src/contract.ts
@@ -1,9 +1,7 @@
-import { ethers } from 'hardhat';
+import { ethers, artifacts, Artifact } from 'hardhat';
 import { Contract } from 'ethers';
 import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/dist/src/signer-with-address';
 
-import { Artifacts } from 'hardhat/internal/artifacts';
-import { Artifact } from 'hardhat/types';
 import path from 'path';
 import { Dictionary } from 'lodash';
 
@@ -30,7 +28,7 @@ export async function deploy(
 
   const artifact = getArtifact(contract);
 
-  const factory = await ethers.getContractFactoryFromArtifact(artifact, { signer: from, libraries });
+  const factory = await ethers.getContractFactory(artifact.abi, artifact.bytecode, from);
   const instance = await factory.deploy(...args);
 
   return instance.deployed();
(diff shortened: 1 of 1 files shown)


pvt/helpers/src/contract.ts
```diff
@@ -30,7 +28,7 @@ export async function deploy(
 
   const artifact = getArtifact(contract);
 
-  const factory = await ethers.getContractFactoryFromArtifact(artifact, { signer: from, libraries });
+  const factory = await ethers.getContractFactory(artifact.abi, artifact.bytecode, from);
```
jubeira, line 31: I think we also would need to send the libraries to link here: https://v2.hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-ethers

pvt/helpers/src/contract.ts
```diff
@@ -67,7 +65,6 @@ export function getArtifact(contract: string): Artifact {
     artifactsPath = `${packagePath}/artifacts`;
   }
 
-  const artifacts = new Artifacts(artifactsPath);
   return artifacts.readArtifactSync(contract.split('/').slice(-1)[0]);
```
jubeira, line 68: So we don't need the `artifactsPath` anymore? how is the global `artifacts` linked to the path in this case, considering we have multiple workspaces in the monorepo?

jubeira (commented):
Thanks for sending this @defiberrys.

Curious how you got there. The helpers under `pvt` might not be perfectly written, but they don't seem to have any apparent issues either. Do you see any specific problem with the current implementation that brings in internal imports?

defiberrys:
> Curious how you got there. The helpers under `pvt` might not be perfectly written, but they don't seem to have any apparent issues either. Do you see any specific problem with the current implementation that brings in internal imports?

I switched to the public Hardhat API to avoid relying on internal modules, which are unstable and not guaranteed to maintain backward compatibility. using `artifacts.readArtifactSync` and `ethers.getContractFactory` ensures the code works with officially supported APIs, making it more robust and future-proof. the old helpers may work now, but they could break on Hardhat updates, whereas the public API is stable and officially supported.

PR #2603, opened 2025-08-19
fix: correct JSDoc parameter name in exitExactBPTInForOneTokenOut

Hopium21:
The function exitExactBPTInForOneTokenOut has parameter named exitTokenIndex in the signature, but JSDoc incorrectly referenced it as enterTokenIndex. This creates inconsistency between the code and documentation.

- Changes @param enterTokenIndex to @param exitTokenIndex
- Aligns with the function's exit operation purpose
- Matches the pattern used in stable pool encoder


EndymionJkb (approved):
Thanks

PR #2604, opened 2025-08-22
Fix grammar in L2 NatSpec

Hopium21:
Description:
Replace “in a L2 chain” with “on an L2 chain” across comments/NatSpec in:

pkg/interfaces/contracts/liquidity-mining/IL2LayerZeroDelegation.sol
pkg/liquidity-mining/contracts/L2LayerZeroBridgeForwarder.sol


EndymionJkb:
> Description: Replace “in a L2 chain” with “on an L2 chain” across comments/NatSpec in:

Looks like it still says "on a L2" instead of "on an L2"

EndymionJkb (approved):
Thanks!

PR #2605, opened 2025-08-27
docs: add AMM math reference link for pools

mmdavenport:
Adds a concise link to the Balancer math overview to help new contributors orient. No code changes.


EndymionJkb:
This link doesn't go anywhere. Also, this is V2, so all doc references should be docs-v2.balancer.fi. I don't think there is a general "AMM math" section, because the math is different for every pool type.

EndymionJkb:
> Thanks! Updated to docs-v2 and linked the specific math pages for each pool type (Weighted, Stable, Linear). No general “AMM math” page in v2, so this should be clearer.

Sounds good, but doesn't look like it got pushed to the fork. Still just see the original commit on this end.

Where-Mental-Wellness-Meets-Metaverse:
Pushed the docs-v2 links for Weighted, Stable, and Linear pool math pages as discussed.

PR #2606, opened 2025-08-28
Fix typos in comments

Hopium21:
Changes

LinearPool.sol:
care must be take → care must be taken (grammar fix).

LinearPool.test.ts:
it('adapt …') → it('adapts …') (subject–verb agreement).

ComposableStablePoolProtocolFees.test.ts:
balaced → balanced (spelling).
there balance change → their balance change (possessive pronoun).

### DIFF
--- a/pkg/pool-stable/test/ComposableStablePoolProtocolFees.test.ts
+++ b/pkg/pool-stable/test/ComposableStablePoolProtocolFees.test.ts
@@ -281,7 +281,7 @@ describe('ComposableStablePoolProtocolFees', () => {
 
       const PREMINTED_BPT = fp(9e9); // This is the BPT normally stored in the Pool's accounting at the Vault
 
-      // We want a Pool that is relatively balanced so that, with a reasonably high amplification factor, each token has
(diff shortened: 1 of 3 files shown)


pkg/pool-stable/test/ComposableStablePoolProtocolFees.test.ts
```diff
@@ -281,7 +281,7 @@ describe('ComposableStablePoolProtocolFees', () => {
 
       const PREMINTED_BPT = fp(9e9); // This is the BPT normally stored in the Pool's accounting at the Vault
 
-      // We want a Pool that is relatively balaced so that, with a reasonably high amplification factor, each token has
+      // We want a Pool that is relatively balanced so that, with a reasonably high amplification factor, each token has
       // similar prices
```
EndymionJkb, line 285:
```suggestion
      // We want a Pool that is relatively balanced so that, with a reasonably high amplification factor, each token
      // has similar prices
```
Change will go over the line length limit.

EndymionJkb (commented):
A couple suggestions so we don't have lint errors

PR #2608, opened 2025-08-30
fix: correct balance comparison in ComposablePoolLib test

Hopium21:
Fix incorrect address casting in dropBptFromBalances test. The test was trying to compare uint256 balance values as addresses, which would cause compilation errors. Now correctly compares balance values directly.


EndymionJkb (approved):
Well it was apparently working anyway, but it is odd and technically wrong. No harm in fixing it. Thanks!

PR #2611, opened 2025-09-02
Fix typos in comments & docs

Hopium21:
Changes

ManagedPool.test.ts
“so is represented” → “are/is represented” (grammar fix in test comments).

BaseSplitCodeFactory.sol
“crated” → “created” (typo in NatSpec).

WordCodec.sol
“accesing” → “accessing” (typo in comment).

### DIFF
--- a/pkg/pool-weighted/test/ManagedPool.test.ts
+++ b/pkg/pool-weighted/test/ManagedPool.test.ts
@@ -471,7 +471,7 @@ describe('ManagedPool', function () {
             const joinSwapResult = await queryExitSwap();
             const joinResult = await queryEquivalentExit();
 
-            // Tokens are leaving the Vault and so is represented as a negative value.
+            // Tokens are leaving the Vault and so are represented as a negative value.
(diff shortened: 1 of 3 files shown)


pkg/pool-weighted/test/ManagedPool.test.ts
```diff
@@ -471,7 +471,7 @@ describe('ManagedPool', function () {
             const joinSwapResult = await queryExitSwap();
             const joinResult = await queryEquivalentExit();
 
-            // Tokens are leaving the Vault and so is represented as a negative value.
+            // Tokens are leaving the Vault and so are represented as a negative value.
```
EndymionJkb, line 474:
```suggestion
            // Tokens are leaving the Vault, so they're represented as negative values.
```
Still had a number mismatch; might as well completely fix it.

EndymionJkb (commented):
We didn't use spell checkers in v2. One suggestion on the first comment.

PR #2624, opened 2025-09-16
Generalized swap fee helper

### LINKED ISSUE #2620: Swap fee helper for weighted pools
We need a swap fee setter for certain pools that come from the weighted pool factory (EZKL managed).

We should apply the same pattern as V3: a contract with an allowlist that allows the partner to set the static swap fee percentage only on allowlisted pools (i.e. not all the pools that come from the same factory).

EndymionJkb:
# Description

This parallels PR 1537 in v3, generalizing the swap fee helper to support multiple partners in a single contract. It works generally the same; the main difference is that pools have control of swap fees in v2, while in v3 it's all centralized in the Vault. While in v3 we just need to grant a single permission for the contract to call a function on the Vault, in v2 governance must grant the contract permission to call the function on each pool type (the actionId is the same for all pools from a given factory).

This upgrades the OZ infrastructure to support the new functionality for parity with v3. In keeping with the rest of v2, the OZ files are not imported, but copied into the relevant library.

Since we don't have custom errors in v2, this adds new errors to BalancerErrors. In order to reuse the tests from v3, equivalent test infrastructure had to be added as well.

It also uses bytes32 poolIds instead of pool addresses, as that is how v2 identifies pools. In order to test the "unregistered pool" functionality, the MockVault now needed to implement registration and poolId handling.

Otherwise, the interface is the same as v3.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #2620


jubeira (approved):
Awesome job @EndymionJkb, this was very clean!

I had a few comments that I had to delete upon a closer look, so this seems good to go.

The only thing (outside the diff) is that `standalone-utils` doesn't run forge tests in CI. Can we add that before merging? (check `test-pool-weighted` for reference if needed).

Thanks for pushing this!

PR #2636, opened 2025-10-04
chore: remove unused @rollup/plugin-commonjs devDependency

ruslan0012:
Remove @rollup/plugin-commonjs from devDependencies as it's not used anywhere in the rollup config. 

The plugin is only needed for bundling CommonJS modules, but this project:
- Uses TypeScript/ESM modules only
- Marks all dependencies as external (not bundled)
- Has no CommonJS code to transform

Verified by checking rollup.config.ts (no import/usage) and running full build + tests successfully without it.


EndymionJkb (approved):
Thanks; this does look like it's unused.

